PR #36, opened 2010-12-01
Pluggable broadcast implementations

mosharaf:
# Summary

Three different implementations of broadcast that pluggable through the config option "spark.broadcast.Factory":
1. DfsBroadcast:
   - Write to DFS from master and read back from everywhere else
2. ChainedBroadcast:
   - Divide the broadcast variable into blocks and form a O(n)-length chain from the master.
   - Can be and should be generalized to any d-ary tree instead of an 1-ary tree.
3. BitTorrentBroadcast: Stripped-down bit torrent with the master as the seed
   - Very fast and fairly distributed (upload:download=~1:~1) distribution
   - Special case with only a few leechers have been resolved
# Notes
- Code files as well broadcast related config files should be restructured
- There are still lots of logging going on in Chained and BitTorrent implementations. Should be turned OFF at some point.
- conf/\* files are included in the 'mos-bt' branch. Should be ignored/removed before committing the merge with 'master'


mateiz:
This looks pretty good overall, but I noticed a few small things:
- Can you remove the conf files (git rm) to make the merge happen more easily? You're free to copy them and leave them around, just don't have them be under git's control.
- I noticed that you changed the default shuffle implementation in RDD.scala (see the "files changed" view). Change it back to DfsShuffle.

mosharaf:
Done. Couple of things about the config files: before using Streaming or BitTorrent, one must update spark.broadcast.MasterHostAddress with Master's IP address and will want to update spark.broadcast.BlockSize to their liking.

mateiz:
Noticed a few other things:
- Why is sendBroadcast part of the public interface of trait Broadcast? It's only called internally, so it doesn't need to be in the Broadcast trait.
- The default broadcast implementation should be the HDFS one, not the BitTorrent one, until we get some more testing of BT. This is especially true if BT requires you to configure those variables.
- When initializing BitTorrent broadcast on the master, figure out your local IP address using code such as http://www.exampledepot.com/egs/java.net/Local.html and set the spark.broadcast.MasterHostAddress system property to that. It will automatically be transferred to the slaves when they get launched. This way the user doesn't need to configure it manually.
- Your naming of system properties is a bit inconsistent with existing ones. You capitalize the first letter of the last word (e.g. spark.broadcast.BlockSize), whereas all the existing properties don't (e.g. spark.broadcast.blockSize). It would be good to uncapitalize those.

mosharaf:
Done. Let me know if you find anything else.

PR #47, opened 2011-04-28
Cache to disk

ankurdave:
DiskSpillingCache is a cache implementation derived from BoundedMemoryCache that spills entries to disk when it runs out of space.

The location on disk where entries are stored is configurable through the spark.diskSpillingCache.cacheDir property.


mateiz:
Thanks Ankur, I committed this.

PR #48, opened 2011-05-03
Bagel: Large-scale graph processing on Spark

ankurdave:
Bagel is an implementation of the Pregel graph processing framework on Spark.

Bagel currently supports basic graph computation, combiners, and aggregators. Future work includes support for mutating the graph topology. Tests exist but currently don't run due to a Spark bug.

### DIFF
--- a/bagel/src/main/scala/bagel/Pregel.scala
+++ b/bagel/src/main/scala/bagel/Pregel.scala
@@ -1,0 +1,103 @@
+package bagel
+
+import spark._
+import spark.SparkContext._
+import scala.collection.mutable.HashMap
+import scala.collection.mutable.ArrayBuffer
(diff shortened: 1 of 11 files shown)


tjhunter:
I would recommend you refactor your code before merging, it is always harder / less tempting to do after.

mateiz:
This looks great, Ankur, except for two naming things: can you change the package name from bagel to spark.bagel, and can you rename the Pregel class to Bagel?

ankurdave:
Sure, I've done so.

mateiz:
Looks great, thanks. The one thing I should add is that maybe you should write a README documenting the examples, or a wiki page (and put a comment in the code pointing to this location).

PR #52, opened 2011-05-26
Fix deprecations when compiled with Scala 2.8.1

ijuma:
See issue #50. The change for `--` and `-` are the only controversial ones in my opinion and there's no better alternative as far as I know (it's the suggested approach in the deprecated message).


ijuma:
Oh, I also upgraded the sbt-idea plugin to version 0.4.0.

mateiz:
Thanks, looks great!

PR #53, opened 2011-05-27
Use explicit asInstanceOf instead of misleading unchecked pattern matching.

ijuma:
This makes the code a bit uglier, but lets someone easily spot where dangerous things are being done. It also silences the compiler so that real "unchecked" warnings can be spotted.

Also enabled -unchecked warnings in SBT build file and restructured it a bit to make that easier.

### DIFF
--- a/core/src/main/scala/spark/CacheTracker.scala
+++ b/core/src/main/scala/spark/CacheTracker.scala
@@ -96,7 +96,7 @@ class CacheTracker(isMaster: Boolean, theCache: Cache) extends Logging {
   // Get a snapshot of the currently known locations
   def getLocationsSnapshot(): HashMap[Int, Array[List[String]]] = {
     (trackerActor !? GetCacheLocations) match {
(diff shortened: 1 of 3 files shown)


mateiz:
Looks good. It looks like your editor also made some formatting changes in the build file, but we can keep those.

Is there any widely used Scala style guide out there? We've tried to adhere to http://www.codecommit.com/scala-style-guide.pdf but it doesn't cover everything.

ijuma:
Thanks for merging. Sorry for the formatting changes in the build file, it seemed like two styles were being used in the same file so I changed it to be consistent with typical Scala style, but forgot to mention it.

I think http://davetron5000.github.com/scala-style/ScalaStyleGuide.pdf is a more recent version of the style guide that Daniel started. In terms of what's used out there, it varies quite a lot and this variation happens even inside the Scala standard library and compiler. I'll have a look at the Style Guide once again (it's been a while since I looked at it) and try to stick to that for Spark then.

PR #58, opened 2011-06-01
Remove unnecessary toStream calls

ijuma:
It could be that I am missing something, but it looks to me like the toStream calls are unnecessary. Used the scala-2.9 branch as I believe that will become master soon.


mateiz:
I'm pretty sure those were needed in 2.8, but it looks like they aren't in 2.9.

PR #59, opened 2011-06-02
Move managedStyle to SparkProject

ijuma:
I had added it to DepJar by mistake. This is not urgent and it could be merged after the other work you have planned for the coming week. I just put it here as a heads-up.


mateiz:
Thanks for the fix.

PR #63, opened 2011-06-24
Implemented RDD.leftOuterJoin and RDD.rightOuterJoin

ogrisel:
I decided to implement them from scratch on the `join` method model rather than filtering out the output of `groupWith` to avoid emitting intermediate results that are dropped right afterward.

I think it's better to use the scala Option type to handle missing values rather that null values as in SQL or Pig, tell be if you don't aggree.

_Note_: There is no scaladoc on those methods as they all lack such documentation in the RDD class. I really think this is missing though. Maybe the RDD should be fully documented at once in another pull request?

### DIFF
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -242,6 +242,44 @@ extends RDD[Array[T]](prev.context) {
     }
   }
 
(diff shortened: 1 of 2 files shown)


mateiz:
Looks good, thanks. A docs patch for RDD would also be appreciated.

PR #64, opened 2011-06-27
Functionality to save RDDs to Hadoop files

tdas:
Overview of changes
1> Introduced class HadoopFileWriter that uses Hadoop's libraries (OutputFormat, OutputCommitter, etc) to save the records in a RDD, each split of the RDD being a file in the destination directory. While instantiating HadoopFileWriter, one can specify which classes to use for key, value, output-format and output-committer. The way it is written, all subclasses of outputformat and outputcommitter should be applicable.

2> Introduced following functions in RDD
(a) RDD[T].saveAsTextFile(<path>): toString will be used to convert records (type T) to String and saved as a text file.
(b) RDD[T].saveAsObjectFile(<path>): Each record, if serializable, will be converted to bytes and saved as serialized objects.  
(b) RDD[(K,V)].saveAsHadoopFile(<path>): Key-value pair RDDs will be saved as a Hadoop file using whatever OutputFormat and OutputCommitter class that the user specifies. This has a few different overloaded syntaxes. 
(c) RDD[(Writable, Writable)].saveAsSequenceFile(<path>): Key-value pair RDDs with keys and value being subclasses of Writable (like IntWritable) or can be implicitly converted to Writable (Int -> IntWritable, etc) can be saved as a Sequential File.

3> Other changes
(a) SparkContext.objectFile(<path>): opens files that have been saved by saveAsObjectFile
(b) A number of implicit functions to convert basic data types to its Writable counterparts
(c) Introduced class TaskContext, that provides a running Task with the contextual information like stage ID, task ID.
(d) Scheduler and Task classes changed to generate and populate TaskContext information


ijuma:
Did you test this when using s3n Paths?

tdas:
No, I havent tested it using s3n paths. I have tested it only with local and hdfs paths. It would be great if some one is able to test s3n paths.

ijuma:
I tried it and ran into some issues during the set-up of OutputCommitter. I'll file an issue with more details if it hasn't been verified to work then.

tdas:
Great, I will look into it once we have the details. Thanks!

ijuma:
I filed https://github.com/mesos/spark/issues/81

PR #76, opened 2011-08-02
Fix code not to use code deprecated in Scala 2.9

ijuma:
The commits are self-explanatory.


mateiz:
Awesome, thanks a lot for taking the time to do this!

PR #82, opened 2011-08-31
Upgrade to Scala 2.9.1.

ijuma:
Scala 2.9.1 has been released and it includes a lot of improvements, including faster compilation:

http://www.scala-lang.org/node/10780

A couple of small changes in the REPL support were needed in addition to the changes in the build file and run script. Please review them to see if they're correct.

All tests pass and ./spark-shell launches successfully.


mateiz:
Looks good, thanks!

PR #103, opened 2011-12-16
Made improvements to takeSample. Also changed SparkLocalKMeans to SparkKMeans

edisontung:
Instead of calling count on an RDD multiple times, stored the count so it can be used. Also, called the collect method for the array earlier also to reduce the number of times count() is called.


core/src/main/scala/spark/RDD.scala
```diff
@@ -91,6 +91,41 @@ abstract class RDD[T: ClassManifest](@transient sc: SparkContext) extends Serial
...
+  	else {
+  		fraction = Math.min(multiplier*(num+1)/count(), 1.0)
+  		total = num.toInt
+  	}
+	
+  	var r = new SampledRDD(this, withReplacement, fraction, seed)
+    var samples = r.collect()
+	
+  	while (samples.length < total) {
+  		r = new SampledRDD(this, withReplacement, fraction, seed)
```
mateiz: Here you are not calling samples = r.collect() again, so samples.length will always remain less than total.

core/src/main/scala/spark/RDD.scala
```diff
@@ -91,6 +91,41 @@ abstract class RDD[T: ClassManifest](@transient sc: SparkContext) extends Serial
...
+  	}
+	
+  	var arr = new Array[T](total)
+	
+  	for (i <- 0 to total - 1) {
+  		arr(i) = samples(i)
+  	}
```
mateiz: Instead of this you could write val arr = samples.take(total), or just return samples.take(total).

core/src/main/scala/spark/RDD.scala
```diff
@@ -91,6 +91,41 @@ abstract class RDD[T: ClassManifest](@transient sc: SparkContext) extends Serial
...
+  	var multiplier = 3.0
+  	var initialCount = count()
+	
+  	if (num > initialCount) {
+  		total = Math.min(initialCount, Integer.MAX_VALUE)
+  		total = total.toInt
+  		fraction = 1.0
+  	}
+  	else if (num < 0) {
+  		throw(new IllegalArgumentException())
```
mateiz: Add some text in the exception ("Negative number of elements requested")

core/src/main/scala/spark/RDD.scala
```diff
@@ -91,6 +91,41 @@ abstract class RDD[T: ClassManifest](@transient sc: SparkContext) extends Serial
...
+  	}
+  	else {
+  		fraction = Math.min(multiplier*(num+1)/count(), 1.0)
```
mateiz: Here you are still calling count() instead of using initialCount.

core/src/main/scala/spark/RDD.scala
```diff
@@ -91,6 +91,41 @@ abstract class RDD[T: ClassManifest](@transient sc: SparkContext) extends Serial
...
+  		total = num.toInt
+  	}
+	
+  	var r = new SampledRDD(this, withReplacement, fraction, seed)
+    var samples = r.collect()
```
mateiz: You could probably also just say val samples = this.sample(withReplacement, fraction, seed).collect()

mateiz:
Hey Edison,

This looks good but I made a few comments on the diff, including finding a couple of bugs (another extra call to count() and a potential infinite loop).

The other thing I noticed is that your new kmeans_data.txt is huge -- 750K lines. It would be better to replace it with the simple one we had before (or just generate another small one), so that people can look at it more easily.

mateiz:
Looks good. Thanks!

PR #107, opened 2012-01-24
Add support for fixed sized samples

holdenk:
Fix for issue #93, adding support for fixed sized samples.

### DIFF
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -91,6 +91,10 @@ abstract class RDD[T: ClassManifest](@transient sc: SparkContext) extends Serial
   def sample(withReplacement: Boolean, fraction: Double, seed: Int): RDD[T] =
     new SampledRDD(this, withReplacement, fraction, seed)
 
+  def collectSample(withReplacement: Boolean, count: Int, seed: Int): RDD[T] =
+    new CollectSampledRDD(this, withReplacement, count, seed)
+
+
   def takeSample(withReplacement: Boolean, num: Int, seed: Int): Array[T] = {
   	var fraction = 0.0
   	var total = 0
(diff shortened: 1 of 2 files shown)


core/src/main/scala/spark/RDD.scala
```diff
@@ -78,54 +78,29 @@ abstract class RDD[T: ClassManifest](@transient sc: SparkContext) extends Serial
...
-    var samples = this.sample(withReplacement, fraction, seed).collect()
-	
-  	while (samples.length < total) {
-  		samples = this.sample(withReplacement, fraction, seed).collect()
-  	}
-	
-  	val arr = samples.take(total)
-	
-  	return arr
+    val initialCount = count()
```
mateiz: Why is this still here? Why not just return collectSample(...)?

holdenk: Good point, I'll remove this.

mateiz:
Hey, sorry for being a bit late on this, I didn't see you posted a new pull request. Would you like me to grab the SortedRDD too or just the rest? As far as I understand, SortedRDD is not complete. We actually had our own effort to implement one so it might be better to use that instead.

holdenk:
Oh crap I forgot about sortedrdd being in my branch, just the other stuff
would be cool :)
On Feb 5, 2012 8:55 PM, "Matei Zaharia" <
reply@reply.github.com>
wrote:

mateiz:
Okay, so I looked at this more carefully, but unfortunately I think there are some problems:

1) There is an array index out of bounds exception on CollectSampledRDD.scala line 41 (in sampleData.update(i,oldData(i))), because the for loop there goes over indices from 1 to sampleSize instead of 0 to sampleSize-1.

2) It looks like the code is pulling out sampleSize items inside each split (partition) of the dataset, rather than sampleSize total items. This is not going to work as the user expects for datasets with more than 1 partition, unless we explicitly document this as the use of the function, which I don't like because I don't want the user to worry about the number of partitions. More generally, partitions might also have different numbers of items, so some might have fewer than sampleSize while some have more.

I think it might be better to stick with the approach in the current collectSample for now, because it avoids these issues. In particular, it would be really hard to know in advance how many items are in each partition of the dataset, and pass the right number of items to the tasks that work on that object. The current code just tries to take a sample that is slightly bigger than the required number of items using independent coin flips on each element, and then pulls out the right number of items.

mateiz:
I closed the pull request because the older collectSample seems to work okay, but thanks for trying this out!

PR #111, opened 2012-02-11
Adding sorting to RDDs

alupher:
Fix for issue #88, adding sort to RDDs. This adds a sortByKey(ascending) method to RDDs of key/value pairs where keys have an Ordered trait. The 'ascending' argument is a boolean to specify order.

(Note: still getting unchecked / eliminated by erasure warnings for getPartition in RangePartitioner class)

### DIFF
--- a/core/src/main/scala/spark/PairRDDFunctions.scala
+++ b/core/src/main/scala/spark/PairRDDFunctions.scala
@@ -377,7 +377,7 @@ class PairRDDFunctions[K: ClassManifest, V: ClassManifest](
     override val partitioner = prev.partitioner
     override val dependencies = List(new OneToOneDependency(prev))
     override def compute(split: Split) = {
-      prev.iterator(split).toList
+      prev.iterator(split).toArray
         .sortWith((x, y) => if (ascending) x._1 < y._1 else x._1 > y._1).iterator
     }
   } 
--- a/core/src/main/scala/spark/PairRDDFunctions.scala
+++ b/core/src/main/scala/spark/PairRDDFunctions.scala
@@ -359,6 +359,29 @@ class PairRDDFunctions[K: ClassManifest, V: ClassManifest](
   def getValueClass() = implicitly[ClassManifest[V]].erasure
 }
 
+  class OrderedRDDFunctions[K <% Ordered[K]: ClassManifest, V: ClassManifest](
+    self: RDD[(K, V)])
+    extends Logging 
+    with Serializable {
+
+    def sortByKey(ascending: Boolean = true): RDD[(K,V)] = {
+      val rangePartitionedRDD = self.partitionBy(new RangePartitioner(self.splits.size, self, ascending))
+      new SortedRDD(rangePartitionedRDD, ascending)
+    }
+  }
+  
+  class SortedRDD[K <% Ordered[K], V](prev: RDD[(K, V)], ascending: Boolean)
+    extends RDD[(K, V)](prev.context) {
+
+    override def splits = prev.splits
+    override val partitioner = prev.partitioner
+    override val dependencies = List(new OneToOneDependency(prev))
+    override def compute(split: Split) = {
+      prev.iterator(split).toList
+        .sortWith((x, y) => if (ascending) x._1 < y._1 else x._1 > y._1).iterator
+    }
+  } 
+ 
 class MappedValuesRDD[K, V, U](prev: RDD[(K, V)], f: V => U) extends RDD[(K, U)](prev.context) {
   override def splits = prev.splits
   override val dependencies = List(new OneToOneDependency(prev))
--- a/core/src/main/scala/spark/Partitioner.scala
+++ b/core/src/main/scala/spark/Partitioner.scala
@@ -24,3 +24,40 @@ class HashPartitioner(partitions: Int) extends Partitioner {
       false
   }
 }
+
+class RangePartitioner[K <% Ordered[K],V](partitions: Int, rdd: RDD[(K,V)], ascending: Boolean = true) 
+  extends Partitioner {
+
+  def numPartitions = partitions
+
+  val rddSize = rdd.count()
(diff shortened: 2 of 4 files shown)


core/src/main/scala/spark/PairRDDFunctions.scala
```diff
@@ -359,6 +359,29 @@ class PairRDDFunctions[K: ClassManifest, V: ClassManifest](
...
+  }
+  
+  class SortedRDD[K <% Ordered[K], V](prev: RDD[(K, V)], ascending: Boolean)
+    extends RDD[(K, V)](prev.context) {
+
+    override def splits = prev.splits
+    override val partitioner = prev.partitioner
+    override val dependencies = List(new OneToOneDependency(prev))
+    override def compute(split: Split) = {
+      prev.iterator(split).toList
```
mateiz: Here it would be more efficient to use toArray instead of toList

core/src/main/scala/spark/Partitioner.scala
```diff
@@ -23,4 +23,41 @@ class HashPartitioner(partitions: Int) extends Partitioner {
...
+          }   
+        if (ascending) p else numPartitions-1-p
+      }   
+      case _ => 0
+    }
+  }
+
+  override def equals(other: Any): Boolean = other match {
+    case r: RangePartitioner[K,V] =>
+      r.numPartitions == numPartitions
```
mateiz: Should also check whether r.rangeBounds == rangeBounds

core/src/main/scala/spark/Partitioner.scala
```diff
@@ -23,4 +23,41 @@ class HashPartitioner(partitions: Int) extends Partitioner {
...
+
+  def numPartitions = partitions
+
+  val rddSize = rdd.count()
+  val maxSampleSize = partitions*10.0
+  val frac = 1.0.min(maxSampleSize / rddSize)
+  val rddSample = rdd.sample(true, frac, 1).collect.toList
+    .sortWith((x, y) => if (ascending) x._1 < y._1 else x._1 > y._1)
+    .map(_._1)
+  val bucketSize:Float = rddSample.size / partitions
```
mateiz: This is still going to do an integer division (rddSample.size / partitions) and then cast the resulting integer to a float. Not sure if this is what you wanted. Also, please leave a colon between the : and the Float to match the code style in other places.

core/src/main/scala/spark/Partitioner.scala
```diff
@@ -23,4 +23,41 @@ class HashPartitioner(partitions: Int) extends Partitioner {
...
+  val rangeBounds = rddSample.zipWithIndex.filter(_._2 % bucketSize == 0)
+    .map(_._1).slice(1, partitions)
+
+  def getPartition(key: Any): Int = { 
+    key match {
+      case k:K => {
```
mateiz: Here you'll get a type erasure warning because the K type is lost a runtime. Any kind of object will thus match this case. Instead of doing the match statement, just assume that key is a K and cast it to a K (val k = key.asInstanceOf[K]).

core/src/main/scala/spark/Partitioner.scala
```diff
@@ -23,4 +23,41 @@ class HashPartitioner(partitions: Int) extends Partitioner {
...
+              if (k > bound) index + 1 else part
+          }   
+        if (ascending) p else numPartitions-1-p
+      }   
+      case _ => 0
+    }
+  }
+
+  override def equals(other: Any): Boolean = other match {
+    case r: RangePartitioner[K,V] =>
```
mateiz: You might get a type erasure warning because K and V are not known at runtime. Use RangePartitioner[_, _] instead.

mateiz:
Hey Antonio,

I made a few comments on the code (in the Diff tab). Also, do you mind adding a unit test for this? Add a file in core/src/test/spark. We use ScalaTest for testing (http://www.scalatest.org/). You can probably just copy and edit one of the existing tests to get the scaffolding right.

alupher:
Great, thanks for your comments, Matei. I've made the changes in my fork and will add a unit test as well.

alupher:
By the way, that does take care of the type erasure warnings :)

mateiz:
Great. By the way, the fixes look good, but I think you forgot to git add the test file.

Matei

mateiz:
Hey Antonio, have you had a chance to work on a unit test for sorting yet?

Matei

alupher:
Hi Matei, just added some unit tests, let me know what you think.

-Antonio

mateiz:
Looks good. Thanks Antonio!

mateiz:
Hey Antonio,

Just FYI, I just made a commit to fix a few issues with sorting:
- The range partitioner would crash with a division by zero if the RDD was smaller than the number of partitions requested.
- Some stuff was inefficient -- in particular, all the local vals you declared to build up the rangeBounds remained as fields of the RangePartitioner object, which made it huge.
- The unit test with high parallelism was taking way too long; we can speed it up further by replacing the getPartition code with a binary search but for now I've just left it out.

Here's the commit: https://github.com/mesos/spark/commit/c7af538ac160727147eea6a1008dc16a2efd802e. Might be helpful for Shark too.

Matei

alupher:
Hey Matei,

Thanks for your notes on the commit, that was very helpful. We'll make
sure to incorporate them in the version of Spark we're using. Sorry
for missing some of those issues the first time around.

We also recently updated our Spark to use defaultParallelism as the
default number of partitions for the sortedRDD (and defined a
sortByKey that lets users choose this value). Here's the relevant
commit: https://github.com/alupher/spark/commit/13d5ffaf087e06e900ac9168d060a927274ff33e
 Do you think this would be good merge into the spark core?  If so, I
can send a pull request.

Antonio

On Sat, Mar 17, 2012 at 1:54 PM, Matei Zaharia
reply@reply.github.com
wrote:

PR #112, opened 2012-03-01
Changed HadoopRDD to get key and value containers from the RecordReader instead of through reflection

cengle:
This allows users to specify the key or value class of a HadoopRDD as Writable, whereas they needed to use a subclass of Writable before. I ran the FileSuite and everything passed.


mateiz:
Looks great, thanks!

PR #119, opened 2012-04-06
Report entry dropping in BoundedMemoryCache

ankurdave:
When BoundedMemoryCache dropped a key, it previously only logged the event to the local log on the slave. Now it reports that fact to the master as well.

This makes it possible to write tools that process the driver output and visualize cache usage across the cluster. It also will make it easier for Arthur to track this information.


rxin:
This is awesome :)

mateiz:
Thanks!

PR #121, opened 2012-04-10
Added an option (spark.closure.serializer) to specify the serializer for closures.

rxin:
This enables using Kryo as the closure serializer.


mateiz:
Cool! Does this mean that closures can be serialized properly this way?

Since the parameter is called spark.closure.serializer, I would not change Utils.serialize and Utils.deserialize to use it, because the names for those would be misleading. Instead, create an instance of the closure serializer in SparkEnv and change the places that pass a closure or a TaskResult to use it.

rxin:
I will make the change accordingly.

I tested it and it worked. Previously the problem was that Class[_] was not serialized properly. This was solved by using a ClassSerializer in Kryo. (In 2.0, this becomes automatic.)

rxin:
Updated. PTAL.

mateiz:
Looks good. Thanks!

PR #127, opened 2012-04-23
End task instead of just exiting in LocalScheduler for tasks that throw exceptions

alupher:
(our Shark testing suite driver was hanging on tests for which an exception was thrown on a worker)


mateiz:
Two comments:
- Don't leave commented-out code in the repo (the //System.exit(1)); we can always see the old code using version control.
- Do Spark's own unit tests (sbt/sbt test) still pass? Some of them do test exceptions thrown in tasks (but use the LocalScheduler setting that allows some failures).

alupher:
Yep, the tests pass. I've removed the commented-out line.

mateiz:
Looks good, thanks!

PR #129, opened 2012-04-24
Force serialize/deserialize task results in local execution mode.

rxin:
In Shark, we were caught by surprise when we tested map join implementation on a cluster that it'd failed because of serialization. This commit forces a serialization/deserialization of task result data even for local modes to avoid such surprises.


rxin:
I added a new test for this too. Note that the test passing is contingent upon Antonio's patch (otherwise System.exit wouldn't give scala test a chance to intercept the exception).

mateiz:
Looks good, thanks!

PR #131, opened 2012-05-15
Return size estimation, cache usage, and cache capacity from slave nodes to CacheTracker

rxin:
Also updated the log messages to give information on cache capacity when entries are added/dropped.


rxin:
I added another commit into this. It is the one that reduces the amount of data to serialize in closure serialization by making dependencies field transient.

mateiz:
Great, thanks!

PR #132, opened 2012-05-20
Little refactoring and unit tests for CacheTrackerActor

Benky:


core/src/main/scala/spark/Utils.scala
```diff
@@ -85,13 +62,8 @@ object Utils {
...
   {
     val buf = new Array[Byte](8192)
-    var n = 0
-    while (n != -1) {
-      n = in.read(buf)
-      if (n != -1) {
-        out.write(buf, 0, n)
-      }
-    }
+    Iterator.continually(in.read(buf)).takeWhile(_ != -1).foreach(out.write(buf, 0, _))
```
mateiz: Sorry, but while this is elegant, it's going be slightly slower than the while loop, and confusing to readers due to the mix of side-effects and FP.

core/src/main/scala/spark/Utils.scala
```diff
@@ -35,29 +35,6 @@ object Utils {
...
-        buf += s.substring(i, j);
-      }
-      i = j
-      while (i < s.length && !isAlpha(s.charAt(i))) {
-        i += 1
-      }
-    }
-    return buf
-  }
-
```
mateiz: Don't delete these because we use them in some tests and benchmarks.

mateiz:
Thanks a lot for these fixes! I had some comments on Utils.scala though -- in particular, leave the fast splitString method in because we've used it in some benchmarks, and revert the simplification of copyStream because it will be confusing to people without a lot of knowledge of how those FP operations work. If you make those changes, I'll merge this in.

core/src/main/scala/spark/Utils.scala
Benky: Ok. Could you please commit these tests?

mateiz:
Thanks for the updates. I'll add in a program to use the splitString thing soon (we basically used it to build a fast WordCount program, but it's mostly been in demos we typed at the command line).

PR #133, opened 2012-05-20
BoundedMemoryCache.put should fail when estimated size of 'value' is larger than cache capacity

Benky:
Current implementation allows you to store object larger than BoundedMemoryCache.getCapacity into cache, which seems to be wrong.


mateiz:
Looks good, thanks!

PR #134, opened 2012-05-27
ShuffleManager & RDD refactored. Some unit tests added.

Benky:


mateiz:
Thanks for posting this. Unfortunately this is a rather busy week for me, but I hope to look at it in the next few days.

core/src/main/scala/spark/RDD.scala
```diff
@@ -206,22 +180,12 @@ abstract class RDD[T: ClassManifest](@transient sc: SparkContext) extends Serial
...
-        result += 1L
-        iter.next
-      }
-      result
-    }).sum
+    val results = sc.runJob(this, (iter: Iterator[T]) => iter.aggregate(zeroValue)(cleanSeqOp, cleanCombOp))
+    results.fold(zeroValue)(cleanCombOp)
   }
 
+  def count(): Long = sc.runJob(this, (iter: Iterator[T]) => iter.size).sum
```
mateiz, line 187: This is a bug because it will cause integer overflow for datasets with more than 2^31 - 1 elements. That's why we were returning a Long before. You should probably convert the sizes to Longs. Also, I'm not positive about this, but it's possible that Iterator.size() is slower than the while loop there. It would be good to test that.

core/src/main/scala/spark/RDD.scala
```diff
@@ -82,9 +67,8 @@ abstract class RDD[T: ClassManifest](@transient sc: SparkContext) extends Serial
   // Transformations (return a new RDD)
   
   def map[U: ClassManifest](f: T => U): RDD[U] = new MappedRDD(this, sc.clean(f))
-  
-  def flatMap[U: ClassManifest](f: T => Traversable[U]): RDD[U] =
-    new FlatMappedRDD(this, sc.clean(f))
+
+  def flatMap[U: ClassManifest](f: T => Traversable[U]): RDD[U] = new FlatMappedRDD(this, sc.clean(f))
```
mateiz, line 71: We tried to stick to a 100-character line length limit. If it's not much hassle and you can change stuff back, it would be good to stay with that.

core/src/test/scala/spark/RDDSuite.scala
```diff
@@ -1,26 +1,143 @@
...
+  test("sample with replacement") {
+    val sc = new SparkContext("local", "test")
+
+    val nums = (1 to 100).toArray
+    assert(sc.makeRDD(nums).sample(true, 1, 20).collect().distinct.diff(nums) === Array())
+    assert(sc.makeRDD(nums).sample(true, 0, 20).collect().size < 100)
+
+    sc.stop()
+  }
+
```
mateiz, line 140: Thanks for adding all of these, this is awesome.

mateiz:
Finally got a chance to look at this. It looks great except for some formatting and a bug and possible performance regression in count() (you should check how Iterator.size is implemented -- I think it uses foreach() and ends up being slower than what I wrote).

core/src/test/scala/spark/RDDSuite.scala
```diff
@@ -1,26 +1,143 @@
 package spark
 
-import scala.collection.mutable.HashMap
 import org.scalatest.FunSuite
-import SparkContext._
+import util.Random
+import collection.mutable.HashMap
```
mateiz, line 5: One other thing: please prefix the scala.\* packages with scala.*, because it's clearer for new readers and that's what the rest of the codebase does.

mateiz:
Closing this pull request because the code has diverged very far now. Let me know if you'd still like some of these to be added.

PR #142, opened 2012-06-29
Scalacheck groupId has changed https://github.com/rickynils/scalacheck/i...

rrmckinley:
...ssues/24. Necessary to build with scalaVersion 2.9.2. Works with 2.9.1 too.


mateiz:
Thanks for the fix!

PR #145, opened 2012-07-12
add Accumulatable, add corresponding docs & tests for accumulators

squito:
Preliminary version of some updates to accumulators -- it works, but probably requires some discussion around the exact API.

I didn't want to break all existing accumulators, so to add an `addInPlace(T,Y)`
method meant creating a whole new type, which I've called `Accumulatable`. Also, I had to distinguish the old `+=` method from the new one.  I chose `+:=`, but that was rather arbitrary, I'm no expert on scala idioms.  (It seems that the old operator is really closer to `++=`, and the new one should be `+=`, but I guess we can't change that.)

I've also added tests for accumulators.  These include an example where the current value of an accumulator is read during a task.  The example application is stochastic gradient descent, where seeing the current value is really critical.  Should `accumulator.value` throw an exception (so the user doesn't think they're getting the global value), but maybe instead we expose `accumulator.localValue`?

### DIFF
--- a/core/src/main/scala/spark/Accumulators.scala
+++ b/core/src/main/scala/spark/Accumulators.scala
@@ -11,7 +11,7 @@ class Accumulable[T,R] (
   
   val id = Accumulators.newId
   @transient
-  var value_ = initialValue // Current value on master
+  private var value_ = initialValue // Current value on master
   val zero = param.zero(initialValue)  // Zero value to be passed to workers
   var deserialized = false
 
@@ -30,7 +30,13 @@ class Accumulable[T,R] (
    * @param term the other Accumulable that will get merged with this
    */
   def ++= (term: T) { value_ = param.addInPlace(value_, term)}
-  def value = this.value_
+  def value = {
+    if (!deserialized) value_
+    else throw new UnsupportedOperationException("Can't use read value in task")
+  }
+
+  private[spark] def localValue = value_
+
   def value_= (t: T) {
     if (!deserialized) value_ = t
     else throw new UnsupportedOperationException("Can't use value_= in task")
@@ -124,7 +130,7 @@ private object Accumulators {
   def values: Map[Long, Any] = synchronized {
     val ret = Map[Long, Any]()
     for ((id, accum) <- localAccums.getOrElse(Thread.currentThread, Map())) {
-      ret(id) = accum.value
+      ret(id) = accum.localValue
     }
     return ret
   }
--- a/core/src/main/scala/spark/Accumulators.scala
+++ b/core/src/main/scala/spark/Accumulators.scala
@@ -35,9 +35,40 @@ class Accumulator[T] (
   override def toString = value_.toString
 }
 
+class Accumulatable[T,Y](
+    @transient initialValue: T,
+    param: AccumulatableParam[T,Y]) extends Accumulator[T](initialValue, param) {
+  /**
+   * add more data to the current value of the this accumulator, via
+   * AccumulatableParam.addToAccum
+   * @param term
+   */
+  def +:= (term: Y) {value_ = param.addToAccum(value_, term)}
+}
+
+/**
+ * A datatype that can be accumulated, ie. has a commutative & associative +
+ * @tparam T
+ */
 trait AccumulatorParam[T] extends Serializable {
   def addInPlace(t1: T, t2: T): T
   def zero(initialValue: T): T
+}
+
+/**
+ * A datatype that can be accumulated.  Slightly extends [[spark.AccumulatorParam]] to allow you to
+ * combine a different data type with value so far
+ * @tparam T the full accumulated data
+ * @tparam Y partial data that can be added in
+ */
+trait AccumulatableParam[T,Y] extends AccumulatorParam[T] {
+  /**
+   * Add additional data to the accumulator value.
+   * @param t1 the current value of the accumulator
+   * @param t2 the data to be added to the accumulator
+   * @return the new value of the accumulator
+   */
+  def addToAccum(t1: T, t2: Y) : T
 }
 
 // TODO: The multi-thread support in accumulators is kind of lame; check
--- a/core/src/main/scala/spark/Accumulators.scala
+++ b/core/src/main/scala/spark/Accumulators.scala
@@ -19,7 +19,7 @@ class Accumulable[T,R] (
 
   /**
    * add more data to this accumulator / accumulable
-   * @param term
+   * @param term the data to add
    */
   def += (term: R) { value_ = param.addAccumulator(value_, term) }
 
@@ -27,7 +27,7 @@ class Accumulable[T,R] (
    * merge two accumulable objects together
    * <p>
    * Normally, a user will not want to use this version, but will instead call `+=`.
-   * @param term
+   * @param term the other Accumulable that will get merged with this
    */
   def ++= (term: T) { value_ = param.addInPlace(value_, term)}
   def value = this.value_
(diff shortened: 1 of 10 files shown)


core/src/test/scala/spark/AccumulatorSuite.scala
```diff
@@ -0,0 +1,233 @@
...
+        assert (assertVal < 100)
+      }
+    }
+  }
+
+}
+
+
+
+//ugly copy and paste from examples ...
```
mateiz: This is definitely unfortunate. Maybe we should just move the Vector class to a spark.util package in core, so that both the examples and tests can depend on it.

mateiz:
Hey, so one major comment on this. Instead of making Accumulable a subclass of Accumulator, I think it would be better to do things the other way around. That is, Accumulable[T, Y] should be the main class that Spark cares about, and Accumulator[T] should extend Accumulable[T, T]. This will require changing a bit more stuff in the rest of the Spark code that depends on accumulators, but it's worth it. Also, if you do this, then the +:= method can just be called +=. The += in Accumulator will be a special case for when T and Y are the same. Does this sound good?

core/src/main/scala/spark/Accumulators.scala
```diff
@@ -35,11 +35,42 @@ class Accumulator[T] (
   override def toString = value_.toString
 }
 
+class Accumulatable[T,Y](
```
mateiz: I would just call this "Accumulable" instead of "Accumulatable", because it's a bit shorter, and because I think Accumulatable is not actually a real word (though I do see some other APIs use it). Also, rename the types to something like [T, R], where T is the type of values you add and R is the result.

squito:
this is a great suggestion.  I'll make the change and update the pull request

On Thu, Jul 12, 2012 at 10:46 AM, Matei Zaharia
reply@reply.github.com
wrote:

core/src/main/scala/spark/Accumulators.scala
```diff
@@ -17,7 +17,19 @@ class Accumulator[T] (
 
   Accumulators.register(this, true)
 
-  def += (term: T) { value_ = param.addInPlace(value_, term) }
+  /**
+   * add more data to this accumulator / accumulable
+   * @param term
```
mateiz: Remove these @param and @return annotations if you aren't filling in the information about them

core/src/main/scala/spark/Accumulators.scala
```diff
@@ -17,7 +17,19 @@ class Accumulator[T] (
...
+   */
+  def += (term: R) { value_ = param.addToAccum(value_, term) }
+
+  /**
+   * merge two accumulable objects together
+   * <p>
```
mateiz: Also remove the <p>, I think ScalaDoc deals with blank lines by itself

core/src/main/scala/spark/Accumulators.scala
```diff
@@ -35,8 +47,49 @@ class Accumulator[T] (
...
+    @transient initialValue: T,
+    param: AccumulatorParam[T]) extends Accumulable[T,T](initialValue, param)
+
+/**
+ * A simpler version of [[spark.AccumulableParam]] where the only datatype you can add in is the same type
+ * as the accumulated value
+ * @tparam T
+ */
+trait AccumulatorParam[T] extends AccumulableParam[T,T] {
+  def addToAccum(t1: T, t2: T) : T = {
```
mateiz: I think you should leave the name of this as addInPlace so that people's old code will still compile. Instead, rename the new addToAccum method to something like addAccumulator.

core/src/main/scala/spark/Accumulators.scala
```diff
@@ -35,8 +47,49 @@ class Accumulator[T] (
...
+ * <p>
+ * You must define how to add data, and how to merge two of these together.  For some datatypes, these might be
+ * the same operation (eg., a counter).  In that case, you might want to use [[spark.AccumulatorParam]].  They won't
+ * always be the same, though -- eg., imagine you are accumulating a set.  You will add items to the set, and you
+ * will union two sets together.
+ *
+ * @tparam T the full accumulated data
+ * @tparam R partial data that can be added in
+ */
+trait AccumulableParam[T,R] extends Serializable {
```
mateiz: Swap the meaning of T and R so that R is the result and T is the partial data (this will be more consistent with other code we have); or leave T as the accumulated type and change R to V for Value.

mateiz:
This looks good! I made a few changes on the diff but once you have those I think it's good to merge in. The most important is keeping the name of the old addInPlace method on accumulator so that people don't have to modify their code.

core/src/main/scala/spark/Accumulators.scala
squito: I think this works as is.  Since AccumulableParam is the base trait, every AccumulatorParam has to implement both addInPlace and addToAccum.  My version of AccumulatorParam just provides a default definition of addToAccum, so the user doesn't have to -- existing code will still work.

You can see I didn't change IntAccumulatorParam at all (it only defines addInPlace, as before) and the tests in AccumulatorSuite verify it still works.

I thought this is what you were thinking of when you suggested changing the hierarchy around -- did you have something else in mind?

core/src/test/scala/spark/AccumulatorSuite.scala
```diff
@@ -0,0 +1,123 @@
...
+            if (trueClassProb <= 0.95)
+              badErrs += 1
+          }
+        }
+        println("Iteration " + itr + " had badErrs = " + badErrs.value)
+        weights += weightDelta.value
+        println(weights)
+        //TODO I should check the number of bad errors here, but for some reason spark tries to serialize the assertion ...
+//        val assertVal = badErrs.value
+//        assert (assertVal < 100)
```
squito: I should have mentioned this earlier -- I don't know why, but I get serialization exceptions if I uncomment these assertions.  This is outside the calls to spark, so I am not sure why its getting serialized.  Can you suggest a fix here?  It would be nice if the unit test actually asserted something ...

core/src/main/scala/spark/Accumulators.scala
mateiz: Ah, makes sense. I misread the code and thought you'd replaced addInPlace. However, can you rename it to something like addAccumulator? We've tried not to have abbreviated names in public APIs.

core/src/test/scala/spark/AccumulatorSuite.scala
mateiz: Weird; I'll take a look into this, but I'm traveling in the next couple of days so I may be slow to respond.

mateiz: So I finally had a chance to look into this, and I think it's an artifact of how Scala creates the closure objects, which leads to a reference to the outer class existing in the closure. Unfortunately, it's not easy to fix this, but if you're curious, there is a workaround:

1) Move DataPoint to be a top-level class in the file, instead of a nested class

2) Right before the foreach, add local variables that reference the variables from the outer scopes:

```
    val weightDelta_ = weightDelta
    val data_ = data
    val weights_ = weights
```

Then use these variables in your closure instead of the outer ones.

This helps because the body of each for() loop is actually a closure, and doing this prevents us from having to reference the outer for() loop. Similarly, with DataPoint being an inner class inside the method, each DataPoint instance could actually hold a reference to the whole Suite.

However, I would actually not include the test, because I realize that the value being sent across is an artifact of local mode, and is not going to work when you run on a cluster. This is because value is transient in Accumulable. Looking back at why we did this, reason was actually that in some cases the value is large (e.g. a hash table) and is not used on the worker nodes, so this will waste a lot of network bandwidth. So I'd rather make it an error to access value on the cluster, and recommend that for applications that do want to use it, they assign it to a local variable. For example:

```
val accum = sc.accumulator(...)
for (i <- 1 to ITERATIONS) {
  val oldValue = accum.value
  data.foreach( /* do stuff with oldValue */ )
}
```

Does this sound okay? If so, I'd like to merge this pull request, because there's a lot of useful stuff here, and we can figure out the value thing later. (Maybe we can add another class for when you do want to see the old value).

mateiz:
Hey, not sure if you saw, but I posted a line comment about the closure failing in the test. I think it's best to remove the whole test because the value being readable on slave nodes is actually not the intended behavior right now, and only happens in local mode, not on a cluster. See the explanation there. (Just posting here in case you don't get notified on line comments).

squito:
Hi,

sorry I hadn't gotten to this yet, just got sidetracked (I did get
notified of the inline comments too, btw).

I can definitely make this change and resubmit.  I'd like to also make
reading from accumulator.value throw an exception on slave nodes (the
same way that assigning a new value does), so that this is more clear.
 I know this will break some existing code (including some of our
own), but really this is just silently waiting to explode when someone
goes from local mode to a cluster.  I think we'd prefer to prevent
those as soon as possible.

I still don't really understand why the value shouldn't be readable,
though.  Its not any more efficient to make the value a local variable
and send that workers, is it?  You'd just rather the user have to do
it explicitly, so they realize how much data they are sending around?

I'll try to write the SGD using mapPartitions example at some point in
a separate set of commits ...

thanks

On Mon, Jul 23, 2012 at 8:26 PM, Matei Zaharia
reply@reply.github.com
wrote:

mateiz:
Looks good, thanks a lot!

The reason that making value readable is tricky is because we actually try to initialize it to "zero" in each thread in readObject(), in order to add back all the changes at the end. This just didn't happen in local mode with one thread because we already have that accumulator in our thread. So we'd need to have two fields -- the "real" value and the "accumulating" value on this worker thread -- and two separate ways to read them. Anyway, I'll think more about including this. Having the exception in there will help a lot until then.

PR #147, opened 2012-07-15
Broadcast refactoring/cleaning up

mosharaf:
- Removed DfsBroadcast completely.
- Removed ChainedBroadcast (TreeBroadcast is a superset of this)
- Refactored code for TreeBroadcast and BitTorrentBroadcast (still there seems to be a lot of code overlapped, but they have differences in the details). The common object is called MultiTracker that basically acts as the torrent indexing service for ongoing broadcasts.
- Added stop() for the broadcast subsystem. Haven't added it to SparkContext though
- Merged several configuration options to one for simplicity. The ones that remain are setup to work without any changes. Just changing the broadcastFactory should do.
- Changed "isLocal" in SparkContext to not consider localhost as local. To use local you have to use "local" and not "localhost". The reasoning is that using the localhost IP still acts like that, but localhost doesn't; plus, it helps in checking broadcast in the same machine without having to use the actual IP.

TODO
- Add a HttpBroadcast backup for TreeBroadcast and BitTorrentBroadcast methods just in case they fail. TreeBroadcast shouldn't. BitTorrentBroadcast has a weird corner case that will take some time in figuring out and hasn't been addressed in this patch.


mateiz:
Hey Mosharaf, did you ever get this to work with the block manager? Apart from that, the other thing we should do is make the broadcast objects be classes instead of singleton objects. You can then initialize and stop them in SparkContext. I've recently done that for the BlockManagerMaster too.

mosharaf:
BroadcastManager is a class now with init() and stop() methods that are called from SparkEnv. Individual broadcast mechanisms still have their own objects though.

mateiz:
Thanks. Are you planning to remove the individual objects too or is that too much work for right now?

mateiz:
(Just wondering whether this is ready to merge)

mateiz:
I've merged this to make it easier to develop new stuff. Thanks! But we should still remove the standalone objects for MultiTracker & co, for people who want to have multiple SparkContexts in the same app.

PR #151, opened 2012-07-18
Examples ship to to cluster

dennybritz:
- Fixes #118
- I wasn't sure how to best get the path to the example JAR, I didn't want to do it manually in every examples. I added another variable in the ./run script that's being used by the examples. Let me know if that's not the best way.
- Should update the example documentation to tell people to run "sbt/sbt package" before running the examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,8 @@ will need to set the `SCALA_HOME` environment variable to point to where
 you've installed Scala. Scala must be accessible through one of these
 methods on Mesos slave nodes as well as on the master.
 
-To run one of the examples, use `./run <class> <params>`. For example:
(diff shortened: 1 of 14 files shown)


mateiz:
This looks good, but can you also update the README to say that you should run sbt package instead of sbt compile? Otherwise the examples JAR won't get built. Also, rename the EXAMPLES_JAR variable to SPARK_EXAMPLES_JAR so that it doesn't conflict with other environment variables the user might make.

mateiz:
Hey Denny, not sure if you saw my comment here, but can you rename the environment variable as mentioned above, and update the README to say sbt package? Otherwise it looks good.

dennybritz:
I made the changes you suggested.

mateiz:
Great, thanks!

PR #152, opened 2012-07-18
Always destroy SparkContext in after block for the unit tests.

dennybritz:
- Fixes #148
- Tested. Tests still passing, except for one that wasn't passing before, see #150
- I intentionally didn't instantiate a `SparkContext` in a before{} block, because every test may instantiate one with different parameters, or not use one at all.

### DIFF
--- a/core/src/test/scala/spark/FileSuite.scala
+++ b/core/src/test/scala/spark/FileSuite.scala
@@ -11,12 +11,12 @@ import org.apache.hadoop.io._
 
 import SparkContext._
 
-class FileSuite extends FunSuite with BeforeAndAfter{
+class FileSuite extends FunSuite with BeforeAndAfter {
   
   var sc: SparkContext = _
   
-  after{
-    if(sc != null){
+  after {
+    if(sc != null) {
       sc.stop()
     }
   }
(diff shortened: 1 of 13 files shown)


core/src/test/scala/spark/FileSuite.scala
```diff
@@ -6,13 +6,23 @@ import scala.io.Source
 
 import com.google.common.io.Files
 import org.scalatest.FunSuite
+import org.scalatest.BeforeAndAfter
 import org.apache.hadoop.io._
 
 import SparkContext._
 
-class FileSuite extends FunSuite {
+class FileSuite extends FunSuite with BeforeAndAfter{
```
mateiz: This looks great but please add spaces before the curly braces you use (I've noticed them missing in a few places, such as this one). Also, did you double-check that BeforeAndAfter runs the "after" part after every test instead of after the whole suite? There used to be two versions of it.

dennybritz: Got it, changed the braces. And yes, the BeforeAndAfter trait is basically same as BeforeAndAfterEach.

mateiz:
Thanks! Merged this.

PR #153, opened 2012-07-19
Java API

JoshRosen:
An initial version of a Java API for Spark.  In addition to the Java API, this commit adds a `distinct()` method to RDDs and fixes some issues related to `DoubleRDDFunctions`.


mateiz:
This looks really good, but a couple of comments:
- Can you rename the package to just spark.java? Seems a bit simpler.
- It would be better to replace the JavaLR with a JavaHdfsLR that reads from a file. The reason is that this will be more common for people wanting to use Spark, so we might as well include it as an example.
- In the WordCount example, I would inline the functions as anonymous inner classes just to show that syntax.
- In your Scala function definitions (def), always add a return type instead of relying on type inference. This will ensure that we notice when we break binary compatibility by changing a type. We've tried to do this in the rest of the code.

core/src/main/scala/spark/api/java/JavaRDD.scala
```diff
@@ -0,0 +1,34 @@
+package spark.api.java
+
+import spark._
+import spark.api.java.function.{Function => JFunction}
+
+class JavaRDD[T](val rdd: RDD[T])(implicit val classManifest: ClassManifest[T]) extends
+JavaRDDLike[T, JavaRDD[T]] {
+
+  def wrapRDD = JavaRDD.fromRDD
```
mateiz: Here is an example of where to add a return type. Actually all the methods in this file should have one.

JoshRosen:
> Can you rename the package to just `spark.java`? Seems a bit simpler.

Scala package references are relative, so import statements like

``` scala
package spark
import java.util.List
```

would be resolved against the `spark.java` package, causing compile errors.  A workaround is to use the top-level [`_root_`](http://stackoverflow.com/questions/687071/what-is-the-root-package-in-scala) package to import from the correct top-level `java` package:

``` scala
package spark
import _root_.java.util.List
```

It's easy to add the `_root_` prefixes to the existing code, but the change would affect a lot of files.  Do you still want to rename the package?

mateiz:
Oh, got it. Let's not rename it then.

Matei

JoshRosen:
I pushed some changes that address your comments; let me know what you think.

core/src/main/scala/spark/api/java/JavaRDDLike.scala
```diff
@@ -0,0 +1,149 @@
...
+import spark.storage.StorageLevel
+
+import java.util.{List => JList}
+
+import scala.collection.JavaConversions._
+import java.lang
+import scala.Tuple2
+
+trait JavaRDDLike[T, This <: JavaRDDLike[T, This]] extends Serializable {
+  def wrapRDD: (RDD[T] => This)
```
mateiz: Why does this need to be a method returning a function, instead of just `def wrapRDD(rdd: RDD[T]): This`? Haven't looked into it much so maybe I'm missing something. If I am, then add a comment for it too.

core/src/main/scala/spark/api/java/JavaRDDLike.scala
```diff
@@ -0,0 +1,149 @@
...
+
+  implicit def classManifest: ClassManifest[T]
```
mateiz: Why is this implicit? Just so that the manifest is available when you create RDDs later?

I didn't know you could have an implicit def that is used to pass implicit objects around. Maybe making it an implicit val would be clearer. I think the val doesn't have to have a value given to it in the trait.

core/src/main/scala/spark/api/java/JavaDoubleRDD.scala
```diff
@@ -0,0 +1,67 @@
...
+
+import java.lang.Double
+
+class JavaDoubleRDD(val srdd: RDD[scala.Double]) extends JavaRDDLike[Double, JavaDoubleRDD] {
+
+  val classManifest: ClassManifest[Double] = implicitly[ClassManifest[Double]]
+
+  lazy val rdd: RDD[Double] = srdd.map(x => Double.valueOf(x))
+
+  def wrapRDD: (RDD[Double]) => JavaDoubleRDD = rdd => new JavaDoubleRDD(rdd.map(_.doubleValue))
```
mateiz: When you're overriding from a trait, add the override keyword in front just to make this work well if we add/remove/rename methods in the trait. It's not obvious but we've tried to do this throughout the code.

examples/src/main/java/spark/examples/JavaTC.java
```diff
@@ -0,0 +1,76 @@
...
+    Integer slices = (args.length > 1) ? Integer.parseInt(args[1]): 2;
+    JavaPairRDD<Integer, Integer> tc = sc.parallelizePairs(generateGraph(), slices);
+
+    // Linear transitive closure: each round grows paths by one edge,
+    // by joining the graph's edges with the already-discovered paths.
+    // e.g. join the path (y, z) from the TC with the edge (x, y) from
+    // the graph to obtain the path (x, z).
+
+      // Because join() joins on keys, the edges are stored in reversed order.
+    JavaPairRDD<Integer, Integer> edges = tc.map(new PairFunction<Tuple2<Integer, Integer>,
```
mateiz, line 56: Looks mis-indented.. is there a tab by any chance?

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -48,10 +48,13 @@ import spark.storage.BlockManagerMaster
...
     frameworkName: String,
-    val sparkHome: String = null,
-    val jars: Seq[String] = Nil)
+    val sparkHome: String,
+    val jars: Seq[String])
   extends Logging {
   
+  def this(master: String, frameworkName: String) = this(master, frameworkName, null, Nil)
+
+
```
mateiz: Too many blank lines

examples/src/main/scala/spark/examples/SparkTC.scala
```diff
@@ -0,0 +1,53 @@
...
+    val edges = tc.map(x => (x._2, x._1))
+
+    // This join is iterated until a fixed point is reached.
+    var oldCount = 0L
+    do {
+      oldCount = tc.count()
+      // Perform the join, obtaining an RDD of (y, (z, x)) pairs,
+      // then project the result to obtain the new (x, z) paths.
+      tc = tc.union(tc.join(edges).map(x => (x._2._2, x._2._1))).distinct()
+    } while (tc.count() != oldCount)
```
mateiz: As a minor optimization it might be good to save the new count when you compute it. So

```
var oldCount = 0L
var nextCount = 0L
do {
    oldCount = nextCount
    ...
    nextCount = tc.count()
} while (nextCount != oldCount)
```

examples/src/main/java/spark/examples/JavaTC.java
```diff
@@ -0,0 +1,76 @@
...
+    });
+
+    long oldCount = 0;
+    do {
+      oldCount = tc.count();
+      // Perform the join, obtaining an RDD of (y, (z, x)) pairs,
+      // then project the result to obtain the new (x, z) paths.
+
+      tc = tc.union(tc.join(edges).map(ProjectFn.INSTANCE)).distinct();
+    } while (tc.count() != oldCount);
```
mateiz: Same comment as for the Scala version on doing count() twice

examples/src/main/scala/spark/examples/SparkTC.scala
mateiz: Actually, another important thing is probably to call cache() after each distinct(). Otherwise, the while tree of RDDs will be recomputed again from the start on each iteration, due to the lazy evaluation.

JoshRosen: Would it be useful to add automatic caching/memoization to RDD functions that return scalar values, such as `count()` or`stddev()`?  The memoization code could probably be implemented as a generic higher-order function that wraps the method definitions, similar to a memoizing decorator in Python.

mateiz: Sure, that would certainly help. Feel free to submit that as a separate pull request. See my comment about caching these RDDs in the transitive closure case too though.

Matei

core/src/main/scala/spark/api/java/JavaRDDLike.scala
JoshRosen: I made the ClassManifest fields `implicit` to simplify the code for wrapping RDDs, although it looks like the benefits are marginal here (JavaPairRDD's implicit key and value ClassManifests have more usages).  I was probably thinking of Java when I used `implicit def`, since Java doesn't allow interfaces or classes to specify abstract values/fields.  I'll change it to `implicit val`.

examples/src/main/scala/spark/examples/SparkTC.scala
JoshRosen: My latest commit changes this to

``` scala
var oldCount = 0L
var nextCount = tc.count()
do {
  oldCount = nextCount
  tc = tc.union(tc.join(edges).map(x => (x._2._2, x._2._1))).distinct().cache();
  nextCount = tc.count()
} while (nextCount != oldCount)
```

`nextCount` is initialized to `tc.count()` instead of `0` in order to halt after one round when the input is already a fixed point (when the input is a complete directed graph, for example).

I also added a `cache()` call when constructing the initial `tc` RDD.

JoshRosen:
Thanks for the feedback; I pushed another commit addressing these comments.

examples/src/main/java/spark/examples/JavaWordCount.java
```diff
@@ -0,0 +1,49 @@
...
+      System.exit(1);
+    }
+
+    JavaSparkContext ctx = new JavaSparkContext(args[0], "JavaWordCount");
+    JavaRDD<String> lines = ctx.textFile(args[1], 1);
+
+    JavaPairRDD<String, Integer> counts = lines.flatMap(new FlatMapFunction<String, String>() {
+      public Iterable<String> apply(String s) {
+        StringOps op = new StringOps(s);
+        return Arrays.asList(op.split(' '));
```
mateiz: Probably should use String.split() on this instead of bringing in the Scala StringOps class. Java's strings have a split() function already. It treats the splitter as a regex but it's fine for space.

JoshRosen:
I removed the dependency on `StringOps` and added in a few RDD methods that were missing from the Java API (`persist`, `splits`, `glom`, and `mapPartitions`).

mateiz:
Looks good. I'm going to merge it now, and we can make further changes as smaller commits. I'm also trying to get feedback on the API from a few other people. But this is an awesome start!

PR #155, opened 2012-07-23
Support for external hashing and sorting

harveyfeng:
Also includes a small fix in RangePartitioner - ascending/descending is accounted for in getPartition(), so sorting rddSample doesn't need the ascending check.


harveyfeng:
Sorry, meant this for the 'dev' branch...

PR #156, opened 2012-07-23
Support for external sorting and hashing

harveyfeng:
Also includes a small fix in RangePartitioner - ascending/descending is accounted for in getPartition(), so sorting rddSample doesn't need the ascending check.


core/src/test/scala/spark/ShuffleBucketSuite.scala
```diff
@@ -0,0 +1,66 @@
...
+
+
+import spark.ExternalSorter
+import spark.SizeEstimator
+import scala.collection.mutable.ArrayBuffer
+
+class InternalBucketSuite extends FunSuite {
+
+  test("Combine (key, value) using put and iterate through combiners") {
+
```
mateiz: Are you still working on these tests or did you mistakenly submit some empty ones?

core/src/main/scala/spark/ShuffleBucket.scala
```diff
@@ -0,0 +1,216 @@
...
+
+      def next() = {
+        val entry = iter.next()
+        (entry.getKey.asInstanceOf[K], entry.getValue.asInstanceOf[C])
+      }
+    }
+  }
+}
+
+trait ShuffleBucket[K, V, C] {
```
mateiz: It would help to add a short Javadoc above this and other classes. Also, can you put ExternalBucket and InternalBucket into their own files? Actually I'd suggest making a spark.shuffle package and putting them in there.

core/src/main/scala/spark/ShuffleBucket.scala
```diff
@@ -0,0 +1,216 @@
...
+    () => new Object2LongOpenHashMap[K].asInstanceOf[JMap[K, C]]
+
+  implicit def createObject2ShortMap[K <: AnyRef, C <: Short] = 
+    () => new Object2ShortOpenHashMap[K].asInstanceOf[JMap[K, C]]
+
+  implicit def createObject2ObjectMap[K <: AnyRef, C <: AnyRef] = 
+    () => new Object2ObjectOpenHashMap[K, C].asInstanceOf[JMap[K, C]]
+    
+  // Cast to JMap[Any, Any] to get put() to work.
+  def makeMap[K, C] (implicit m: () => JMap[K, C] = () => new JHashMap[K, C]) = {
```
mateiz: Are you sure that this works as a way of passing a default? If not, maybe you can just bite the bullet and ask for a keyClass and valueClass as before.

core/src/test/scala/spark/ShuffleBucketSuite.scala
harveyfeng: Still working on them - makeMake() called in ShuffleRDD always chooses the default JavaHashMap right now =\

core/src/main/scala/spark/ShuffleBucket.scala
harveyfeng: It works as the default, but since makeMap() is called in ShuffleRDD, the implicits here don't kick in since they are out of scope. Importing these in ShuffleRDD won't work since the types passed to ShuffleRDD are erased anyway. So, the default is always used. I'll change it to ask for key and value classes and update the ShuffleBucket tests.

rxin:
For a change of this size, it would be nice to have a high level writeup for it (1 - 5 pages) to explain the general structure / components. It'd also be very useful for reviewers and others to read when they want to modify this in the future.

core/src/main/scala/spark/ShuffleBucket.scala
mateiz: Yeah, I guess either you have to do the implicits all the way where the types are still known, or use the classes. Using the classes seems a lot simpler.

PR #157, opened 2012-07-30
Logging Throwables in Info and Debug

paulcavallaro:
Logging Throwables in logInfo and logDebug instead of swallowing them.


mateiz:
Thanks for catching this! Looks like a useful fix.

PR #159, opened 2012-08-02
Spark WebUI

dennybritz:
- Uses spray.twirl (https://github.com/spray/twirl)
- CSS and JS files are included twice, once for the worker and once for the master resources. We may want to have just one shared resources folder since most people won't separate the code.
- One problem that (I think, untested) remains is that the code uses the machine's local IP address (same as the spark server) in the URL links, even though the WebUI server is bound to 0.0.0.0/0. That means if you're running on a cluster with both a private and public ip address the private one will be used and you won't be able to follow the links from outside the cluster.
- Jobs are always marked as completed. We need to add conditions under which a job counts as FAILED.


mateiz:
Thanks! Just to respond to the comments you raised:
- It would be nice to add a shared "static" folder for the CSS and JS. I believe you can change the paths in Spray so that anything starting with /static is served out of that directory.
- The IP address thing is tricky. For the Mesos scripts, we let the users set a "public hostname" parameter for each machine that was used in the web UI. Maybe we can do the same here. It will definitely matter when we run on EC2.
- Figuring out what's FAILED is tricky as well due to disconnections. I think that right now, the client just disconnects. Ideally you would add an "unregister" message for a successful job, and add a shutdown hook on the client (Runtime.addShutdownHook) to send it. You'd also need a version of "unregister" that reports failure, because the Spark master can decide it's failed if a task fails too many times. And finally, for other disconnects, you would count them as failures.

dennybritz:
- I'll look into the static folder.
- I guess we don't need to worry about the public ip address too much. Let's instead modify the scripts to give the --ip argument to the master and workers. The user is then responsible for giving a public ip.

dennybritz:
Files are now in a static folder.

core/src/main/scala/spark/deploy/worker/Worker.scala
```diff
@@ -108,25 +110,34 @@ class Worker(ip: String, port: Int, webUiPort: Int, cores: Int, memory: Int, mas
...
-      executors(jobId + "/" + execId).kill()
-      executors -= fullId
-      finishedExecutors += fullId
+      executor.kill()
 
     case Terminated(_) | RemoteClientDisconnected(_, _) | RemoteClientShutdown(_, _) =>
       masterDisconnected()
+      
+    case RequestWorkerState => {
+      sender ! WorkerState(ip + ":" + port, workerId, executors.values.toList, finishedExecutors.values.toList, masterUrl, cores, memory, coresUsed, memoryUsed, masterWebUiUrl)
```
mateiz: This line is a bit long -- keep them limited to 100 characters

project/SparkBuild.scala
```diff
@@ -71,7 +72,7 @@ object SparkBuild extends Build {
       "cc.spray" % "spray-can" % "1.0-M2.1",
       "cc.spray" % "spray-server" % "1.0-M2.1"
     )
-  ) ++ assemblySettings ++ extraAssemblySettings ++ mergeSettings
+  ) ++ assemblySettings ++ extraAssemblySettings ++ mergeSettings ++ Seq(Twirl.settings: _*)
```
mateiz: I think you can do just `++ Twirl.settings` instead of `++ Seq(Twirl.settings: _*)`

core/src/main/scala/spark/deploy/DeployMessage.scala
```diff
@@ -44,4 +49,19 @@ case class JobKilled(message: String)
...
+// Master to MasterWebUI
+
+case class MasterState(uri : String, workers: List[WorkerInfo], activeJobs: List[JobInfo], completedJobs: List[JobInfo])
+
+//  WorkerWebUI to Worker
+case object RequestWorkerState
+
+// Worker to WorkerWebUI
+
+case class WorkerState(uri: String, workerId: String, executors: List[ExecutorRunner], finishedExecutors: List[ExecutorRunner], masterUrl: String, cores: Int, memory: Int, coresUsed: Int, memoryUsed: Int, masterWebUiUrl: String)
```
mateiz: Ditto on the overly long line

mateiz:
This looks good to merge except for a few formatting things (overly long lines of text).

core/src/main/scala/spark/deploy/master/MasterWebUI.scala
```diff
@@ -1,17 +1,52 @@
...
+  val STATIC_RESOURCE_DIR = "spark/deploy/static"
 
   val handler = {
     get {
       path("") {
-        getFromResource(RESOURCE_DIR + "/index.html")
+        completeWith {
+          val masterState = getMasterState()
+          // Render the HTML
+          masterui.html.index.render(masterState)
```
mateiz: Just a minor thing here: Instead of blocking the thread, you could actually return when the Future from RequestMasterState returns. I think you'd need to do this:

```
val future = master ? RequestMasterState
future.onComplete(state => completeWith { masterui.html.index.render(state) })
```

dennybritz:
Thanks, I made the changes you suggested.

I couldn't get the non-blocking calls to work with Spray (I had actually tried that before), because of its built-in implicit Marshallers. However, I found that it actually supports implicit conversions of Futures.

mateiz:
Thanks! Just merged it.

PR #160, opened 2012-08-02
Standalone cluster scripts

dennybritz:
- Heavily influenced by Hadoop scripts. Basically works the same way, but with different argument names :) A list of slaves is expected to be in conf/slaves.
- Added a  EC2 script, ec2-standalone, based on the mesos script, that starts a spark cluster in standalone mode. It works the same way as the mesos one (and takes the same arguments). The biggest differences are that there are no zoo keeper nodes, different security groups, and that it executes bin/start-all.sh to setup the cluster.
- Created an AMI for the standalone mode, ami-1ff25976 (https://s3.amazonaws.com/spark-standalone-amis/latest-spark). All it has is the latest spark and scala.
- Tested the scripts locally and on EC2.
- We probably want to write up a tutorial on how to use the scripts, though it should be rather straightforward for people familiar with the hadoop ones.
- One TODO is the configuration. How do we want users to specify things like master port, webUiPort, worker memory, worke cores, and so on? The simplest way would be to put these as env variables into the spark-env.sh, but that can get messy rather quickly.
- Removed explicit netty dependency. This caused trouble because 2 netty jars were included. Akka already has a netty dependency.


mateiz:
Hey, just a couple of comments:
- Regarding the configuration, I think environment variables are the best way to go for now, to be consistent with out other config stuff. Just call them things like SPARK_MASTER_PORT.
- Would it be possible to reuse the spark-ec2 script from before, but just with a different AMI? Since a lot of the setup is handled by the script on the AMI, I believe it should be possible to reuse the client-side code. If necessary, you can add options to it. That's better than having two nearly identical files.

dennybritz:
Yes, I could reuse the existing EC2 script. I'll just have to add another switch to it and make the security groups and ssh commands conditional.

dennybritz:
I merged the two scripts into one. Still working on the configuration, but it's working fine with default settings right now.

dennybritz:
Added configuration variables.

mateiz:
Thanks. I looked at it some more and I have just two other suggestions:
- If nontrivial parts of the code, such as the "rotate logs" bit, are copied from Hadoop, add Hadoop's Apache license notice to the the top of that file. If it's not copied but just based on reading the code, then there's probably no need.
- Why didn't you make the master be configured in a separate "masters" file as well? I thought about this a bit and I think it's good to do it the same way as Hadoop because people will be familiar with that. It's also useful in case someone tries to stop or start the cluster from a machine other than the master (because the scripts will be everywhere).

dennybritz:
- Added the Apache license to the two scripts that are a bit more complex and obviously taken from Hadoop.
- The masters file in Hadoop is actually not what most people expect it to be. In Hadoop the NameNode and DataNode are only started on the machine where the script is executed. The masters file only specifies the locations where a SecondaryNameNode should be started. So, the naming of the _masters_ file in Hadoop is a bit off. Since Spark doesn't have the notion of a SecondaryNameNode (which isn't a backup for the NameNode anyway) I thought a masters file was unnecessary.

mateiz:
Oh, I see. That makes sense then. I'm going to merge this, and I guess the last thing that will be left is to document it.

For the documentation, I was thinking of switching away from the wiki to a set of HTML docs generated from markup files in our source tree, using a system such as Jekyll (https://github.com/mojombo/jekyll). I'll send an email later about that, but it can wait a bit.

PR #158, opened 2012-07-31
Merge Akka reference.conf files in sbt assembly task

JoshRosen:
This fixes an issue when running Spark from the jar generated by the sbt `assembly` task.

To reproduce the issue:

``` java
import spark.SparkContext;

public class Test {
    public static void main(String[] args) {
        new SparkContext("local", "test");
        System.exit(0);
    }
}
```

Place this code in a file (but not in the root of the `SPARK_HOME` directory), then compile and run it against the assembly jar:

```
$ javac -cp ../core/target/spark-core-assembly-0.6.0-SNAPSHOT.jar Test.java 
$ java -cp ../core/target/spark-core-assembly-0.6.0-SNAPSHOT.jar:. Test
Exception in thread "main" com.typesafe.config.ConfigException$Missing: No configuration setting found for key 'akka.version'
    at com.typesafe.config.impl.SimpleConfig.findKey(SimpleConfig.java:115)
    at com.typesafe.config.impl.SimpleConfig.find(SimpleConfig.java:135)
    at com.typesafe.config.impl.SimpleConfig.find(SimpleConfig.java:140)
    at com.typesafe.config.impl.SimpleConfig.find(SimpleConfig.java:108)
    at com.typesafe.config.impl.SimpleConfig.find(SimpleConfig.java:146)
    at com.typesafe.config.impl.SimpleConfig.getString(SimpleConfig.java:188)
    at akka.actor.ActorSystem$Settings.<init>(ActorSystem.scala:116)
    at akka.actor.ActorSystemImpl.<init>(ActorSystem.scala:429)
    at akka.actor.ActorSystem$.apply(ActorSystem.scala:103)
    at spark.util.AkkaUtils$.createActorSystem(AkkaUtils.scala:36)
    at spark.SparkEnv$.createFromSystemProperties(SparkEnv.scala:58)
    at spark.SparkContext.<init>(SparkContext.scala:71)
    at spark.SparkContext.<init>(SparkContext.scala:55)
    at Test.main(Test.java:5)
```

This issue is caused by the way that Akka `reference.conf` files are handled in the `assembly` task; the commit copies the solution provided by the Akka Team Blog: http://letitcrash.com/post/21025950392/howto-sbt-assembly-vs-reference-conf.

PR #161, opened 2012-08-02
Use sbt mergeStrategy for reference.conf files.

JoshRosen:
Cleans up the code as discussed in #158.


mateiz:
Looks good, thanks!

PR #162, opened 2012-08-02
Use maxMemory to better estimate memory available for BlockManager cache

shivaram:


shivaram:
Github seems to add any extra commits to my branch to the same pull request. Let me know if I should close this one and separate them out into two different pull requests.

mateiz:
Yes, you can only have one pull request from each of your branches at a time. Further updates to the branch count as updates to that request. But anyway, it looks good.

PR #165, opened 2012-08-04
Fix test checkpoint to reuse spark context defined in the class

shivaram:
I sometimes see a crash in the after clause calling sc.stop and this seems to fix it

### DIFF
--- a/core/src/test/scala/spark/RDDSuite.scala
+++ b/core/src/test/scala/spark/RDDSuite.scala
@@ -61,9 +61,8 @@ class RDDSuite extends FunSuite with BeforeAndAfter {
   }
 
   test("checkpointing") {
-    val sc = new SparkContext("local", "test")
+    sc = new SparkContext("local", "test")
     val rdd = sc.makeRDD(Array(1, 2, 3, 4), 2).flatMap(x => 1 to x).checkpoint()
     assert(rdd.collect().toList === List(1, 1, 2, 1, 2, 3, 1, 2, 3, 4))
-    sc.stop()
   }
 }


mateiz:
Looks good, but out of curiosity, what kind of crash is it? Is it because sc got reused from the previous test? Maybe we should set sc=null after to prevent this in the future.

shivaram:
The stack trace I saw was:

```
[info] Exception encountered when attempting to run a suite with class name: spark.RDDSuite *** ABORTED ***
[info]   java.lang.NullPointerException:
[info]   at spark.SparkContext.stop(SparkContext.scala:303)
[info]   at spark.RDDSuite$$anonfun$5.apply(RDDSuite.scala:14)
[info]   at org.scalatest.BeforeAndAfter$class.runTest(BeforeAndAfter.scala:180)
[info]   at spark.RDDSuite.runTest(RDDSuite.scala:8)
[info]   at org.scalatest.FunSuite$$anonfun$runTests$1.apply(FunSuite.scala:1304)
[info]   at org.scalatest.FunSuite$$anonfun$runTests$1.apply(FunSuite.scala:1304)
[info]   at org.scalatest.SuperEngine$$anonfun$org$scalatest$SuperEngine$$runTestsInBranch$1.apply(Engine.scala:226)
[info]   at org.scalatest.SuperEngine$$anonfun$org$scalatest$SuperEngine$$runTestsInBranch$1.apply(Engine.scala:215)
[info]   at scala.collection.LinearSeqOptimized$class.foreach(LinearSeqOptimized.scala:59)
[info]   at scala.collection.immutable.List.foreach(List.scala:45)
```

The stack trace indicates that sc.stop was called but the dagScheduler was null (line 303 was dagScheduler.stop in my tree).  My guess is that somehow the spark context (sc) used by the after method was not null as sc had been created by the test "checkpointing".  So sc.stop was executed twice and that led to the NullPointerException.
I am not sure why there could have been any confusion between the function local sc and the class variable sc though.

mateiz:
I think the problem is just that it call the after() part twice, calling stop() twice on an old SparkContext that had been left over by a previous test. Each test initializes the class variable sc at the start.

shivaram:
Do you think its a good idea to use the trait OneInstancePerTest to prevent tests from influencing each other ?

mateiz:
It probably wouldn't hurt for most of our tests, but we should do it consistently, which will take a bit of work. It's also easy to forget to add on new tests, so from that point of view it may not be great.

shivaram:
You were right - I added some debug information and the SparkContext from the previous test was being destructed twice. Adding sc = null in the after clause does handle those cases as well - so it might be safe to have it.

PR #166, opened 2012-08-08
Avoid a copy in ShuffleMapTask 

shivaram:
Avoid a copy in ShuffleMapTask by creating an iterator that will be used by the block manager.

### DIFF
--- a/core/src/main/scala/spark/scheduler/ShuffleMapTask.scala
+++ b/core/src/main/scala/spark/scheduler/ShuffleMapTask.scala
@@ -124,14 +124,18 @@ class ShuffleMapTask(
     val blockManager = SparkEnv.get.blockManager
     for (i <- 0 until numOutputSplits) {
       val blockId = "shuffleid_" + dep.shuffleId + "_" + partition + "_" + i
-      val arr = new ArrayBuffer[Any]
       val iter = buckets(i).entrySet().iterator()
-      while (iter.hasNext()) {
-        val entry = iter.next()
-        arr += ((entry.getKey(), entry.getValue()))
+      // Create an iterator to avoid a copy. blockManager will traverse this
+      // to create an block
+      val entryIter = new Iterator[Any] {
+        override def hasNext() = iter.hasNext()
+        override def next() = {
+          val entry = iter.next()
+          (entry.getKey(), entry.getValue())
+        }
       }
       // TODO: This should probably be DISK_ONLY
-      blockManager.put(blockId, arr.iterator, StorageLevel.MEMORY_ONLY, false)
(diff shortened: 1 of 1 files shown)


mateiz:
Looks good, thanks!

core/src/main/scala/spark/scheduler/ShuffleMapTask.scala
```diff
@@ -124,14 +124,18 @@ class ShuffleMapTask(
...
     for (i <- 0 until numOutputSplits) {
       val blockId = "shuffleid_" + dep.shuffleId + "_" + partition + "_" + i
-      val arr = new ArrayBuffer[Any]
       val iter = buckets(i).entrySet().iterator()
-      while (iter.hasNext()) {
-        val entry = iter.next()
-        arr += ((entry.getKey(), entry.getValue()))
+      // Create an iterator to avoid a copy. blockManager will traverse this
+      // to create an block
+      val entryIter = new Iterator[Any] {
```
rxin, line 130: This is a common enough pattern to warrant its own class. Can you pull this out and make it a class, e.g. JavaMapIteratorWrapper?

JoshRosen, line 130: It looks like this pattern of obtaining a Scala `Iterator[(_, _)]` from a `HashMap` can be expressed using [`JavaConversions.mapAsScalaMap`](http://www.scala-lang.org/api/current/index.html#scala.collection.JavaConversions$); I'm not sure if there's a performance difference, though:

``` scala
import scala.collection.JavaConversions
import java.util.HashMap

object Test {
        val javaMap = new HashMap[Int,Int]()
        javaMap.put(1, 2)
        javaMap.put(3, 4)
        val scalaMap = JavaConversions.mapAsScalaMap(javaMap)
        val iter : Iterator[Any] = scalaMap.toIterable.iterator
        println(iter.next())
        println(iter.next())
    }
}
```

shivaram:
Thanks Josh for pointing this ! I don't think there is a performance concern as it just wraps the java map https://github.com/scala/scala/blob/2.9.x/src/library/scala/collection/JavaConversions.scala#L485 

Making a change to just use the JavaConversions call now...

shivaram:
New pull request at https://github.com/mesos/spark/pull/168

core/src/main/scala/spark/scheduler/ShuffleMapTask.scala
JoshRosen, line 130: Oops, `scalaMap.toIterable.iterator` should be replaced by `scalaMap.iterator`.

PR #167, opened 2012-08-08
Detect non-zero exit status from PipedRDD process

JoshRosen:
Changes PipedRDD to detect a non-zero exit status from its child process, rather than silently failing.

### DIFF
--- a/core/src/main/scala/spark/PipedRDD.scala
+++ b/core/src/main/scala/spark/PipedRDD.scala
@@ -57,7 +57,20 @@ class PipedRDD[T: ClassManifest](
     }.start()
 
     // Return an iterator that read lines from the process's stdout
-    Source.fromInputStream(proc.getInputStream).getLines
+    val lines = Source.fromInputStream(proc.getInputStream).getLines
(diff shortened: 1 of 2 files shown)


core/src/main/scala/spark/PipedRDD.scala
```diff
@@ -57,7 +57,20 @@ class PipedRDD[T: ClassManifest](
...
-    Source.fromInputStream(proc.getInputStream).getLines
+    val lines = Source.fromInputStream(proc.getInputStream).getLines
+    return new Iterator[String] {
+      def next() = lines.next()
+      def hasNext = {
+        if (lines.hasNext) {
+          true
+        } else {
+          if (proc.waitFor() != 0) {
+            throw new Exception("Subprocess exited with non-zero exit status")
```
mateiz: Looks good but do you mind saying what the exit status was as part of the exception? Thanks!

JoshRosen:
I added the exit status to the exception.

mateiz:
Great, thanks.

PR #168, opened 2012-08-08
Use JavaConversion to get a scala iterator

shivaram:


mateiz:
Isn't there an implicit conversion from Java iterators to Scala iterators or Seq? That would be cleaner.

mateiz:
Great, thanks.

PR #169, opened 2012-08-12
Changes to SizeEstimator more accurate

shivaram:
Motivation:
This patch is motivated by an observation that the amount of heap space used by
the BoundedMemoryCache is often much larger than what we account for.
For example running with 10 files from the RITA dataset[1] and 4GB for the cache,
we see the currBytes variable to be 3917.67. However analyzing the memory heap
with Eclipse MAT[2] shows that the BoundedMemoryCache in fact occupies 4360.55 MB.

Changes made:
This patch tries to address the discrepancy by making some changes to the
SizeEstimator. The object size and reference size are changed based on
the architecture in use and if or not CompressedOops are in use by the JVM. This
results in the object size changing from 8 to 12 or 16 and references being
either 4 or 8 bytes long.  We also account for the fact that arrays have an
object header + an int for the length. Lastly, this patch also account for the
fact that fields and objects are aligned to 8-byte boundaries by the JVM.
Changes are based on information from [3,4,5]

Tests:
Changes are verified by comparing the results from spark.SizeEstimator.estimate
to those from MAT. An example can be found in
https://github.com/shivaram/spark/tree/size-estimate-test/res where the first
100 lines from the 1990 dataset was used. The file spark-err-log.txt shows that
the size estimate was 25000 bytes which matches the size of the hashmap entry in
BoundedMemoryCache entry found in spark-bounded-memory.txt.

Also, a simple script that can be used to run such a test with any text file can
be found in the `size-estimate-test` tree as `run_size_estimate_test` With this
patch and the original dataset from the motivation, we get an estimate of
3878.81MB while MAT reports memory usage of 3879.61MB. The difference is 
explained below.

Caveats:
Arrays are still sampled during estimation and this could lead to small
variations as seen with the example above. Also, the patch uses HotSpot
diagnostic MXBean to figure out if compressed oops are being used by the JVM and
this may fail on non-hotspot JVMs. Finally, the patch has been tested only on
64-bit HotSpot JVMs (1.6.0_24 and 1.7.0_147-icedtea). I don't have access to a
32-bit machine to test it, but we can do it on EC2.

[1] http://stat-computing.org/dataexpo/2009/the-data.html
[2] http://eclipse.org/mat
[3] https://wikis.oracle.com/display/HotSpotInternals/CompressedOops
[4] http://kohlerm.blogspot.com/2008/12/how-much-memory-is-used-by-my-java.html
[5] http://lingpipe-blog.com/2010/06/22/the-unbearable-heaviness-jav-strings/

### DIFF
--- a/core/src/main/scala/spark/SizeEstimator.scala
+++ b/core/src/main/scala/spark/SizeEstimator.scala
@@ -63,6 +63,9 @@ object SizeEstimator extends Logging {
   classInfos.put(classOf[Object], new ClassInfo(OBJECT_SIZE, Nil))
 
   private def getIsCompressedOops : Boolean = {
+    if (System.getProperty("spark.test.useCompressedOops") != null) {
+      return System.getProperty("spark.test.useCompressedOops").toBoolean 
+    }
     try {
       val hotSpotMBeanName = "com.sun.management:type=HotSpotDiagnostic";
       val server = ManagementFactory.getPlatformMBeanServer();
--- a/core/src/main/scala/spark/SizeEstimator.scala
+++ b/core/src/main/scala/spark/SizeEstimator.scala
@@ -19,8 +19,6 @@ import it.unimi.dsi.fastutil.ints.IntOpenHashSet
  * http://www.javaworld.com/javaworld/javaqa/2003-12/02-qa-1226-sizeof.html
  */
 object SizeEstimator {
-  private val OBJECT_SIZE  = 8 // Minimum size of a java.lang.Object
-  private val POINTER_SIZE = 4 // Size of an object reference
 
   // Sizes of primitive types
   private val BYTE_SIZE    = 1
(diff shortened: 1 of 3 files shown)


mateiz:
Looks great! One thing on the tests though: can you add a couple of tests with other settings of 64-bit and compressed oops? It does't need to be a lot, but just to cover those code paths.

shivaram:
Added tests for the 32-bit and no-compressed oops cases. I had to move some of the variables to be initialized inside a method to get the tests to work correctly. 

Also, this change should be applicable to the dev branch as well. Do you usually port patches from master to dev or should I use another pull request for that ?

mateiz:
Thanks. If you have a pull request for dev that will save me some trouble, otherwise I'll cherry-pick these commits into it. (Unfortunately right now you can't merge master into dev.)

shivaram:
I will open another pull request as I am in the process of applying these changes to dev.

PR #170, opened 2012-08-13
Launching Spark over YARN

dennybritz:
**Usage Example:** 
SPARK_JAR=./core/target/spark-core-assembly-0.6.0-SNAPSHOT.jar ./run spark.deploy.yarn.Client **--jar** examples/target/scala-2.9.1/spark-examples_2.9.1-0.6.0-SNAPSHOT.jar **--class** spark.examples.SparkPi **--args** standalone **--num-workers** 3 **--worker-memory** 512M

**Notes:**
- Need to run sbt-assembly to generate the core-assembly jar used above.
- YARN does not support requesting containers/resources by the number of cores. Using `Runtime.getRuntime.availableProcessors()` for each slave for now. 
- The user must use a the String "standalone" as the master's URL. That starts the scheduler without trying to connect to a cluster.
- There are several TODOs. In particular 1. Checking for and handling container/app failure, 2. Letting the user set the context priority, and some code refactoring.
- Tested locally, but not on a real cluster.


core/src/main/scala/spark/deploy/yarn/ApplicationMaster.scala
```diff
@@ -0,0 +1,197 @@
...
+    // Start the user's JAR
+    val userThread = startUserClass()
+    
+    // This is pretty hacky, but we need to wait until the spark.master.port property has
+    // been set by the Thread executing the user class.
+    while (System.getProperty("spark.master.port") == null 
+      || System.getProperty("spark.master.port").toInt == 0) {
+      Thread.sleep(10)
+    }
+    logInfo("Port: " + System.getProperty("spark.master.port"))
```
mateiz: This might not always work well, because the user may have set spark.master.port from outside (e.g. as a JVM option), so you go ahead and launch workers, but those workers might connect _before_ the Spark master is started. Ideally you would keep looking for that property and try to open a connection to that port once it's set, and wait for the master to really be up. How hard would it be to do that?

core/src/main/scala/spark/deploy/yarn/WorkerRunnable.scala
```diff
@@ -0,0 +1,120 @@
...
+  }
+  
+  def prepareEnvironment : HashMap[String, String] = {
+    val env = new HashMap[String, String]()
+    env("CLASSPATH") = "$CLASSPATH:./*:"
+    System.getenv().filterKeys(_.startsWith("SPARK")).foreach { case (k,v) => env(k) = v }
+    return env
+  }
+  
+  def conntectToCM : ContainerManager = {
```
mateiz: This is misspelled

core/src/main/scala/spark/deploy/yarn/ApplicationMaster.scala
```diff
@@ -0,0 +1,197 @@
...
+        // Run each of the allocated containers
+        for (container <- allocatedContainers) {
+          val masterUrl = "akka://spark@%s:%s/user/%s".format(
+            System.getProperty("spark.master.host"), System.getProperty("spark.master.port"),
+            StandaloneSchedulerBackend.ACTOR_NAME)
+          val workerId = numWorkersRunning.intValue.toString
+          val workerHostname = container.getNodeId().getHost()
+          // YARN does not support requesting resources by the number of cores yet.
+          // TODO: How do we handle this?
+          // val workerCores = 1
```
mateiz: Good question. I would go for asking the number of cores per worker as a command-line argument instead of setting it to Runtime.getNumberOfProcessors as you do now, just because the latter might be extremely high. Shouldn't be too hard to change.

core/src/main/scala/spark/deploy/yarn/ApplicationMaster.scala
dennybritz: Good point, I'll fix that.

core/src/main/scala/spark/deploy/yarn/ApplicationMaster.scala
dennybritz: Okay, I'll add that. However, if the user resource request cannot be satisfied (there are not enough free cores) we won't be able to tell the user in advance.

core/src/main/scala/spark/deploy/yarn/WorkerRunnable.scala
dennybritz: Good catch.

core/src/main/scala/spark/deploy/yarn/ApplicationMaster.scala
mateiz: Yeah, this is fine, but I just want to give users a way to limit the number of cores. For example, with the old code, many users might fit into the memory of one machine, and they would all try to use the max number of cores.

mateiz:
Hey Denny,

Let me know when you've made the changes above and I'll merge this. It would be nice to have this in the main repo.

dennybritz:
I made the changes. One thing that I couldn't quite figure out is why the application is marked as "FAILED" by YARN even though the ApplicationMaster exits with exit code 0. I am even setting the status to FinalApplicationStatus.SUCCEEDED explicitly. The error message isn't very enlightening: `Application application_1346087295431_0014 failed 1 times due to AM Container for appattempt_1346087295431_0014_000001 exited with  exitCode: 0 due to: .Failing this attempt.. Failing the application.` Any ideas?

mateiz:
Thanks. Let's look into that later.

PR #171, opened 2012-08-13
Size estimator changes for dev

shivaram:
Based on pull request #169 - Additionally change BlockManagerSuite to make sure tests pass.


mateiz:
Thanks Shivaram!

rxin:
BTW I think we should open source just this size estimator - others will find it useful in other projects.

mateiz:
Yeah, it's a good idea.

Matei

PR #172, opened 2012-08-14
Rsync root directory in EC2 script

dennybritz:
- Will modify the Standalone AMI to allow for root login as well.


dennybritz:
Modified Standalone AMI to use root instead of ec2-user.

mateiz:
Great, thanks.

PR #173, opened 2012-08-14
make accumulator.localValue public, add tests

squito:


rxin:
This seems to break the abstraction as the user can abuse it and make changes that are not associative. 

For this usage, I think you can just implement an AccumulableParam for sets

class SetAccumulableParam[T] extends AccumulableParam[Set, T] {
  def addAccumulator(t1: R, t2: T) : R = { t1 += t2 } 
  def addInPlace(t1: R, t2: R): R = { t1 ++= t2 }
  def zero(initialValue: R): R = new Set
}

mateiz:
I think Imran had other usages, right? The main one is to read the value?

I agree that the test shouldn't be using += on localValue though. That just encourages potentially error-prone use of accumulators.

squito:
Yeah, the use of Set is not really a compelling reason for this feature, it was just the easiest to write a test case to verify the behavior.

As Matei said, one reason is to be able to read the current value of the accumulator, eg. something like SGD.

I agree that the user certainly _could_ abuse localValue, but I feel like the name & comment are a fair warning.

I don't really understand the argument against calling += directly on localValue, though.  In fact, pretty much the only thing that calling += on most (all?) accumulators does is delegate to += on localValue.

Another reason to allow access to localValue is that you may get access to a richer api through localValue.  Eg., adding to a Set can tell you whether the element already existed or not, and adding to a Map can give you the previously stored value for the key.  That functionality isn't available via the accumulator interface.

The final reason is that if you want to call some method on localValue which takes multiple arguments, its just kind of pain (and requires needless object creation) to have to (a) define a helper "update" type and (b) create instances of the update type.  Eg., imagine you had:

class PartsOfSpeechCounts {
  def addCounts(nouns:Int, verbs: Int) {...}
...
}

to use this in an accumulator, you now have to create another object
class PartsOfSpeechUpdate(val nouns: Int, val verbs: Int)

and then change all your updates from

acc.localValue.addCounts(nouns,verbs)
to
acc += new PartsOfSpeechUpdate(nouns, verbs)

It isn't the end of the world, but seems like a headache with no benefit.

core/src/main/scala/spark/Accumulators.scala
```diff
@@ -35,7 +35,16 @@ class Accumulable[T,R] (
     else throw new UnsupportedOperationException("Can't use read value in task")
   }
 
-  private[spark] def localValue = value_
+  /**
+   * get the current value of this accumulator from within a task.
```
mateiz: Capitalize this

core/src/test/scala/spark/AccumulatorSuite.scala
```diff
@@ -79,4 +79,18 @@ class AccumulatorSuite extends FunSuite with ShouldMatchers {
...
 
+  test ("localValue readable in tasks") {
+    import SetAccum._
+    val maxI = 1000
+    for (nThreads <- List(1, 10)) { //test single & multi-threaded
+    val sc = new SparkContext("local[" + nThreads + "]", "test")
+      val acc: Accumulable[mutable.Set[Any], Any] = sc.accumulable(new mutable.HashSet[Any]())
+      val d = sc.parallelize(1 to maxI)
+      d.foreach {
+        x => acc.localValue += x
```
mateiz: It would be nice to change this example to do something more interesting than +=; for example, maybe you can use ++=, a method that adds multiple items to a Set and would not be available on a normal accumulator.

mateiz:
Looks good, thanks!

PR #175, opened 2012-08-17
add accumulators for mutable collections, with correct typing!

squito:
add accumulator params for all "mutable collection" types (anything that is growable and traversable).

Rather than create a bunch of implicit objects, there is a helper method which just creates the AccumulableParam needed.  This way, you also get correct typing of the accumulator.  This does require a different method, though -- I called it accumulableCollection, not really tied to that name.

(this is completely independent from the other pull request on accumulable.localValue)

### DIFF
--- a/core/src/main/scala/spark/SparkContext.scala
+++ b/core/src/main/scala/spark/SparkContext.scala
@@ -4,7 +4,6 @@ import java.io._
 import java.util.concurrent.atomic.AtomicInteger
 
 import scala.actors.remote.RemoteActor
-import scala.collection.mutable.ArrayBuffer
 
 import org.apache.hadoop.fs.Path
 import org.apache.hadoop.conf.Configuration
@@ -31,6 +30,7 @@ import org.apache.hadoop.mapreduce.{Job => NewHadoopJob}
 import org.apache.mesos.MesosNativeLibrary
 
 import spark.broadcast._
+import collection.generic.Growable
 
 class SparkContext(
     master: String,
@@ -253,6 +253,10 @@ class SparkContext(
   def accumulable[T,R](initialValue: T)(implicit param: AccumulableParam[T,R]) =
     new Accumulable(initialValue, param)
 
+  def accumlableCollection[R <% Growable[T] with TraversableOnce[T] with Serializable, T](initialValue: R) = {
+    val param = new GrowableAccumulableParam[R,T]
+    new Accumulable(initialValue, param)
+  }
 
   // Keep around a weak hash map of values to Cached versions?
   def broadcast[T](value: T) = Broadcast.getBroadcastFactory.newBroadcast[T] (value, isLocal)
--- a/core/src/main/scala/spark/SparkContext.scala
+++ b/core/src/main/scala/spark/SparkContext.scala
@@ -253,7 +253,13 @@ class SparkContext(
   def accumulable[T,R](initialValue: T)(implicit param: AccumulableParam[T,R]) =
     new Accumulable(initialValue, param)
 
-  def accumlableCollection[R <% Growable[T] with TraversableOnce[T] with Serializable, T](initialValue: R) = {
+  /**
+   * create an accumulator from a "mutable collection" type.
+   * 
+   * Growable and TraversableOnce are the standard apis that guarantee += and ++=, implemented by
+   * standard mutable collections.  So you can use this with mutable Map, Set, etc.
+   */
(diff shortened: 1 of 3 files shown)


core/src/main/scala/spark/SparkContext.scala
```diff
@@ -253,6 +253,10 @@ class SparkContext(
   def accumulable[T,R](initialValue: T)(implicit param: AccumulableParam[T,R]) =
     new Accumulable(initialValue, param)
 
+  def accumlableCollection[R <% Growable[T] with TraversableOnce[T] with Serializable, T](initialValue: R) = {
```
mateiz: This is misspelled -- accumlable instead of accumulable. However, I'd also like to find a way to not have a special name for this. We might be able to do something like this:

```
def accumulable[R <% ..., T](initialValue: R)(implicit paramMaker: () => AccumulatorParam[R, T]) {
   new Accumulator(R, paramMaker())
}

implicit def growableParamMaker[R <% ...](): AccumulatorParam[R] = new GrowableAccumulatorParam[R]
```

mateiz: Also, I'd suggest dropping `with Serializable` to simplify the signature a bit.

squito: sorry about the typo -- I like your suggestion about the implicit
paramMaker, I will make that change.

How do you feel about taking this one step further, and adding another
implicit paramMaker for anything with a += and ++= ?  I think we would need
to add a new trait, but then you also automatically get the accumulator
param for anything with the trait declared.  I suppose implementing
Growable and Traversable is only a little bit more work, so maybe its best
to just push the standard scala traits, but I wanted to see what you
thought.

squito: On Sat, Aug 18, 2012 at 9:54 PM, Matei Zaharia notifications@github.comwrote:

> Also, I'd suggest dropping with Serializable to simplify the signature a
> bit.

Really?  I'd much rather get an compile error than a runtime error.  (In
fact I wanted to change the types for all sc.accumulable method, but
realized that you can't b/c you also want to accumulate primitives.)

I know just declaring the type as Serializable doesn't really save you from
all of the NotSerializableExceptions, since the contained types might not
be Serializable.  But IMO, it still helps to have at least some
compile-time check.

squito: unfortunately I haven't been able to figure out a way to get the method name to be accumulable.  I end with errors like this when I try to use it:

[error] /Users/imran/spark/spark/core/src/test/scala/spark/AccumulatorSuite.scala:33: ambiguous reference to overloaded definition,
[error] both method accumulable in class SparkContext of type [T, R](initialValue: T)(implicit paramMaker: () => spark.AccumulableParam[T,R])spark.Accumulable[T,R]
[error] and  method accumulable in class SparkContext of type [T, R](initialValue: T)(implicit param: spark.AccumulableParam[T,R])spark.Accumulable[T,R]
[error] match argument types (scala.collection.mutable.HashSet[Int]) and expected result type spark.Accumulable[scala.collection.mutable.Set[Int],Int]
[error]       val acc: Accumulable[mutable.Set[Int], Int] = sc.accumulable(new mutable.HashSet[Int]())
[error]                                                        ^

I could just make it a two arg method where the paramMaker is not implicit.  Or else maybe somebody can figure out the right scala magic to make this work, I am a little stuck though ...

mateiz: Alright. Give me a chance to look at this myself in the next few days, and if I don't find a way to do it, I'll merge it as is.

squito:
Hi Matei,

any update on this?  I have some real use cases for this, so it would be nice to get it merged in.

thanks!

mateiz:
Sorry, I haven't had a ton of time to look at it, but I also haven't found a better way to do it, so we should probably just go with your approach (accumulableCollection). Can you update the pull request to be based on the latest dev branch though? Just git pull in your own repo. Right now it's not automatically mergeable.

mateiz:
Oh actually I see your pull request is against master. Update it for the latest master code then, and I'll copy the commits into the dev branch too.

mateiz:
Thanks for the update -- I've committed it. Just to let you know though, there was a bug that actually caused the test to fail sometimes: GrowableAccumulatorParam.zero() returned the initial collection, so if we had added some values into it before sending out the last few tasks, we ended up with multiple copies of those values. I've fixed it in commit 2498f95199bc642119ab52981d1c6508f71ff2ff.

PR #179, opened 2012-08-26
Cache points in SparkLR example

JoshRosen:
For consistency with the SparkHDFSLR example, the SparkLR example should cache the data points used to train its model.


mateiz:
Looks good, thanks.

PR #181, opened 2012-08-28
Removed the deserialization cache for ShuffleMapTask

rxin:
Removed the deserialization cache for ShuffleMapTask becuase it was causing concurrency problems (some variables in Shark get set to null). The cost of task deserialization on slaves is trivial compared with the
execution time of the task anyway.


mateiz:
Thanks, looks good.

PR #182, opened 2012-08-29
Add a limit on the number of parallel fetches in the reduce stage

harveyfeng:
Made the numer of parallel fetches per reduce task 
System.getProperty("spark.default.parallelism", "8")
by default, but it can be set by setting the 'spark.blockManager.parallelFetches' property.

### DIFF
--- a/core/src/main/scala/spark/storage/BlockManager.scala
+++ b/core/src/main/scala/spark/storage/BlockManager.scala
@@ -2,16 +2,14 @@ package spark.storage
 
 import java.io._
 import java.nio._
-import java.nio.channels.FileChannel.MapMode
-import java.util.{HashMap => JHashMap}
-import java.util.LinkedHashMap
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.LinkedBlockingQueue
 import java.util.Collections
 
 import akka.dispatch.{Await, Future}
 import scala.collection.mutable.ArrayBuffer
-import scala.collection.mutable.HashMap
+import scala.collection.mutable.{HashMap, HashSet}
+import scala.collection.mutable.Queue
 import scala.collection.JavaConversions._
 
 import it.unimi.dsi.fastutil.io._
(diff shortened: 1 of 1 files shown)


core/src/main/scala/spark/storage/BlockManager.scala
```diff
@@ -595,6 +613,12 @@ class BlockManager(val master: BlockManagerMaster, val serializer: Serializer, m
 }
 
 object BlockManager {
+
+  def getNumParallelFetchesFromSystemProperties(): Int = {
+    System.getProperty("spark.blockManager.parallelFetches",
+      System.getProperty("spark.default.parallelism", "8")).toInt
```
mateiz: Don't make the default spark.default.parallelism, just set it to 8. The reason is that _each_ reduce task fetches this many blocks, so that could be a lot of total fetches.

mateiz:
Hey Harvey,

Just to understand, is this going to request the first 8 blocks, then wait until they've _all_ been fetched, then request the next 8, etc? It would be nicer to start fetching the next block as soon as the previous one was obtained, so that you don't have times of idleness when waiting for the last one. The way you'd do this is by somehow creating one big queue of the block requests, and sending the next request whenever a Future completes and you've got the current one.

harveyfeng:
Hi Matei,

Yeah, the parallel fetching is doing 8 requests at a time right now - I'll
change it.
Also, to confirm what you mentioned earlier, the serializing bug is
because the callbacks are being executed on the same thread, so two
deserializeStreams are using the same ThreadLocal objectBuffer...I'll
include a fix to this too.

Thanks!

mateiz:
Hey Harvey,

Just FYI, I pushed a fix to the deserializer issue, by doing all the deserialization in the caller's thread for getMultiple. It shouldn't affect your work in a major way but make sure you merge that in.

mateiz:
Thanks Harvey! This looks good.

PR #184, opened 2012-08-30
Disable running combiners on map tasks when mergeCombiners function is not specified by the user. 

rxin:
I haven't done much testing yet beyond running sbt/sbt test. Want to get some early feedback.

Basically if the mergeCombiners field is not set (null), combiners are not applied on map side. It doesn't break any existing API.


rxin:
Ok I added a test. Ready for review/commit.

mateiz:
Looks good, thanks.

PR #186, opened 2012-08-30
Spark HTTP FileServer

dennybritz:
A general fileserver for serving both regular as well as JAR files.
- I modified the existing JAR file server code to put everything into one fileserver.
- Modified the LocalScheduler to mimic the ClusterScheduler/Executor in terms of getting files. That's useful for testing.
- Test with standalone mode locally and local mode, but not on a mesos cluster.


core/src/test/scala/spark/FileServerSuite.scala
```diff
@@ -0,0 +1,43 @@
...
+import org.scalatest.BeforeAndAfter
+import java.io.{File, PrintWriter}
+
+class FileServerSuite extends FunSuite with BeforeAndAfter {
+  
+  var sc: SparkContext = _
+  
+  before {
+    // Create a sample text file
+    val pw = new PrintWriter(System.getProperty("java.io.tmpdir") + "FileServerSuite.txt")
```
mateiz: You should use Files.createTempDir() here to get a temp directory, as in FileSuite.scala. It's in the com.google.common.io.Files class and it ensures that you get a unique directory, so other people don't accidentally overwrite your file.

core/src/test/scala/spark/FileServerSuite.scala
```diff
@@ -0,0 +1,43 @@
...
+    // Clean up downloaded file
+    val tmpFile = new File("FileServerSuite.txt")
+    if (tmpFile.exists) {
+      tmpFile.delete()
+    }
+  }
+  
+  test("Distributing files") {
+    sc = new SparkContext("local[4]", "test")
+    sc.addFile(System.getProperty("java.io.tmpdir") + "FileServerSuite.txt")
```
mateiz: Ditto on the temp dir

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -74,7 +75,10 @@ class SparkContext(
     true,
     isLocal)
   SparkEnv.set(env)
-
+  
+  // Used to store a URL for each static file together with the file's local timestamp
+  val files = HashMap[String, Long]()
```
mateiz: Call this something like filesAdded to make its meaning clearer

core/src/main/scala/spark/scheduler/Task.scala
```diff
@@ -8,4 +13,21 @@ abstract class Task[T](val stageId: Int) extends Serializable {
   def preferredLocations: Seq[String] = Nil
 
   var generation: Long = -1   // Map output tracker generation. Will be set by TaskScheduler.
+  
+  // Stores file dependencies for this task. 
+  var fileSet : HashMap[String, Long] = new HashMap[String, Long]()
```
mateiz, line 18: I think you'll need to modify ShuffleMapTask's custom serialization code to make sure this is sent along. Have you tested it in a job with a shuffle yet? The problem is that ShuffleMapTask serializes the fields of the parent Task manually and then caches the serialization of some of them.

core/src/test/scala/spark/FileServerSuite.scala
```diff
@@ -0,0 +1,43 @@
...
+    val testRdd = sc.parallelize(List(1,2,3,4))
+    val result = testRdd.map { x => 
+      val in  = new java.io.BufferedReader(new java.io.FileReader("FileServerSuite.txt"))
+      val fileVal = in.readLine().toInt
+      in.close()
+      fileVal
+    }.reduce(_ + _)
+    assert(result == 400)
+  }
+
```
mateiz, line 63: Add a test with a groupByKey or reduceByKey to check that ShuffleMapTask sends the files correctly.

core/src/main/scala/spark/scheduler/Task.scala
```diff
@@ -8,4 +13,21 @@ abstract class Task[T](val stageId: Int) extends Serializable {
...
+  
+  // Downloads all file dependencies from the Master file server
+  def downloadFileDependencies(currentFileSet : HashMap[String, Long]) {
+    // Find files that either don't exist or have an earlier timestamp
+    val missingFiles = fileSet.filter { case(k,v) => 
+      !currentFileSet.isDefinedAt(k) || currentFileSet(k) <= v
```
mateiz: Instead of isDefinedAt, you can use contains.

mateiz:
I pointed out a few small formatting issues but I think there is also a problem with ShuffleMapTask's special serialization -- check that out and add a test for it. Otherwise it looks pretty good.

dennybritz:
I made the changes you suggested and added the functionality of dynamically adding JAR files. Like you said, I had to modify the ShuffleMapTask serialization to make it work.

mateiz:
This looks okay except for one big thing: creating a new ClassLoader for each task in the Executor is going to be very expensive, and is going to result in many copies of each task being loaded. Can you instead reuse the same ClassLoader, but give it more paths to search? One easy way would be to have a custom subclass of ClassLoader that holds an array of URLClassLoaders, and queries them to find the first one that contains a file.

As a more minor suggestion, in ShuffleMapTask, can you cache the serialized versions of the JAR and file HashMaps the same way we cache the RDDs? That serialization code becomes a bottleneck when you have a lot of tasks, and these maps will not change for subsequent runs of the task (or at least we shouldn't let a user use a JAR _before_ they add it to the system). You may need to change the Executor to ignore older JARs as well in case it gets an old task. Also, in the serialization part, you may be better off doing a toArray on these and serializing that instead of serializing a HashMap; then, just do a toMap on the Executor.

dennybritz:
Yep, that make sense. I'll make the changes.

dennybritz:
I am closing this pull request here because I want to submit the request from another, clean, branch.

PR #189, opened 2012-09-05
Simulating a Spark standalone cluster locally

dennybritz:
Title is pretty self-explanatory. The URL scheme is spark-cluster[N, coresPerSlave, memoryPerSlave]. There currently seems to be a problem with the Shutdown of the Executors, apparently they keep the JVM running when the user exits the Spark shell. I will fix that (let me know if you know a good way), but I submitted already because I wanted to get a quick code review.


rxin:
I think it is better to separate the master / slaves into multiple JVM processes so the local cluster setup better approximates a real cluster mode, which makes testing much easier.

Updated: never mind. I thought this was for the executor. As long as the executor is launched as a separate JVM from the master node, I am happy.

mateiz:
I actually disagree with Reynold on that -- I think running the spark.deploy master and slaves in the same process is fine. They will still communicate over Akka and see the same issues they would in separate JVMs. I'd prefer to avoid spawning lots of JVMs so that we can use this effectively in unit tests.

For the executor JVMs, is the problem just that they stay running when you Ctrl-C? You should add a shutdown hook (Runtime.addShutdownHook) inside the spark.deploy.Worker that stops its child processes when the JVM is stopped, or even add a way to explicitly stop a Worker from outside. I thought I'd actually already added such a shutdown hook.

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -81,9 +83,11 @@ class SparkContext(
     val LOCAL_N_REGEX = """local\[([0-9]+)\]""".r
     // Regular expression for local[N, maxRetries], used in tests with failing tasks
     val LOCAL_N_FAILURES_REGEX = """local\[([0-9]+),([0-9]+)\]""".r
+    // Regular expression for simulating a Spark cluster of [N, cores, memory] locally
+    val SPARK_LOCALCLUSTER_REGEX = """spark-cluster\[([0-9]+)\,([0-9]+),([0-9]+)]""".r
```
mateiz: We should come up with a better name for this; maybe local-cluster.

rxin:
Oops - updated my comment. I didn't realize the executors would've been launched as separate JVM processes.

core/src/main/scala/spark/SparkContext.scala
JoshRosen: Maybe support optional whitespace, so that `spark-cluster[1, 1, 1014]` works (I guess this would also apply to `LOCAL_N_FAILURES_REGEX`)

dennybritz:
I looked into the executor problem. They exit fine if I quit the repl via "ctrl+c", but not when I type "exit" or ":quit". Any ideas? Maybe that's not even a problem with the new code.

JoshRosen:
I think that Ctrl-C is sending SIGINT to the REPL and all of its children (see http://www.vidarholen.net/contents/blog/?p=34).  If I send SIGINT directly to the REPL process using `kill -SIGINT`, then the executors keep running after the REPL exits.

mateiz:
So shutdown hooks seem to work fine in the Scala and Spark shells:

   scala> val r = Runtime.getRuntime
   r: java.lang.Runtime = java.lang.Runtime@67439515

   scala> r.addShutdownHook(new Thread() { override def run() {println("!!!")} })

   scala> [Ctrl-D] !!!

Try printing some stuff in our shutdown hook and making sure it's called. You may also need to do a process.waitFor after killing it with kill().

dennybritz:
Working now, see the commit above.

mateiz:
Looks good, but instead of having the shutdown hook kill the master actor system, can you just add a shutdown hook in ExecutorRunner that does process.destroy()? I don't see that here.

Also, it would be nice if calling SparkContext.stop() when there is a local Spark cluster would shutdown that cluster. I think this would require some kind of wrapper or flag in SparkSchedulerBackend that lets you run some code when stop() is called, which in this case would stop the ActorSystems in the LocalSparkCluster, or send them a message. Basically the reason is that I'd like to use the local-cluster mode in unit tests, which will each start and stop a SparkContext.

mateiz:
Looks good; thanks!

PR #190, opened 2012-09-05
Log cache add/remove messages in block manager.

rxin:
Removed cache add/remove log messages from CacheTracker.
Added log messages on BlockManagerMaster to reflect block add/remove.
Also did some minor cleanup of storage package code.


mateiz:
Looks good, thanks.

PR #195, opened 2012-09-10
Spark HTTP FileServer

dennybritz:
A general fileserver for serving both regular as well as JAR files.
- I modified the existing JAR file server code to put everything into one fileserver.
- Modified the LocalScheduler to mimic the ClusterScheduler/Executor in terms of getting files. That's useful for testing.
- Includes test with local-cluster and local mode.
- Modified serialization of ShuffleMapTask
- I wasn't quite sure about the caching in ShuffleMapTask, is using the hashCode safe? There may be a better way to do this. Also, it doesn't seem like the serializedInfoCache (and fileCache) is cleared except when the SparkContext shuts down? What if it's a really long running task?


dennybritz:
Thinking about it, computing the hash code probably takes a long time as well, should I index the cache with the stageID?

mateiz:
Yeah, do index it by shuffleID.

mateiz:
Looks good; thanks!

PR #196, opened 2012-09-11
Log entire exception (including stack trace) in BlockManagerWorker.

rxin:
Previously only the error message is logged. This is not super helpful when you see an error...


core/src/main/scala/spark/storage/BlockManagerWorker.scala
```diff
@@ -34,7 +34,7 @@ class BlockManagerWorker(val blockManager: BlockManager) extends Logging {
           /*logDebug("Processed block messages")*/
           return Some(new BlockMessageArray(responseMessages).toBufferMessage)
         } catch {
-          case e: Exception => logError("Exception handling buffer message: " + e.getMessage)
+          case e: Exception => logError("Exception handling buffer message: " + e)
```
mateiz: Change it to just logError("Exception handling buffer message", e). Then it will print the stack trace too.

rxin:
PTAL

mateiz:
Great, thanks.

PR #197, opened 2012-09-12
Adds a "docs" directory containing existing Spark documentation and doc build instructions

andyk:
This change adds a directory called docs to the root Spark project directory. I used Jekyll as the framework for directory structure and compiling the docs, and used initializr.com for out-of-the-box generic look-and-feel (initializr generates pretty html wrapper leveraging Bootstrap/jquery, modernizr, ...). The Scaladoc API link in the global nav menu doesn't yet work.

To build the docs (install and) run `jekyll` and look in docs/_site or run `jekyll --server` and look at localhost:4000


mateiz:
Thanks Andy!

mateiz:
By the way, is it possible to change the way this generates links so that they use relative paths? Right now if you open docs/_site/index.html in a browser, none of the links works because they have absolute paths.

PR #198, opened 2012-09-13
Fix links and make things a bit prettier.

andyk:
Fixed links so that viewing html without running `jekyll --server` works better.
Added code syntax highlighting via pygments.
Updated the stylesheet a bit to make things a little easier to read.
Removed a duplicate doc file (EC2-Scripts.html) which existed because of capitalization weirdness (was a duplicate of ec2-scripts.html).


mateiz:
Thanks!

PR #199, opened 2012-09-13
YARN and standalone documentation

dennybritz:
YARN and standalone documentation


dennybritz:
Also, the navbar style was being messed up by the line-height property. I restricted the line-height to the body text.

mateiz:
Great, thanks!

PR #200, opened 2012-09-14
Updates to docs

andyk:
You can see what the generated site currently looks like here: http://www.cs.berkeley.edu/~andyk/spark-docs
- Added jekyll plugin (in _plugin directory, written in Ruby) that makes sure scaladoc is built and copies over Scaladoc from Spark subprojects into docs directory before generating site. Also updated links to the scaladoc in api.md (which generates api.html) to point to these scaladoc directories.
- Make logo look nicer (created a vector copy of it using illustrator magic and then used that to get a small version that looks a lot nicer)
- Changed navbar to not collapse when page gets narrower (which was not super helpful for a source code documentation site) but left responsive css in so that the page looks good when the window gets smaller (turning responsive off in bootstrap unfortunately makes things somewhat ugly with narrow windows)
- Adds back a file (ec2-scripts.md) that was mistakenly removed in previous commit.


mateiz:
Thanks!

PR #204, opened 2012-09-17
When a file is downloaded, make it executable.

dennybritz:
That's neccsary for scripts (e.g. in Shark).


mateiz:
Sounds like a good idea.

PR #208, opened 2012-09-19
Separated ShuffledRDD into multiple classes.

rxin:
RepartitionShuffledRDD, ShuffledSortedRDD, and ShuffledAggregatedRDD.

Mostly for performance reasons. In the future, we can also add Sort based aggregation to the code.

### DIFF
--- a/core/src/main/scala/spark/Aggregator.scala
+++ b/core/src/main/scala/spark/Aggregator.scala
@@ -9,9 +9,9 @@ package spark
   *                       known as map-side aggregations. When set to false, 
   *                       mergeCombiners function is not used.
   */
-class Aggregator[K, V, C] (
+case class Aggregator[K, V, C] (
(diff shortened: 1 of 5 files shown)


mateiz:
Hey Reynold,

This looks fine, but just to double-check, do we need new tests to verify all the new RDD subclasses? Or are they covered by the existing operations? 

Matei

rxin:
They are mostly covered by existing tests if you assume the sort test pretty much covers RepartitionShuffledRDD.

Basically there are 3 RDDs now:

RepartitionShuffledRDD: Used by ShuffledSortedRDD

ShuffledSortedRDD: tested by sortByKey

ShuffledAggregatedRDD: tested by anything combineByKey as before.

mateiz:
Alright, I've merged it. Let's see if we run into any issues.

PR #210, opened 2012-09-19
Set a limited number of retry in standalone deploy mode.

rxin:
It took Denny and me 2 hrs today to figure out I had the wrong scala home set and the test (which uses the standalone cluster) goes into infinite retry and everything just hangs.

This commit limits the number of retries allowed.

### DIFF
--- a/core/src/main/scala/spark/deploy/master/JobState.scala
+++ b/core/src/main/scala/spark/deploy/master/JobState.scala
@@ -4,4 +4,6 @@ object JobState extends Enumeration("WAITING", "RUNNING", "FINISHED", "FAILED") 
   type JobState = Value
 
   val WAITING, RUNNING, FINISHED, FAILED = Value
+
+  val MAX_NUM_RETRY = 10
 }
--- a/core/src/main/scala/spark/deploy/master/Master.scala
+++ b/core/src/main/scala/spark/deploy/master/Master.scala
@@ -1,19 +1,18 @@
 package spark.deploy.master
 
-import scala.collection.mutable.{ArrayBuffer, HashMap, HashSet}
-
 import akka.actor._
-import spark.{Logging, Utils}
-import spark.util.AkkaUtils
+import akka.actor.Terminated
+import akka.remote.{RemoteClientLifeCycleEvent, RemoteClientDisconnected, RemoteClientShutdown}
+
 import java.text.SimpleDateFormat
 import java.util.Date
-import akka.remote.RemoteClientLifeCycleEvent
+
+import scala.collection.mutable.{ArrayBuffer, HashMap, HashSet}
+
 import spark.deploy._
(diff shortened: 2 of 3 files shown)


mateiz:
Good idea! Just ran into this with someone else actually.

core/src/main/scala/spark/deploy/master/Master.scala
```diff
@@ -81,12 +80,22 @@ class Master(ip: String, port: Int, webUiPort: Int) extends Actor with Logging {
...
-            schedule()
+
+            // Only retry certain number of times so we don't go into an infinite loop.
+            if (jobInfo.incrementRetryCount <= JobState.MAX_NUM_RETRY) {
+              schedule()
+            } else {
+              val e = new SparkException("Job %s wth ID %s failed %d times.".format(
+                jobInfo.desc.name, jobInfo.id, jobInfo.retryCount))
+              logError(e.getMessage, e)
+              throw e
```
JoshRosen, line 96: Is there a reason why you don't  call `removeJob()` instead of throwing an exception?  Removing the job would prevent it from being scheduled.

I'm trying to fix [SPARK-643](https://spark-project.atlassian.net/browse/SPARK-643), which is related to this exception.

core/src/main/scala/spark/deploy/master/Master.scala
```diff
@@ -81,12 +80,22 @@ class Master(ip: String, port: Int, webUiPort: Int) extends Actor with Logging {
...
+            jobInfo.removeExecutor(exec)
             exec.worker.removeExecutor(exec)
-            // TODO: the worker would probably want to restart the executor a few times
-            schedule()
+
+            // Only retry certain number of times so we don't go into an infinite loop.
+            if (jobInfo.incrementRetryCount <= JobState.MAX_NUM_RETRY) {
+              schedule()
+            } else {
+              val e = new SparkException("Job %s wth ID %s failed %d times.".format(
```
JoshRosen, line 93: Maybe it would be clearer to say "Job %s with ID %s  had %d executor failures; aborting".

core/src/main/scala/spark/deploy/master/JobState.scala
```diff
@@ -4,4 +4,6 @@ object JobState extends Enumeration("WAITING", "RUNNING", "FINISHED", "FAILED")
   type JobState = Value
 
   val WAITING, RUNNING, FINISHED, FAILED = Value
+
+  val MAX_NUM_RETRY = 10
```
JoshRosen, line 8: Should this be configurable?

mateiz, line 8: This is an old pull request, Josh -- if you want to change stuff you should submit a new one. I agree that making this configurable (and lower by default!) would be good.

PR #215, opened 2012-09-21
HTTP File server fixes

dennybritz:
The server returned the local file path instead of the URL, thus it was working when testing locally, but not on a cluster.


mateiz:
Thanks for the fix.

PR #217, opened 2012-09-24
Added a method to RDD to expose the ClassManifest.

rxin:


mateiz:
Minor thing.. rename it to something like elementManifest to make it more obscure. It would've also been nice to make it private[spark] but I guess your Shark code is not in the Spark package, right?

rxin:
Renamed it to elementClassManifest. PTAL.

mateiz:
OK, thanks.

PR #218, opened 2012-09-25
Scripts to start Spark under windows

rnpandya:
Windows Command Shell versions of run and sbt scripts


mateiz:
Awesome, thanks Ravi!

PR #222, opened 2012-09-27
Added MapPartitionsWithSplitRDD.

rxin:


mateiz:
Awesome, thanks! One small thing I might change after is to rename this to just mapPartitions, since the compiler should be able to infer the type of function passed based on the number of arguments it expects.

rxin:
Talked to Matei offline - it is best to keep the names separate because if they are the same name, scala compiler can be confused by situations like

rdd.mapPartitions(myfunction)

When the arguments are not specified to myfunction, it cannot reliably infer the type (whether it is a 2 arg closure or a 1 arg closure).

PR #225, opened 2012-09-28
Log message which records RDD origin

pwendell:
This adds tracking to determine the "origin" of an RDD. Origin is defined by
the boundary between the user's code and the spark code, during an RDD's
instantiation. It is meant to help users understand where a Spark RDD is
coming from in their code.

This patch also logs origin data when stages are submitted to the scheduler.

Finally, it adds a new log message to fix an inconsitency in the way that
dependent stages (those missing parents) and independent stages (those
without) are logged during submission.

### DIFF
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -12,7 +12,6 @@ import scala.collection.mutable.ArrayBuffer
 import scala.collection.Map
 import scala.collection.mutable.HashMap
 import scala.collection.JavaConversions.mapAsScalaMap
-import scala.util.control.Breaks._
 
 import org.apache.hadoop.io.BytesWritable
 import org.apache.hadoop.io.NullWritable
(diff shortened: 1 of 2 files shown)


core/src/main/scala/spark/RDD.scala
```diff
@@ -124,6 +128,37 @@ abstract class RDD[T: ClassManifest](@transient sc: SparkContext) extends Serial
...
+    // transformation, a SparkContext function (such as parallelize), or anything else that leads
+    // to instantiation of an RDD. We also track the first (deepest) user method, file, and line.
+    var lastSparkMethod = "<not_found>"
+    var firstUserMethod = "<not_found>"
+    var firstUserFile = "<not_found>"
+    var firstUserLine = -1
+
+    breakable {
+      for (el <- trace) {
+        if (el.getClassName().contains("spark") && !el.getClassName().contains("spark.examples")) {
```
mateiz: Use startsWith instead of contains here

core/src/main/scala/spark/RDD.scala
```diff
@@ -12,6 +12,7 @@ import scala.collection.mutable.ArrayBuffer
 import scala.collection.Map
 import scala.collection.mutable.HashMap
 import scala.collection.JavaConversions.mapAsScalaMap
+import scala.util.control.Breaks._
```
mateiz: Thanks Patrick but try not to use this if possible, because the rest of our code doesn't; the break stuff in Scala is confusing and not very efficient

core/src/main/scala/spark/RDD.scala
```diff
@@ -124,6 +128,37 @@ abstract class RDD[T: ClassManifest](@transient sc: SparkContext) extends Serial
...
+        }
+        else {
+          firstUserMethod = el.getMethodName()
+          firstUserLine = el.getLineNumber()
+          firstUserFile = el.getFileName()
+          break
+        }
+      }
+    }
+    "%s called in %s (%s:%s)".format(lastSparkMethod, firstUserMethod, firstUserFile, firstUserLine)
```
mateiz: Might be better to just change this to "[sparkMethod] at [file]:[line]" to make the log messages shorter

core/src/main/scala/spark/RDD.scala
pwendell: Okay - are you against doing this in an imperative style in general (there are other ways to create an analog of break without using this package)?

mateiz: Just replace it with some while loops or for loops -- I don't care exactly how, as long as it doesn't rely on break. It's probably possible to do something functional too where you recursively search down the list, but it's not necessary.

mateiz:
Thanks for the fixes!

core/src/main/scala/spark/RDD.scala
```diff
@@ -124,6 +127,38 @@ abstract class RDD[T: ClassManifest](@transient sc: SparkContext) extends Serial
     }
   }
   
+  // Describe which spark and user functions generated this RDD. Only works if called from
+  // constructor.
+  def getOriginDescription : String = {
```
squito, line 132: should this method be private?

mateiz, line 132: Good idea -- will fix this.

PR #227, opened 2012-09-29
Allow controlling number of splits in distinct().

JoshRosen:
Also documents `distinct()` in the Scala programming guide.

### DIFF
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -168,7 +168,8 @@ abstract class RDD[T: ClassManifest](@transient sc: SparkContext) extends Serial
   
   def filter(f: T => Boolean): RDD[T] = new FilteredRDD(this, sc.clean(f))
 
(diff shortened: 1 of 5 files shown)


core/src/main/scala/spark/RDD.scala
```diff
@@ -168,7 +168,8 @@ abstract class RDD[T: ClassManifest](@transient sc: SparkContext) extends Serial
   
   def filter(f: T => Boolean): RDD[T] = new FilteredRDD(this, sc.clean(f))
 
-  def distinct(): RDD[T] = map(x => (x, "")).reduceByKey((x, y) => x).map(_._1)
+  def distinct(numSplits: Int = splits.size): RDD[T] =
+    map(x => (x, "")).reduceByKey((x, y) => x, numSplits).map(_._1)
```
mateiz: One minor suggestion: use null instead of "" here, to avoid object allocation when we deserialize it. (!)

JoshRosen: Good catch; fixed.

mateiz:
Alright, thanks!

PR #233, opened 2012-10-01
Fixed #232: DirectBuffer's cleaner was empty and Spark tried to invoke clean on it.

rxin:


mateiz:
Thanks Reynold! I actually saw that check in some of the samples I read but never saw it happen on my box.

PR #235, opened 2012-10-01
publish-local should go to maven + ivy by default

pwendell:
This makes publishing to ivy and maven the default behavior. Worked for me on a clean build.

### DIFF
--- a/project/SparkBuild.scala
+++ b/project/SparkBuild.scala
@@ -22,7 +22,7 @@ object SparkBuild extends Build {
 
   // A configuration to set an alternative publishLocalConfiguration
   lazy val MavenCompile = config("m2r") extend(Compile)
-  lazy val publishLocalBoth = TaskKey[Unit]("pl", "publish local for m2 and ivy")
+  lazy val publishLocalBoth = TaskKey[Unit]("publish-local", "publish local for m2 and ivy")
 
   def sharedSettings = Defaults.defaultSettings ++ Seq(
     organization := "org.spark-project",


mateiz:
Great, thanks.

ghost:
FWIW, I don't see why the default should be to publish everything twice.  For  project builds SBT allows one to specify whether a particular dependency is M2 or Ivy.  No particular need to double publish locally.  Not arguing against having the build system support double local publishing, just making it the default behavior.  I am not aware of any other SBT / Scala project that follows this approach.

mateiz:
Are you worried that the build will just take too long? I think that in that case we should do Maven only; the problem is that more tools seem to be aware of Maven than Ivy. But I'm not sure how easy it is to change the rules in sbt so that publish-local means Maven only. We'd probably have to give it a different name (e.g. maven-local).

pwendell:
I think the ideal setup would be

maven-local
ivy-local
publish-local (goes to both)

With regards to publish-local going to both: For people who have
expectations based on other projects, I don't see that they are
particularly harmed by having it publish to maven as well - it takes
only negligible time. For people who do not have expectations from other
projects, I think it's most intuitive that publish-local would publish to
any local repo's that are publish-able. This project supports a Java API,
so we expect to have a lot of people building for the purpose of hooking
into maven, including people who aren't active developers on several scala
projects.
- Patrick

rxin:
Patrick is best to respond to this, but I think the extra time is measured

## in seconds (just a few seconds) ....

Reynold Xin
AMPLab, UC Berkeley
http://www.cs.berkeley.edu/~rxin/

ghost:
Sorry, guys.  Did not intend to generate a tempest in a teapot on this one.

Certainly there is a tacit expectation (principle of least surprise) that a standard Scala project built with SBT upon execution of publish-local will Ivy publish locally at a minimum.  I don't think that should change, which it has not.

Extending the behavior to double publish on the surface may affect those who already customize their local-publish command (which I do).   In my case the double publishing had no deleterious impact.  (Changing publish-local to maven only would have.)

After that the question comes down to time and space considerations.  In my case for example, I routinely disable ScalaDoc and Source(?) packaging in publish local to save time in my development cycle.

All-in-all, this is a minor concern.  I'd have left publish-local to default behavior and extended behavior with publish-local-maven.  It's a one-liner to put back single Ivy local publishing and I've already made that change in my Spark clone.  So all is well.  I withdraw my concerns.

PR #236, opened 2012-10-02
A Spark "Quick Start" example

pwendell:
This commit includes a quick start example that covers:
1) Basic usage of the Spark shell
2) A simple Spark job in Scala
3) A simple Spark job in Java


mateiz:
This looks pretty good overall! A few comments though:
- For the standalone jobs, show them how to package their job into a JAR and pass that to SparkContext. Don't use the two-arguent SparkContext constructor. In fact it might be good to end the guide with a section called "running on a cluster" that points them to the standalone cluster doc and tells them to try running the job against that.
- At the beginning, include a line on how to compile Spark (it's just sbt package)
- Link to the programming guide when describing the operations. It might also be good to add an example of reduceByKey that does a word count, without fully explaining it, and say "look in the programming guide for details of these operations".
- Separate the "A Spark Job" into two top-level sections: Writing a Standalone Scala Job and Writing a Standalone Java Job

pwendell:
Hey Matei,

I want to make sure I understand the first point, since I haven't used that
constructor yet. The job would look the same as it does now, except that it
would pass a reference to a packaged jar file (containing the job itself)
to the constructor. We'd walk through pacaking the jar file in a way that's
consistent with the name passed in the constructor.
- Patrick

mateiz:
Yes, exactly. Look at the part on Linking with Spark in the programing guide.

pwendell:
This latest commit includes substantial changes based on Matei's feedback. It addresses all 4 comments.

mateiz:
Thanks Patrick! Might do a few minor edits to this but it looks good.

pwendell:
Sounds good. In your court now.

PR #240, opened 2012-10-03
Package-Private Classes

dennybritz:
Made all core classes except for *RDD and what's used by Shark package private. Haven't yet figured out how to make the HTML templates for the standalone UI private. 

I'll do the same for the other subprojects, just wanted to get comments.

### DIFF
--- a/core/src/main/scala/spark/Accumulators.scala
+++ b/core/src/main/scala/spark/Accumulators.scala
@@ -18,7 +18,7 @@ import scala.collection.generic.Growable
  * @tparam R the full accumulated data
  * @tparam T partial data that can be added in
  */
-private[spark] class Accumulable[R, T] (
+class Accumulable[R, T] (
     @transient initialValue: R,
     param: AccumulableParam[R, T])
   extends Serializable {
@@ -73,7 +73,7 @@ private[spark] class Accumulable[R, T] (
  * @tparam R the full accumulated data
  * @tparam T partial data that can be added in
  */
-private[spark] trait AccumulableParam[R, T] extends Serializable {
+trait AccumulableParam[R, T] extends Serializable {
   /**
    * Add additional data to the accumulator value.
    * @param r the current value of the accumulator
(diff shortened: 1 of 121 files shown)


mateiz:
Good start, but the following should _not_ be private:
- Accumulable, AccumulableParam, Accumulator, AccumulatorParam
- Broadcast

mateiz:
Also PairRDDFunctions, DoubleRDDFunctions, etc (all the implicit conversions).

mateiz:
By the way, there's no need to do the same in repl and examples. We are not providing ScalaDoc for those. Maybe do it for internal classes in Bagel, but there aren't any of those AFAIK.

mateiz:
One last thing: as a matter of style, maybe put the private[spark] on a separate line above the class name when the class name is long or has many parameters. Also, no need to include it for inner classes in a class that is already private[spark].

dennybritz:
Made the changes. I thought I didn't mark any inner classes, at least not purposefully.

rxin:
Denny - can you check if Shark still compiles with this?

dennybritz:
Yes, Shark compiles. There are some classes (like OneToOneDependency) that I left purposefully public for Shark even though they probably shouldn't be.

JoshRosen:
What about the Java API's classes?  The abstract traits `WrappedFunction1` and `WrappedFunction2` should probably be private, since they aren't part of the user-facing API.

I think that `JavaRDDLike` should be private.  It doesn't really function like a proper RDD interface/base class since some common methods like `filter()` and `cache()` aren't defined in it, so leaving it public could be confusing.  This situation is due to a weird interaction between a Scala compiler bug and Java, which prevents me from completely following the [`TraversableLike` approach](http://docs.scala-lang.org/overviews/core/architecture-of-scala-collections.html) and having abstract methods that return RDDs of the instance's type.

dennybritz:
Yeah, I wasn't sure about those. Seems like making them private is the right thing to do then.

JoshRosen:
On the other hand, `JavaRDDLike` might be useful when implementing other language APIs on top of the Java API; removing it would force you to have variables of type `Object`  to hold Java RDDs of any type.  Both approaches would require casting, but `JavaRDDLike` provides slightly more type safety.  I suppose a third option would be to define an empty trait like `JavaRDDBase` for the sole purpose of being a superclass/interface for the Java RDDs.

dennybritz:
Yeah, that makes sense. I left JavaRDDLike public for now.

mateiz:
Looks good. I think we should leave the Dependency stuff public in case people want to subclass RDD themselves. We might instead make some of the RDD subclasses private (e.g. MappedRDD) but let's leave that for later. I'll merge this in now.

mateiz:
I just made a few more things package-private, so try Shark again. Doubt it's affected though. I left all the shuffle RDDs Reynold added in.

PR #241, opened 2012-10-03
First cut at adding documentation for GC tuning

shivaram:
Created new sub-headings for cache-size tuning and GC tuning.


mateiz:
Thanks Shivaram!

PR #243, opened 2012-10-03
Removes the included mesos-0.9.0.jar and pulls it from Maven Central instead

andyk:
Removes the included mesos-0.9.0.jar and adds a libraryDependency to the build file so that mesos-0.9.0-incubating.jar (which contains the same class files, but has a silightly different name) will be pulled down from Maven Central instead.


mateiz:
Sweet! Thanks Andy.

PR #249, opened 2012-10-06
Move RDD classes/files to their own package/directory

andyk:
Refactors RDD classes/files to their own package/directory. All RDD files except for RDD.scala
are now in src/main/scala/spark/rdd. Also fixes a bug in AccumulatorSuite.scala that was causing
the test named "localValue readable in tasks" to fail.


mateiz:
Awesome, thanks!

mateiz:
By the way just FYI, your Eclipse seemed to have added one tab, so you might want to change it to spaces only. I'll fix that in another commit.

PR #250, opened 2012-10-07
Fixed a bug in addFile that if the file is specified as "file:///", the symlink is created incorrectly for local mode.

rxin:


mateiz:
Looks good, thanks.

PR #251, opened 2012-10-07
Document Dependency classes and make minor interface improvements

JoshRosen:
These commits add API documentation for the `Dependency` classes and make two interface changes:
- Remove the unused `isShuffle` field from `Dependency`
- Change the type of `ShuffleDependency.aggregator` to an `Option` to communicate that the aggregator is optional.  Currently, the absence of an aggregator is expressed using `Aggregator[K, V, V](null, null, null, false)`, which is a bit of a hack.

Will these two changes introduce backwards-compatibility problems in the API?

### DIFF
--- a/core/src/main/scala/spark/Dependency.scala
+++ b/core/src/main/scala/spark/Dependency.scala
@@ -1,8 +1,8 @@
 package spark
 
-abstract class Dependency[T](val rdd: RDD[T], val isShuffle: Boolean) extends Serializable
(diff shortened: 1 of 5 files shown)


mateiz:
Cool, thanks!

mateiz:
I'm okay changing the RDD internal API for now, though this might affect Shark. We will lock it down more when we get to a version 1.0.

PR #256, opened 2012-10-08
Adding new download instructions

pwendell:


pwendell:
I updated this just now to reflect the templated versions. I also changed the wording a bit.

mateiz:
Thanks but I think this is too long for an overview page. Just say "this documentation corresponds to version X" and they can figure out how to get others. Also, cut the part about Git.

pwendell:
sure now this just says the version and the downloads page.

mateiz:
OK, thanks.

PR #257, opened 2012-10-08
Adds special version variables to docs templating system

andyk:
Adds liquid variables to docs templating system so that they can be used throughout the docs: SPARK_VERSION, SCALA_VERSION, and MESOS_VERSION.

To use them, e.g. use {{site.SPARK_VERSION}}.

Also removes uses of {{HOME_PATH}} which were being resolved to ""
by the templating system anyway.


mateiz:
Can you pull the latest change from dev into your branch to make it mergeable? One of the earlier commits made it incompatible.

andyk:
Ok, I think I've fixed the conflict. i also updated the README with instructions for building without scaladoc (because this feature can save people lots of time).

mateiz:
Great, thanks!

PR #258, opened 2012-10-08
Readding yarn-standalone scheduler scheme

dennybritz:
Also merged the latest dev branch in.


mateiz:
Thanks. Please reply to that thread on spark-developers too.

PR #259, opened 2012-10-08
Synchronization bug fix in broadcast implementations

mosharaf:
- Synchronization bugs found during debugging the broadcast problem faced by one of the users (their problem couldn't be reproduced and still remain unresolved)
- Removed some logging comments
- Changed the broadcast example to call broadcast on multiple consecutive iterations, instead of just one.


mateiz:
Thanks Mosharaf. It would still be good to investigate and fix the full bug sometime though.

PR #260, opened 2012-10-09
Updates docs to use the new version num vars and adds Spark version in nav bar

andyk:
Updating lots of docs to use the new special version number variables, and adding the version to the navbar so it is easy to tell which version of Spark these docs were compiled for.

See what the generated docs look like after this commit at: http://www.cs.berkeley.edu/~andyk/spark-docs-e1a724f39/


mateiz:
Thanks Andy.

PR #263, opened 2012-10-09
Add m1.medium node option to cluster management script.

ghost:
EC2 m1.medium node option is missing in the spark_ec2.py script.


mateiz:
Awesome, thanks.

PR #264, opened 2012-10-09
Support for Hadoop 2 distributions such as cdh4

tomdz:
Hadoop 2 unfortunately introduced backwards-incompatible changes in classes used by Spark, `JobContext` and `TaskAttemptContext` in both the `mapred` and `mapreduce` packages. In the new codebase, these classes are now interfaces and thus cannot be instantiated directly anymore. Instead, you are now required to use the `JobContextImpl` and `TaskAttemptContextImpl` classes, respectively.

This pull requests extracts the code in Spark where it uses these, into separate traits that are located in Hadoop major version specific source folders that the sbt build then includes/excludes as necessary.

In addition, for Hadoop 2, Spark now also needs to depend on the `hadoop-client` artifact.

### DIFF
--- a/core/src/hadoop1/scala/org/apache/hadoop/mapred/HadoopMapRedUtil.scala
+++ b/core/src/hadoop1/scala/org/apache/hadoop/mapred/HadoopMapRedUtil.scala
@@ -1,0 +1,7 @@
+package org.apache.hadoop.mapred
+
+trait HadoopMapRedUtil {
+  def newJobContext(conf: JobConf, jobId: JobID): JobContext = new JobContext(conf, jobId)
+
+  def newTaskAttemptContext(conf: JobConf, attemptId: TaskAttemptID): TaskAttemptContext = new TaskAttemptContext(conf, attemptId)
+}
--- a/core/src/hadoop1/scala/org/apache/hadoop/mapreduce/HadoopMapReduceUtil.scala
+++ b/core/src/hadoop1/scala/org/apache/hadoop/mapreduce/HadoopMapReduceUtil.scala
@@ -1,0 +1,9 @@
+package org.apache.hadoop.mapreduce
+
+import org.apache.hadoop.conf.Configuration
+
+trait HadoopMapReduceUtil {
+  def newJobContext(conf: Configuration, jobId: JobID): JobContext = new JobContext(conf, jobId)
+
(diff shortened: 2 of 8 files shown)


mateiz:
Thanks Tom, but we actually already have these fixes in the `yarn` branch. I don't think we want to merge them into master because that still relies on Hadoop 0.20.205. You can't use a Hadoop 2 client to connect to an older cluster, right?

I think we'll just recommend using the `yarn` branch to people who use Hadoop 2, and we'll merge them when more people have moved to Hadoop 2.

tomdz:
I don't know much about the yarn branch, but the name implies that it requires yarn, which we're not using with our cdh4 installation (we only use MRv1). The changes in this pull request seem to be sufficient to run current spark against the hdfs in our installation.

mateiz:
My question is just, do these changes work when connecting to HDFS clusters on Hadoop 1.0 or 0.20.x? Since there's an API change in Hadoop, it seems that they wouldn't work.

We should probably just rename the "yarn" branch to "hadoop2". You're right that the name of that branch makes it sound like it requires YARN, but actually it runs with Mesos, the standalone deploy mode, etc as well.

tomdz:
The change are supposed to make Spark work with the Hadoop/hdfs version that it is compiled against. I.e. if I use `2.0.0-mr1-cdh4.1.0` as the Hadoop version (and set the `HADOOP_MAJOR_VERSION` to 2 since cdh4 is a Hadoop 2 distribution), then the resulting Spark artifact will work with cdh4 (and presumably everything that is API compatible to that Hadoop version). The mechanism would allow to handle other Hadoop backwards-incompatible changes as well.

On a related note, we have a Maven POM that defines profiles for each of the popular Hadoop distributions (Hadoop 1 & 2) and generates Spark jars with corresponding classifiers. The benefit of that is that in my project I can depend on the right version of Spark (using a classifier in the dependency). If there is interest in that, I could create a pull request for it.

mateiz:
Thanks, I didn't see that you're building differently based on the major version. I'll merge this in but probably after the release of 0.6.0 (in the next minor release after that). Maybe we can use the same mechanism to bring the YARN stuff into the main branch and only compile it conditionally.

tomdz:
Regenerated the pull request against 0.6 dev branch in #285.

PR #267, opened 2012-10-10
Fixed bug when fetching Jar dependencies.

dennybritz:
Instead of checking currentFiles check currentJars. Came across this when testing some Shark queries.


mateiz:
Good catch; I probably miscopied this when I moved this code around to fetch JARs before deserializing the task.

PR #268, opened 2012-10-11
Adding code for publishing to Sonatype.

pwendell:
By default - I'm leaving this commented out. This is because
there is a bug in the PGP signing plugin which causes it to active
even duing a publish-local. So we'll just uncomment when we decide
to publish.

### DIFF
--- a/project/SparkBuild.scala
+++ b/project/SparkBuild.scala
@@ -47,34 +47,34 @@ object SparkBuild extends Build {
     pomExtra := (
       <url>http://spark-project.org/</url>
       <licenses>
-	<license>
-	  <name>BSD License</name>
-	  <url>https://github.com/mesos/spark/blob/master/LICENSE</url>
-	  <distribution>repo</distribution>
-	</license>
+        <license>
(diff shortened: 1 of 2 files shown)


project/SparkBuild.scala
```diff
@@ -33,7 +35,51 @@ object SparkBuild extends Build {
...
+      "sonatype-staging" at "https://oss.sonatype.org/service/local/staging/deploy/maven2/"),
+
+    publishMavenStyle := true,
+
+    useGpg in Global := true,
+
+    pomExtra := (
+      <url>http://spark-project.org/</url>
+      <licenses>
+	<license>
```
mateiz: Are there tabs here? Looks weird.

pwendell: good catch - I fixed with a new commit.

mateiz:
Okay, thanks.

mateiz:
By the way, the same issue seems to affect publish-local on branch 0.5. I'll fix it there too.

PR #271, opened 2012-10-11
Change block manager to accept a ArrayBuffer 

shivaram:
Attempt to fix #269. Existing tests pass, but I haven't added a new test to make sure things are fixed. This is partly because I was not sure if I should extend CacheTrackerSuite or add something to the Block Manager test cases.

### DIFF
--- a/core/src/main/scala/spark/storage/BlockManager.scala
+++ b/core/src/main/scala/spark/storage/BlockManager.scala
@@ -237,7 +237,10 @@ class BlockManager(val master: BlockManagerMaster, val serializer: Serializer, m
             diskStore.getValues(blockId) match {
               case Some(iterator) =>
                 // Put the block back in memory before returning it
-                memoryStore.putValues(blockId, iterator, level, true).data match {
+                // TODO: Consider creating a putValues that also takes in a iterator ?
+                val elements = new ArrayBuffer[Any]
+                elements ++= iterator
+                memoryStore.putValues(blockId, elements, level, true).data match {
                   case Left(iterator2) =>
                     return Some(iterator2)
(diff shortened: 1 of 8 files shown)


core/src/main/scala/spark/storage/BlockManager.scala
```diff
@@ -529,11 +532,18 @@ class BlockManager(val master: BlockManagerMaster, val serializer: Serializer, m
     }
   }
 
+  def put(blockId: String, values: Iterator[Any], level: StorageLevel, tellMaster: Boolean)
```
mateiz, line 535: Why didn't you make tellMaster have a default value there? Did the compiler complain due to the other put?

shivaram, line 535: Yes - The compiler gives an error that multiple overloaded methods cannot have a default value.

mateiz:
This looks good to me. I would add a test in DistributedSuite actually, to just launch a cluster where the memory limit per worker is low (e.g. set spark.storage.memoryFraction = 0.01) and try caching and using some data on it.

core/src/main/scala/spark/storage/BlockManager.scala
mateiz, line 535: OK, makes sense. Just add the test as I mentioned and it should be good to go.

shivaram:
Added the test with the assumption that 512MB of memory times 0.0001 is around 50KB and the array should be larger than that.

mateiz:
Thanks. It's a good bet because we do set the JVMs to exactly 512 MB of memory. I might also add a test where half the data fits and half doesn't.

PR #273, opened 2012-10-13
Let the user specify environment variables to be passed to the Executors

dennybritz:


core/src/main/scala/spark/SparkContext.scala
```diff
@@ -92,6 +92,12 @@ class SparkContext(master: String, jobName: String, val sparkHome: String, jars:
   private[spark] val addedFiles = HashMap[String, Long]()
   private[spark] val addedJars = HashMap[String, Long]()
 
+  // Environment variables to pass to our executors
+  private[spark] val executorEnvs = HashMap[String, String]()
+  Seq("SPARK_MEM", "SPARK_CLASSPATH", "SPARK_LIBRARY_PATH", 
+    "SPARK_JAVA_OPTS", "SPARK_TESTING").foreach { key => executorEnvs.put(key, System.getenv(key)) }
```
mateiz, line 98: Only put each key if System.getenv(key) is not null

mateiz:
Hey Denny,

Are you sure this will work for the standalone and Mesos coarse-grain modes? In those cases, it the Scheduler will launch executors _before_ you get the chance to call putExecutorEnv on the SparkContext, so they won't see the new variables.

I suggest that, instead, you make SparkContext's constructor take an `env` map as a fifth argument (and add a separate 4-argument constructor). That will also make it clear that you can't add or change the env after you start a job.

Matei

mateiz:
By the way if you do this, change JavaSparkContext in the same way.

dennybritz:
Ah, you're right. I won't be back before tonight though so if you want to release today please feel free to make the changes yourself. If not I will change it tomorrow morning.

mateiz:
Alright, I'll merge it and fix this myself; thanks.

PR #276, opened 2012-10-14
Disable gpg by default.

rxin:


mateiz:
I've already done this separately, so closing the pull request, but thanks!.

PR #281, opened 2012-10-16
Added a method to report slave memory status; force serialize accumulator update in local mode.

rxin:


mateiz:
I've merged this, thanks. But actually thinking a bit more about it I would prefer if the map returned (amountInUse, totalAmount) pairs instead of (total, remaining) as a more sensible public API. Feel free to submit another request with that if you can also change Shark to take it into account.

PR #284, opened 2012-10-18
Structure SBT build file into modules.

ghost:
_Build Changes_

This does a bit of housekeeping for the SparkBuild SBT file.  It should be considered a first cut and not an ideal.  Please feel free to modify to taste.
- It is intended to be "neutral" in effect in the sense no build semantics were meant to be changed.
- No dependency versions were changed.
- It does publish-local a Spark build successfully on my machine.

_Breakage_
- The Maven stuff is certainly not working and I hope this is something that can be repaired quickly.  Sorry about that.

_Modifications_
- I added an additional file ShellPrompt.scala  This modifies the SBT prompt to show the prompt as 
  project:branch:project-version>
- As a result of that, I changed the root projects name from "root" to "spark" as seeing a console prompt starting with the word "root" gives me palpitations.
- I took the liberty of adding two additional scalac compiler settings -Xlint and -Xcheckinit.

Ray


ghost:
I added a comment about something I noted in the build file.
  /\* Workaround for issue #206 (fixed after SBT 0.11.0) */
  // Is this still applicable?  Remove? RPR
  watchTransitiveSources <<= Defaults.inDependencies[Task[Seq[File]]](watchSources.task,
                                                                              const %28std.TaskExtra.constant %28Nil%29%29, 
                                                                              aggregate = true, 
                                                                              includeRoot = true) apply { _.join.map (_.flatten) },

Since Spark is now on SBT 11.3 (maybe heading to 12.2) does anyone recall if this is still necessary?

mateiz:
Hey Ray,

That watchTransitiveSources part probably isn't needed. It's just in there from a while back.

Unfortunately I recently committed another change which affected the build file to do a conditional build based on Hadoop version (https://github.com/mesos/spark/pull/285). Do you want to integrate that here or would you prefer for me to do it?

Also, send these to the master branch instead of dev from now on, although they are currently identical.

Matei

mateiz:
Hi Ray,

I'm going to close this since we decided to switch to Maven. Hopefully the Maven build will be relatively sane. It's certainly better organized than the SBT file was.

Matei

PR #285, opened 2012-10-18
Support for Hadoop 2 distributions such as cdh4

tomdz:
Hadoop 2 unfortunately introduced backwards-incompatible changes in classes used by Spark, `JobContext` and `TaskAttemptContext` in both the `mapred` and `mapreduce` packages. In the new codebase, these classes are now interfaces and thus cannot be instantiated directly anymore. Instead, you are now required to use the `JobContextImpl` and `TaskAttemptContextImpl` classes, respectively.

This pull requests extracts the code in Spark where it uses these, into separate traits that are located in Hadoop major version specific source folders that the sbt build then includes/excludes as necessary.

In addition, for Hadoop 2, Spark now also needs to depend on the `hadoop-client` artifact.

This pull request is a new version of #264 against the 0.6 dev branch.

### DIFF
--- a/core/src/hadoop1/scala/org/apache/hadoop/mapred/HadoopMapRedUtil.scala
+++ b/core/src/hadoop1/scala/org/apache/hadoop/mapred/HadoopMapRedUtil.scala
@@ -1,0 +1,7 @@
+package org.apache.hadoop.mapred
+
+trait HadoopMapRedUtil {
+  def newJobContext(conf: JobConf, jobId: JobID): JobContext = new JobContext(conf, jobId)
+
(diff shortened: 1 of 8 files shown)


tomdz:
This approach will require that there are two Spark artifacts published to the maven repo (one for Hadoop 1 and one for Hadoop 2). I have not made those changes to the `SparkBuild` file primarily because I'm not very familar with sbt (we use Maven to generate our Spark build).

Also worth noting that we played around with an alternate approach where Spark deals with the Hadoop incompatibilities at runtime by using reflection. The benefit of this would be a single Spark artifact. However this introduces a runtime overhead for every task/job that spark generates in HDFS, plus it removes type safety and overall feels less clean.

mateiz:
Thanks for submitting this. I've added it in dev, and I will also add it in branch 0.5 soon.

PR #286, opened 2012-10-19
Allow EC2 script to stop/destroy cluster after master/slave failures

JoshRosen:
This commit solves a problem where `spark-ec2` could not destroy or stop a cluster if its master or all of its slaves had been destroyed.

It also prints the total count of master and slave nodes before printing error messages.


JoshRosen:
This fix is useful when running spot instances, where all of the slaves may die due to the spot price exceeding the maximum bid.  In this case, the unpatched script prevents me from terminating the cluster because it cannot find any slaves.

mateiz:
Thanks Josh, looks good.

PR #294, opened 2012-10-26
Fix minor typos in quickstart and Scala programming guides

JoshRosen:


mateiz:
Good catches, thanks!

PR #295, opened 2012-10-29
A UI for the BlockManager and RDD Stats

dennybritz:
It's a WebUI to see how much memory and disk space an RDD and its blocks are taking up. By default started on port 9080. Wanted to get some feedback for the code.

Some notes:
- Added a method to name an RDD. The names are stored locally in the SparkContext.
- Added an Map[Int, RDD] to the SparkContext, because I couldn't find an easy way to get to the RDD properties (like StorageLevel) given an RDD ID.


core/src/main/scala/spark/SparkContext.scala
```diff
@@ -102,10 +103,19 @@ class SparkContext(
...
+
   // Used to store a URL for each static file/jar together with the file's local timestamp
   private[spark] val addedFiles = HashMap[String, Long]()
   private[spark] val addedJars = HashMap[String, Long]()
 
+  // Keeps track of all persisted RDDs
+  private[spark] val persistentRdds = new ConcurrentHashMap[Int, RDD[_]]()
+  // A HashMap for friendly RDD Names
+  private[spark] val rddNames = new ConcurrentHashMap[Int, String]()
+
```
JoshRosen, line 116: Would it be simpler to make the name a property of the `RDD` itself?  It looks like you only fetch the names for RDDs that are members of `persistentRdds`, so you could just fetch the RDD then access its name.

dennybritz, line 116: Good catch. I added the name feature before I added the RDD fetching, so I didn't think about it at that time.

JoshRosen, line 116: On a related note, I don't know whether it makes more sense to define a default name for an RDD in the RDD itself, or to define the name as `Option[String]` and leave it up to the caller to pick an appropriate default.

rxin:
For the detailed view, it might make more sense to first group by slaves, and then on each slave what rdd blocks are on them. A pretty critical piece of information is how balanced the cached data are on slaves, the current view doesn't show that information.

squito:
Whats the status of this?  If nobody is working on it, I might try to pick it up -- we'd love a feature like this.  But I'm not sure I know what the remaining issues are ...

rxin:
There was a refactoring I did to block manager that basically would require rewriting some of this pull request. If you want to take it over, that'd be great too.

pwendell:
I was looking at this a few weeks ago, but it got shelved in favor of some
other work. It would be great if this could be:
1. Brought up to the most recent version of the block manager code.
2. Be slightly generalized so it can support other UI-data that is on a
   per-executor basis. Mostly I just mean that the BlockMangerUI could be
   renamed, or encapsulated in a broader ExecutorUI which could be extended to
   add other things like JVM metrics and debugging information.

squito:
I couldn't update this pull request, so opened a new one with the code brought up to date
https://github.com/mesos/spark/pull/401

pwendell:
This is subsumed by the new pull request from squito.

PR #296, opened 2012-11-01
Remove unnecessary hash-map put in MemoryStore

shivaram:
The putBytes method in the MemoryStore was inserting twice into the entries HashMap. I don't think there was a correctness issue with this, but it was just redundant code.


mateiz:
Good catch, thanks.

PR #297, opened 2012-11-01
Cancel spot instance requests when exiting spark-ec2

JoshRosen:
This commit cancels spot instance requests if `spark-ec2` is interrupted while waiting for the requests to be fulfilled.  When exiting, the script checks whether any of the spot requests were fulfilled and logs a warning if any instances were launched (it does not automatically kill those instances).


mateiz:
Thanks for posting this.

PR #298, opened 2012-11-04
Fix check for existing instances during spark-ec2 launch

JoshRosen:
This fixes a bug in how `spark-ec2` checks for running instances before launching a cluster.  The old code compared reservation group ids instead of names; this bug allowed multiple clusters to be launched with the same name.


mateiz:
Weird; that definitely worked in the past, but I guess some stuff about Boto changed.

PR #300, opened 2012-11-08
fix bug in getting slave id out of mesos

squito:
Fixed this error:

12/11/05 11:22:13 INFO mesos.MesosSchedulerBackend: Mesos slave lost: 201211021432-3171224074-5050-18217-5
Exception in thread "Thread-338" java.util.NoSuchElementException: key not found: value: "201211021432-3171224074-5050-18217-5"

b/c we used to toString a protobuf instead of getting the value out of it.


mateiz:
Great, thanks!

PR #302, opened 2012-11-09
Blockmanager fix

tdas:
Fixed BlockManager deadlock.
1. Changed the lock structure of BlockManager by replacing the 337 coarse-grained locks to use BlockInfo objects as per-block fine-grained locks.
2. Changed the MemoryStore lock structure by making the block putting threads lock on a different object (not the memory store) thus making sure that putting threads minimally contends with the getting threads.
3. Added spark.storage.ThreadingTest to stress test the BlockManager using 5 block producer and 5 block consumer threads.

Tested with 5 producer (consumer) threads each doing 50k puts (gets), took 15 minutes to run, no errors or deadlocks.

### DIFF
--- a/core/src/main/scala/spark/storage/BlockManager.scala
+++ b/core/src/main/scala/spark/storage/BlockManager.scala
@@ -50,16 +50,6 @@ private[spark]
 case class BlockException(blockId: String, message: String, ex: Exception = null)
 extends Exception(message)
 
-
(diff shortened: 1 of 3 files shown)


mateiz:
Thanks TD, looks good!

tdas:
We should run the performance tests on this fixed version. It will be interesting to see how much difference this makes in real workloads.

PR #303, opened 2012-11-12
Add capabilities to RDD's to show the DAG structure for debugging

nkronenfeld:
This is my attempt to output the DAG structure of an RDD, as per the "RDD Task Introspection" thread I started on spark-users.

You use it by just adding a .debug() call to any RDD in the chain about to be calculated, i.e.,

val foo = sc.textFile(...).map(...).debug().collect()

The DAGScheduler checks the stage, before it runs it, to see if any of its tasks are marked debug, and if they are, it prints out their structure.

I've only run this in a local setup, but if I understand things correctly, the scheduler is only run locally anyway, so I don't think there should be any issues there.

This also introduces a level in the RDD class tree in which to consolidate previous-RDD methods, which I suspect will end up being needed anyway eventually; I'm not sure if I've done this the best way or not, and welcome comments.


mateiz:
Cool! Give me a bit of time to review this because I have a few other things I need to deal with in the next few days, but this is very welcome.

mateiz:
Hi Nathan,

Sorry for taking a while to get back on this. I think this functionality is useful, but changing every RDD to extend SingleParentRDD or something similar is unwieldy. I would prefer a version that just looks at rdd.dependencies and prints a tree directly. Also, I would make this a method in SparkContext (e.g. SparkContext.printRDDTree(rdd)) instead of one on RDD, so that users can call it on just the RDD they want, without having to submit a job.

If you'd like to do it that way, start with the code in the `streaming` branch, because the RDD class has changed a little there. It still lists dependencies in a similar way but the fields are named differently.

nkronenfeld:
Ok, I've taken all your suggestions, and pushed them up to github in a streaming branch on my repository.  I had a memory from months ago about finding different output at execution and pre-execution - something to do with ShuffleRDDs.  But, either you changed it, or I was imagining things, because I can't find it now.  Therefore, I just got rid of the execution-time stuff in the stage and scheduler entirely - and with using dependencies like I should have to begin with, that leaves the only real change in RDD itself (no need to pull in the context at all, really).  I've left it as RDD.description, but if you want to change it to RDD.tree, or whatever, I certainly won't be offended :-)

I also have a second commit in the same branch reinstating the -deprecation flag, and fixing most of the deprecation warnings.  There are only 4 left - 3 dealing with hadoop (like the comment about removing -deprecation said), and one seemingly unfixable thread.stop warning, but since there were around 6 legitimate ones that were easily fixable, it seemed to me like it was hiding more real things than it was spam.

Do I need to initiate a new push request, or just notify you?  I notice this request references the debug branch, and I'm now on the streaming branch - is this a problem?

mateiz:
I think you need to initiate a new pull request to master from your streaming branch. Pull requests on GitHub do auto-update but only if you keep making commits to the same branch in your repo (in this case debug). By the way, you may also have to merge the public master branch into yours to do this, though I'm not sure. It will tell you whether the pull request can merge cleanly.

Anyway, both this and the deprecation fix sound great. I didn't realize there were that few deprecation warnings -- we should leave the flag on.

stephenh:
Nathan, I saw the commit in your streaming branch; it looked useful, so I adapted it locally for some debugging I'm doing. After fiddling, I come up with adding this to RDD:

``` scala
  /** A user-readable description of this RDD */
  def toDebugString(): String = {
    def debugString(rdd: RDD[_], prefix: String): Seq[String] = {
      Seq(prefix + rdd.toString) ++ rdd.dependencies.flatMap(d => debugString(d.rdd, prefix + "  "))
    }
    debugString(this, "").mkString("\n")
  }

  override def toString() = "%s[%d] at %s".format(getClass.getSimpleName, id, origin)
```

Sample output:

```
MapPartitionsRDD[4] at main at <unknown>:0
  ShuffledRDD[3] at main at <unknown>:0
    MapPartitionsRDD[2] at main at <unknown>:0
      MappedRDD[1] at main at <unknown>:0
        ParallelCollection[0] at main at <unknown>:0
```

This is less generic, i.e. startDelim/endDelim/prefix aren't configurable, but that seemed a bit like gold plating.

I also changed used "toDebugString" because I would see "description" ending up being used as a one-liner type description like you would show in regular console logs/the web UI.

I can submit a pull request with my version if you like, attributing you with the original idea, or you can adapt my approach into your commit, again if you like, or obviously you can just submit yours as is too. :-)

stephenh:
Nathan, I submitted my variation on your approach:

https://github.com/mesos/spark/pull/425

JoshRosen:
Since #425 was merged, I'm going to close this.  The deprecation fixes should be submitted as a separate pull request.

PR #304, opened 2012-11-15
SPARK-624: make the default local IP customizable

mbautin:
It would be great to be able to specify the default local IP for Spark to run servers on. This way we would not have to modify /etc/hosts on our servers (we have reasons not to). https://spark-project.atlassian.net/browse/SPARK-624

### DIFF
--- a/core/src/main/scala/spark/Utils.scala
+++ b/core/src/main/scala/spark/Utils.scala
@@ -200,7 +200,7 @@ private object Utils extends Logging {
    * Get the local host's IP address in dotted-quad format (e.g. 1.2.3.4).
    */
   def localIpAddress(): String = {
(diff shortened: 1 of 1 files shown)


mateiz:
Actually do you mind renaming this to just SPARK_LOCAL_IP? Seems a bit shorter if this is a thing that people will have to configure often.

mbautin:
Matei: I've updated the environment variable name to be SPARK_LOCAL_IP.

mateiz:
Great, thanks!

mbautin:
Matei: could you also merge this into 0.6, or should I open a separate pull request? Thanks!

mateiz:
I've merged them myself. They'll appear in the next minor version.

PR #305, opened 2012-11-16
Set default uncaught exception handler to exit

charlesreiss:
This patch sets a global default uncaught exception handler in SparkEnv which logs the exception and then exits with exit status 1, or an exit status of 2 if an exception occurs while trying to log the exception. This should prevent utility threads in Spark from failing (for example, due to OutOfMemoryErrors or IOExceptions caused by running out disk space) but merely causing Spark overall to stall.

This patch assumes that no Spark code or user code relies on uncaught exceptions causing normal thread termination. I looked at where Spark created utility threads and didn't see any obvious instances of this problem, though  I wonder what PipedRDD is supposed to do in the case of a broken pipe.

I'm a bit uncomfortable setting this JVM-wide setting rather than something that might be limited to Spark only. An alternate approach would be to create utility threads within Spark using some wrapper for Thread that sets an uncaught exception handler (or wraps run() appropriately), but this would be more code and some cases would likely be missed.

### DIFF
--- a/core/src/main/scala/spark/SparkEnv.scala
+++ b/core/src/main/scala/spark/SparkEnv.scala
@@ -68,7 +68,6 @@ object SparkEnv extends Logging {
       isMaster: Boolean,
       isLocal: Boolean
     ) : SparkEnv = {
-
     val (actorSystem, boundPort) = AkkaUtils.createActorSystem("spark", hostname, port)
 
     // Bit of a hack: If this is the master and our port was 0 (meaning bind to any free port),
--- a/core/src/main/scala/spark/executor/Executor.scala
+++ b/core/src/main/scala/spark/executor/Executor.scala
@@ -43,6 +43,21 @@ private[spark] class Executor extends Logging {
     urlClassLoader = createClassLoader()
     Thread.currentThread.setContextClassLoader(urlClassLoader)
 
+    // Make any thread terminations due to uncaught exceptions kill the entire
+    // executor process to avoid surprising stalls.
+    Thread.setDefaultUncaughtExceptionHandler(
+      new Thread.UncaughtExceptionHandler {
+        override def uncaughtException(thread: Thread, exception: Throwable) {
+          try {
+            logError("Uncaught exception in thread " + thread, exception)
+            System.exit(1)
+          } catch {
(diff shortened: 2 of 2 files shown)


mateiz:
Looks good, but can you make it happen only on worker nodes (in Executor)?

I think PipedRDD will work fine because our task thread in the Executor catches the exception.

charlesreiss:
Updated to only happen when spark.executor.Executors are created.

My concern with PipedRDD is that writer thread created at https://github.com/mesos/spark/blob/master/core/src/main/scala/spark/rdd/PipedRDD.scala#L54 will get an IOException if the piped program terminates prematurely causing writes to fail. (As long as the program closes its output normally, this will not cause the PipedRDD's iterator proper to fail.) I tried to reproduce this for a test, however, and failed -- it seemed like the PrintWriter just stalled indefinitely instead of producing an IOException for the broken pipe.

mateiz:
Ah, weird. We should probably just catch the exception in that thread in case this can happen sometimes.

mateiz:
Anyway, thanks for this, and feel free to send another pull request for PipedRDD.

hamzaasad:
i'm facing a problem while executing a simple join query on Shark terminal running on spark over hadoop. i have a table which contains data of approximately 12.5 GB and another table which contains only 10 rows. When i execute the query, after few minutes of execution, an exception occurs "java.lang.OutOfMemoryError: GC overhead limit exceeded". How can i solve this problem?? i'm using system having 4GB ram in which i have assigned SPARK_MEM=3g, SHARK_MASTER_MEM=3g and HADOOP_CLIENT_OPTS="-Xmx3072m"... can any one help how can i get rid off with this

mateiz:
Hi Hamza,

Please send questions to the Spark and Shark mailing lists https://groups.google.com/forum/?fromgroups#!forum/spark-users / https://groups.google.com/forum/?fromgroups#!forum/shark-users, not by posting them as comments on pull requests. Anyway, for this question, you probably need to configure your number of reduce tasks higher. Use set mapred.reduce.tasks=200 for example. This will require less memory per task. In addition, you might be able to use a map-side join, but I'm not sure whether that's available in Shark 0.2 (it is in the Git version of Shark though).

PR #306, opened 2012-11-16
Delete security groups when destroying cluster

pas256:
Delete security groups when deleting the cluster. As many operations are done on instances in specific security groups, and this scripts creates those groups (if they don't exist already because they were left over), this seems like a reasonable thing to clean up.

### DIFF
--- a/ec2/spark_ec2.py
+++ b/ec2/spark_ec2.py
@@ -509,6 +509,20 @@ def main():
         print "Terminating zoo..."
         for inst in zoo_nodes:
           inst.terminate()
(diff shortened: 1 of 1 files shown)


mateiz:
Looks good, thanks.

JoshRosen:
I think that the deletion of security groups should be optional: I've opened an issue at https://spark-project.atlassian.net/browse/SPARK-628

JoshRosen:
Also, the deletion occasionally results in an error: https://spark-project.atlassian.net/browse/SPARK-626

PR #307, opened 2012-11-16
Ignore .pydevproject

pas256:
Adding 1 more file to the list of files for Git to ignore


pas256:
Sorry, this included more than it should have... please ignore

PR #309, opened 2012-11-19
Improved use of Boto 

pas256:
Boto supports a number of ways to set AWS credentials, including a local `.boto` file which is what I use. Here are the others: 

http://boto.cloudhackers.com/en/latest/boto_config_tut.html

I also added a wrapper around the connection creation to gracefully handle exceptions such as  AuthFailure when bad credentials are used. Better than a Python traceback.

### DIFF
--- a/ec2/spark_ec2.py
+++ b/ec2/spark_ec2.py
@@ -19,7 +19,6 @@
 
 from __future__ import with_statement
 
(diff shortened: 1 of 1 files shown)


mateiz:
Thanks, looks good.

PR #292, opened 2012-10-22
Tweaked run file to live more happily with typesafe's debian package

tomdz:
The `run` script assumes a single-directory Scala installation structure with a `bin` and `lib` subdirectory. Unfortunately, that does not work well with Typesafe's Scala debian package which has this layout:

```
$ dpkg -L scala
/.
/usr
/usr/bin
/usr/bin/scalap
/usr/bin/scalac
/usr/bin/scala
/usr/bin/scaladoc
/usr/bin/fsc
/usr/share
/usr/share/scala
/usr/share/scala/plugins
/usr/share/scala/plugins/continuations.jar
/usr/share/java
/usr/share/java/scala-partest.jar
/usr/share/java/scala-swing.jar
/usr/share/java/scala-dbc.jar
/usr/share/java/jline.jar
/usr/share/java/scalacheck.jar
/usr/share/java/scala-library.jar
/usr/share/java/scala-compiler.jar
/usr/share/java/scalap.jar
/usr/share/doc
/usr/share/doc/scala
/usr/share/doc/scala/changelog.gz
/usr/share/doc/scala/copyright
/usr/share/man
/usr/share/man/man1
/usr/share/man/man1/fsc.1.gz
/usr/share/man/man1/scalap.1.gz
/usr/share/man/man1/scalac.1.gz
/usr/share/man/man1/scaladoc.1.gz
/usr/share/man/man1/scala.1.gz
```

This pull request changes the run script to:
- Use the `scala` binary as is if it is in the path, otherwise require `SCALA_HOME`
- Require either `SCALA_LIBRARY_PATH`, or if that is not defined, `SCALA_HOME` for running with `java` instead of `scala`

PR #310, opened 2012-11-21
Maven build setup

tomdz:
Some notes about this pull request:

It requires Maven 3 (we use 3.0.4) and Java 1.6 or newer (works fine with Java 1.7 which is what we use).

It includes two profiles for building for Hadoop 1 or Hadoop 2. One of these **has** to be chosen explicitly, there is no default:

```
$ mvn -Phadoop1 clean install
```

for Hadoop 1 (using 0.20.205.0), or

```
$ mvn -Phadoop2 clean install
```

for Hadoop 2 (using 2.0.0-mr1-cdh4.1.1).

It uses the [scala-maven-plugin](http://davidb.github.com/scala-maven-plugin/) which supports incremental and continuous compilation. E.g.

```
mvn -Phadoop2 scala:cc
```

should run continuous compilation (i.e. wait for changes). We have not tested this extensively however.

This setup works fine in IntelliJ IDEA 11.1.4. After opening the project via the `pom.xml` file in the project root folder, you only need to activate either the `hadoop1` or `hadoop2` profile in the "Maven Properties" popout. We have not tried Eclipse/Scala IDE with this.

It includes support for building a Debian package containing a 'fat-jar' which includes the repl, the examples and bagel. This can be created by specifying the `deb` profile:

```
$ mvn -Phadoop2,deb clean install
```

The debian package can then be found under `repl/target`. We added the short commit hash to the file name so that we can distinguish individual packages build for SNAPSHOT versions.

Tests are run via the [scalatest-maven-plugin](http://www.scalatest.org/user_guide/using_the_scalatest_maven_plugin). With this you can do things like:

```
$ mvn -DskipTests -Phadoop2 clean install
```

for skipping test execution (but not compilation), and

```
repl$ mvn -Phadoop2 -Dsuites=spark.repl.ReplSuite test
```

for running a specific test suite.

Unfortunately the Twirl bits only work as a SBT plugin, so I had to extract the API source code into a separate artifact (https://github.com/tomdz/twirl-api) and create a Maven plugin (https://github.com/tomdz/twirl-maven-plugin). Both of these are released to the Sonatype's OSS repository and should show up in Maven central within the next day or so. I also have an open issue with the Spray project about a shared twirl-api (and maybe twirl-compiler) project (https://github.com/spray/twirl/issues/8), but I haven't heard back from them yet.

Last but not least I included support for deploying to OSS Sonatype (which is the official way to get into the Maven central repository) as per https://docs.sonatype.org/display/Repository/Sonatype+OSS+Maven+Repository+Usage+Guide. More specifically, the `pom.xml` files should be setup for deployment, and only the JIRA & GPG steps in that document should be what's left to do in order to be able deploy Spark to the Sonatype's OSS maven repository (which is automatically sync'd to Maven central once a day or so) .


bagel/pom.xml
```diff
@@ -0,0 +1,100 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.spark-project</groupId>
+    <artifactId>parent</artifactId>
+    <version>0.6.1-SNAPSHOT</version>
```
massie: Please add...

```
<relativePath>../pom.xml</relativePath>
```

... in order to make the directory structure explicit.

bagel/pom.xml
```diff
@@ -0,0 +1,100 @@
...
+  <profiles>
+    <profile>
+      <id>hadoop1</id>
+      <dependencies>
+        <dependency>
+          <groupId>org.spark-project</groupId>
+          <artifactId>spark-core</artifactId>
+          <version>${project.version}</version>
+          <classifier>hadoop1</classifier>
+        </dependency>
```
massie, line 52: I like the use of classifiers here.

examples/pom.xml
```diff
@@ -0,0 +1,100 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.spark-project</groupId>
+    <artifactId>parent</artifactId>
+    <version>0.6.1-SNAPSHOT</version>
+  </parent>
```
massie, line 9: Please add a relativePath here...

```
<relativePath>../pom.xml</relativePath>
```

bagel/pom.xml
tomdz: I left it out because that is the default value for `relativePath` and the pom's are big enough as they are :)

massie: Ha!  Agree with you there but it is only a line.  If the relative path for this subproject ever changes, having an explicit path helps.

pom.xml
```diff
@@ -0,0 +1,511 @@
...
+        <groupId>com.github.scala-incubator.io</groupId>
+        <artifactId>scala-io-file_${scala.version}</artifactId>
+        <version>0.4.1</version>
+      </dependency>
+      <!-- In the root dir: mvn install:install-file -Dfile=core/lib/mesos-0.9.0.jar -DgroupId=org.apache.mesos -DartifactId=mesos -Dversion=0.9.0 -Dpackaging=jar -->
+      <dependency>
+        <groupId>org.apache.mesos</groupId>
+        <artifactId>mesos</artifactId>
+        <version>0.9.0</version>
+      </dependency>
```
massie, line 237: You can drop the comment and update the version to be 0.9.0-incubating in order to fetch the artifact from maven central (http://search.maven.org/#search|ga|1|mesos).   It also makes sense to create a mesos.version in the properties.

pom.xml
```diff
@@ -0,0 +1,511 @@
...
+      <dependency>
+        <groupId>cc.spray</groupId>
+        <artifactId>spray-server</artifactId>
+        <version>${spray.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.tomdz.twirl</groupId>
+        <artifactId>twirl-api</artifactId>
+        <version>1.0.2</version>
+      </dependency>
```
massie, line 227: When I run a `mvn clean install`, I get the following error:

```
[ERROR] Failed to execute goal org.tomdz.twirl:twirl-maven-plugin:1.0.0:generate (default) on project spark-core: Execution default of goal org.tomdz.twirl:twirl-maven-plugin:1.0.0:generate failed: Unable to load the mojo 'generate' in the plugin 'org.tomdz.twirl:twirl-maven-plugin:1.0.0' due to an API incompatibility: org.codehaus.plexus.component.repository.exception.ComponentLookupException: org/tomdz/twirl/maven/TwirlMojo : Unsupported major.minor version 51.0
```

I'm using JDK 6

```
$ java -version
  java version "1.6.0_37"
  Java(TM) SE Runtime Environment (build 1.6.0_37-b06)
  Java HotSpot(TM) 64-Bit Server VM (build 20.12-b01, mixed mode)
```

Does the 'twirl-api' artifact only work with JDK 7?

tomdz, line 227: Err, right, I didn't pay attention when I built that. I'll push a new version build with Java 6 (same for the maven plugin).

massie, line 227: Ping me when it's available.

I'm really impressed by your maven work and look forward to getting it into the master branch.

tomdz, line 227: Turns out that the `scala-maven-plugin` doesn't enforce the same classfile version for java code it compiles (in incremental mode) as for scala (scalac generates 1.5 classfiles currently, i.e. 49.0). This affects the twirl maven plugin (not the twirl-api artifact which is pure scala) and the spark project itself, so I'll update both accordingly.

repl/pom.xml
```diff
@@ -0,0 +1,291 @@
...
+                <configuration>
+                  <transformers>
+                    <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
+                    <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
+                      <mainClass>spark.repl.Main</mainClass>
+                    </transformer>
+                  </transformers>
+                </configuration>
+              </execution>
+            </executions>
```
massie, line 136: You're shading the jar here without relocating the classes.  Do you see any namespace issues if any of the dependent jars are on the classpath?

tomdz, line 136: This is the "fat-jar" for the repl specifically, which we use to run Spark as a repl (see the debian package below). I was assuming that nobody would depend on the repl "fat-jar" artifact as a library. This pom should generate a separate normal repl jar that people can depend on if they want.

massie, line 136: That makes perfect sense.  Thanks, Tom.

pom.xml
tomdz, line 227: I released a 1.0.2 version of the maven plugin to Sonatype's OSS repo which fixes this, should be in central shortly.

core/pom.xml
```diff
@@ -0,0 +1,259 @@
...
+        <executions>
+          <execution>
+            <phase>compile</phase>
+            <goals>
+              <goal>run</goal>
+            </goals>
+            <configuration>
+              <exportAntProperties>true</exportAntProperties>
+              <tasks>
+                <property name="spark.classpath" refid="maven.test.classpath"/>
```
massie, line 124: Please change this to...

```
          <tasks>
            <property name="spark.classpath" refid="maven.test.classpath"/>
            <property environment="env"/>
            <fail unless="env.SCALA_HOME" message="Please set the SCALA_HOME environment variable and retry."/>
          </tasks>
```

I was using the CentOS 6.3 scala package which doesn't set the SCALA_HOME env variable.  It took me a while to realize what was happening so I hope this simple  check saves others the trouble.

massie:
As far as I'm concerned, this work is ready to go.

I was able to run all tests in Maven using both the "hadoop1" and "hadoop2" profiles and build the REPL debian package.

Thanks for the contribution, Tom.  Great stuff!  Really well done.

core/pom.xml
tomdz, line 124: I'll add it though I'll make a slight tweak since for CentOS (and Ubuntu/Debian which has the same problem), `SCALA_LIBRARY_PATH` would be enough since #292 was merged.

tomdz, line 124: Added.

mateiz:
Thanks Tom for this great effort! I'm excited to have this in.

mateiz:
By the way, Tom, have you tried this with IntelliJ IDEA 12? It imports the project for me but it doesn't seem to respect the hadoop1 profile (it doesn't add the hadoop1 folder to the source path). It does, however, build fine if I add that manually.

tomdz:
@mateiz: I haven't tried IDEA 12 (haven't upgraded yet), but in IDEA 11 you'll need to activate the profile in a view that shows all profiles in the pom, on the right sidebar.

PR #311, opened 2012-11-28
Fix NullPointerException when map output unregistered from MapOutputTracker twice

charlesreiss:
Apparently, if two fetch failed messages are in flight (for the same map output), DAGScheduler will call unregisterMapOutput twice for that output. MapOutputTracker will currently give a NullPointerException in that case, killing the DAGScheduler thread and causing spark to hang. This patch fixes the NPE and adds tests for MapOutputTracker (incl. one which triggers this bug).


mateiz:
Hey Charles, I'm just going to check with Josh about whether this is related to a thing he had fixed for Shark. I think it might not be though.

mateiz:
Actually looks like Josh also fixed it in his branch, but I'll merge this.

PR #312, opened 2012-11-28
For size compression, compress non zero values into non zero values.

rxin:
This is important because we want to differentiate 1 from 0. If the size of a block is 0, the shuffle fetcher can safely skip it.


mateiz:
Looks good.

PR #315, opened 2012-11-30
Counters

squito:
This is a first pass at addressing https://spark-project.atlassian.net/browse/SPARK-603
I don't think this is ready to be merged in, but ready for a discussion of api & implementation (it works, at least)

Two major changes here:
1) Every RDD can set a "completion function" -- a function that will be run whenever the RDD is actually generated.  When execution actually happens, the completion function is run for _all_ RDDs that were executed (eg., 3 maps in a row can specify 3 completion functions, all are executed).  By default, the value of all accumulators are printed out.
2) "Counters" are a simple accumulator that are automatically created for every job that is run, and that can be easily incremented.

I don't really like the name "completionFunction", but not sure what else to call it.  Not really sure what will happen when if task is killed and gets re-run -- I think it will only get called once, but I'm not positive.

Getting Counters to work was much harder than I expected.  b/c the closure doesn't contain a reference to the accumulator, they aren't serialized via normal accumulator serialization, so required special logic.


mateiz:
Hey Imran, sorry for the delay in getting to this, but I've finally had a chance to look at it. Just to understand, is the purpose of this just debugging (to print out the accumulators after each stage)?

The other thing this change would enable, which I actually like, is to print the accumulator updates for each task only once. Right now, updates that happen as part of a transformation (e.g. map) will happen again if that map gets re-run, which is problematic. Do you want to add that as well?

Anyway, I'd be glad to talk sometime about how to best accomplish this, and what the semantic should be (does each counter run just once, or many times).

core/src/main/scala/spark/scheduler/Task.scala
```diff
@@ -68,12 +72,15 @@ private[spark] object Task {
    *
    * @return (taskFiles, taskJars, taskBytes)
    */
-  def deserializeWithDependencies(serializedTask: ByteBuffer)
+  def deserializeWithDependencies(serializedTask: ByteBuffer, serializer: SerializerInstance)
```
mateiz, line 75: If I recall correctly, passing the Serializer here was dangerous, because the task might contains some JARs it wants to add to the classpath that are necessary for deserialization, added using SparkContext.addJAR. For example, say that the user adds a JAR that contains class Foo, and then uses a Foo in the task; we need to add the JAR onto the executor's classpath _before_ deserializing the task body, because that body will use Foo. I guess this might not be a problem for Counters, since it is part of Spark, but it would be nicer to deserialize them after, as part of the task, instead of here.

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -395,10 +396,10 @@ class DAGScheduler(taskSched: TaskScheduler) extends TaskSchedulerListener with
     val task = event.task
     val stage = idToStage(task.stageId)
     event.reason match {
-      case Success =>  
+      case Success =>
         logInfo("Completed " + task)
-        if (event.accumUpdates != null) {
-          Accumulators.add(event.accumUpdates) // TODO: do this only if task wasn't resubmitted
+        if (stage.rdd.executionCount == 0 && event.accumUpdates != null) {
```
squito, line 402: I think this change already makes sure that accumulator updates only happen once -- if the task is re-run, the executionCount will be > 0.

But now that I think about this more, I don't know if that will work if the final rdd hasn't been executed before, but a dependent one has.  probably need a test for that.  Though, if this doesn't do the trick, I'm not sure how to fix it ...

squito:
yeah, the purpose is debugging (certainly doesn't enable any computation you couldn't do before), but debugging in the most general sense.  Its really nice to add some basic sanity checks to your code, to make sure things work as you expected, even if you don't know of any bugs.  I'd like that to be as painless as possible.

Also, its sometimes useful to use historical values of these stats later on.  eg., plot a graph of how many records failed to parse over time, so you can find anomalies.  I've seen Hadoop setups where the counters get logged to a DB for this reason -- that's why I added the ability for the user to define their own "completion function".

Another use case is "system" counters for profiling -- how long was spent in serialization, number of bytes of data & task io, histogram of task run times, etc. A lot of that info is available in the logs now, but a nice clean summary of it at the end of every stage would make it a lot easier for users to profile & improve performance.

squito:
this is from pretty far back, so I'll abandon it.
still would be great if we got something like this in the future though

PR #316, opened 2012-12-02
Use external addresses in standalone web UI when running on EC2

JoshRosen:
This patch modifies the standalone cluster launch scripts to detect when they are run on EC2 so that links in the web UI use machines' externally-accessible addresses. This is a first pass at fixing [SPARK-613](https://spark-project.atlassian.net/browse/SPARK-613); I'm submitting it here for review.

This solution is modeled on the `MESOS_PUBLIC_DNS` environment variable used in the Mesos EC2 scripts.

In Spark's scripts, I pattern-match the machine's hostname to determine whether it is running on EC2 before attempting to resolve its public address.  Is there a better way to do this?  I don't want to use a method that relies on contacting an external service, since this might slow or break the script for users who run standalone clusters that do not have internet access.  Relying on the hostname is fine for clusters created with our EC2 scripts, but custom EC2 clusters might use a different hostname scheme.

I used a `SPARK_PUBLIC_DNS` environment variable to set each slave's public address; should this be a command-line option to the standalone worker instead?

### DIFF
--- a/bin/start-slave.sh
+++ b/bin/start-slave.sh
@@ -1,13 +1,9 @@
 #!/usr/bin/env bash
 
-# Starts the master on the machine this script is executed on.
-
 bin=`dirname "$0"`
 bin=`cd "$bin"; pwd`
 
-. "$bin/spark-config.sh"
-
-# Set SPARK_PUBLIC_DNS so the master report the correct webUI address to the slaves
+# Set SPARK_PUBLIC_DNS so slaves can be linked in master web UI
 if [ "$SPARK_PUBLIC_DNS" = "" ]; then
     # If we appear to be running on EC2, use the public address by default:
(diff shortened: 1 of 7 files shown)


mateiz:
The SPARK_PUBLIC_DNS thing seems fine as is. For custom EC2 clusters with different hostnames, people could set it in spark-env.sh (e.g. export <code>SPARK_PUBLIC_DNS=`wget ...`</code>) so I think that's fine. Thanks for adding this!

JoshRosen:
It looks like there are addresses of the form `domU-12-31-39-09-F5-02.compute-1.internal` which aren't recognized as EC2 machines by the script.  I could add another regex to match these, but this approach seems brittle.  Any suggestions?

PR #317, opened 2012-12-06
Implement block manager heartbeat

charlesreiss:
Implement heart beats for the BlockManagers. BlockManagers periodically send HeartBeat messages to the master, and the master periodically deregisters slaves which have not sent a heart beat in about 3 heart beat periods. The BlockManagerMaster will now indicate when a BlockManager updates a block or sends a heart beat but is not registered. In this case, the BlockManager will reregister and resend block updates for all its blocks.

Other changes made along the way:
- deregistering a BlockManager and forgetting about its blocks is implemented in BlockManagerMaster;
- the old "HeartBeat" message is renamed to "BlockUpdate";
- the BlockManagerMaster tracks the BlockManagerId for each hostname. If a new BlockManagerMaster registers for a hostname, the old one is deregistered. This tracking incidentally fixes a bug where notifyADeadHost() would try to remove a BlockManagerId that usually wouldn't exist;
- local-cluster[...] mode uses distinct hostnames -- 127.100.0.(slave num) (which should be a valid name for localhost like all addresses in 127/8), which permits BlockManager to behave correctly in spite of multiple executors on the local node;

### DIFF
--- a/core/src/main/scala/spark/storage/BlockManager.scala
+++ b/core/src/main/scala/spark/storage/BlockManager.scala
@@ -1,7 +1,9 @@
 package spark.storage
 
+import akka.actor.{ActorSystem, Cancellable}
 import akka.dispatch.{Await, Future}
 import akka.util.Duration
+import akka.util.duration._
 
 import it.unimi.dsi.fastutil.io.FastByteArrayOutputStream
 
@@ -12,7 +14,7 @@ import java.util.concurrent.{ConcurrentHashMap, LinkedBlockingQueue}
 import scala.collection.mutable.{ArrayBuffer, HashMap, HashSet, Queue}
 import scala.collection.JavaConversions._
 
-import spark.{CacheTracker, Logging, SizeEstimator, SparkException, Utils}
+import spark.{CacheTracker, Logging, SizeEstimator, SparkEnv, SparkException, Utils}
 import spark.network._
 import spark.serializer.Serializer
 import spark.util.ByteBufferInputStream
@@ -45,13 +47,13 @@ private[spark] class BlockManagerId(var ip: String, var port: Int) extends Exter
   }
 }
 
-
 private[spark] 
 case class BlockException(blockId: String, message: String, ex: Exception = null)
 extends Exception(message)
 
 private[spark]
-class BlockManager(val master: BlockManagerMaster, val serializer: Serializer, maxMemory: Long)
+class BlockManager(actorSystem: ActorSystem, val master: BlockManagerMaster,
+                   val serializer: Serializer, maxMemory: Long)
   extends Logging {
 
   class BlockInfo(val level: StorageLevel, val tellMaster: Boolean) {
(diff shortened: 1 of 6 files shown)


core/src/main/scala/spark/storage/BlockManager.scala
```diff
@@ -104,8 +104,33 @@ class BlockManager(val master: BlockManagerMaster, val serializer: Serializer, m
...
 
+  @volatile private var shuttingDown = false
+
+  private def heartBeat() {
+    if (!master.mustHeartBeat(HeartBeat(blockManagerId))) {
+      reregister()
+    }
+  }
+
+  val heartBeatThread = new Thread("BlockManager heartbeat") {
```
mateiz: This could probably be an Akka actor instead, to save a thread; did you look into that or was there a reason not to do it?

core/src/main/scala/spark/storage/BlockManager.scala
```diff
@@ -855,6 +931,12 @@ object BlockManager extends Logging {
     (Runtime.getRuntime.maxMemory * memoryFraction).toLong
   }
 
+  def getHeartBeatFrequencyFromSystemProperties: Long =
+    System.getProperty("spark.storage.blockManagerHeartBeatMs", "2000").toLong
```
mateiz: I would increase this to 5000 to be conservative; sometimes GC can take a long time

mateiz:
This looks pretty good! I had a few small comments but otherwise it seems fine. Thanks for putting this together.

core/src/main/scala/spark/storage/BlockManager.scala
charlesreiss: Done.

core/src/main/scala/spark/storage/BlockManager.scala
charlesreiss: Done (unfortunately passing actorSystem to the BlockManager constructor in a bunch places); I didn't do it originally because I wasn't very familiar with Akka.

mateiz:
Great, thanks a lot.

core/src/main/scala/spark/storage/BlockManagerMaster.scala
```diff
@@ -436,27 +524,49 @@ private[spark] class BlockManagerMaster(actorSystem: ActorSystem, isMaster: Bool
     }
   }
 
-  def mustHeartBeat(msg: HeartBeat) {
-    while (! syncHeartBeat(msg)) {
-      logWarning("Failed to send heartbeat" + msg)
+  def mustHeartBeat(msg: HeartBeat): Boolean = {
+    var res = syncHeartBeat(msg)
+    while (!res.isDefined) {
```
JoshRosen, line 529: It looks like `mustHeartBeat` becomes stuck in an infinite loop if the call to `syncHeartBeat` fails and returns None.  Perhaps this is missing a `res = syncHeartBeat(msg)` line inside the loop?

charlesreiss, line 529: Yes; can't believe I missed that.  Incidentally pull request #327 fixes this.

PR #318, opened 2012-12-10
Minor tweaks to the debian build

tomdz:
- Use the correct name for the the package file according to the debian rules.
- Don't attach the debian package to the maven build.


mateiz:
Thanks Thomas!

PR #321, opened 2012-12-11
Fix FailureSuite failure (expecting the right exception type)

mbautin:
FailureSuite fails because it expects the wrong exception type.


mbautin:
Actually I'm not yet sure if this fixes the correct issue. Will debug a bit more.

mateiz:
What was the issue? The test works for me.

Sometimes there are some nondeterministic failures in the tests due to GC causing a timeout somewhere. If that's what you saw then the right thing is to find where that was and increase the timeout.

mbautin:
Yes, I think it was a nondeterministic failure. The test is passing now. Closing this for now.

PR #323, opened 2012-12-11
SPARK-626: Remove security groups properly

pas256:
SPARK-626: Remove rules before removing security groups, with a pause in between so wait for AWS eventual consistency to catch up.

### DIFF
--- a/ec2/spark_ec2.py
+++ b/ec2/spark_ec2.py
@@ -30,6 +30,7 @@ import time
 import urllib2
 from optparse import OptionParser
 from sys import stderr
+import boto
 from boto.ec2.blockdevicemapping import BlockDeviceMapping, EBSBlockDeviceType
 from boto import ec2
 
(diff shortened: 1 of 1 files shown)


mateiz:
In that case, can you make deleting the groups an option, and make it off by default? I don't want all users to wait 30 seconds to turn off the cluster. The ones that want this feature can add in the option.

pas256:
Yes I can.

pas256:
It turns out this is far more error prone than I first though. There are multiple dependencies in play:
- dependencies between rules in security groups
- instances may not have been terminated completely when trying to delete a security group
- AWS back-end eventual consistency
- group.revoke() returns True even when it fails

The unfortunate result is the code is somewhat messy. This is as clean as I can make - there are 3 retries to delete the all groups, and exceptions are caught.

By default, groups will not be deleted, so most people won't experience this. You can delete groups when destroying a cluster by adding `--delete-groups`

mateiz:
Okay, thanks a lot for adding this! Looks good.

PR #325, opened 2012-12-11
Mark hadoop dependencies provided in all library artifacts

tomdz:
The only exception is the repl, which I'm assuming won't be used as a library for now.


mateiz:
Great, thanks.

PR #326, opened 2012-12-12
Call slaveLost on executor death for standalone clusters.

charlesreiss:
In standalone mode, currently the Spark deploy cluster client receives messages about executor death, but these are not forwarded to the Spark scheduler. This patch calls slaveLost on the ClusterScheduler on executor death (most likely from a crash), which mirrors the behavior of the Mesos scheduler backend.


mateiz:
Closing this because it was incorporated as part of #329.

PR #327, opened 2012-12-13
Added the ability in block manager to remove blocks.

rxin:
As part of doing this, I also updated BlockManagerMasterActor's BlockManagerInfo class to track the size of each blocks. This will make the block manager memory dashboard simpler (some changes are pulled in from there).

I can take over the dashboard implementation if Denny can't finish it before he leaves.


JoshRosen:
Calling `removeBlock()` can cause a lot of warning messages to be printed to the log if blocks are not in the MemoryStore:

```
12/12/13 14:37:34 INFO BlockManager: Removing block shuffle_7_1_9
12/12/13 14:37:34 WARN MemoryStore: Block shuffle_7_1_9 could not be removed as it does not exist
```

The DiskStore issues no warnings when removing nonexistent blocks.

For consistency, we should either remove the warning from MemoryStore or add it to DiskStore.

A better approach would be to change `BlockStore.remove()` to return a boolean indicating success or failure, which would allow us to issue warnings only if the block cannot be found in either store.

rxin:
Ok the changes just got bigger. I merged TD's changes from streaming manually into this too.

With this commit, we should be able to replace the storage folder in streaming (radlab/dev) with this. 

@mateiz  and @woggling and @tdas, can you guys take a look at this too? Thanks.

rxin:
Testing information:

sbt/sbt test passes

./run spark.storage.ThreadTesting passes too

core/src/main/scala/spark/util/GenerationIdUtil.scala
```diff
@@ -0,0 +1,19 @@
+package spark.util
+
+import java.util.concurrent.atomic.AtomicInteger
+
```
mateiz: This should be documented. Also, the name "generation ID" is confusing here because it is the same as the "generation" for map output locations (which workers use to tell when to invalidate their cache of map output locations because some nodes have failed). I would just call this object IdGenerator. Also, I would place an IdGenerator instance on object BlockManager instead of having one "util" class where all the possible sequences of IDs are generated.

core/src/main/scala/spark/util/MetadataCleaner.scala
```diff
@@ -0,0 +1,35 @@
+package spark.util
+
+import java.util.concurrent.{TimeUnit, ScheduledFuture, Executors}
+import java.util.{TimerTask, Timer}
+import spark.Logging
+
```
mateiz, line 6: Will this prevent the JVM from exiting until cancelled? Or does Java use daemon threads for TimerTask? Basically it would be nice if programs will still exit even if they don't call SparkContext.stop(). Try that out, and if it doesn't work, switch this to using a DaemonThread with a sleep loop.

core/src/main/scala/spark/util/TimeStampedHashMap.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  override def foreach[U](f: ((A, B)) => U): Unit = {
+    val iterator = internalMap.entrySet().iterator()
+    while(iterator.hasNext) {
+      val entry = iterator.next()
+      val kv = (entry.getKey, entry.getValue._1)
+      f(kv)
+    }
+  }
+
```
mateiz, line 74: Add a comment explaining threshTime. I'd also be okay renaming this to cleanOlderValues, but maybe after we merge streaming.

core/src/main/scala/spark/storage/BlockManager.scala
```diff
@@ -865,6 +846,52 @@ class BlockManager(actorSystem: ActorSystem, val master: BlockManagerMaster,
     }
   }
 
+  /**
+   * Remove a block from both memory and disk. This one doesn't report to the master
+   * because it expects the master to initiate the original block removal command, and
+   * then the master can update the block tracking itself.
```
mateiz: Is there any reason why you made the master remove the entry right away instead of waiting for a message from the worker? The latter seems simpler and safer unless the master will immediately need to check whether it's removed some particular dataset.

core/src/main/scala/spark/storage/BlockManager.scala
```diff
@@ -865,6 +846,52 @@ class BlockManager(actorSystem: ActorSystem, val master: BlockManagerMaster,
...
+      val removedFromMemory = memoryStore.remove(blockId)
+      val removedFromDisk = diskStore.remove(blockId)
+      if (!removedFromMemory && !removedFromDisk) {
+        logWarning("Block " + blockId + " could not be removed as it was not found in either " +
+          "the disk or memory store")
+      }
+      blockInfo.remove(blockId)
+    } else {
+      // The block has already been removed; do nothing.
+      logWarning("Block " + blockId + " does not exist.")
```
mateiz: Maybe say "Asked to remove block [blockId], which does not exist", so users don't just see an unexplained "block does not exist" message.

core/src/main/scala/spark/storage/BlockManagerMasterActor.scala
```diff
@@ -0,0 +1,415 @@
...
+          logInfo("Got another registration for host " + blockManagerId)
+          managers += blockManagerId
+        case None =>
+          blockManagerIdByHost += (blockManagerId.ip -> ArrayBuffer(blockManagerId))
+      }
+
+      blockManagerInfo += (blockManagerId -> new BlockManagerMasterActor.BlockManagerInfo(
+        blockManagerId, System.currentTimeMillis(), maxMemSize, slaveActor))
+    }
+    logDebug("Got in register 1" + tmp + Utils.getUsedTimeMs(startTimeMs))
```
mateiz: Probably good to remove these timing messages while you're at it

core/src/main/scala/spark/storage/BlockManagerMaster.scala
```diff
@@ -1,676 +1,182 @@
...
-      }
-      res += peers(index % peers.size)
-    }
-    sender ! res.toSeq
+  /**
+   * Remove a block from the slaves that have it. This can only be used to remove
+   * blocks that the master knows about.
+   */
+  def removeBlock(blockId: String) {
+    askMaster(RemoveBlock(blockId))
```
mateiz: This seems to be the only use of askMaster instead of askMasterWithRetry. Should we just do withRetry every time? TBH I'm not sure Akka could even fail on one try and then succeed on the next one but I guess it's worth it to retry.

core/src/main/scala/spark/storage/BlockManagerMaster.scala
```diff
@@ -1,676 +1,182 @@
...
-      if (info.lastSeenMs < minSeenTime) {
-        logWarning("Removing BlockManager " + info.blockManagerId + " with no recent heart beats")
-        toRemove += info.blockManagerId
-      }
-    }
-    // TODO: Remove corresponding block infos
-    toRemove.foreach(removeBlockManager)
-  }
+  val AKKA_RETRY_ATTEMPS: Int = System.getProperty("spark.akka.num.retries", "5").toInt
+  val AKKA_RETRY_INTERVAL_MS: Int = System.getProperty("spark.akka.retry.wait", "100").toInt
```
mateiz: This is way too small, needs to be at least 3000. Otherwise GC on the master can easily cause it to fail.

core/src/main/scala/spark/storage/BlockManagerMaster.scala
```diff
@@ -1,676 +1,182 @@
...
-    for (info <- blockManagerInfo.values) {
-      if (info.lastSeenMs < minSeenTime) {
-        logWarning("Removing BlockManager " + info.blockManagerId + " with no recent heart beats")
-        toRemove += info.blockManagerId
-      }
-    }
-    // TODO: Remove corresponding block infos
-    toRemove.foreach(removeBlockManager)
-  }
+  val AKKA_RETRY_ATTEMPS: Int = System.getProperty("spark.akka.num.retries", "5").toInt
```
mateiz: I'd make this only 3.

mateiz:
Hey Reynold, this looks pretty good overall. It's great to see the block manager code becoming cleaner! I made a few comments in the code, mainly on how to report when the block has gone away (I'd prefer to keep the "end-to-end" way of doing it, which is to have the workers send an update, to minimize the number of code paths that can be taken). I'll let you know if I have any other comments.

rxin:
I updated the changes based on your comment. Take another look.

Note that TimeStampedHashMap.scala and MetadataCleaner.scala are based on code from @tdas. Probably make more sense to merge this one first, and ask @tdas to update those two files.

mateiz:
Alright, looks good. Thanks!

PR #328, opened 2012-12-13
SPARK-635: Pass a TaskContext object to compute() interface and use that to close Hadoop input stream.

rxin:


core/src/main/scala/spark/rdd/HadoopRDD.scala
```diff
@@ -77,6 +74,9 @@ class HadoopRDD[K, V](
     val fmt = createInputFormat(conf)
     reader = fmt.getRecordReader(split.inputSplit.value, conf, Reporter.NULL)
 
+    // Register an on-task-completion callback to close the input stream.
+    taskContext.registerOnCompleteCallback(Unit => reader.close())
```
mateiz: I think the more idiomatic way to write this is () => reader.close(), or _ => reader.close()

rxin:
Updated - PTAL.

mateiz:
Awesome, thanks!

JoshRosen:
This appears to have broken one of the Java API tests:

```
[error] /Users/joshrosen/Documents/spark/spark/core/src/test/scala/spark/JavaAPISuite.java:386: iterator(spark.Split,spark.TaskContext) in spark.api.java.JavaRDD<java.lang.Integer> cannot be applied to (spark.Split)
[error]     Assert.assertEquals(1, rdd.iterator(rdd.splits().get(0)).next().intValue());
[error]                               ^
[error] Note: /Users/joshrosen/Documents/spark/spark/core/src/test/scala/spark/JavaAPISuite.java uses unchecked or unsafe operations.
[error] Note: Recompile with -Xlint:unchecked for details.
[error] 1 error
```

PR #329, opened 2012-12-14
Executor exit status codes

charlesreiss:
[Note: This incorporates pull request #326.]

Currently, Mesos and the Spark deploy mode provide executor exit statuses, which could diagnose some executor crashes. This patch assigns exit codes for deliberate executor exits (e.g. when an executor can't create a temporary directory; or an OutOfMemoryError in a non-task thread) and logs them from the master at the error level when executor death with an exit status is detected.

### DIFF
--- a/core/src/main/scala/spark/executor/ExecutorExitCode.scala
+++ b/core/src/main/scala/spark/executor/ExecutorExitCode.scala
@@ -1,0 +1,40 @@
+package spark.executor
+
+/**
+ * These are exit codes that executors should use to provide the master with information about
+ * executor failures assuming that cluster management framework can capture the exit codes (but
+ * perhaps not log files). The exit code constants here are chosen to be unlikely to conflict
+ * with "natural" exit statuses that may be caused by the JVM or user code. In particular,
+ * exit codes 128+ arise on some Unix-likes as a result of signals, and it appears that the
+ * OpenJDK JVM may use exit code 1 in some of its own "last chance" code.
+ */
+private[spark]
+object ExecutorExitCode {
+  /** The default uncaught exception handler was reached. */
+  val UNCAUGHT_EXCEPTION = 50
+  /** The default uncaught exception handler was called and an exception was encountered while
+      logging the exception. */
+  val UNCAUGHT_EXCEPTION_TWICE = 51
+  /** The default uncaught exception handler was reached, and the uncaught exception was an 
+      OutOfMemoryError. */
+  val OOM = 52
+  /** DiskStore failed to create a local temporary directory after many attempts. */
+  val DISK_STORE_FAILED_TO_CREATE_DIR = 53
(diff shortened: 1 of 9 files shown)


mateiz:
This looks awesome, thanks! Just a few minor formatting things:
- Rename ExecutorLostReason to ExecutorLossReason (to make the first part a noun, kind of like TaskEndReason)
- One spacing fix I will comment on in the code.

core/src/main/scala/spark/executor/ExecutorExitCode.scala
```diff
@@ -0,0 +1,40 @@
...
+ * executor failures assuming that cluster management framework can capture the exit codes (but
+ * perhaps not log files). The exit code constants here are chosen to be unlikely to conflict
+ * with "natural" exit statuses that may be caused by the JVM or user code. In particular,
+ * exit codes 128+ arise on some Unix-likes as a result of signals, and it appears that the
+ * OpenJDK JVM may use exit code 1 in some of its own "last chance" code.
+ */
+private[spark]
+object ExecutorExitCode {
+  /** The default uncaught exception handler was reached. */
+  val UNCAUGHT_EXCEPTION = 50
```
mateiz, line 14: Add some newlines after these values to make them easier to read

core/src/main/scala/spark/scheduler/cluster/SparkDeploySchedulerBackend.scala
```diff
@@ -65,9 +66,27 @@ private[spark] class SparkDeploySchedulerBackend(
...
   def executorAdded(id: String, workerId: String, host: String, cores: Int, memory: Int) {
+    executorIdToSlaveId += id -> workerId
     logInfo("Granted executor ID %s on host %s with %d cores, %s RAM".format(
        id, host, cores, Utils.memoryMegabytesToString(memory)))
   }
 
-  def executorRemoved(id: String, message: String) {}
+  def executorRemoved(id: String, message: String) {
+    var reason: ExecutorLostReason = SlaveLost(message)
+    if (message.startsWith("Command exited with code ")) {
```
mateiz, line 76: Actually it might be good to modify the standalone deploy client to pass an exit code. Maybe in a later pull request.

core/src/main/scala/spark/executor/ExecutorExitCode.scala
charlesreiss, line 14: Done.

mateiz:
Alright, looks good. Thanks!

It looks like GitHub lost a line comment that I made, but one other thing that it would be nice to fix is the message.substring("Command exited with code ".length).toInt. We should make the standalone cluster pass the code as a separate parameter. I've opened https://spark-project.atlassian.net/browse/SPARK-639 to report this.

PR #330, opened 2012-12-14
Use spark-env.sh to configure standalone master

JoshRosen:
The standalone cluster's `start-master.sh` script does not read `spark-env.sh` and uses different defaults than `start-slaves.sh`.  This can cause problems because the slaves may connect to the master using the wrong address (see [SPARK-638](https://spark-project.atlassian.net/browse/SPARK-638)).

This commit fixes this issue by modifying the master startup script to configuration options from `spark-env.sh` and use proper defaults.

I also fixed a typo in the documentation and removed some debug output that was accidentally included in an earlier commit.

### DIFF
--- a/bin/start-all.sh
+++ b/bin/start-all.sh
@@ -11,7 +11,7 @@ bin=`cd "$bin"; pwd`
 . "$bin/spark-config.sh"
 
 # Start Master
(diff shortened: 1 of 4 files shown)


mateiz:
Thanks Josh.

PR #331, opened 2012-12-14
Have standalone cluster report exit codes to clients

charlesreiss:
Add explicit exit status field to executor state update messages, and argument to executorRemoved callback in standalone cluster. Addresses SPARK-639.

### DIFF
--- a/core/src/main/scala/spark/scheduler/cluster/SparkDeploySchedulerBackend.scala
+++ b/core/src/main/scala/spark/scheduler/cluster/SparkDeploySchedulerBackend.scala
@@ -72,7 +72,10 @@ private[spark] class SparkDeploySchedulerBackend(
   }
 
   def executorRemoved(id: String, message: String, exitStatus: Option[Int]) {
-    var reason: ExecutorLossReason = exitStatus.map(ExecutorExited).getOrElse(SlaveLost(message))
+    val reason: ExecutorLossReason = exitStatus match {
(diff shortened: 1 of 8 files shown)


core/src/main/scala/spark/scheduler/cluster/SparkDeploySchedulerBackend.scala
```diff
@@ -71,15 +71,8 @@ private[spark] class SparkDeploySchedulerBackend(
...
-    var reason: ExecutorLossReason = SlaveLost(message)
-    if (message.startsWith("Command exited with code ")) {
-      try {
-        reason = ExecutorExited(message.substring("Command exited with code ".length).toInt)
-      } catch {
-        case nfe: NumberFormatException => {}
-      }
-    }
+  def executorRemoved(id: String, message: String, exitStatus: Option[Int]) {
+    var reason: ExecutorLossReason = exitStatus.map(ExecutorExited).getOrElse(SlaveLost(message))
```
mateiz: Kind of funky-looking -- I would do

```
exitStatus match {
  case Some(code) => ExecutorExited(code)
  case None => SlaveLost(message)
}
```

charlesreiss: Changed.

mateiz:
Great, thanks!

PR #332, opened 2012-12-14
Add try-finally to handle MapOutputTracker timeouts

JoshRosen:
This is a fix for hangs caused by timeouts when contacting the MapOutputTracker; see [SPARK-607](https://spark-project.atlassian.net/browse/SPARK-607) for more details.

On a local standalone cluster, it successfully prevented the hangs created by the test case attached to that issue; the failed tasks were retried and ran to completion.


mateiz:
Looks good, thanks Josh!

PR #333, opened 2012-12-14
Fixed the broken Java unit test from SPARK-635.

rxin:


mateiz:
Thanks for fixing this.

PR #336, opened 2012-12-18
Fix issues with the branch to allow it to work on non trivial YARN clusters

mridulm:
This is an export of our private svn to github which fixes and resolves issues we faced while running spark on a hadoop yarn cluster.
Lot of the changes are intertwined, and so not split into different patches.

1) Use host:port instead of host in the codebase everywhere (patch use of preferredLocations to revert to host)
Since multiple containers can be started on same node, using only "host" fails.

2) Handle container allocation and maintenance - re-request containers which were lost, release more containers than required, etc.

3) Handle data locality at allocation time (not task dispatch time) - that is, request for containers base of \* : so that the subsequent launch of tasks (based on preferredLocations) can be dispatched to nodes which are data-local.

4) Add support for rack locality of task (at allocation and dispatch time).

5) Move to using host everywhere instead of ip. iirc hadoop expects this anyway - this is probably not required, but otherwise we end up with conflicts between IP vs host at various places where we compare.
So consistently using hostname solves that issue : though I did see a bugfix from you regarding picking right IP - so we might have issues to resolve in case of multi-honed machines, etc.
I dont have a good solution for this problem : atleast current solution works for our clusters fine !

Simple and/or Localized changes -

6) Run the yarn job as the user launching the job - acl's, hadoop base directory, etc depend on this.
Note that I am yet to support kerberos authorization/tokens, etc - this will need to be added later.

7) Support long runing tasks - essentially periodically report progress to yarn : else it kills job after idle timeout (10 mins usually).

8) There are classpath issues which needed to be fixed.

9) Fix file handle leak

10) Changes to arg parsing (so that multiple arg's can be passed to the job - via repeated --args : trying to tokenize and parse was causing issues.

11) Add support for Queue

12) Do a best case effort to stop the spark context and notify yarn about application exit.
TODO: Propagate actual application success/failure to yarn.

Please note that we are not very familiar with scala (or spark for that matter) - so any and all suggestions to make this patch better are welcome !


mridulm:
Please note that we are still working on task scheduling part : so that will change in a later pull request.

vinoduec:
Any case of changes in YARN itself that you needed?

mridulm:
You mean to hadoop yarn codebase ? No.
Though we plan to initiate a discussion on issues with running application masters on commodity hardware.

core/src/main/scala/spark/ClosureCleaner.scala
```diff
@@ -8,12 +8,20 @@ import scala.collection.mutable.Set
...
-      cls.getName.replaceFirst("^.*\\.", "") + ".class"))
+    // Copy data over, before delegating to ClassReader - else we can run out of open file handles.
+    val className = cls.getName.replaceFirst("^.*\\.", "") + ".class"
+    val resourceStream = cls.getResourceAsStream(className)
+    // todo: Fixme - continuing with earlier behavior ...
+    if (null == resourceStream) return new ClassReader(resourceStream)
+
+    val baos : ByteArrayOutputStream = new ByteArrayOutputStream(128)
+    Utils.copyStream(resourceStream, baos, true)
+    new ClassReader(new ByteArrayInputStream(baos.toByteArray))
```
mateiz, line 24: Hmm, isn't it possible to just call close() on the reader at a later point? I guess this works but it's a little hacky.

core/src/main/scala/spark/scheduler/TaskScheduler.scala
```diff
@@ -10,6 +10,10 @@ package spark.scheduler
 private[spark] trait TaskScheduler {
   def start(): Unit
 
+  // Invoked after system has successfully initialized (typically in spark context).
+  // Yarn uses this to bootstrap allocation of resources based on preferred locations, wait for slave registerations, etc.
+  def postStartHook(): Unit
```
mateiz: You can actually add an empty implementation here to be the default. Scala's traits are not pure interfaces -- they can have implementations for some methods. So just

```
def postStartHook() {}
```

And now only YarnScheduler has to override it.

core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
```diff
@@ -117,20 +120,60 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
...
       val availableCpus = offers.map(o => o.cores).toArray
       var launchedTask = false
+
+
       for (manager <- activeTaskSetsQueue.sortBy(m => (m.taskSet.priority, m.taskSet.stageId))) {
+
+        // Split offers based on host local, rack local and off-rack tasks.
+	val hostLocalOffers : HashMap[String, ArrayBuffer[Int]] = new HashMap[String, ArrayBuffer[Int]]()
+	val rackLocalOffers : HashMap[String, ArrayBuffer[Int]] = new HashMap[String, ArrayBuffer[Int]]()
+	val otherOffers : HashMap[String, ArrayBuffer[Int]] = new HashMap[String, ArrayBuffer[Int]]()
```
mateiz: I think you might have some tabs here.. please just use two spaces. Also, in terms of code style:
- Don't put a space before ":"
- Don't even put types in situations like this, where the type is clear from the initializer; same thing for local variables inside a method
- Keep lines below 100 characters (these may be fine but it's hard to tell)

We've been trying to use the Scala style guide at http://docs.scala-lang.org/style/ with some small tweaks (http://spark-project.org/docs/latest/contributing-to-spark.html).

core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
```diff
@@ -117,20 +120,60 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
...
+
+        for (i <- 0 until offers.size) {
+          val hostPort = offers(i).hostPort
+          // DEBUG code
+          Utils.checkHostPort(hostPort)
```
mateiz, line 140: Ditto on the weird indentation / possible tabs

mateiz:
Hey Mridul, I've started reviewing this, and I think the changes are great, but it will definitely take a while to review. One thing I noticed is that there's still some debug code in -- can you remove that when you're done? I also commented on the code style and some Scala stuff in a few places. But I'm going to keep looking at it in more detail.

The other tricky thing will be getting this back into the master branch, but I guess we can do 0.6 first. To get it into master, you'll probably end up having to produce one big patch (all your commits on this branch) that we can then try to apply there. There are some changes to the BlockManager there, but hopefully it will not be too bad to merge.

core/src/main/scala/spark/SparkEnv.scala
```diff
@@ -77,6 +77,16 @@ object SparkEnv extends Logging {
...
     }
 
+    // set only if unset until now.
+    if (null == System.getProperty("spark.hostname", null)) {
+      if (!isMaster){
+        // unexpected
+        Utils.logErrorWithStack("Unexpected NOT to have spark.hostname set")
+      }
+      Utils.checkHost(hostname)
+      System.setProperty("spark.hostname", hostname + ":" + boundPort)
```
mateiz: This is a little hacky, reading spark.hostname and then appending a port to it. I would just create a new system property for hostPort and rename usages of spark.hostname to it.

core/src/main/scala/spark/Utils.scala
```diff
@@ -218,6 +223,152 @@ private object Utils extends Logging {
...
+  // the available list - everything else being same.
+  // That is, we we first consume data local, then rack local and finally off rack nodes. So the 
+  // prioritization from this method applies to within each category
+  def prioritizeContainers[K, T] (map: HashMap[K, ArrayBuffer[T]]): List[T] = {
+    val _keyList : ArrayBuffer[K] = new ArrayBuffer[K](map.size)
+    _keyList ++= map.keys
+
+    // order keyList based on population of value in map
+    val keyList = _keyList.sortWith(
+      (left: K, right: K) => map.get(left).getOrElse(Set()).size > map.get(right).getOrElse(Set()).size
```
mateiz: No need to specify types here (: K)

core/src/main/scala/spark/ClosureCleaner.scala
mridulm, line 24: The ClassReader keeps reading from the inputstream passed in - while not exposing the underlying stream (for us to 'get' it and close the stream).
So everywhere it is used, we will need to pass along the inputstream - so that it can be closed in case use of the classreader is over or it fails (via exceptions, etc).

That being said - since this was a bug observed by a scientist here as part of using yarn port, I was more trying to get a quick fix in : and probably there might be better ways to resolve this !

core/src/main/scala/spark/scheduler/TaskScheduler.scala
mridulm: Should have checked if trait's can have default impl - I went and added {} to all implementations except for yarn scheduler !
Will do.

core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
mridulm: I think vi + intellij might have caused tabs to creep in - will grep the source tree and replace them, my apologies.

1) I think I am violating that all over the place - will correct.

2) I started explicitly introducing types since intellij was not able to infer the types properly - so that I can get sufficient contextual information while coding (code completion, lookups, etc) ... and then it became a general practise I guess.

3) I think we made an effort to ensure all lines were <= 100 chars (unless it is overflowing by a few chars and splitting will make the line really messy). Hopefully I did not miss any.

core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
mridulm, line 140: Yes, modified all tab's to spaces

core/src/main/scala/spark/SparkEnv.scala
mridulm: We are not appending a port to it, but setting it if not already set.
Typically, this should not happen : but not having the property set messes up all sorts of assumptions within spark (with and without yarn branch iirc).
Currently I am using it more as debug code than anything else !

core/src/main/scala/spark/Utils.scala
mridulm: Done.

core/src/main/scala/spark/SparkEnv.scala
mridulm: To add, as part of renaming all host to hostPort (where relevant) within code, this was the only part which I did not change - and left as spark.hostname : even though it is actually a hostport now.
I was not sure if this was an exposed environment variable which can be set by users customizing spark execution : cant for yarn, but was not sure for other deployment modes.

If that is not the case and this system property is not part of exposed interface of spark, I would love to remove all references of spark.hostname and replace with spark.hostport !

mridulm:
Hi Matei,

Thanks a lot for the review - I look forward to more comments to further improve the patch !

About debug code. Ideally I would have removed them before pushing, except for two reasons :
1) Since we are still actively working on this branch (or rather, the svn which was exported to this branch), and are experimenting on task scheduling, caching etc improvements : hence the Utils.check\* methods still in the exported code to ensure that no constraints/assumptions were violated in our changes.

2) We have tested only in local and yarn mode (and probably not in all environments of local mode too).
So I was not sure if our assumption (w.r.t host, etc) were holding in other deployment modes (mesos, etc for example).
Hence, the Utils.check\* and debug code were left around (clearly marked I hope !) to ensure that they can be proactively caught.
Ofcourse, once testing is done, this can (and must - for perf reasons) be removed.

About merging to trunk.
I am unsure of what the best path would be - would defer to your and other spark developer's suggestions on how to accomplish it best. We would love to get it merged as soon as possible - so that the yarn branch does not diverge too much from trunk where I can see some interesting work already getting merged !

Thanks,
Mridul

core/src/main/scala/spark/InputFormatInfo.scala
```diff
@@ -0,0 +1,214 @@
...
+import scala.collection.immutable.Set
+import org.apache.hadoop.mapred.{FileInputFormat, JobConf}
+import org.apache.hadoop.util.ReflectionUtils
+import org.apache.hadoop.mapreduce.Job
+import org.apache.hadoop.conf.Configuration
+import scala.collection.mutable.{ArrayBuffer, HashMap, HashSet}
+import scala.collection.JavaConversions._
+
+
+// information about a specific split instance : handles both split instances.
```
mateiz: Capitalize this

core/src/main/scala/spark/InputFormatInfo.scala
```diff
@@ -0,0 +1,214 @@
...
+  // So unless there is identity equality between underlyingSplits, it will always fail even if it 
+  // is pointing to same block.
+  override def equals(other: Any): Boolean = other match {
+    case that: SplitInfo => {
+      this.hostLocation == that.hostLocation && 
+        this.inputFormatClazz == that.inputFormatClazz &&
+        this.path == that.path &&
+        this.length == that.length &&
+        // other split specific checks (like start for FileSplit)
+        this.underlyingSplit == that.underlyingSplit
```
mateiz: Indent is weird here, maybe there are some tabs?

core/src/main/scala/spark/InputFormatInfo.scala
```diff
@@ -0,0 +1,214 @@
...
+import org.apache.hadoop.util.ReflectionUtils
+import org.apache.hadoop.mapreduce.Job
+import org.apache.hadoop.conf.Configuration
+import scala.collection.mutable.{ArrayBuffer, HashMap, HashSet}
+import scala.collection.JavaConversions._
+
+
+// information about a specific split instance : handles both split instances.
+// So that we do not need to worry about the differences.
+class SplitInfo(val inputFormatClazz: Class[_], val hostLocation: String, val path: String,
```
mateiz: Put this into a separate .scala file since it's a fairly large class

core/src/main/scala/spark/InputFormatInfo.scala
```diff
@@ -0,0 +1,214 @@
...
+      return prefLocsFromMapredInputFormat()
+    }
+  }
+}
+
+
+
+
+object InputFormatInfo {
+  /*
```
mateiz: Sounds good, but make this a Javadoc comment

core/src/main/scala/spark/InputFormatInfo.scala
mateiz: Also, I would move both of these into the spark.scheduler package

core/src/main/scala/spark/Utils.scala
```diff
@@ -218,6 +223,158 @@ private object Utils extends Logging {
...
 
+  // expensive ? cache ?
+  def getAddressHostName(address: String): String = {
+    InetAddress.getByName(address).getHostName
+  }
+
+
+
+  def localHostPort(): String = {
+    val retval = System.getProperty("spark.hostname", null)
```
mateiz: Let's just rename this to spark.hostPort; it's not really used by users as spark.hostname

core/src/main/scala/spark/Utils.scala
```diff
@@ -218,6 +223,158 @@ private object Utils extends Logging {
     customHostname.getOrElse(InetAddress.getLocalHost.getHostName)
   }
 
+  // expensive ? cache ?
```
mateiz: Pretty sure that the OS or even Java caches this

core/src/main/scala/spark/deploy/master/WorkerInfo.scala
```diff
@@ -20,6 +20,10 @@ private[spark] class WorkerInfo(
   def coresFree: Int = cores - coresUsed
   def memoryFree: Int = memory - memoryUsed
 
+  def hostPort: String = {
+    if (port > 0) host + ":" + port else host
+  }
```
mateiz, line 26: Why would workers give us a port of 0? Might be better to always require them to give a port.

core/src/main/scala/spark/deploy/yarn/ApplicationMaster.scala
```diff
@@ -70,115 +88,150 @@ class ApplicationMaster(args: ApplicationMasterArguments, conf : Configuration)
...
+      // so that the loop (in ApplicationMaster.sparkContextInitialized) breaks
+      ApplicationMaster.incrementAllocatorLoop(ApplicationMaster.ALLOCATOR_LOOP_WAIT_COUNT)
     }
     logInfo("All workers have launched.")
+
+    // Launch a progress reporter thread, else app will get killed after expiration (def: 10mins) timeout
+    if (userThread.isAlive){
+      // ensure that progress is sent before YarnConfiguration.RM_AM_EXPIRY_INTERVAL_MS elapse.
+
+      val timeoutInterval = yarnConf.getInt(YarnConfiguration.RM_AM_EXPIRY_INTERVAL_MS, 120)
```
mateiz: If this is in milliseconds, 120 seems like a pretty low value; maybe you meant 120*1000?

core/src/main/scala/spark/deploy/yarn/ApplicationMasterArguments.scala
```diff
@@ -1,50 +1,58 @@
...
-    case ("--worker-cores") :: IntParam(value) :: tail =>
-      workerCores = value
-      parse(tail)
-    
-    case Nil =>
-      if (userJar == null || userClass == null) {
+  def parseArgs(args: List[String]): Unit = {
+    val userArgsSeq = new ArrayBuffer[String]()
+    parseImpl(userArgsSeq, args)
+    userArgs = userArgsSeq.readOnly
```
mateiz: This userArgsSeq stuff is slightly strange -- why not make userArgs an ArrayBuffer directly, and then just append to it in parseArgs? ArrayBuffer is still a subclass of Seq so you can pass it outside to code that expects a Seq.

core/src/main/scala/spark/deploy/yarn/ClientArguments.scala
```diff
@@ -2,56 +2,78 @@ package spark.deploy.yarn
...
+  // TODO
+  var inputFormatInfo: List[InputFormatInfo] = null
+
+  parseArgs(args.toList)
+
+  def parseArgs(args: List[String]): Unit = {
+    val userArgsSeq: ArrayBuffer[String] = new ArrayBuffer[String]()
+    val inputFormatMap: Map[String, InputFormatInfo] = Map[String, InputFormatInfo]()
+    parseImpl(userArgsSeq, inputFormatMap, args)
+    userArgs = userArgsSeq.readOnly
```
mateiz: Ditto on the UserArgsSeq here

core/src/main/scala/spark/Utils.scala
```diff
@@ -218,6 +223,158 @@ private object Utils extends Logging {
...
+  // K = number to be user). This is to ensure that if we have multiple eligible allocations,
+  // they dont end up allocating all containers on a small number of hosts - increasing probability of
+  // multiple container failure when a host goes down.
+  // Note, there is bias for keys with higher number of entries in value to be picked first (by design)
+  // Also note that invocation of this method is expected to have containers of same 'type' 
+  // (host-local, rack-local, off-rack) and not across types : so that reordering is simply better from 
+  // the available list - everything else being same.
+  // That is, we we first consume data local, then rack local and finally off rack nodes. So the 
+  // prioritization from this method applies to within each category
+  def prioritizeContainers[K, T] (map: HashMap[K, ArrayBuffer[T]]): List[T] = {
```
mateiz: This shouldn't be in Utils IMO since it's specific to scheduling. I would put it in ClusterScheduler or the ClusterScheduler static object, or maybe add a SchedulingUtils in the scheduler package.

core/src/main/scala/spark/executor/StandaloneExecutorBackend.scala
```diff
@@ -55,6 +56,12 @@ private[spark] class StandaloneExecutorBackend(
     case LaunchTask(taskDesc) =>
       logInfo("Got assigned task " + taskDesc.taskId)
       executor.launchTask(this, taskDesc.taskId, taskDesc.serializedTask)
+
+    // Else we get a tonne of errors in the logs at shutdown
+    case Terminated(actorRef) =>
+      logError("Actor " + actorRef + " terminated")
+      // right ?
```
mateiz, line 63: Cool, makes sense

core/src/main/scala/spark/network/ConnectionManager.scala
```diff
@@ -194,8 +197,13 @@ private[spark] class ConnectionManager(port: Int) extends Logging {
...
       logInfo("Removing ReceivingConnection to " + remoteConnectionManagerId)
-      
-      val sendingConnectionManagerId = connectionsById.keys.find(_.host == remoteConnectionManagerId.host).orNull
+
+      // There must be a better way to phrase this in scala !
+      val sendingConnectionManagerId = connectionsById.keys.find(
+        _ match {
+          case k: ConnectionManagerId if k.host == remoteConnectionManagerId.host && k.port == remoteConnectionManagerId.port => true
+          case _ => false
+        } ).orNull
```
mateiz: The ConnectionManagerId class actually has a hashCode and equals since it's a case class, so you can just do `if (!connectionsById.contains(remoteConnectionManagerId))` for the check below and not keep this null stuff around.

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -71,8 +74,9 @@ class DAGScheduler(taskSched: TaskScheduler) extends TaskSchedulerListener with
   val cacheTracker = env.cacheTracker
   val mapOutputTracker = env.mapOutputTracker
 
-  val deadHosts = new HashSet[String]  // TODO: The code currently assumes these can't come back;
-                                       // that's not going to be a realistic assumption in general
+  val deadHostPorts = new HashSet[String]  // TODO: The code currently assumes these can't come back;
+                                       // that's not going to be a realistic assumption in general - though practically,
+                                       // (for yarn) since it is associated with a port, this should hold.
```
mateiz, line 79: Yeah, we've fixed this in the master branch

core/src/main/scala/spark/storage/BlockManager.scala
```diff
@@ -853,6 +861,23 @@ class BlockManager(val master: BlockManagerMaster, val serializer: Serializer, m
     diskStore.clear()
     logInfo("BlockManager stopped")
   }
+
+
+  // Do we actually need this ? I am still investigating !
```
mateiz, line 867: The local dir should not be in HDFS... unless YARN can somehow ask jobs to put their scratch space on HDFS. It should be just a directory on the machine, for fast access. We don't assume that it's reliable. Usually in YARN I think you can just use the working directory of the process.

core/src/main/scala/spark/executor/Executor.scala
```diff
@@ -30,9 +30,10 @@ private[spark] class Executor extends Logging {
 
   initLogging()
 
-  def initialize(slaveHostname: String, properties: Seq[(String, String)]) {
+  def initialize(slaveHostPort: String, properties: Seq[(String, String)]) {
```
mateiz: This will probably not work with the MesosExecutorBackend now unless you make that pass a hostPort too. Maybe you can add code there that creates a dummy ActorSystem and figures out its port similar to in the standalone one. Also do the same in the CoarseGrainedMesos one.

mateiz:
Hey Mridul,

Sorry for the delay, but I've looked at this some more now, and I think it looks really good overall. I've made a few line comments throughout.

For merging to trunk, I think you should create a diff with all these commits once they're in, and try applying that. The one thing that I'd consider doing before that though is renaming the hostPort variables back to host, since a lot of the changed lines are just this rename. We can try it without that though and see how it goes. It's fine to leave them like this in the 0.6 YARN branch but it might be a good way to make this easier to merge to master.

Anyway, thanks again for taking the time to do this!

core/src/main/scala/spark/InputFormatInfo.scala
mridulm: I will move InputFormatInfo to spark.scheduler package.
What about SplitInfo ? It is used in SparkContext (will introduce cyclic dependency between packages).

core/src/main/scala/spark/Utils.scala
mridulm: great !

core/src/main/scala/spark/InputFormatInfo.scala
mateiz: It's okay to move that too. I don't think we've avoided cyclic dependencies between packages at this point.

core/src/main/scala/spark/Utils.scala
mridulm: That is an outdated comment - will remove.
I did not add an explicit cache and decided to rely on dns cache for that reason.

core/src/main/scala/spark/deploy/master/WorkerInfo.scala
mridulm, line 26: The if condition (which might be there in other places) was added to incrementally change the codebase.
Going over the code now, I think you are right - port is always specified.

Will change to host + ":" + port

core/src/main/scala/spark/deploy/yarn/ApplicationMaster.scala
mridulm: Yes, was supposed to be 120k, that is a lousy bug ! (though subsequent line was alleviating the effects of it)

core/src/main/scala/spark/deploy/yarn/ApplicationMasterArguments.scala
mridulm: The idea was to ensure that userArgs is readonly once the cli optoins were parsed. I was passing userArgsSeq (better name perhaps) to the parse function - and once done, make it read only (for rest of spark) to read from.
Basically, we were minimizing changes to existing code.
The same thing is done to ClientArguments too

Is there a better way to do it ?

core/src/main/scala/spark/deploy/yarn/ClientArguments.scala
mridulm: there is a inputFormatMap also passed along here. Please see above : if there is a better way to refactor it, please do let us know !

core/src/main/scala/spark/Utils.scala
mridulm: Will move to ClusterScheduler static object.
Currently it is used in ClusterScheduler and YarnAllocationHandler (spark.deploy.yarn package)

core/src/main/scala/spark/InputFormatInfo.scala
mridulm: You mean the lines subsequent to hostLocation ? I had indented that two spaces on purpose to identify that as continuation of the condition.

Or are you referring to something else I am missing ?

core/src/main/scala/spark/network/ConnectionManager.scala
mridulm: You are right, will change - this slipped by. Will check if there are other instances of this sort of issue.

core/src/main/scala/spark/storage/BlockManager.scala
mridulm, line 867: This snippet actually escaped my cleanup :-) was not supposed to be part of the external commit request, but now that it is out here ...

So the reason why I was investigating this was - what happens to recovery when one or more nodes fail (or a node with multiple containers fails) : and there are joins/cogroups/etc which require all keys from previous iteration to recompute the missing rdd splits.
If I am not wrong, the current approach will cause reset back to initial state even if there is a checkpoint (since the checkpoint was on the node which is now lost).

Particularly for ML applications we are considering spark for, where the computation goes for hours on large cluster with  input of 100's of TB, this would be non trivial amount of wasted computation.

I hope my understanding is wrong ! We are yet to test these aspects - hence why the BlockManager changes are still empty.

core/src/main/scala/spark/executor/Executor.scala
mridulm: Actually, the method uses only the hostname and not hostPort - I can change StandaloneExecutorBackend to pass the hostname to Executor - and leave mesos changes intact.

mridulm:
Thanks a lot for the review Matei. I will try to attach the changed diffs sometime today.
Hopefully once it is reviewed, it will be merged to yarn branch.

Once all the changes are finalized, I will attempt to do the same to trunk and see what the changes are - I expect some amount of work there, should be fun :-)

core/src/main/scala/spark/deploy/yarn/ApplicationMasterArguments.scala
mateiz: I think it's fine to change the type of userArgsSeq to an ArrayBuffer. It won't affect the rest of the code that much -- it's only used in one place -- and it will be easier to follow.

mridulm: userArgsSeq is an ArrayBuffer (though badly named) - or am I missing something ?

mateiz: Sorry, what I meant is that instead of doing `userArgs = userArgsSeq.readOnly`, change the type of `userArgs` to an ArrayBuffer. That is a subclass of Seq so it won't affect the existing code that tries to use it.

mateiz: After this you could also just include parseImpl's code in here instead of having two methods.

mridulm: Ah ok, the reason for two methods and arraybuffer -> read only seq change was: 
I was trying to ensure that the result of the parse is enforced to be read only - so that the args passed in by the invoker do not get modified once it is parsed. Reasoning was, if immutable, then it is one less surface for bugs (also, was continuing with existing contract of the class)

If that is not a concern, I can change it to ArrayBuffer - I do not know the performance characterstics of .readOnly : so I assume there is a good reason to drop it !

Btw, parse (and parseImpl) should be private, sigh

mateiz: I guess it doesn't matter much either way, so up to you. I just wanted to make the class a little simpler.

mridulm: Can you please review the changes I made ? Not sure if it makes it cleaner and simpler.
I am still coming to grips with scala, so I am sure there must be better ways to do it !

mateiz:
Hey Mridul, are you still working on any changes here or do you think this is ready to be merged? I'm excited to get it into the code.

mridulm:
Hi Matei,

  The main pending changes are : 

a) The cache related changes - investigating effects and alleviating the impact of what happens when multiple containers/nodes fail in an iterative workflow which has one or more wide dependencies (joins, etc).

b) Authorization changes - support for kerberos tokens in hadoop.

Both of these are orthogonal to this path for now - and as it stands, this patch should be complete for the subset of functionality it provides ((b) is related, but I dont think I will be working on it for a couple of months due to other deliverables).

Gives this, would be great if the patch can be comitted in case there are no further review changes !

rxin:
Hi Mridul,

Do you think if there is a way for the yarn to simply launch a standalone cluster, and then spits out a Spark master url? 

The reason being that would make both spark-shell and shark command line mode work, given for those programs, the spark master needs to run on the user machines.

mridulm:
Hi Reynold,

  I have not tried yarn in pseudo distributed mode yet - though hadoop has always supported one.
Cloudera documentation (specifically https://ccp.cloudera.com/display/CDH4DOC/Installing+CDH4+on+a+Single+Linux+Node+in+Pseudo-distributed+Mode#InstallingCDH4onaSingleLinuxNodeinPseudo-distributedMode-InstallingCDH4withYARNonaSingleLinuxNodeinPseudodistributedmode) points to how it can be done - though unfortunately, we have not tried it yet.

Traditionally, there is no difference between a full cluster and a pseudo distributed cluster (except for resource constraints, etc which are limitations of pseudo cluster) - and iirc pig dropped its own standalone mode in favour of hadoop's pseudo distrib mode a while back (not sure if it still has own standalone for backward compatibility ...)

Thanks,
Mridul

mateiz:
Alright, thanks Mridul -- I've merged this now. Thanks a lot for taking the time to do this!

I think what Reynold was asking is whether you can launch an instance of Spark on a real YARN cluster, but then pass the master URL outside so that client programs running on the user's machine could connect to it. This would be nice for things like spark-shell and shark that run on the user's machine (because they have a console interface). It's probably not too bad -- just a matter of launching an app that only creates a SparkContext and then communicates its chosen host/port back to the YARN client somehow.

Anyway, I guess the next step is to do this same thing for the master branch. The best way might be for you to start with a patch of the yarn branch as it is now versus before your commits, but I imagine it will still be quite a bit of work. However, one really nice thing about master is that we now have conditional compilation of Hadoop 1 vs Hadoop 2 code. If you put your files in the core/src/hadoop2 folder, they will only be built when people are compiling against Hadoop 2. So there's no longer a need for a separate branch.

Let me know how your timeline looks for this. We'd like to release Spark 0.7 in 1.5-2 weeks and although YARN might not be available for it at that point, we can always add it in a minor version later.

mridulm:
Hi Matei,

  Thanks for the review, glad to hear it is committed !

Please see more inline.

On Fri, Jan 11, 2013 at 7:20 AM, Matei Zaharia notifications@github.com wrote:

> Alright, thanks Mridul -- I've merged this now. Thanks a lot for taking the time to do this!
> 
> I think what Reynold was asking is whether you can launch an instance of Spark on a real YARN cluster, but then pass the master URL outside so that client programs running on the user's machine could connect to it. This would be nice for things like spark-shell and shark that run on the user's machine (because they have a console interface). It's probably not too bad -- just a matter of launching an app that only creates a SparkContext and then communicates its chosen host/port back to the YARN client somehow.

Apologies to Reynold, looks like I totally misunderstood the intent of
the query ...
In theory, I dont see any problem with it - we could initialize only
the spark context with no slaves, and return a master url to clients
(the how of which would be interesting, but we could leverage the
appTrackingUrl or some such to communicate that back.

In practise, particularly right now, the entire initialization is a dirty hack.
SparkContext initialization takes in preferred locations for the job -
which is passed in to YarnAllocationHandler (via YarnClusterScheduler
created in SparkContext) to set the preferred locations.
Since maintaining the allocation state is not a one time operation
(but a persistent background task to ensure yarn does not time out our
app, all containers continue to be active and re-allocate 'gone'
hosts, etc) - resetting that would be interesting.
I dont see why in theory we cant do it - but in practise, I am not
very sure how it might play out.

Having said that, if this is a usecase which I missed out, I can
definitely try to patch it in - I do not want any feature regression
on this !

> Anyway, I guess the next step is to do this same thing for the master branch. The best way might be for you to start with a patch of the yarn branch as it is now versus before your commits, but I imagine it will still be quite a bit of work. However, one really nice thing about master is that we now have conditional compilation of Hadoop 1 vs Hadoop 2 code. If you put your files in the core/src/hadoop2 folder, they will only be built when people are compiling against Hadoop 2. So there's no longer a need for a separate branch.

That sounds like a good plan.

> Let me know how your timeline looks for this. We'd like to release Spark 0.7 in 1.5-2 weeks and although YARN might not be available for it at that point, we can always add it in a minor version later.

I would love to get yarn support into master as soon as possible,
though I am slightly tied up with other efforts : hopefully I should
get back to porting this as soon as possible.

Thanks to you and other spark developers for all the patience and
invaluable review comments !

Regards,
Mridul

PR #337, opened 2012-12-18
SPARK-616: Logging dead workers in Web UI.

pwendell:
This patch keeps track of which workers have died and marks them
as such in the master web UI. It also handles workers which die and
re-register using different actor ID's.

### DIFF
--- a/core/src/main/scala/spark/deploy/master/Master.scala
+++ b/core/src/main/scala/spark/deploy/master/Master.scala
@@ -156,7 +156,8 @@ private[spark] class Master(ip: String, port: Int, webUiPort: Int) extends Actor
     if (spreadOutJobs) {
       // Try to spread out each job among all the nodes, until it has all its cores
       for (job <- waitingJobs if job.coresLeft > 0) {
-        val usableWorkers = workers.toArray.filter(canUse(job, _)).sortBy(_.coresFree).reverse
+        val usableWorkers = workers.toArray.filter(_.state == WorkerState.ALIVE)
+          .filter(canUse(job, _)).sortBy(_.coresFree).reverse
         val numUsable = usableWorkers.length
         val assigned = new Array[Int](numUsable) // Number of cores to give on each node
         var toAssign = math.min(job.coresLeft, usableWorkers.map(_.coresFree).sum)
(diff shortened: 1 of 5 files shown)


mateiz:
Did you add a file called WorkerState?

mateiz:
Oh never mind, I missed it on GitHub for some reason... Just wanted to make sure everything is added with git add.

core/src/main/scala/spark/deploy/master/WorkerState.scala
```diff
@@ -0,0 +1,7 @@
+package spark.deploy.master
+
+private[spark] object WorkerState extends Enumeration("ALIVE", "DEAD", "DECOMMISSIONED") {
+  type WorkerState = Value
+
+  val ALIVE, DEAD, DECOMMISSIONED = Value
```
JoshRosen, line 6: It doesn't look like the `DECOMMISSIONED` state is used anywhere?

pwendell, line 6: I added it so that in the future we can support decommissioning nodes. Actually adding a deccomission option was out of scope for this JIRA.

Can remove it if you think it's confusing to have here.

core/src/main/scala/spark/deploy/master/Master.scala
```diff
@@ -203,6 +204,8 @@ private[spark] class Master(ip: String, port: Int, webUiPort: Int) extends Actor
 
   def addWorker(id: String, host: String, port: Int, cores: Int, memory: Int, webUiPort: Int,
     publicAddress: String): WorkerInfo = {
+    // There may be one or more refs to dead workers on this same node (w/ different ID's), remove them.
+    workers.filter(w => (w.host == host) && (w.state == WorkerState.DEAD)).foreach(workers -= _)
```
mateiz, line 208: I'd prefer not to do this because it will mess up local testing. Can't we just rely on them to time out?

pwendell, line 208: Matei can you be more specific what the problem is in local mode? This is really just to garbage collect the dead workers. If you are running locally, it will work just fine if you have multiple workers on the same host, it will just decide to garbage collect them all at once.

If we rely on a timeout, what would it be? In my experience people will leave nodes off for a long time (e.g. if they are swapping out hardware, sometimes for days)... and they'd want to keep seeing that node as dead until they have a replacement.

mateiz, line 208: Oh I see, I didn't realize that this only removed dead workers. I thought it removed all other workers on this host.

mateiz:
Alright, merged this. Thanks Patrick!

PR #338, opened 2012-12-18
Fixed repl maven build

tomdz:
This fixes two issues:
- The repl jar was not built with a hadoop classifier
- The repl pom did not have the hadoop dependencies in provided scope

Both of these were problems for e.g. shark which depends on the repl artifact.


mateiz:
Thanks Tom.

PR #339, opened 2012-12-19
Show JVM status in the Standlone WebUI.

pwendell:


core/src/main/scala/spark/deploy/JVMInfo.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import management.ManagementFactory
+import collection.JavaConversions._
+
+class GCInfo(val name: String, val gcCount: Long, val gcTime: Long)
+
+/** Exposes various internal JVM metrics for use in Web UI. */
+object JVMInfo {
+  def getAllThreads : Seq[Thread] = {
+    var tg : ThreadGroup  = Thread.currentThread().getThreadGroup
```
mateiz, line 11: Some code style things:
- Don't put a space before the ":"
- Don't put types for local variables like this unless it's not obvious what they are
- Use longer variable names (I'd just call these threadGroup and parentGroup)
- Limit lines to 100 characters (applies below)

In general we've been trying to use the Scala style guide (http://docs.scala-lang.org/style/) with some tweaks (http://spark-project.org/docs/latest/contributing-to-spark.html).

core/src/main/scala/spark/deploy/JVMInfo.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  def getClasspath:Seq[String] = {
+    val pathSep : String = Option(System.getProperty("path.seperator")) match {
+      case Some(s) => s
+      case None => ":"
+    }
+    ManagementFactory.getRuntimeMXBean.getClassPath.split(pathSep)
+  }
+
+  def getSystemProperties: Map[String, String] = Map(ManagementFactory.getRuntimeMXBean.getSystemProperties.toList: _*)
```
mateiz, line 38: This is the line that seems too long

mateiz, line 38: Also, you may be able to just call ManagementFactory.getRuntimeMXBean.getSystemProperties.toMap here, though I'm not sure.

core/src/main/scala/spark/deploy/JVMInfo.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  def getRuntimeDescription: String = "%s %s".format(
+    ManagementFactory.getRuntimeMXBean.getVmName,
+    ManagementFactory.getRuntimeMXBean.getVmVersion)
```
mateiz, line 42: Minor thing but for these longer methods, put curly braces around the body

core/src/main/twirl/spark/deploy/common/classpath_table.scala.html
```diff
@@ -0,0 +1,16 @@
+@(classpath: Seq[String])
+
+<table class="table table-bordered table-striped table-condensed sortable">
+  <thead>
+    <tr>
+     <th>Path Entry</th>
+    </tr>
+  </thead>
+  <tbody>
+    @for(e <- classpath) {
```
mateiz, line 10: Call this "entry" instead of "e" maybe

core/src/main/twirl/spark/deploy/master/index.scala.html
```diff
@@ -17,6 +20,21 @@
...
     </div>
+
+    <!-- JVM Status -->
+     <div class="span6">
+      <h3>JVM Status</h3>
+      <ul class="unstyled">
+        <li><strong>JRE:</strong> @jre</li>
+        <li><strong>JVM:</strong> @jvm</li>
+        <li><strong>Active GCs:</strong>
+          @gcs.map{gc => @gc.name: @gc.gcCount collections, @gc.gcTime ms}
```
mateiz, line 31: Rather weird Twirl formatting here, though I can sort of understand what it does. I'd prefer something like

```
@{gcs.map(gc => "%s: %s collections, %s ms".format(whatever))}
```

That is, keep it as a Scala expression rather than this mix of text and Scala which I guess is okay with Twirl.

mateiz:
Hey Patrick, this is some pretty cool info to collect, but as we discussed offline this will only give you the info for the standalone Master and Worker daemons, not the job JVMs themselves. I think we need to add a HTTP service inside Spark jobs, both on the master side and on the executor side, to expose this kind of info. Denny's RDD dashboard patch might be a good start. Getting this info for the cluster manager is secondary (I'd almost consider not having it, unless we can factor this code in a way where it can be used by both the standalone cluster and Spark jobs themselves).

I also made some comments on the code style since this was a bigger commit with some stuff that was inconsistent with the rest of the code.

tomdz:
I'd like to recommend the metrics library (http://metrics.codahale.com/) for this, I've been using it on several projects with great success. It includes servlets that expose jvm and custom metrics as json, and it has built-in support for sending the stats to a monitoring system like ganglia or graphite.

pwendell:
Tom - thanks for the pointer. I'm going to look into just using the servlets provided there as an alternative to this.

I think ideally we'd have some Spark specific web interface (for things like RDD information), but it would link to these other servlets as auxiliary debugging tools.

mateiz:
Erich from Quantifind also suggested this library: https://github.com/twitter/ostrich. Seems pretty good too, though I haven't checked the differences versus Coda Hale's.

ryanlecompte:
The ostrich library from Twitter that @enachb mentioned is great. In the
past I've used Coda Hale's metrics library (in a pure Java project) and it
worked really well too. It has a nice built-in admin servlet that exposes
JVM metrics (by default) along with any custom gauges/timers/meters that
you define. I haven't used it in a Scala project yet, but it seems to have
fairly good support:

http://metrics.codahale.com/manual/scala/
https://github.com/erikvanoosten/metrics-scala

pwendell:
I spent a while looking at the Coda's Metrics library. It has a small servlet that exposes JVM and user-specified metrics as JSON. Most of the logic there is around formatting the JVM properties nicely in JSON - not in collecting the metrics themselves (which is mostly trivial because MBean's contain everything you'd want). So it wouldn't really replace the logic of templating these metrics nicely in our existing UI.

The more interesting question is whether we'd want to start tracking custom metrics using one of these libraries. Then it might make sense to re-write this patch to access the JVM properties through that library rather than directly, since it would extend naturally to displaying our own metrics.

I haven't looked closely at Ostrich, but it looks like it has some similar functionality (though it doesn't seem to have built in metrics for the JVM properties). This project also doesn't have it's own UI, instead favoring a REST-like JSON interface. Not sure how easy it is to access the metrics on your own in Ostrich, which we'd want to do.

enachb:
Coda's Metrics and Ostrich are very similar. Both abstract the collection from the representation/reporting of metrics and collect standard JVM metrics by default. We use Ostrich and have a custom reporter that pushes the metrics into Kafka. It does come with a HTTP server to display/graph stats. However while this is nice for local stuff, it is really inconvenient if you have more than one node.

As a Spark user the most important aspect would be to make the reporting aspect pluggable, so I can push the metrics into my standard system (Graphite in my case). This can be done with either, so I'm not particular about which lib makes it. IMHO the biggest drawback of Ostrich are the limited docs.

The other aspect would be to allow users to use and expose their own job metrics. I think exposing something similar to Hadoop Job Counters in the UI would be great.

pwendell:
I'm closing this. I'm going to redactor this a bit using Coda Metrics and resubmit later...

PR #340, opened 2012-12-19
Tweaked debian packaging to be a bit more in line with debian standards

tomdz:
This also allows to have the hadoop1 and hadoop2 versions installed side-by-side if somebody wants to do that.


mateiz:
Cool, thanks!

PR #341, opened 2012-12-21
Kryo2 update against Spark master

rxin:


core/src/main/scala/spark/KryoSerializer.scala
```diff
@@ -9,153 +9,80 @@ import scala.collection.mutable
...
+    } catch {
+      // DeserializationStream uses the EOF exception to indicate stopping condition.
+      case e: com.esotericsoftware.kryo.KryoException => throw new java.io.EOFException
+    }
   }
 
-  def close() { in.close() }
+  def close() {
+    input.close()
+    inStream.close()
```
ryanlecompte, line 50: Should each of these be wrapped in a try/catch in case one fails?

rxin, line 50: The exception will propagate upwards anyway. But I looked into Kryo source, and looks like input.close will also close the input stream. I am going to remove the second close from this.

PR #342, opened 2012-12-24
Allow distinct() to be called without parentheses

markhamstra:
When using the default number of splits, distinct() should also work without the parentheses -- e.g., rdd.distinct.collect


mateiz:
Cool, thanks Mark!

PR #343, opened 2012-12-24
lookup() needn't fail when there is no partitioner

markhamstra:
Falling back to filter-map-collect is an option when the optimal path using a single partitioner is not available.


mateiz:
Looks good, thanks.

PR #344, opened 2012-12-26
Log preferred hosts

mbautin:
Log preferred host names in case when running a task on a non-preferred one, because this sometimes indicates an inconsistency between the hostnames Spark workers and Hadoop datanodes use.

### DIFF
--- a/core/src/main/scala/spark/scheduler/cluster/TaskSetManager.scala
+++ b/core/src/main/scala/spark/scheduler/cluster/TaskSetManager.scala
@@ -201,7 +201,9 @@ private[spark] class TaskSetManager(
           val taskId = sched.newTaskId()
           // Figure out whether this should count as a preferred launch
           val preferred = isPreferredLocation(task, host)
-          val prefStr = if (preferred) "preferred" else "non-preferred"
+          val prefStr = if (preferred) "preferred" else
+            "non-preferred, not one of " +
+            task.preferredLocations.mkString(", ")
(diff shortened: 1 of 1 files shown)


core/src/main/scala/spark/scheduler/cluster/TaskSetManager.scala
```diff
@@ -201,7 +201,9 @@ private[spark] class TaskSetManager(
           val taskId = sched.newTaskId()
           // Figure out whether this should count as a preferred launch
           val preferred = isPreferredLocation(task, host)
-          val prefStr = if (preferred) "preferred" else "non-preferred"
+          val prefStr = if (preferred) "preferred" else
```
tomdz: Scala preferred style is to wrap the two lines in `{}` for the else branch, or combine them into one line and move the else in front of it.

mateiz: Agreed with Tom on this. Otherwise it looks good.

mbautin:
This should be ready to merge.

mateiz:
Great, thanks!

mateiz:
By the way, the style you used was still not quite standard; here's what Tom meant for the right style for multi-line conditionals:

```
      val prefStr = if (preferred) {
        "preferred"
      } else {
        "non-preferred, not one of " + task.preferredLocations.mkString(", ")
      }
```

I will change it to this style myself.

PR #345, opened 2012-12-29
Fix deletion of files in current working directory by clearFiles()

JoshRosen:
This addresses an issue where Spark could delete files in the current working directory that were added to the job using `addFile()`.  I encountered this issue when working on PySpark's code deployment mechanism, which is based on `addFile()`.

From user-code's perspective (e.g. UDFs), files added through `addFile()` are assumed to be in the current working directory.  For jobs that are run locally using `DAGScheduler.runLocally()`, tasks run with the driver's current working directory.  As a result, files added through `addFile()` must be copied to the driver's current working directory.  There's no mechanism to change the CWD in Java.

`clearFiles()` and `clearJars()` clean up these files when the driver exits.  This can be a problem if the original files that were added were in the driver's current working directory, because this will cause them to be deleted.

A long-term fix would be to hide the location of fetched files from user code by requiring it to access files through an API like `SparkFiles.get("my-file-name.txt")`.  This will require changes to user code and may require changes to Shark.

As a short-term fix, this pull request removes the code that deletes files in the current working directory and adds checks to `Utils.fetchFiles()` to avoid overwriting existing local files with new data.  The one downside of this change is that it may add junk to the current working directory.  This is preferable to accidentally deleting files.

I've also added `addFile()`/`addJar()` to the Java API.

I also added synchronization to `LocalScheduler.updateDependencies` to avoid performing multiple parallel fetches for the same file.

### DIFF
--- a/core/src/main/scala/spark/Utils.scala
+++ b/core/src/main/scala/spark/Utils.scala
@@ -128,6 +128,9 @@ private object Utils extends Logging {
   /**
    * Download a file requested by the executor. Supports fetching the file in a variety of ways,
    * including HTTP, HDFS and files on a standard filesystem, based on the URL parameter.
+   *
+   * Throws SparkException if the target file already exists and has different contents than
+   * the requested file.
    */
   def fetchFile(url: String, targetDir: File) {
     val filename = url.split("/").last
@@ -142,9 +145,9 @@ private object Utils extends Logging {
         val out = new FileOutputStream(tempFile)
         Utils.copyStream(in, out, true)
         if (targetFile.exists && !Files.equal(tempFile, targetFile)) {
-          logWarning("File " + targetFile + " exists and does not match contents of " + url +
-            "; using existing version")
           tempFile.delete()
+          throw new SparkException("File " + targetFile + " exists and does not match contents of" +
+            " " + url)
         } else {
           Files.move(tempFile, targetFile)
         }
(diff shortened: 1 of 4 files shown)


core/src/main/scala/spark/Utils.scala
```diff
@@ -130,28 +131,47 @@ private object Utils extends Logging {
...
       case "http" | "https" | "ftp" =>
-        logInfo("Fetching " + url + " to " + targetFile)
+        logInfo("Fetching " + url + " to " + tempFile)
         val in = new URL(url).openStream()
-        val out = new FileOutputStream(targetFile)
+        val out = new FileOutputStream(tempFile)
         Utils.copyStream(in, out, true)
+        if (targetFile.exists && !Files.equal(tempFile, targetFile)) {
+          logWarning("File " + targetFile + " exists and does not match contents of " + url +
+            "; using existing version")
```
mateiz: Turn this into an error and throw an exception here; it's too risky to continue running IMO.

core/src/main/scala/spark/scheduler/local/LocalScheduler.scala
```diff
@@ -108,22 +108,24 @@ private[spark] class LocalScheduler(threads: Int, maxFailures: Int, sc: SparkCon
...
-      logInfo("Fetching " + name + " with timestamp " + timestamp)
-      Utils.fetchFile(name, new File("."))
-      currentJars(name) = timestamp
-      // Add it to our class loader
-      val localName = name.split("/").last
-      val url = new File(".", localName).toURI.toURL
-      if (!classLoader.getURLs.contains(url)) {
-        logInfo("Adding " + url + " to class loader")
-        classLoader.addURL(url)
+    this.synchronized {
```
mateiz: You can just say synchronized instead of this.synchronized; it will match the style of the existing code more that way.

core/src/main/scala/spark/Utils.scala
JoshRosen: I agree;  I pushed a commit to change this to SparkException.

mateiz:
Thanks Josh!

PR #346, opened 2012-12-30
Python API (PySpark)

JoshRosen:
This is a first version of the Spark Python API (PySpark).

This pull request contains the PySpark code, plus a Python Programming Guide, Epydoc API docs, and additions to the Quick Start guide.

I would appreciate any feedback, but I'm especially interested in any comments on these topics:
- **Testing**: I've tested this code locally and on the Spark EC2 image, but it would be useful to see if the out-of-the-box configuration works on other systems.
- **Python Programming Guide**: I'd appreciate additional examples or suggestions for this guide.
- **Code clarity**: PySpark's code has a few complicated parts, particularly at the interfaces between Java and the Python worker processes.  I'd love feedback on which parts are confusing and require comments.
- **Fixing the LR example**: I'd like to fix [SPARK-646](https://spark-project.atlassian.net/browse/SPARK-646) so that I can run the LR example on a large dataset.

If you want to try out PySpark, follow the instructions in the documentation included in this branch.


mateiz:
Hey Josh, maybe add a warning to the script in this case. Or, is there any reason not to make it launch with `scala` by default? The reason I added the ability to launch with java was that for worker processes in the standalone mode, killing the `scala` shell script did not kill the JVM subprocess it created. However, for spark-shell for example, I still launch with `scala`, and it works fine.

JoshRosen:
@mateiz I'll modify `./run-pyspark` to always launch using `scala`.  There's no reason why we have to launch using `java`.

core/src/main/scala/spark/api/python/PythonPartitioner.scala
```diff
@@ -0,0 +1,39 @@
...
+  override def getPartition(key: Any): Int = {
+    if (key == null) {
+      return 0
+    }
+    else {
+      val hashCode = {
+        if (key.isInstanceOf[Array[Byte]]) {
+          Arrays.hashCode(key.asInstanceOf[Array[Byte]])
+        }
+        else
```
mateiz: Add some curly braces here

core/src/main/scala/spark/api/python/PythonRDD.scala
```diff
@@ -0,0 +1,222 @@
...
+import scala.io.Source
+
+import spark.api.java.{JavaSparkContext, JavaPairRDD, JavaRDD}
+import spark.broadcast.Broadcast
+import spark._
+import spark.rdd.PipedRDD
+
+
+private[spark] class PythonRDD[T: ClassManifest](
+  parent: RDD[T], command: Seq[String], envVars: java.util.Map[String, String],
```
mateiz: Put these arguments one per line, and indented with two spaces, to match the rest of the code

core/src/main/scala/spark/api/python/PythonRDD.scala
```diff
@@ -0,0 +1,222 @@
...
+  override val partitioner = if (preservePartitoning) parent.partitioner else None
+
+  override def compute(split: Split, context: TaskContext): Iterator[Array[Byte]] = {
+    val SPARK_HOME = new ProcessBuilder().environment().get("SPARK_HOME")
+
+    val pb = new ProcessBuilder(Seq(pythonExec, SPARK_HOME + "/pyspark/pyspark/worker.py"))
+    // Add the environmental variables to the process.
+    val currentEnvVars = pb.environment()
+
+    envVars.foreach {
```
mateiz: You could do for ((variable, value) <- envVars) instead

core/src/main/scala/spark/api/python/PythonRDD.scala
```diff
@@ -0,0 +1,222 @@
...
+      var _nextObj = read()
+
+      def hasNext = _nextObj.length != 0
+    }
+  }
+
+  val asJavaRDD : JavaRDD[Array[Byte]] = JavaRDD.fromRDD(this)
+}
+
+private class PairwiseRDD(prev: RDD[Array[Byte]]) extends
```
mateiz, line 133: Maybe add a comment saying what this means (just that it passes two elements at a time)

core/src/main/scala/spark/api/python/PythonRDD.scala
```diff
@@ -0,0 +1,222 @@
...
+
+  /** Strips the pickle PROTO and STOP opcodes from the start and end of a pickle */
+  def stripPickle(arr: Array[Byte]) : Array[Byte] = {
+    arr.slice(2, arr.length - 1)
+  }
+
+  /**
+   * Write strings, pickled Python objects, or pairs of pickled objects to a data output stream.
+   * The data format is a 32-bit integer representing the pickled object's length (in bytes),
+   * followed by the pickled data.
```
mateiz, line 155: Add a reference to the Pickle specification for other people who want to look into it

core/src/main/scala/spark/api/python/PythonRDD.scala
```diff
@@ -0,0 +1,222 @@
...
+    val file = new DataOutputStream(new FileOutputStream(filename))
+    for (item <- items) {
+      writeAsPickle(item, file)
+    }
+    file.close()
+  }
+}
+
+private object Pickle {
+  def b(x: Int): Byte = x.asInstanceOf[Byte]
```
mateiz: Instead of this, you can just do 0x80.toByte, 0x02.toByte, etc. I think that `val PROTO: Byte = 0x80` doesn't work because `0x80` is an integer bigger than 127 (namely 128), but others will work (for example `val TWO: Byte = 0x02` does).

docs/python-programming-guide.md
```diff
@@ -0,0 +1,74 @@
...
+
+# Key Differences in the Python API
+
+There are a few key differences between the Python and Scala APIs:
+
+* Python is dynamically typed, so RDDs can hold objects of different types.
+* PySpark does not currently support the following Spark features:
+    - Accumulators
+    - Special functions on RRDs of doubles, such as `mean` and `stdev`
+    - Approximate jobs / functions, such as `countApprox` and `sumApprox`.
```
mateiz: Remove this, since it's an undocumented feature for Spark itself until we decide to finalize this API

docs/python-programming-guide.md
```diff
@@ -0,0 +1,74 @@
...
+This guide will show how to use the Spark features described there in Python.
+
+# Key Differences in the Python API
+
+There are a few key differences between the Python and Scala APIs:
+
+* Python is dynamically typed, so RDDs can hold objects of different types.
+* PySpark does not currently support the following Spark features:
+    - Accumulators
+    - Special functions on RRDs of doubles, such as `mean` and `stdev`
```
mateiz: Should be RDDs

docs/python-programming-guide.md
```diff
@@ -0,0 +1,74 @@
...
+
+PySpark requires Python 2.6 or higher.
+PySpark jobs are executed using a standard cPython interpreter in order to support Python modules that use C extensions.
+We have not tested PySpark with Python 3 or with alternative Python interpreters, such as [PyPy](http://pypy.org/) or [Jython](http://www.jython.org/).
+By default, PySpark's scripts will run programs using `python`; an alternate Python executable may be specified by setting the `PYSPARK_PYTHON` environment variable in `conf/spark-env.sh`.
+
+All of PySpark's library dependencies, including [Py4J](http://py4j.sourceforge.net/), are bundled with PySpark and automatically imported.
+
+Standalone PySpark jobs should be run using the `run-pyspark` script, which automatically configures the Java and Python environmnt using the settings in `conf/spark-env.sh`.
+The script automatically adds the `pyspark` package to the `PYTHONPATH`.
```
mateiz, line 66: I would rename this script to something else, since otherwise people will confuse it with the `run` script, which is not for running user's code (it's for running Spark's internal classes). Could you make one `pyspark` script that can launch either the shell or a standalone program? If that's confusing, having a separate `pyspark` for jobs and `pyspark-shell` is fine.

docs/python-programming-guide.md
```diff
@@ -0,0 +1,74 @@
...
+
+
+# Interactive Use
+
+PySpark's `pyspark-shell` script provides a simple way to learn the API:
```
mateiz: Somewhere in this section, explain how to connect `pyspark-shell` to a non-local cluster. Also it's not just for learning the API, but also for exploring data interactively. Maybe say `pyspark-shell` lets you use PySpark to explore data interactively, and is also a simple way to learn the API.

pyspark/examples/pi.py
```diff
@@ -0,0 +1,20 @@
+import sys
+from random import random
+from operator import add
+from pyspark.context import SparkContext
+
+
+if __name__ == "__main__":
+    if len(sys.argv) == 1:
+        print >> sys.stderr, \
+            "Usage: PythonPi <host> [<slices>]"
```
mateiz: Call the `<host>` parameter `<master>` in the examples; we should eventually do that in the Java and Scala ones too

pyspark/examples/tc.py
```diff
@@ -0,0 +1,49 @@
```
mateiz: Might be good to rename this file to transitive_closure, and the LR one to logistic_regression

pyspark/examples/kmeans.py
```diff
@@ -0,0 +1,49 @@
+import sys
+
+from pyspark.context import SparkContext
```
mateiz: Would it be easy to make it so that you just import `pyspark.SparkContext` instead of `pyspark.context.SparkContext`? It's a minor thing but I think it would be a bit nicer. Just not sure whether Python modules will work that way.

docs/index.md
```diff
@@ -7,11 +7,11 @@ title: Spark Overview
 TODO(andyk): Rewrite to make the Java API a first class part of the story.
 {% endcomment %}
 
-Spark is a MapReduce-like cluster computing framework designed for low-latency iterative jobs and interactive use from an 
-interpreter. It provides clean, language-integrated APIs in Scala and Java, with a rich array of parallel operators. Spark can 
-run on top of the [Apache Mesos](http://incubator.apache.org/mesos/) cluster manager, 
+Spark is a MapReduce-like cluster computing framework designed for low-latency iterative jobs and interactive use from an interpreter.
+It provides clean, language-integrated APIs in Scala, Java, and Python, with a rich array of parallel operators.
```
mateiz: Make the Scala/Java/Python be links to the corresponding programming guides.

docs/index.md
```diff
@@ -72,7 +73,7 @@ of `project/SparkBuild.scala`, then rebuilding Spark (`sbt/sbt clean compile`).
 
 * [Configuration](configuration.html): customize Spark via its configuration system
 * [Tuning Guide](tuning.html): best practices to optimize performance and memory use
-* [API Docs (Scaladoc)](api/core/index.html)
+* API Docs: [Java/Scala (Scaladoc)](api/core/index.html) and [Python (Epydoc)](api/pyspark/index.html)
```
mateiz: Maybe make API Docs a new sub-heading before Deployment guides, to make this more visible

pyspark/examples/kmeans.py
mateiz: Actually, I noticed that you did do this, by referencing the object in pyspark/**init**.py. Can you change the examples to `import pyspark.SparkContext` and export RDD the same way? It's also unfortunate that the Epydoc page doesn't turn these top-level objects into links, but I guess we have to live with that.

docs/python-programming-guide.md
```diff
@@ -0,0 +1,74 @@
...
+* PySpark does not currently support the following Spark features:
+    - Accumulators
+    - Special functions on RRDs of doubles, such as `mean` and `stdev`
+    - Approximate jobs / functions, such as `countApprox` and `sumApprox`.
+    - `lookup`
+    - `mapPartitionsWithSplit`
+    - `persist` at storage levels other than `MEMORY_ONLY`
+    - `sample`
+    - `sort`
+
```
mateiz, line 25: Somewhere around here you should write which features of Python are supported in lambdas, and best practices for doing more complicated functions (for example, if you don't want to express your function as a lambda, you can call a global function). Also explain that PySpark will automatically ship your code to workers if you give it the right files.

mateiz:
Hey Josh, I've tried this out now, and it works well, but I made a few comments on the docs and code. I also have a few other questions/comments:
- It would be nice to have one `pyspark` script that launches either the shell or a standalone program. How hard is that?
- Even you don't do this, can you move the Pyspark scripts (`pyspark-shell`, etc) to the top level?
- How can I run the unit tests? Maybe we should have a shell script to run them.

pyspark/examples/kmeans.py
```diff
@@ -0,0 +1,49 @@
...
+from numpy import array, sum as np_sum
```
mateiz: I would just import numpy as np and prefix the calls to array and sum below, to make it more obvious that we're using NumPy in the code.

core/src/main/scala/spark/api/python/PythonRDD.scala
JoshRosen, line 133: `PairwiseRDD` is used to form `RDD[(Array[Byte], Array[Byte])]` from key-value pairs returned by the Python worker process; it doesn't affect how values are passed to the worker process.  I'll add a comment to clarify this.

pyspark/examples/kmeans.py
JoshRosen: Epydoc will display the module docstring in `pyspark/__init__.py` on the mani API page, so I'll I can add links there.

docs/python-programming-guide.md
JoshRosen, line 66: By launching Python with a custom [PYTHONSTARTUP](http://docs.python.org/2/tutorial/interpreter.html#the-interactive-startup-file) file, it's possible to create a `pyspark` script that functions like the regular `python` but creates a SparkContext when used interactively.  The master would be specified by setting the MASTER environment variable.

Unfortunately, we can't name this script `pyspark` because it will conflict with the `pyspark` directories.  Do you have any other naming suggestions?

mateiz, line 66: Ah, rename the folder to something else then (maybe just `python`).

JoshRosen:
Hi Matei,

Thanks for the comments.  I incorporated all of the small typo fixes and style changes.

I replaced the `pyspark-shell` and `run-pyspark` scripts by a single top-level `pyspark` script, which behaves like a  regular `python` executable but creates a SparkContext when used interactively.

I also added a `run-tests` script to run the tests.  I'd like to use a real test runner like `nose` or `py.test`, but I'm running into trouble with the module-level setup that I perform in `rdd.py`'s `__main__()` method.  I use this method to add `sc` to the `extraglobs` dict so that I don't have to create a new SparkContext in each docstring.  Unfortunately, the `nose` doctest runner can't pick this up.  Nose has [doctest fixtures](https://nose.readthedocs.org/en/latest/doc_tests/test_doctest_fixtures/doctest_fixtures.html), which should solve this problem, but I can't seem to get them to work.  Other nose users have run into the same problem: https://groups.google.com/d/topic/nose-users/MFmQywrZQDE/discussion; I posted a message to bump that thread to see if there's a solution (my message is still pending moderation).

The k-means example is currently broken because I haven't ported the `takeSample` code to PySpark (the old `takeSample` code won't work with PySpark's batching).  When I tried replacing `takeSample` with `take`, I ran into a `key not found` error in the cache tracker actor, so there might be a bug lurking somewhere.  I'll see about fixing this.

JoshRosen:
I found the cause of the NoSuchElementException in `take()`.  My current implementation of `take()` calls `RDD.iterator()` directly; I'm not sure that `iterator()` was designed to be called outside of an existing `iterator()` implementation.

Calling `iterator()` on a non-cached RDD works fine, but calling it on a cached RDD will trigger an error in the CacheTrackerActor:

``` scala
scala> val rdd = sc.parallelize(1 to 100).cache()
rdd: spark.RDD[Int] = spark.ParallelCollection@51037076

scala> val rdd = sc.parallelize(1 to 100)
rdd: spark.RDD[Int] = spark.ParallelCollection@339d035e

scala> rdd.iterator(rdd.splits(0), new TaskContext(0, 0, 0))
res8: Iterator[Int] = non-empty iterator

scala> rdd.cache()
res9: spark.RDD[Int] = spark.ParallelCollection@339d035e

scala> rdd.iterator(rdd.splits(0), new TaskContext(0, 0, 0))
13/01/02 15:27:46 INFO CacheTracker: Cache key is rdd_8_0
13/01/02 15:27:46 INFO CacheTracker: Computing partition spark.ParallelCollectionSplit@7d9
13/01/02 15:27:46 INFO MemoryStore: ensureFreeSpace(2952) called with curMem=12608, maxMem=349922918
13/01/02 15:27:46 INFO MemoryStore: Block rdd_8_0 stored as values to memory (estimated size 2.9 KB, free 333.7 MB)
13/01/02 15:27:46 INFO BlockManagerMasterActor$BlockManagerInfo: Added rdd_8_0 in memory on joshs-mbp.local:59701 (size: 2.9 KB, free: 333.7 MB)
13/01/02 15:27:46 INFO BlockManagerMaster: Updated info of block rdd_8_0
13/01/02 15:27:46 ERROR CacheTrackerActor: key not found: 8
java.util.NoSuchElementException: key not found: 8
    at spark.CacheTrackerActor$$anonfun$receive$1.apply(CacheTracker.scala:58)
    at spark.CacheTrackerActor$$anonfun$receive$1.apply(CacheTracker.scala:45)
    at akka.actor.Actor$class.apply(Actor.scala:318)
    at spark.CacheTrackerActor.apply(CacheTracker.scala:31)
    at akka.actor.ActorCell.invoke(ActorCell.scala:626)
    at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:197)
    at akka.dispatch.Mailbox.run(Mailbox.scala:179)
    at akka.dispatch.ForkJoinExecutorConfigurator$MailboxExecutionTask.exec(AbstractDispatcher.scala:516)
    at akka.jsr166y.ForkJoinTask.doExec(ForkJoinTask.java:259)
    at akka.jsr166y.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:975)
    at akka.jsr166y.ForkJoinPool.runWorker(ForkJoinPool.java:1479)
    at akka.jsr166y.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:104)
```

The problem is that the cached RDD is never registered with the cache tracker because it is never registered in a stage with the DAGScheduler.  It looks like `CacheTracker.registerRDD()` is idempotent, so I can solve this by adding an extra call from PySpark's `take()` method.  Does this sound reasonable, or is there an easier way to implement `take()`?

mateiz:
Ah, you aren't meant to call Iterator before passing the RDD through the DAGScheduler. It will register it properly when you do. If you can't use the original take() in this, use DAGScheduler.runJob directly. There are different versions and they let you do whatever you want with the iterator.

mateiz:
Hey Josh,

This looks good. Can run-tests be made to print something at the end though? Otherwise the output is kind of confusing (hard to tell if something failed).

MLnick:
One thing I noticed is that the new `pyspark` script loses the ipython capability.

I've created an amended shell script to deal with this. Basically it just uses getopts to add the `-i` argument, and set `PYSPARK_PYTHON="ipython"` if it is set (as well as launching ipython to run `PYTHONSTARTUP` explicitly since it doesn't use this env variable).

This is probably preferable to doing it in shell.py actually.

One upshot of this is one cannot then use ipython to run a script, only for the shell (well it is possible but turned out to look very messy) - this is probably ok as mostly it would be used for interactive work. 

I can submit a PR when this one is merged (it's a very small change).

JoshRosen:
@MLnick I dropped ipython support because I'd like to have a `pyspark` script that behaves exactly like a regular `python` executable except for creating a SparkContext.

I was able to use ipython with the following command:

```
PYSPARK_PYTHON="ipython" ./pyspark -i $SPARK_HOME/python/pyspark/shell.py
```

Note that the `PYSPARK_PYTHON` executable is used by both the master and workers, so this will require ipython to be installed on all machines in the cluster.

JoshRosen:
@mateiz I pushed a commit that prints success or failure (with colored text) after running the PySpark tests.

The only major remaining tasks are to port takeSample() / sample() and to improve error reporting and propagation of exceptions from Python workers to the Python driver.

MLnick:
@JoshRosen Makes sense. That is effectively what my amended script does. I guess people who want it can just use that approach.

Really looking forward to using this more!

python/pyspark/context.py
```diff
@@ -0,0 +1,159 @@
...
+        self.sparkHome = sparkHome or None # None becomes null in Py4J
+        self.environment = environment or {}
+        self.batchSize = batchSize  # -1 represents a unlimited batch size
+
+        # Create the Java SparkContext through Py4J
+        empty_string_array = self.gateway.new_array(self.jvm.String, 0)
+        self._jsc = self.jvm.JavaSparkContext(master, jobName, sparkHome,
+                                              empty_string_array)
+
+        self.pythonExec = os.environ.get("PYSPARK_PYTHON_EXEC", 'python')
```
MLnick: `pyspark` script references `PYSPARK_PYTHON` whereas this is `PYSPARK_PYTHON_EXEC`, causing sc.pythonExec to always be `python` regardless of setting the env var on the commandline.

MLnick:
Successfully tested this PR out on a small Hadoop cluster running CDH4.0.0. Was as simple as git clone, change Hadoop version appropriately, sbt/sbt package, ship the spark directory to the rest of the nodes, add slaves and start-all, and run ./pyspark with appropriate env vars.

Impressed with the ease of getting up & running!

python/pyspark/context.py
JoshRosen: Ah, thanks for catching this.  Fixed.

mateiz:
Thanks Josh! I've merged this so that people can more easily contribute to it.

PR #347, opened 2012-12-30
Changes to make Spark run using IBM J9

shivaram:
The changeset contains the following changes:
a. Fix WorkerArguments to use reflection to get total physical memory using the appropriate MBean

b. While modifying the SizeEstimator to work with IBM J9, I found that the project `java-sizeof` already has a RamUsageEstimator class which uses very similar techniques to what we have. Additionally it is pulled from the Apache Lucene/Solr source code and is available on maven. To avoid having to maintain SizeEstimator in spark, I have replaced it with java-sizeof from https://github.com/dweiss/java-sizeof.  
More details on what it does can be found at 
http://www.slideshare.net/DawidWeiss/sizeofobject-how-much-memory-objects-take-on-jvms-and-when-this-may-matter (Also the slides show why it is tricky to estimate size across JVMs)

c. Upgrade default Hadoop version to 1.0.3. This is because Hadoop 0.20 has Sun-specific dependencies. Also fix a bug in saving SequenceFiles to infer the correct Writable type

d. Remove unused BoundedMemoryCache and associated test file. These tests fail on IBM J9 as string size varies across JVMs. The class was unused in the current version of Spark as it has been replaced by MemoryStore.

Tested using OpenJDK 6,7 on 32 and 64-bit arch and using IBM J9 1.7 on 32-bit arch.


mateiz:
Hey Shivaram, a few things:
- Have you checked that java-sizeof produces reasonable behavior on the tests we used to have in SizeEstimatorSuite? I don't like to just delete tests.
- Does java-sizeof use the sampling trick that SizeEstimator had for large arrays (where it just pulls out 100 elements and extrapolates the total size from that)?
- Add the new dependency and change the Hadoop version in the Maven POM file as well

shivaram:
Hey Matei

Thanks for taking a look
a. Yes, I checked the java-sizeof procedures and they do produce reasonable behavior on the tests we had. In fact for the OpenJDK cases, they almost produce the exact same results that we had.  I agree that deleting tests is not very healthy, but while changing the code, I realized that we had almost the same functionality as java-sizeof and thought its better to reuse an existing module than maintain our own copy -- (well almost see (b))

b. java-sizeof does the caching trick so we don't have to not revisit the same class again, but I forgot about the array sampling technique we had. java-sizeof  does not sample arrays and walks through each of its elements.

I'm not sure whats the best way around this --  The two main changes we don't have right now (that java-sizeof does have) are accounting for field offsets and array base offsets using sun.misc.Unsafe. On the other hand java-sizeof doesn't have the array sampling trick. Do you think its better to add stuff to our SizeEstimator or modify their code ?

c. Will do the Maven fix.

mateiz:
Hey Shivaram,

Could we maybe just keep the Sun JVM code on IBM's to begin with? How far off are the sizes? I think it's good to keep the array sampling trick because the objects we cache can be pretty large arrays. I don't have numbers off the top of my head but I do vaguely remember size estimation taking a while without it.

shivaram:
I agree that we should keep the sampling trick as I don't think its good to degrade performance with this change.

Regd. keeping our code, the differences are in how fields are packed in a class and how array headers are aligned etc. From what I have seen, our existing code will overestimate the size on J9 -- so while this means we won't use all the available memory, it will be conservative and I guess that is okay.

The trouble is that our SizeEstimator test cases might not pass on J9 -- Would that be okay ? I can create a JIRA issue to track that separately.

mateiz:
Why would the tests not pass? They don't know the true size, they just check that it matches what we computed, at least as far as I know. It's true that we would use less memory than we really can but that's not the end of the world if the difference is small.

mateiz:
Basically, maybe we can make the tests and/or SizeEstimator take the information from some kind of object that we can mock, so that we can test the size returned for any JVM on any other JVM.

shivaram:
Sorry I forgot to mention this in the previous post -- The reason the tests fail on J9 is because the String class in J9 has an additional field called hashCode32. I can't find any documentation about that online, but this means we have an extra int in that class. Also J9 sometimes allocates extra elements in the char array (this may be due to some object re-use ?) resulting in single or two character strings having 4 or 8 length char arrays within them. (This is before accounting for padding -- reflect.Array.getLength returns 4 or 8)

The problems mostly stem from strings -- I like your idea about having some kind of mock JVM properties that can be used for tests.  Here is one such proposal, we could split the tests into two parts:  
The first part could use a mock-String class that is the same as say OpenJDK7's String class (for which we have source etc.). The second part could more of a benchmark which uses the JVM's string and tries to summarize memory usage for different string sizes. The second one will be primarily intended as a tool that can be used by developers to get an idea of memory usage. In fact this could support reading from a file etc. as well.

mateiz:
Ah, cool, that sounds good for the String stuff. We might still not match object sizes exactly on J9, but at least the tests will work. I'd go with that.

shivaram:
I tried to rebase and remove the RamUsageEstimator in this branch, but that leads github to believe that the pull request contains 76 commits. I'm opening another pull request based on a new branch at https://github.com/shivaram/spark/compare/ibm-jdk-fixes  -- Sorry for the trouble.

shivaram:
Opened https://github.com/mesos/spark/pull/354 - Closing this pull request

PR #348, opened 2013-01-01
Raise exception when hashing Java arrays (SPARK-597)

JoshRosen:
This commit adds exceptions in situations where hashing Java arrays may lead to incorrect behavior (e.g. performing reduceByKey() with Java arrays as keys will produce incorrect output); see [SPARK-597](https://spark-project.atlassian.net/browse/SPARK-597).

We're able to catch some cases where Java arrays will lead to problems, but we can't catch all of them: what if a user has a `Tuple2[Int, Array[Int]]` key?

Ultimately, the burden is on the user to know that it's dangerous to rely on arrays' hashcodes.  But it doesn't hurt to add low-cost checks for common pitfalls.


mateiz:
Looks good, thanks.

PR #349, opened 2013-01-01
Avoid stalls when computation of cached RDD throws exception

charlesreiss:
CacheTracker.getOrCompute() had a try {} finally block wrapping too little, so a transient failure in RDD.compute() would leave it marked as loading after the resulting task failed. When computing the RDD was retried on the same slave, the resulting task would stall indefinitely. This patch fixes this and adds a test.


mateiz:
Good catch, thanks!

PR #350, opened 2013-01-02
Spark Streaming

tdas:
Happy New Year !!!    

[Need i say more :P]


mateiz:
Great, 12,000+ lines of pending pull requests, thanks! :P

ryanlecompte:
Wow! This is awesome. I can't wait to review this.

core/src/main/scala/spark/MapOutputTracker.scala
```diff
@@ -126,7 +129,7 @@ private[spark] class MapOutputTracker(actorSystem: ActorSystem, isMaster: Boolea
 
   // Called on possibly remote nodes to get the server URIs and output sizes for a given shuffle
   def getServerStatuses(shuffleId: Int, reduceId: Int): Array[(BlockManagerId, Long)] = {
-    val statuses = mapStatuses.get(shuffleId)
+    val statuses = mapStatuses.get(shuffleId).orNull
```
ryanlecompte, line 132: Minor nit: Might be better avoid null if possible, e.g.:

``` scala
mapStatuses.get(shuffleId) match {
  case Some(status) => ...
  case _ => ...
}
```

core/src/main/scala/spark/broadcast/HttpBroadcast.scala
```diff
@@ -123,4 +130,21 @@ private object HttpBroadcast extends Logging {
...
+
+  def cleanup(cleanupTime: Long) {
+    val iterator = files.internalMap.entrySet().iterator()
+    while(iterator.hasNext) {
+      val entry = iterator.next()
+      val (file, time) = (entry.getKey, entry.getValue)
+      if (time < cleanupTime) {
+        try {
+          iterator.remove()
+          new File(file.toString).delete()
```
ryanlecompte, line 142: Probably should check the return value from File.delete(), since it returns a boolean true/false if file was successfully deleted.

tdas, line 142: I thought of it. But I could not think of a good use of the return value. If it could not delete it then most probably the file is somehow already deleted and I dont need to do anything else anyways.

core/src/main/scala/spark/util/RateLimitedOutputStream.scala
```diff
@@ -0,0 +1,56 @@
+package spark.util
+
+import java.io.OutputStream
+
+class RateLimitedOutputStream(out: OutputStream, bytesPerSec: Int) extends OutputStream {
```
ryanlecompte, line 5: Nice class!

streaming/src/main/scala/spark/streaming/Interval.scala
```diff
@@ -0,0 +1,51 @@
+package spark.streaming
+
+private[streaming]
+case class Interval(beginTime: Time, endTime: Time) {
```
ryanlecompte, line 4: Any reason not to extend scala.math.Ordering here?

streaming/src/main/scala/spark/streaming/StreamingContext.scala
```diff
@@ -0,0 +1,389 @@
...
+   */
+  def start() {
+    if (checkpointDir != null && checkpointInterval == null && graph != null) {
+      checkpointInterval = graph.batchDuration
+    }
+
+    validate()
+
+    val networkInputStreams = graph.getInputStreams().filter(s => s match {
+        case n: NetworkInputDStream[_] => true
```
ryanlecompte, line 347: Minor nit: This could be written more succinctly using Traversable.collect():

``` scala
val networkInputStreams = graph.getInputStreams.collect {
  case n: NetworkInputDStream[_] => n.toArray
}
```

streaming/src/main/scala/spark/streaming/Time.scala
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * This is a simple class that represents time. Internally, it represents time as UTC.
+ * The recommended way to create instances of Time is to use helper objects
+ * [[spark.streaming.Milliseconds]], [[spark.streaming.Seconds]], and [[spark.streaming.Minutes]].
+ * @param millis Time in UTC.
+ */
+
+case class Time(private val millis: Long) {
+
```
ryanlecompte, line 11: Any thoughts on using Joda Time for this type of time/interval things?

conf/streaming-env.sh.template
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
mateiz: Why is this different from spark-env.sh? Let's just have one config file, it will be less confusing to users.

sentences.txt
```diff
@@ -0,0 +1,3 @@
+Hello world!
```
mateiz: You can delete this file, doesn't seem like it's used for anything

core/src/main/scala/spark/util/MetadataCleaner.scala
```diff
@@ -4,9 +4,10 @@ import java.util.concurrent.{TimeUnit, ScheduledFuture, Executors}
 import java.util.{TimerTask, Timer}
 import spark.Logging
```
mateiz, line 5: A comment on this file from the version that was merged in by Reynold: The MetadadataCleaner seems like it may be creating non-daemon threads by using TimerTask. Will this prevent the JVM from exiting until this is stopped? Or does Java use daemon threads for TimerTask? Basically it would be nice if programs will still exit even if they don't call SparkContext.stop(). Try that out, and if it doesn't work, switch this to using a DaemonThread with a sleep loop instead of TimerTask.

core/src/main/scala/spark/util/TimeStampedHashMap.scala
```diff
@@ -1,8 +1,8 @@
 package spark.util
```
mateiz, line 1: In this class, rename cleanup() to cleanOlderValues, and add a Javadoc comment for it explaining what threshTime is.

core/src/main/scala/spark/RDDCheckpointData.scala
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  // Get iterator. This is called at the worker nodes.
+  def iterator(split: Split, context: TaskContext): Iterator[T] = {
+    rdd.firstParent[T].iterator(split, context)
+  }
+}
+
+private[spark] object RDDCheckpointData {
+  def checkpointCompleted() {
```
mateiz: Rename this to something like clearCaches to make clearer what it does in the code above

core/src/main/scala/spark/RDDCheckpointData.scala
```diff
@@ -0,0 +1,97 @@
...
+ * Enumeration to manage state transitions of an RDD through checkpointing
+ * [ Initialized --> marked for checkpointing --> checkpointing in progress --> checkpointed ]
+ */
+private[spark] object CheckpointState extends Enumeration {
+  type CheckpointState = Value
+  val Initialized, MarkedForCheckpoint, CheckpointingInProgress, Checkpointed = Value
+}
+
+/**
+ * This class contains all the information of the regarding RDD checkpointing.
```
mateiz: This needs a bit more detail -- basically say that it's associated with an RDD and that it handles checkpointing for it.

mateiz:
Hey TD, a couple of things from a quick look:

1) I made some comments on the code, though nothing super detailed yet

2) Can you move the streaming examples into the "examples" project instead of "streaming"? Maybe put them in package spark.examples.streaming there. It doesn't make sense to package them with the streaming JAR and Maven artifact.

core/src/main/scala/spark/RDDCheckpointData.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  import CheckpointState._
+
+  var cpState = Initialized
+  @transient var cpFile: Option[String] = None
+  @transient var cpRDD: Option[RDD[T]] = None
+
+  // Mark the RDD for checkpointing
+  def markForCheckpoint() {
+    RDDCheckpointData.synchronized {
```
mbautin, line 38: Does this have to be synchronized on the global `RDDCheckpointData` object, or could it be synchronized on `this`? (also applies to other synchronized blocks in this class).

streaming/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,11 @@
+# Set everything to be logged to the file streaming-tests.log
```
mateiz: Why isn't this logging just to spark-tests.log like the other projects?

mateiz:
One last thing: it would be nice to add a pom.xml so that this can be built with Maven. Ask Reynold or Matt for help if you need it. It would probably just be similar to the pom.xml for examples.

core/src/main/scala/spark/util/MetadataCleaner.scala
tdas, line 5: It will exit fine because I configured the Timer object to use a daemon thread (see line 12 below). I also tested it, just to be sure, and the program exited without calling SparkContext.stop()

core/src/main/scala/spark/util/RateLimitedOutputStream.scala
tdas, line 5: To be honest, this class was written by Matei :)

streaming/src/main/scala/spark/streaming/Interval.scala
tdas, line 4: I agree it would be cool. But I dont think I ever needed the functionality. Furthermore, it is a little unclear how to compare intervals, especially when they are overlapping. So I dont think we should have anything like that until we really need it. Note that this is priavate[streaming], so it is not exposed to the developer.

streaming/src/main/scala/spark/streaming/Time.scala
tdas, line 11: I have got this question multiple times! It is true that JodaTime could have been used. However, I wanted something far more simple with just a few additional convenience functions, and JodaTime seemed a little complex for the simple requirements. In fact I had considered using just Long instead of having the wrapper class Time. But I chose to keep the thin wrapper just in case we need to switch to a different internal format in the future, maybe JodaTime instead of Long. So for the time being this was simple, sufficient, while having the flexibility of extending it later.

conf/streaming-env.sh.template
tdas: Removed it. I had kept it thinking that I would put recommended settings for SparkStreaming in it. But now all the setting recommendations have gone into the Programming Guide, so removing this is fine. I agree that it will be less confusing.

sentences.txt
tdas: Removed.

core/src/main/scala/spark/RDDCheckpointData.scala
tdas, line 38: Very good question! The whole point of synchronizing this at all is so that while the RDD and its splits are being serialized for being sent as a Task, the switch in the dependencies should not take place. Doing a finer granularity locking by locking individual RDDs/RDDCheckpoints requires all the RDDs involved in a task to be locked. This would be slightly tricky to implement, as one must traverse through the RDD dependency graph to figure out which RDDs will be serialized along with the task and lock only those RDDs. Locking multiple objects always leads to a possibility of deadlocks, etc (though my gut feeling says it would be fine as we locking along a ordered DAG of objects). To play it safe, I chose to keep it simple by having a global lock. Furthermore, we dont expect the checkpoint operation to be performed very frequently (maybe once a second at max), so most of the time, only the task launching thread is going to take the lock. So lock contention is not much of a performance issue here. Hence I felt that keeping it simple was fine for now. In future we may have rethink this.

streaming/src/test/resources/log4j.properties
tdas: So that one can look at the streaming test logs separately from spark-only test logs.

mateiz: In that case we should do the same for the other sub projects. Personally I'd prefer to have just one common file since people can always run only one module's tests with sbt.

tdas: Okay, I think I am cool with keeping it spark-tests.log.

On that note, I had kept the examples of Spark Streaming with <repo>/streaming/ because the examples of the bagel projects were inside bagel and not in <repo>/examples. But I agree that no point having the examples in the spark streaming jar. I am moving them to examples.

core/src/main/scala/spark/util/TimeStampedHashMap.scala
tdas, line 1: Done.

core/src/main/scala/spark/RDDCheckpointData.scala
tdas: Done.

core/src/main/scala/spark/RDDCheckpointData.scala
tdas: Done.

streaming/src/test/resources/log4j.properties
tdas: The streaming-test.log should be kept separate from spark-tests.log. Otherwise the streaming tests overwrite the spark tests logs as log4j is configured to not append. In fact, right now, the bagel and repl tests overwrite the logs of the core tests. I think we should keep it as "not append" (because it would be confusing to have older logs mixed with new logs) and have separate log files for core, streaming, bagel, repl (so that they dont overwrite each other).

mateiz: Oh okay, in that case they should each make their own files. Maybe put those files under the target/ directory for each project so they don't pollute the top-level one.

tdas: Should the output of core tests be also under core/target/spark-tests.log?

mateiz: Yeah, maybe for each project call the file <project>/target/unit-tests.log.

tdas: Done. 
I am having trouble compiling Spark with maven (as first step to writing the pom.xml for streaming), so I plan to do it after Reynold is back. But that does not stop us in any way in doing the pull. Is there anything else to be done for the pull request to be accepted?

mateiz:
Alright, I think this is good! We can do a bit more reviewing when we merge this into master, but let's have this in the "streaming" branch now to do further streaming development here.

Do you think there are any major things missing in the interaction with the rest of Spark before we merge into master? If not, I'd like to do that sooner rather than later, since there are a few other things I want to fix there.

tdas:
YAY! THANKS!

I dont think there are any more Spark related changes left that needs to be pushed / merged. All subsequent changes that Patrick and I have been working on are within Spark Streaming. 

Regarding merging into master, there are few things that should be tried. 1. We should run the spark-perf to see the impact of these changes. 2. Changes to the RDD API affect the Shark and systems with custom RDDs. Those needs to be figured out.

streaming/src/main/scala/spark/streaming/dstream/KafkaInputDStream.scala
```diff
@@ -0,0 +1,201 @@
...
+  topics: Map[String, Int], initialOffsets: Map[KafkaPartitionKey, Long], 
+  storageLevel: StorageLevel) extends NetworkReceiver[Any](streamId) {
+
+  // Timeout for establishing a connection to Zookeper in ms.
+  val ZK_TIMEOUT = 10000
+
+  // Handles pushing data into the BlockManager
+  lazy protected val dataHandler = new DataHandler(this, storageLevel)
+  // Keeps track of the current offsets. Maps from (broker, topic, group, part) -> Offset
+  lazy val offsets = HashMap[KafkaPartitionKey, Long]()
```
ryanlecompte, line 115: This should probably be a java.util.ConcurrentHashMap since multiple MessageHandler worker threads modify the offsets map.

PR #351, opened 2013-01-05
Add PairRDDFunctions.keys and values.

stephenh:


core/src/test/scala/spark/ShuffleSuite.scala
```diff
@@ -216,6 +216,13 @@ class ShuffleSuite extends FunSuite with ShouldMatchers with BeforeAndAfter {
     // Test that a shuffle on the file works, because this used to be a bug
     assert(file.map(line => (line, 1)).reduceByKey(_ + _).collect().toList === Nil)
   }
+
+  test("kesy and values") {
```
mateiz: Looks good but there's a typo here

stephenh: Ack, thanks; I started out with just values and added keys quickly after the fact. Fixed.

mateiz:
Great, thanks!

PR #353, opened 2013-01-05
Add RDD.tupleBy.

stephenh:
Given Spark requires joins/cogroups to be done on Tuple2s with the same key, it seems like a common operation is to take an RDD[YourRecordType] and turn it in to RDD[JoinKey, YourRecordType].

I had been using groupByKey as a short hand for doing this succinctly, but really didn't need/want the semantics of turning it into RDD[JoinKey, Seq[YourRecordType]], as that ended up complicating some logic downstream where I had to unrolling/de-Seq the RDD.

So, the suggested tupleBy is like groupBy, but without the group.


mateiz:
I've thought about this a bit and I think it might be better to call it something like keyBy or addKey, since tupleBy is not a very standard name. Any thoughts? I just want it to be really obvious what it does.

pwendell:
I prefer keyBy the most. This will be nice to have.

stephenh:
I like keyBy too; I'll amend the pull request.

mateiz:
Alright, sounds good. Do you mind also adding this into the JavaRDDLike class to make it available in the Java API? It's in core/src/main/scala/spark/api/java/JavaRDDLike.scala. Let me know if it's not clear how to do it... basically you just need to add a method that returns `JavaPairRDD.fromRDD(rdd.keyBy(function))`.

mateiz:
Also, it looks like you need to merge the latest master branch into your branch for the pull request to work, since we added a few methods.

stephenh:
Okay, I merged and updated JavaRDDLike. I found casting ClassManifest[AnyRef] to ClassManifest[K] kind of odd (I don't understand how that works), but I copied it from some other code in JavaRDDLike, and it passes the test.

mateiz:
Alright, thanks. The ClassManifest thing is indeed weird but it really just needs to be different for primitive types. It doesn't matter what kind of object it is if it's an object. ClassManifest is used by Scala to instantiate arrays of a given type, creating primitive arrays instead of arrays of wrapper values for types like Int.

PR #355, opened 2013-01-08
Activate hadoop1 profile by default for maven builds

shivaram:
While testing maven builds today, I realized that we don't have a default hadoop profile and its a bit clumsy to always use -Phadoop1. This patch makes hadoop1 the default profile for all the projects. hadoop2 can still be used with the -P switch.

### DIFF
--- a/bagel/pom.xml
+++ b/bagel/pom.xml
@@ -47,7 +47,7 @@
       <id>hadoop1</id>
       <activation>
         <property>
-          <name>!hadoop</name>
+          <name>!hadoopVersion</name>
         </property>
       </activation>
       <dependencies>
@@ -79,7 +79,7 @@
       <id>hadoop2</id>
       <activation>
         <property>
-          <name>hadoop</name>
+          <name>hadoopVersion</name>
           <value>2</value>
         </property>
       </activation>
--- a/bagel/pom.xml
+++ b/bagel/pom.xml
@@ -45,6 +45,9 @@
   <profiles>
     <profile>
       <id>hadoop1</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
+      </activation>
       <dependencies>
         <dependency>
           <groupId>org.spark-project</groupId>
--- a/bagel/pom.xml
+++ b/bagel/pom.xml
@@ -46,7 +46,9 @@
     <profile>
       <id>hadoop1</id>
       <activation>
-        <activeByDefault>true</activeByDefault>
+        <property>
+          <name>!hadoop</name>
+        </property>
       </activation>
       <dependencies>
         <dependency>
--- a/bagel/pom.xml
+++ b/bagel/pom.xml
@@ -77,6 +77,12 @@
     </profile>
     <profile>
       <id>hadoop2</id>
+      <activation>
+        <property>
+          <name>hadoop</name>
+          <value>2</value>
+        </property>
+      </activation>
       <dependencies>
         <dependency>
           <groupId>org.spark-project</groupId>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -159,6 +159,9 @@
   <profiles>
     <profile>
       <id>hadoop1</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
+      </activation>
       <dependencies>
         <dependency>
           <groupId>org.apache.hadoop</groupId>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -160,7 +160,9 @@
     <profile>
       <id>hadoop1</id>
       <activation>
-        <activeByDefault>true</activeByDefault>
+        <property>
+          <name>!hadoop</name>
+        </property>
       </activation>
       <dependencies>
         <dependency>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -216,6 +216,12 @@
     </profile>
     <profile>
       <id>hadoop2</id>
+      <activation>
+        <property>
+          <name>hadoop</name>
+          <value>2</value>
+        </property>
+      </activation>
       <dependencies>
         <dependency>
           <groupId>org.apache.hadoop</groupId>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -161,7 +161,7 @@
       <id>hadoop1</id>
       <activation>
         <property>
-          <name>!hadoop</name>
+          <name>!hadoopVersion</name>
         </property>
       </activation>
       <dependencies>
@@ -218,7 +218,7 @@
       <id>hadoop2</id>
       <activation>
         <property>
-          <name>hadoop</name>
+          <name>hadoopVersion</name>
           <value>2</value>
         </property>
       </activation>
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -47,7 +47,7 @@
       <id>hadoop1</id>
       <activation>
         <property>
-          <name>!hadoop</name>
+          <name>!hadoopVersion</name>
         </property>
       </activation>
       <dependencies>
@@ -79,7 +79,7 @@
       <id>hadoop2</id>
       <activation>
         <property>
-          <name>hadoop</name>
+          <name>hadoopVersion</name>
           <value>2</value>
         </property>
       </activation>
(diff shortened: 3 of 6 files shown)


mateiz:
Shivaram, are you sure that if you do this, it doesn't activate _both_ profiles if you do -Phadoop2? I think Maven profiles might be "additive" rather than being one-or-the-other. If it doesn't do that, then it's great.

There was also a feature called environments that might help with specifying a default set of profiles, but I haven't looked into it much.

shivaram:
From http://maven.apache.org/guides/introduction/introduction-to-profiles.html

"This profile will automatically be active for all builds unless another profile in the same POM is activated using one of the previously described methods. All profiles that are active by default are automatically deactivated when a profile in the POM is activated on the command line or through its activation config."

I read that as hadoop2 will get used and hadoop1 will be deactivated with the command line switch.

Is there an easy way to test this ?

mateiz:
Yeah, if it compiles fine with -Phadoop2, then you're probably good. Some of the conditionally compiled code we had would only work with one version of Hadoop (otherwise you'd have two source files with the same class defined).

shivaram:
Actually I'm having some trouble getting repl to compile with -Phadoop2 -- Maven reports an error

[ERROR] Failed to execute goal on project spark-repl: Could not resolve dependencies for project org.spark-project:spark-repl:jar:0.7.0-SNAPSHOT: Could not find artifact org.spark-project:spark-core:jar:hadoop1:0.7.0-SNAPSHOT -> [Help 1]

So the repl is looking for spark-core built with hadoop1 for some reason.  I will update this once I figure out what is wrong...

shivaram:
Actually I couldn't find a way to get this to work: From what I can see, the default profile (hadoop1) compiles correctly. But when we pass -Phadoop2, hadoop2 is activated in each module separately. I think this leads to a situation where while repl is being built, hadoop2 is not activated in core and hadoop1 gets activated by default.  I am not a 100% sure how to get around this, so I am closing this for now.

rxin:
Shiv - Hive has the same thing. Perhaps look into Hive's pom file?

mateiz:
Also, did it work before this change? I guess it must've unless we recently broke it.

tomdz:
See http://jira.codehaus.org/browse/MNG-3229 for a discussion of `activeByDefault`. The simplest way to avoid having to type `-Phadoop1` all the time, is probably to add this to your `settings.xml`:

```
<activeProfiles>
  <activeProfile>hadoop1</activeProfile>
</activeProfiles>
```

shivaram:
Couple of points
- Things work fine before this change -- it looks like setting activeByDefault leads to this error
- I managed to get it to work by activating based on the absence of a system property. I found the sonatype book useful for this - http://www.sonatype.com/books/mvnref-book/reference/profiles-sect-activation.html#profiles-sect-activation-by-absence
  The patch is at https://github.com/shivaram/spark/commit/f7adb382ace7f54c5093bf90574b3f9dd0d35534 . But using hadoop2 is a bit clumsy with this change and the behavior we have is:
  a. mvn package  # builds using hadoop1
  b. mvn -Dhadoop -Phadoop2 # builds using hadoop2
  c. mvn -Dhadoop -Phadoop1 # also builds using hadoop1
  d. mvn -Phadoop2 # fails as both hadoop1 and hadoop2 are active
- `settings.xml` also works, but this is a per-user setting -- so we'll need to track that as something new users should do

mateiz:
Is there a way you could use the property value syntax a bit farther up in that to turn on hadoop2 if you pass -Dhadoop=2 for example? Would make it even simpler.

mbautin:
As an alternative  solution, wouldn't it be cleaner to explicitly tell the user that -Phadoop1 or -Phadoop2 has to be specified? Example in Shark: https://github.com/mbautin/shark/commit/cad0b8e6f807d68935270b0d01a6476c6805a69c

mateiz:
I think we can have both that and a default. I do like having a less scary mvn command by default. I'm going to look into what I mentioned above (a -Dhadoop2 version) and if not, I'll probably merge this.

shivaram:
I added a change to activate hadoop2 with -Dhadoop=2. Also I think the maven activation properties are or'ed, so I guess we could also support -Dhadoop=1 to activate hadoop1 if required.

mateiz:
Cool, let's do that. Also maybe change the property name "hadoop" to something like "hadoopType" to make it sound less mysterious. Thanks for looking into this!

shivaram:
I renamed the variable to hadoopVersion -- I could not add support for both hadoopVersion=1 and !hadoopVersion to trigger the hadoop1 profile -- So only !hadoopVersion triggers hadoop1 right now. Turns out maven does not allow two property blocks in the same profile activation and the or'ing works across different activation techniques like OS, JDK version, environment variable etc.

mateiz:
Alright, thanks for looking into this. Going to merge it in.

mbautin:
Unfortunately, this forces the hadoop1 profile to be selected when importing to IntelliJ IDEA, and there is no way to uncheck it without removing activation on !hadoopVersion. I am not sure what the best fix for this would be yet.

mateiz:
Oh, that's too bad. If you can't find a fix for this we can revert it, but definitely let us know.

shivaram:
This is unfortunate -- I don't use IntelliJ IDEA and only tested builds
from the command line. I will try to take a look to see if I can find a
fix.

Shivaram

mbautin:
Tested with IntelliJ 11.1.5 (IntelliJ 12.0.2's Maven profile handling is still buggy), the same problem happens there. I am using this script to fix pom.xml's locally in the meantime:

``` python
#!/usr/bin/env python

import sys
import os
import re

ACTIVATION_RE = re.compile( ' *' + r'\s*'.join([
  '<activation>',
  '<property>',
  '<name>!hadoopVersion</name>',
  '</property>',
  '</activation>'
]), re.MULTILINE)

def remove_activation(f_path):
  f_in = open(f_path)
  pom_xml = f_in.read()
  f_in.close()
  f_out = open(f_path, 'w')
  f_out.write(ACTIVATION_RE.sub('', pom_xml))
  f_out.close()

for root, dirs, files in os.walk('.'):
  for f in files:
    if f == 'pom.xml':
      remove_activation(os.path.join(root, f))
```

mbautin:
Here is one more problem, this time more serious. When trying to build Shark against a Spark version with auto-activation of hadoop1 in its POM, I get the following error:

[ERROR] Failed to execute goal on project shark: Could not resolve dependencies for project edu.berkeley.cs.amplab:shark:jar:0.3.0-csd-7-SNAPSHOT: The following artifacts could not be resolved: org.spark-project:spark-core:jar:hadoop1:0.7.0-csd-2, org.spark-project:spark-bagel:jar:hadoop1:0.7.0-csd-2, org.spark-project:spark-examples:jar:hadoop1:0.7.0-csd-2, org.spark-project:spark-streaming:jar:hadoop1:0.7.0-csd-2: Could not find artifact org.spark-project:spark-core:jar:hadoop1:0.7.0-csd-2

(Artifacts for the 0.7.0-csd-2 version of Spark for hadoop2 are deployed in our repository.)

The build command was:

``` bash
mvn -Phadoop2 -DskipTests clean package
```

mateiz:
Ah, I see, so somehow the build is still tagged as hadoop1 even though you built with -Phadoop2? Did you try passing -Dhadoop=2 instead of -Phadoop2? it might somehow have tried to activate both profiles.

If this continues to be a problem I'm okay removing this eventually.

shivaram:
Yeah - The way maven profiles are passed across modules seems to be really tricky. The benefit of having a default profile may not be worth it.

mbautin:
Another observation: building Shark with

```
mvn -Phadoop2 -DhadoopVersion=2 -DskipTests clean package
```

works, so it looks like Maven evaluates transitive dependencies using the profile activation logic based on `hadoopVersion` since it is specified in Spark's pom files. The `hadoopVersion` property was not set in my previous attempt, so Maven added a dependency on `hadoop1`. If I am interpreting this correctly, with these changes any project depending on Spark would have to specify `hadoopVersion` in addition to using the correct classifier with its Spark dependencies.

PR #356, opened 2013-01-08
Fix an issue in ConnectionManager where sendMessage may create too many unnecessary connections

shane-huang:
In ConnectionManager, sendMessage creates a new SendingConnection when connection host key is not found in connectionsById.  But there might be too many unnecessary connections created before connectionsById is updated in the connection-manager-thread.run. In out test ConnectionMangerTest fails on "connection reset by peer" or "timeout" when there're too many message sending threads. This patch reuse the connections and fix this issue.   Reported in issue tracker as SPARK-647 https://spark-project.atlassian.net/browse/SPARK-647

### DIFF
--- a/core/src/main/scala/spark/network/Connection.scala
+++ b/core/src/main/scala/spark/network/Connection.scala
@@ -136,10 +136,10 @@ extends Connection(SocketChannel.open, selector_) {
           if (chunk.isDefined) {
             messages += message  // this is probably incorrect, it wont work as fifo
             if (!message.started) {
-		logDebug("Starting to send [" + message + "]")
- 		message.started = true
-	 	message.startTime = System.currentTimeMillis
-	    }
+              logDebug("Starting to send [" + message + "]")
+              message.started = true
+              message.startTime = System.currentTimeMillis
+            }
             return chunk 
           } else {
             /*logInfo("Finished sending [" + message + "] to [" + remoteConnectionManagerId + "]")*/
--- a/core/src/main/scala/spark/network/Connection.scala
+++ b/core/src/main/scala/spark/network/Connection.scala
@@ -135,8 +135,11 @@ extends Connection(SocketChannel.open, selector_) {
           val chunk = message.getChunkForSending(defaultChunkSize)
           if (chunk.isDefined) {
             messages += message  // this is probably incorrect, it wont work as fifo
-            if (!message.started) logDebug("Starting to send [" + message + "]")
-            message.started = true
+            if (!message.started) {
+		logDebug("Starting to send [" + message + "]")
+ 		message.started = true
+	 	message.startTime = System.currentTimeMillis
+	    }
             return chunk 
           } else {
             /*logInfo("Finished sending [" + message + "] to [" + remoteConnectionManagerId + "]")*/
--- a/core/src/main/scala/spark/network/ConnectionManager.scala
+++ b/core/src/main/scala/spark/network/ConnectionManager.scala
@@ -43,12 +43,12 @@ private[spark] class ConnectionManager(port: Int) extends Logging {
   }
   
   val selector = SelectorProvider.provider.openSelector()
-  val handleMessageExecutor = Executors.newFixedThreadPool(4) 
+  val handleMessageExecutor = Executors.newFixedThreadPool(20) 
   val serverChannel = ServerSocketChannel.open()
   val connectionsByKey = new HashMap[SelectionKey, Connection] with SynchronizedMap[SelectionKey, Connection] 
   val connectionsById = new HashMap[ConnectionManagerId, SendingConnection] with SynchronizedMap[ConnectionManagerId, SendingConnection]
   val messageStatuses = new HashMap[Int, MessageStatus] 
-  val connectionRequests = new SynchronizedQueue[SendingConnection]
+  val connectionRequests = new HashMap[ConnectionManagerId, SendingConnection] with SynchronizedMap[ConnectionManagerId, SendingConnection] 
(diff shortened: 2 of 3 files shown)


core/src/main/scala/spark/network/Connection.scala
```diff
@@ -135,8 +135,11 @@ extends Connection(SocketChannel.open, selector_) {
           val chunk = message.getChunkForSending(defaultChunkSize)
           if (chunk.isDefined) {
             messages += message  // this is probably incorrect, it wont work as fifo
-            if (!message.started) logDebug("Starting to send [" + message + "]")
-            message.started = true
+            if (!message.started) {
```
rxin, line 138: Can you indent this, and trim trailing whitespaces?

core/src/main/scala/spark/network/ConnectionManager.scala
```diff
@@ -43,12 +43,12 @@ private[spark] class ConnectionManager(port: Int) extends Logging {
   }
   
   val selector = SelectorProvider.provider.openSelector()
-  val handleMessageExecutor = Executors.newFixedThreadPool(4) 
+  val handleMessageExecutor = Executors.newFixedThreadPool(20)
```
rxin: Perhaps this is a good chance to make this configurable as a system property? Same for the 10 secs wait time.

core/src/main/scala/spark/network/ConnectionManager.scala
```diff
@@ -78,11 +78,12 @@ private[spark] class ConnectionManager(port: Int) extends Logging {
 
   def run() {
     try {
-      while(!selectorThread.isInterrupted) {
-        while(!connectionRequests.isEmpty) {
-          val sendingConnection = connectionRequests.dequeue
+      while(!selectorThread.isInterrupted) {   
+        for( (connectionManagerId, sendingConnection) <- connectionRequests) {
```
rxin, line 82: trim trailing whitespace

rxin:
Hi Shane - thanks for doing this. I left a couple nitpicks.

core/src/main/scala/spark/network/ConnectionManager.scala
```diff
@@ -78,11 +78,12 @@ private[spark] class ConnectionManager(port: Int) extends Logging {
...
+          //val sendingConnection = connectionRequests.dequeue
```
tdas: Any reason to keep this line around?

shane-huang: should remove the commented line. Sorry didn't notice.

core/src/main/scala/spark/network/ConnectionManagerTest.scala
```diff
@@ -13,8 +13,8 @@ import akka.util.duration._
 
 private[spark] object ConnectionManagerTest extends Logging{
   def main(args: Array[String]) {
-    if (args.length < 2) {
-      println("Usage: ConnectionManagerTest <mesos cluster> <slaves file>")
+    if (args.length < 5) {
+      println("Usage: ConnectionManagerTest <mesos cluster> <slaves file> <num of tasks> <size of msg> <count>")
```
tdas: Since you have added 2 extra parameters, can you specify in some comment above what values should be used for these 2 parameters?

core/src/main/scala/spark/network/ConnectionManager.scala
shane-huang: I agree it's good to make thread pool size a system property. However I think it's not worthy of making10 secs wait time as system property in ConnectionManager's object main.  It seems this small unit test is not quite useful for real testing (as it only sends message to itself) and there's many fixed arguments such as count 10 and size 10MB. I will change it back to 1 and make it less confusing. i'd rather make the arguments settable in ConnectionManagerTest.scala which demonstrates better a cluster's network connection throughput.

core/src/main/scala/spark/network/ConnectionManagerTest.scala
```diff
@@ -56,13 +56,13 @@ private[spark] object ConnectionManagerTest extends Logging{
           logInfo("Sending [" + bufferMessage + "] to [" + slaveConnManagerId + "]")
           connManager.sendMessageReliably(slaveConnManagerId, bufferMessage)
         })
-        val results = futures.map(f => Await.result(f, 1.second))
+        val results = futures.map(f => Await.result(f, 999.second))
```
shane-huang: i'll change the 999 wait time as settable. Because the wait time should differ based on different transfer size.

core/src/main/scala/spark/network/ConnectionManager.scala
rxin: Ah yes - sorry I missed the part that it was in test code; and I was thinking about something else (the map output tracker wait time).

shane-huang: BTW, I noticed the configurations are scattered everywhere in source code. Why not have a place to gather all settable system properties (something like mapred-site.xml/hdfs-site.xml in hadoop/conf) or at least in some documentation so that people don't have to look into code to know which arguments are tunable.  Any plans for that?

rxin: I thought it was a great idea and created a ticket for this a while ago
https://spark-project.atlassian.net/browse/SPARK-529

I don't think anyone has got around to implement it yet. If you are
interested, it might make sense to do it after the streaming code is merged
into master.

shane-huang:
The new commit modified whitespace, use configurable system properties and add comments in ConnectionManagerTest

mateiz:
Hey Shane,

This looks good -- thanks! I'm going to add it into the 0.6 branch too.

PR #357, opened 2013-01-09
JSON support added to WebUI

tysonjh:
Accessing either the Worker or Master WebUI with a query parameter "?format=json" will yield a response with content-type "application/json" and a properly formatted JSON object.

### DIFF
--- a/project/SparkBuild.scala
+++ b/project/SparkBuild.scala
@@ -134,7 +134,6 @@ object SparkBuild extends Build {
       "cc.spray" % "spray-can" % "1.0-M2.1",
       "cc.spray" % "spray-server" % "1.0-M2.1",
       "cc.spray" %%  "spray-json" % "1.1.1",
-      "io.spray" %%  "spray-json" % "1.2.3",
       "org.apache.mesos" % "mesos" % "0.9.0-incubating"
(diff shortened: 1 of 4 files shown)


project/SparkBuild.scala
```diff
@@ -133,6 +133,8 @@ object SparkBuild extends Build {
       "colt" % "colt" % "1.2.0",
       "cc.spray" % "spray-can" % "1.0-M2.1",
       "cc.spray" % "spray-server" % "1.0-M2.1",
+      "cc.spray" %%  "spray-json" % "1.1.1",
+      "io.spray" %%  "spray-json" % "1.2.3",
```
mateiz: Why do you add both of these in here? If you want we can change Spray altogether to io.spray.

tysonjh: There is no need for both here, you are correct. 

Using io.spray would mean spray-server would be replaced with spray-routing. Currently they are on release 1.1-M7 which is a "double upgrade" for supporting new Scala (2.9.2/2.10) and Akka (2.0.4) versions so may be worth waiting for a stable release to make the change.

mateiz:
Cool, thanks! Looks like this doesn't show executor info but it's still useful.

tysonjh:
I'm adding support right now for the ExecutorRunner, JobDescription classes so we can add the list of running/finished executors.

PR #358, opened 2013-01-09
Retrieve jars to a flat directory so * can be used for the classpath.

stephenh:
This is somewhat trivial, but when doing "ps aux | grep java" to see what Spark processes were running, I noticed the classpath line is incredibly long.

Since JVM 6, the classpath arg can take "dir/*" and load all of the jars in that directory, which results in a much shorter command line in the ps output.

"dir/*" is not recursive though, so it requires changing Ivy to retrieve the jars in a flat manner.

I think this is a improvement, granted its minor, so understand if you're fine with the way things work now.


squito:
I agree that the really long classpath is a pain for "ps aux | grep java", so I'm in favor of the change.  (can't comment on the code in particular)

mateiz:
Interesting, didn't know this. It sounds good.

Maybe we can also set it up so that if you build with Maven, you only get their equivalent of the assembly JAR on the path (it's called spark-repl-shaded-something).

PR #359, opened 2013-01-09
Add file:// jars to the shell as well as the SparkContext.

stephenh:
It would be nice to handle non-file:// URIs, e.g. Hadoop URIs, as well, but I couldn't see how to do that off the top of my head.

I can look in to that more, if that is a requirement for merging the pull request.


mateiz:
Ah cool, this is an elegant solution. I think you can use Utils.downloadFile to fetch a HDFS or HTTP or FTP URL as well. Take a look at it.. it's what the addJar, etc files are fetched with on the worker nodes.

mateiz:
Sorry, it's actually called Utils.fetchFile.

stephenh:
Nice, thanks, fetchFile looks handy.

It wants a targetDir to put the fetched file in...trying to think what a good option for that is...java.io.tmp or . or something else. Also thinking that maybe it could skip targetDir and just return the tempFile directory.

What do you think?

stephenh:
Er, still not ready. Looks like post-addClasspath closures aren't getting shipped around.

mateiz:
Ah, too bad; did you ever find a way to make it work?

It might be because Spark's JavaSerializer or ClosureCleaner gets a reference to the ClassLoader at some point and assumes the classpath will not change.

stephenh:
Turns out each addClasspath call leads to the interpreter getting recreated, which was stopping the old class server and started a new class server, without informing the workers of the new port.

I fixed that by extracting/reusing the class server across interpreter instances:

https://github.com/stephenh/spark/compare/master...addjartoshell

This makes closures defined in the REPL work again...but now something else weird is going on, I'm getting ClassNotFoundExceptions for classes that are in the jar that I sc.addJar'd. Not sure why they aren't getting to the workers, given I call super.addJar and, AFAIK, didn't change anything.

We'll see; I'm still looking in to it.

stephenh:
Okay, I had to reorder the class loaders so that REPL lines that referenced addJar'd classes could see them.

I'm going to reopen the pull request, but I'd appreciate a thorough review/testing before being merged.

mateiz:
Hey Stephen,

So a few comments on this:
- When SparkContext.addJar is called in the REPL, warn the user that it will replay _all_ their commands so far, and give them a yes/no prompt to confirm whether they want to do this.
- Don't print the whole classpath each time, since it looks really daunting.
- Add a note to the addJar Scaladoc comment to warn about this behavior.

To be honest, because of the restart issue, I'm wondering whether it's better to just let users add the JARs one, statically, through some command-line options to spark-shell, and disable addJar in it. It's a little disappointing but it would be simpler to maintain.

stephenh:
> To be honest, because of the restart issue...

Yeah, I agree. This was not as simple as I thought it was going to be, especially with the replay commands.

I agree a separate environment variable sounds like a good idea, and then override addJar to be a no-opt that tells the spark-shell user to use that instead. ...or maybe does an addJar, but outputs a warning that the shell's classpath won't be affected.

I can work on this; ideas for the name of the environment variable (that would do shell.addClasspath as well as sc.addJar,  but only on init)? SPARK_SHELL_ADD_JAR? Or just ADD_JAR, so it'd be easy to type:

ADD_JAR=/home/blah/foo.jar ./spark/spark-shell

mateiz:
I would call it ADD_JARS and allow it to be a comma-separated list of JARs. Having no prefix is fine because it will make it clearer that this only works for the shell, similarly to MASTER. (It's a little odd that our shell takes parameters this way; it's mostly because any arguments are passed to spark.repl.Main, but it might be better to modify the latter to support our own arguments.)

PR #360, opened 2013-01-11
Changed CoGroupRDD's hash map from Scala to Java.

rxin:
Experiences with both Shark and GraphLab have shown that using the Java hashmap improved performance.

### DIFF
--- a/core/src/main/scala/spark/rdd/CoGroupedRDD.scala
+++ b/core/src/main/scala/spark/rdd/CoGroupedRDD.scala
@@ -1,8 +1,8 @@
 package spark.rdd
 
 import java.util.{HashMap => JHashMap}
+import scala.collection.JavaConversions
 import scala.collection.mutable.ArrayBuffer
-import scala.collection.JavaConversions._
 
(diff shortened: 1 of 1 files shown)


mateiz:
Yes it does, I think we replaced it in other places before. Are you sure that getOrElseUpdate is fast on this though? Maybe it would be better to do map.get / map.put.

rxin:
Yes - the Shark one uses get/put explicitly. Let me do that.

mateiz:
By the way, for these use cases where a table just grows, we may be even better off with fastutil's Object2ObjectOpenHashMap: http://fastutil.dsi.unimi.it/docs/it/unimi/dsi/fastutil/objects/Object2ObjectOpenHashMap.html. Might be fun to try sometime. It's an array-based map so it has much fewer objects.

rxin:
Done.

I will do the fastutil hashmap next time in a separate pull request.

mateiz:
Great, thanks. The fastutil thing would need testing anyway to make sure it's faster. And it might be best to jump straight to our own HashMap implementation anyway (using a single Array[Object] instead of the two that fastutil has for better cache locality).

PR #361, opened 2013-01-11
Merge Streaming into master

mateiz:


mateiz:
Alright, I've merged this now. Some of this code still needs to be reviewed more carefully, but this should make it easier to keep all our changes in sync.

PR #362, opened 2013-01-11
Add SparkContext.hadoopConfiguration

stephenh:
The hadoopConfiguration also pulls in any spark.hadoop.\* system properties and explicitly checks for the AWS creds for Hadoop's s3/s3n file systems (based on Shark's code).

I've been using this in our cluster to avoid using "s3n://user:pass@..."  style URLs, with the AWS creds being appended to the spark-env.sh file, and it's working well.


JoshRosen:
- A `JobConf` is also created in `PairRDDFunctions.saveAsHadoopFile()`, so we should probably use the default there, too.
- For consistency's sake, should we also add an optional `jobConf` argument to `RDD.saveAsObjectFile()` and `RDD.saveAsTextFile()`?  I noticed that this option was missing while trying to use a different HDFS block size in `saveAsTextFile()`.
- Before this pull request, default JobConf objects weren't shared across RDD transformations (each call created a fresh JobConf).  Is it safe to share the same JobConf instance across all RDD transformations?  Should we clone or copy the configuration before modifying it in those calls?  Or is there a reason why that would be overly defensive?

stephenh:
Hi Josh,
- I updated PairRDDFunctions.saveAsHadoopFile JobConf to use the default hadoopConfiguration, good call, thanks.
- For the optional jobConf params, I'm ambivalent--I haven't yet fully understood the method overloading approach for the various RDD methods (vs. having default parameters), so can't really comment. Maybe consider that as a separate change for now?
- I believe JobConfs are still not shared across RDD transformations--hadoopConfiguration itself is (after my latest tweak) not shared directly, but passed into new JobConfs/new Configurations each time, which leads to its contents being cloned (in the Configuration base class). So I think we should be okay.

mateiz:
Hi Stephen,

Thanks for doing this -- the approach seems pretty good, although I might like to eventually move all Spark configuration into some kind of "SparkConf" object that has a Hadoop configuration within it. But I think that shouldn't preclude us merging this.

The other thing I notice, though, is that this only deals with the "old Hadoop API" JobConf (the one in the org.apache.mapred package), and not the new API one (org.apache.hadoop.mapreduce). We have a few methods, such as SparkContext.newAPIHadoopFile and PairRDDFunctions.saveAsNewAPIHadoopFile, that can use the new API. Can you pass this configuration object to them as well when the user hasn't set one?

stephenh:
Hi Matei, sorry for the delay, just got behind last week.

I believe I addressed the newAPI methods as you had wanted, and then merged in the latest master.

Let me know if I missed something or misunderstood.

mateiz:
Great, thanks a lot! Merged it in now.

JoshRosen:
This broke a test in the JavaAPISuite:

```
[error] Test spark.JavaAPISuite.writeWithNewAPIHadoopFile failed: null
[error]     at spark.PairRDDFunctions.saveAsNewAPIHadoopFile(PairRDDFunctions.scala:497)
[error]     at spark.PairRDDFunctions.saveAsNewAPIHadoopFile(PairRDDFunctions.scala:497)
[error]     at spark.PairRDDFunctions.saveAsNewAPIHadoopFile(PairRDDFunctions.scala:497)
[...]
```

The latest commit introduced infinite recursion.

stephenh:
Er...okay, looking in to it.

stephenh:
Yeah, shoot, I was trusting the compiler too much to shunt to the method with the default job conf, should have a fix soon.

core/src/main/scala/spark/PairRDDFunctions.scala
```diff
@@ -494,7 +494,7 @@ class PairRDDFunctions[K: ClassManifest, V: ClassManifest](
       keyClass: Class[_],
       valueClass: Class[_],
       outputFormatClass: Class[_ <: NewOutputFormat[_, _]]) {
-    saveAsNewAPIHadoopFile(path, keyClass, valueClass, outputFormatClass, new Configuration)
+    saveAsNewAPIHadoopFile(path, keyClass, valueClass, outputFormatClass)
```
JoshRosen, line 497: This is the source of the infinite recursion.

JoshRosen, line 497: We can just delete this method, since its functionality is subsumed by the method with the default job conf.

stephenh, line 497: Cool, I was going to restore it to be conservative, but agreed that cutting down on the method overloads would be great. Opening a pull request...

PR #363, opened 2013-01-11
add repositories section to quick-start SimpleJob pom.xml

heuermh:
The java SimpleJob quick start example will not build without the repositories section in the pom.xml

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -200,6 +200,16 @@ To build the job, we also write a Maven `pom.xml` file that lists Spark as a dep
   <name>Simple Project</name>
   <packaging>jar</packaging>
   <version>1.0</version>
+  <repositories>
+    <repository>
+      <id>Spray.cc repository</id>
+      <url>http://repo.spray.cc</url>
+    </repository>
+    <repository>
+      <id>Typesafe repository</id>
+      <url>http://repo.typesafe.com/typesafe/releases</url>
+    </repository>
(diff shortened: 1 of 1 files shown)


pwendell:
What version of maven are you using? We have these repositories in the scala quickstart, but I was able to get maven building without specifying the repositories directly. At least in some versions, Maven will follow transitive repository references that are present in our own maven pom.

I'm all for adding this in any case, but just wondering.

heuermh:
$ mvn -version
Apache Maven 3.0.4 (r1232337; 2012-01-17 02:44:56-0600)
Maven home: /Users/xxx/bin/maven-latest
Java version: 1.7.0_12-ea, vendor: Oracle Corporation
Java home: /Library/Java/JavaVirtualMachines/jdk1.7.0_12.jdk/Contents/Home/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "10.7.5", arch: "x86_64", family: "mac"

pwendell:
Ah maybe it was a change in maven 3. LGTM!

heuermh:
Do you have a Nexus proxy set up to mirror those repositories?  That would hide third party repository resolution failures from maven.

pwendell:
That would be really nifty to have in the future, but no, we don't have
that yet.

mateiz:
Thanks for the fix. Let's just have this repos section for now.

PR #364, opened 2013-01-11
Executor and JobDescription JSON support added

tysonjh:
I added the implicit json writers for Executor and JobDescription and modified the jobId query. 

The updated jobId query is optimal because:
1. Array.find method is called on MasterState's completedJobs only if no matching job is found in activeJobs
2. Creating a new collection from the ++ operation is avoided

The downside is that the updated version is arguably less concise and more lines of code. Overall I believe a positive.

### DIFF
--- a/core/src/main/scala/spark/deploy/JsonProtocol.scala
+++ b/core/src/main/scala/spark/deploy/JsonProtocol.scala
@@ -1,6 +1,7 @@
 package spark.deploy
 
 import master.{JobInfo, WorkerInfo}
+import worker.ExecutorRunner
 import cc.spray.json._
 
 /**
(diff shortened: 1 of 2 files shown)


core/src/main/scala/spark/deploy/JsonProtocol.scala
```diff
@@ -51,7 +70,9 @@ private[spark] object JsonProtocol extends DefaultJsonProtocol {
       "cores" -> JsNumber(obj.cores),
       "coresused" -> JsNumber(obj.coresUsed),
       "memory" -> JsNumber(obj.memory),
-      "memoryused" -> JsNumber(obj.memoryUsed)
```
tysonjh, line 54: Not sure why this shows up

mateiz:
Great, thanks Tyson!

PR #365, opened 2013-01-11
Improve sleeping algorithm for rate limiting output streams

ryanlecompte:
This pull request contains minor cleanup, and also a more accurate approach to sleeping when we determine that we need to throttle down the rate at which we are writing bytes. This is based on the way Kafka does it via their Throttler utility class. Previously we were always sleeping a static 5ms when throttling down.

### DIFF
--- a/core/src/main/scala/spark/util/RateLimitedOutputStream.scala
+++ b/core/src/main/scala/spark/util/RateLimitedOutputStream.scala
@@ -31,7 +31,7 @@ class RateLimitedOutputStream(out: OutputStream, bytesPerSec: Int) extends Outpu
   def waitToWrite(numBytes: Int) {
     while (true) {
       val now = System.nanoTime
-      val elapsedSecs = SECONDS.convert(max(now - lastSyncTime, 1), NANOSECONDS)
+      val elapsedSecs = SECONDS.convert(math.max(now - lastSyncTime, 1), NANOSECONDS)
       val rate = bytesWrittenSinceSync.toDouble / elapsedSecs
       if (rate < bytesPerSec) {
         // It's okay to write; just update some variables and return
--- a/core/src/main/scala/spark/util/RateLimitedOutputStream.scala
+++ b/core/src/main/scala/spark/util/RateLimitedOutputStream.scala
@@ -41,6 +41,7 @@ class RateLimitedOutputStream(out: OutputStream, bytesPerSec: Int) extends Outpu
           lastSyncTime = now
           bytesWrittenSinceSync = numBytes
         }
+        return
       } else {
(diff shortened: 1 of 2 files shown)


ryanlecompte:
For reference: https://github.com/kafka-dev/kafka/blob/master/core/src/main/scala/kafka/utils/Throttler.scala

ryanlecompte:
Replaced the missing return. Thanks @karthik7thrn !

ryanlecompte:
FYI - added a unit test for the rate limiting as well.

ryanlecompte:
I was able to get rid of the return after all. :)

core/src/main/scala/spark/util/RateLimitedOutputStream.scala
```diff
@@ -1,9 +1,14 @@
...
 
+import scala.annotation.tailrec
+
 import java.io.OutputStream
+import java.util.concurrent.TimeUnit._
 
 class RateLimitedOutputStream(out: OutputStream, bytesPerSec: Int) extends OutputStream {
-  var lastSyncTime = System.nanoTime()
+  val SyncIntervalNs = NANOSECONDS.convert(10, SECONDS)
+  val ChunkSize = 8192
```
mateiz: Call this CHUNK_SIZE since we've used uppercase constant names in the rest of the code

core/src/test/scala/spark/util/RateLimitedOutputStreamSuite.scala
```diff
@@ -0,0 +1,23 @@
...
+
+class RateLimitedOutputStreamSuite extends FunSuite {
+
+  private def benchmark[U](f: => U): Long = {
+    val start = System.nanoTime
+    f
+    System.nanoTime - start
+  }
+
+  test("write") {
```
mateiz, line 15: I'm a little wary to have tests that depend on real-world time in the code, since things like garbage collection can cause them to fail nondeterministically. Ideally we would use a Clock abstraction here, but we can keep this for now and see whether it breaks.

ryanlecompte:
I just updated to use uppercase/snake-cased constants. Sorry about that, I'm still getting familiar with the Spark code base and had used a pattern that I picked up from the Akka code base.

core/src/test/scala/spark/util/RateLimitedOutputStreamSuite.scala
ryanlecompte, line 15: I thought about using a clock abstraction here also but wasn't sure if Spark had a pattern/way to annotate time-sensitive tests like this or what the general pattern Spark uses for time-sensitive tests. I agree that we can change this to use a Clock abstraction if we see that it breaks, etc.

mateiz:
Great, thanks!

PR #367, opened 2013-01-13
fix spelling mistake in LocalLR example

ericzhang-cn:
fix spelling mistake


mateiz:
Good catch, thanks for the fix!

PR #369, opened 2013-01-14
Java API For Spark Streaming

pwendell:
This patch adds a Java API for Spark Streaming. It is modeled after, and fully integrated with, the existing Spark Java API. This merge includes full API functionality for both normal streams and paired streams. It also includes an example in Java and a comprehensive test suite. The test suite improves overall streaming test coverage while also validating the Java <=> Scala type conversions which constitute most of the logic in the API code.

This currently does not support creating a custom DStream or InputDStream in Java. I'm planning to include that in a subsequent merge that is independently reviewed.

It would be great to get some feedback on this in the next few days as I add the additional tests.


streaming/src/main/scala/spark/streaming/api/java/JavaPairDStream.scala
```diff
@@ -0,0 +1,329 @@
...
+      mergeValue: JFunction2[C, V, C],
+      mergeCombiners: JFunction2[C, C, C],
+      partitioner: Partitioner
+    ): JavaPairDStream[K, C] = {
+    implicit val cm: ClassManifest[C] =
+      implicitly[ClassManifest[AnyRef]].asInstanceOf[ClassManifest[C]]
+    dstream.combineByKey(createCombiner, mergeValue, mergeCombiners, partitioner)
+  }
+
+  def countByKey(numPartitions: Int): JavaPairDStream[K, JLong] = {
```
JoshRosen: We should perform a similar conversion in the regular Spark Java API.  We could do this in a separate pull request, though.

pwendell: I think this may works fine in the normal API, because count() is an action so it directly returns a Long type. In Streaming, count() returns another stream, so it has this issue where Long is actually a parameter of the outer type and the compiler doesn't automate a cast.

JoshRosen: It's still a problem in countByKey() (we already have the fix in countByValue()).

pwendell: oh right, it will be a problem whenever it is a transformation.

JoshRosen:
Could you add checkpoint(), isCheckpointed(), and getCheckpointFile() to JavaRDDLike and add setCheckpointDir() and checkpointFile() to JavaSparkContext?

mateiz:
Did you coordinate with TD on merging this? I don't want us to be merging stuff here that hasn't been reviewed by TD, especially as he was in the process of merging the latest changes in master into the streaming branch.

PR #370, opened 2013-01-14
Added more documentation and minor change in API for NetworkReceiver

tdas:


pwendell:
The rest of the Spark codebase has a very consistent style for comments and the style of the comments here doesn't quite match. Since this is already merged, I just fixed it the streaming-java-api branch. Here are some of the key issues:
1. Spark uses imperative rather than descriptive commenting style. It should say "Create a new DStream" rather than "Creates a new DStream". I've updated all of the streaming docs to use the former.
2. Some comments violate the 100 character style guide.
3. Some of the comments do the following.

```
Class DStream {
  /** Do something to `this` DStream`. */
  def doSomething() {
  }
}
```

Saying "to `this` DStream" is not necessary. It is implicitly true that all methods are being executed on the object in question.
1. Many of the PairDStream docs re-state the core streaming abstraction. That is, the comment itself repeats the fact that stream is created by a sequence of RDD's. The description of DStream's in the class-level docstring is sufficient.

In the API branch I've changed the comments to make them more consistnt with the Spark style. In general, this makes for much shorter comments. E.g:

```
== OLD ==
/**
 * Creates a new DStream by applying `groupByKey` on each RDD of `this` DStream.
 * Therefore, the values for each key in `this` DStream's RDDs are grouped into a
 * single sequence to generate the RDDs of the new DStream. Hash partitioning is
 * used to generate the RDDs with `numPartitions` partitions.
 */

== NEW ==
 /**
  * Create a new DStream by applying `groupByKey` to each RDD. Hash partitioning is used to
  * generate the RDDs with `numPartitions` partitions.
  */
```

PR #371, opened 2013-01-14
Scala 2.10.0

folone:
Moved the project to scala 2.10.0. Changes are:
- akka 2.0.3 → 2.1.0
- spray 1.0-M1 → 1.1-M7
  For now the repl subproject is commented out, as scala reflection api changed very much since the introduction of macros.
  Some discussion on the ML: https://groups.google.com/forum/?fromgroups=#!topic/spark-developers/kauDJdR0rYQ

### DIFF
--- a/core/src/main/scala/spark/CacheTracker.scala
+++ b/core/src/main/scala/spark/CacheTracker.scala
@@ -5,12 +5,12 @@ import scala.collection.mutable.HashMap
 import scala.collection.mutable.HashSet
 
 import akka.actor._
-import akka.dispatch._
+import scala.concurrent.Await
 import akka.pattern.ask
 import akka.remote._
-import akka.util.Duration
+import scala.concurrent.duration.Duration
 import akka.util.Timeout
-import akka.util.duration._
+import scala.concurrent.duration._
 
 import spark.storage.BlockManager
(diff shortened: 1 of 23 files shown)


tomdz:
It would be great if Spark could keep generating 2.9.2 binaries for a little while longer (maybe alongside 2.10 binaries). 2.10 just came out and not everybody is ready to upgrade yet, plus most other libraries don't generate 2.10 binaries yet.

folone:
@tomdz I believe, it will. The branch, I'm sending this pull request to, will not become master any time soon. At least, before the repl subproject is rewritten with the new reflection API. And when it does, I would really like to have a backport/support branch for scala 2.9.

mateiz:
Thanks for doing this, Georgil! It will be very useful to have.

We'll probably keep separate 2.9 and 2.10 branches for a while, and won't make an official release with 2.10 until we fix the REPL (which is kind of complicated in my experience doing this with previous Scala updates, but who knows). But I'll try to keep this branch up to date with changes in master and with the 0.7 release.

mateiz:
The other thing we'll need to do is to fix the Maven build on this branch (the pom.xml files).

folone:
@mateiz I've just altered the maven build in #388.

PR #372, opened 2013-01-15
Removing offset management code that is non-existent in kafka 0.7.0+

mcnamaras:
KafkaInputDStream was referencing offset management code that doesn't exist in Kafka 0.7.0+, and which Kafka does under the hood.  It would cause the KafkaInputDStream to emit a singe message, and would subsequently die.  This fixed the issue for us.

### DIFF
--- a/examples/src/main/scala/spark/streaming/examples/KafkaWordCount.scala
+++ b/examples/src/main/scala/spark/streaming/examples/KafkaWordCount.scala
@@ -13,19 +13,19 @@ import spark.streaming.util.RawTextHelper._
 object KafkaWordCount {
   def main(args: Array[String]) {
     
-    if (args.length < 6) {
-      System.err.println("Usage: KafkaWordCount <master> <hostname> <port> <group> <topics> <numThreads>")
+    if (args.length < 5) {
+      System.err.println("Usage: KafkaWordCount <master> <zkQuorum> <group> <topics> <numThreads>")
       System.exit(1)
     }
 
-    val Array(master, hostname, port, group, topics, numThreads) = args
+    val Array(master, zkQuorum, group, topics, numThreads) = args
 
     val sc = new SparkContext(master, "KafkaWordCount")
     val ssc =  new StreamingContext(sc, Seconds(2))
     ssc.checkpoint("checkpoint")
 
     val topicpMap = topics.split(",").map((_,numThreads.toInt)).toMap
-    val lines = ssc.kafkaStream[String](hostname, port.toInt, group, topicpMap)
+    val lines = ssc.kafkaStream[String](zkQuorum, group, topicpMap)
     val words = lines.flatMap(_.split(" "))
     val wordCounts = words.map(x => (x, 1l)).reduceByKeyAndWindow(add _, subtract _, Minutes(10), Seconds(2), 2)
     wordCounts.print()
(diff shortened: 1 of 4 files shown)


tdas:
Kafka 0.7.2 is included as jar in Spark repository. See ROOT/streaming/lib/kafka-0.7.2.jar . KafkaInputDStream was written and tested by Denny against this version of Kafka. Are you using this jar for running your application or a different Kafka jar and different Kafka version?

mcnamaras:
Ahh yes, that is probably the issue.  We are using kafka 0.7.1, and yes we re using our own packaged jar.  I believe kafka 0.7.1 and 0.7.2 are not compatible, so I'm not sure what we'll have to do to get around this.

mcnamaras:
It makes sense to stay on 0.7.2 since that is forward looking.  So I will self-close this issue.

tdas:
I may still take a look at the code. There were a number of experimental things added to the KafkaInputDStream for master fault tolerance and manipulating offsets was one of them. Since we eliminate a lot of the functionality it maybe that manipulating offsets is just a left over and therefore not necessary for normal operation.

mcnamaras:
Yes, the code looks like it's in a transitive state, and I think what you mention is exactly the case.  We will be doing more with Kafka+spark very soon.  I would be happy to take a closer look at that code in a few days, unless someone beats me to it.  So I think I will re-open this pull request, with the intention of using it as a place for any housekeeping work I may do to it.

tdas:
Alternatively you can post as an issue in the issue tracker. Spark Streaming has its own JIRA tracker now. 
https://spark-project.atlassian.net/browse/STREAMING

mateiz:
I think we should remove the offset code if it's not being used. No point having it in the file -- it will be in the Git history anyway.

tdas:
I completely agree. In fact there are a lot commented code and currently unnecessary classes in that file that should be removed as they are not being used right now. Regarding the particular deletion in this pull request, I think it is safe to remove them for Kafka 0.7.2 as well (obviously its safe for 0.7.0 as Sean has confirmed). I have requested Denny to take a look at this discussion to confirm.

mcnamaras:
I have completed my housekeeping, things look much better now.

Changes:
- Removed all unused/commented out offset management code
- Added kafka jar to ./run script classpath (was preventing kafka example from running)
- Removed zookeeper port from API. It is standard to be inferred from the zkQuorum connect string (see below)
- Previously only a single zookeeper host could be specified.  Quorums are now supported!  Usage examples:
  - zoo
  - zoohost1,zoohost2,...
  - zoo1:2181,zoo2:2181,...
- Updated KafkaWordCount example to reflect changes

I have not yet heard from denny, and he has not responded to e-mail.  I would be happy to go over the code if anyone has any questions or suggestions.

examples/src/main/scala/spark/streaming/examples/KafkaWordCount.scala
```diff
@@ -13,19 +13,19 @@ import spark.streaming.util.RawTextHelper._
 object KafkaWordCount {
```
tdas, line 13: Can you add comments in the file to explain what this example does and how to run this example? You can take a look at the comments in the NetworkWordCount to get an idea of what I am asking for.

tdas:
Sean, this was great too. I just had one minor comment about adding documentation on the KafkaWordCount. I thought its fine to add that later, so I went ahead and merged it. Please do add that documentation though, otherwise its hard for someone to figure out how to run the example.

examples/src/main/scala/spark/streaming/examples/KafkaWordCount.scala
mcnamaras, line 13: Sure! no problem.  I just pushed https://github.com/Reinvigorate/spark/commit/5f4737e4ab764095963c68015df8878d2603cb3e   Do I need to re-open this pull request so that you can pull it in?

mcnamaras:
TD, here is that documentation added:  https://github.com/Reinvigorate/spark/commit/5f4737e4ab764095963c68015df8878d2603cb3e

I am not seeing how to re-open this pull request though.  Should I create a new one so you can get this change merged in?

tdas:
Dont worry, I added the comments. Thank you!

PR #373, opened 2013-01-15
StateDStream changes to give updateStateByKey consistent behavior

mcnamaras:
updateStateByKey only calls updateFunc when data is actively pushed into it.  This was problematic for us when trying to implement session expiration logic using a SynchronizedQueue.  This update fixes the issue, as well as makes the behavior of updateStateByKey consistent.

### DIFF
--- a/streaming/src/main/scala/spark/streaming/StreamingContext.scala
+++ b/streaming/src/main/scala/spark/streaming/StreamingContext.scala
@@ -283,17 +283,31 @@ class StreamingContext private (
   }
 
   /**
-   * Creates a input stream from an queue of RDDs. In each batch,
+   * Creates an input stream from a queue of RDDs. In each batch,
    * it will process either one or all of the RDDs returned by the queue.
    * @param queue      Queue of RDDs
    * @param oneAtATime Whether only one RDD should be consumed from the queue in every interval
-   * @param defaultRDD Default RDD is returned by the DStream when the queue is empty
    * @tparam T         Type of objects in the RDD
    */
   def queueStream[T: ClassManifest](
       queue: Queue[RDD[T]],
-      oneAtATime: Boolean = true,
-      defaultRDD: RDD[T] = null
+      oneAtATime: Boolean = true
+    ): DStream[T] = {
+    queueStream(queue, oneAtATime, sc.makeRDD(Seq[T](), 1))
+  }
+
+  /**
+   * Creates an input stream from a queue of RDDs. In each batch,
+   * it will process either one or all of the RDDs returned by the queue.
+   * @param queue      Queue of RDDs
+   * @param oneAtATime Whether only one RDD should be consumed from the queue in every interval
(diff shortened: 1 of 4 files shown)


tdas:
You are right. The empty Seq()  generates an empty RDD, which will not invoke the None condition as the parentRDD exists, even if null. You need to make sure that no RDD is generated at all. This requires a bit of change. One way this can be done is have null instead of Seq(). And in the TestInputStream's compute function in spark.streaming.TestSuiteBase, make sure that when "selectedInput" is null, None is returned instead of Some(...). That may work. 

Incidentally that leads a different question. Why was your workload seeing None at all?  Ideally an input DStream should produce an RDD every batch, even it is empty as no data was received. What was your input DStream and how was its compute function returning None for some batches?

mcnamaras:
Here is code that exhibits the behavior: https://github.com/Reinvigorate/spark/commit/2e4057f197976ea8be201f69dfd8eadfd7471cd3

It seems to be trigged because QueueInputDStream is returning a None at the very bottom of QueueInputDStream.scala.  So then maybe the issue is only with QueueInputDStream?

tdas:
Yes you are right! QueueInputDStream returns a "default RDD" from compute when the queue has not RDD in it. If you look at the StreamingContext.queueStrema, the defaultRDD is set to null, which causes a None to be generated as from the compute function. I think it is cleaner to ensure that all the input streams generates RDDs at each time interval, even if empty RDD for time when there is not data. So the default RDD should be a empty parallel collection, created by sc.makeRDD(Seq[T](), 1). So I think, the queueStream function should be divided into two functions ... one with one parameter (oneAtATime = true, defaultRDD = sc.makeRDD(...) ) and the second one with the full parameter list. 

And the testcase that I referred to above, would you still be able to write it? 

[BTW, sorry for the delay. I had tried commenting on this 4 days ago but I was having a bad network and it seems that the comment was not posted.]

mcnamaras:
Sounds great!  I will take a stab at getting this and the testcase bits put in.  Thanks TD

> [BTW, sorry for the delay. I had tried commenting on this 4 days ago but I was having a bad network and it seems that the comment was not posted.]

No worries!

mcnamaras:
Please let me know if you have any questions.   What I did in StreamingContext.queueStream makes sense to me, but I think is slightly different than what you described?  Let me know if what I have done makes sense to you.

mcnamaras:
Doesn't look like any work is going off the streaming branch anymore.   I have two pull requests floating out here, should I close and re-issue them on the master branch?

tdas:
Hey Sean, I have been visiting my home country since last week and hence I havent had much time to take a look at stuff. My apologies. I will be back on track starting next week. Streaming is the right branch where I want to stage all the streaming pull request before merging in master. At least for now.

mcnamaras:
TD- ok perfect!  Just wanted to make sure I was still on the same page =)

tdas:
Thanks Sean, this was a good catch and a great fix.

PR #374, opened 2013-01-15
Generate FetchFailedException even for cached missing map outputs

charlesreiss:
Currently, MapOutputTracker generates a FetchFailedException if it fetches new map output locations from the master and finds that at least one map output is missing, but it generates a NullPointerException when the missing map output location is cached locally. (This is likely to occur when two parallel reduce tasks ask for the map output locations at roughly the same time.) This patch generates a FetchFailedException instead (triggering the correct fault-handling code). It also includes some tests which trigger MapOutputTracker's output location fetching code.

### DIFF
--- a/core/src/test/scala/spark/MapOutputTrackerSuite.scala
+++ b/core/src/test/scala/spark/MapOutputTrackerSuite.scala
@@ -105,52 +105,8 @@ class MapOutputTrackerSuite extends FunSuite with BeforeAndAfter {
     masterTracker.incrementGeneration()
     slaveTracker.updateGeneration(masterTracker.getGeneration)
     intercept[FetchFailedException] { slaveTracker.getServerStatuses(10, 0) }
-  }
 
-  test("simulatenous fetch fails") {
-    System.clearProperty("spark.master.host")
-    val dummyActorSystem = ActorSystem("testDummy")
-    val dummyTracker = new MapOutputTracker(dummyActorSystem, true)
-    dummyTracker.registerShuffle(10, 1)
-    // val compressedSize1000 = MapOutputTracker.compressSize(1000L)
-    // val size100 = MapOutputTracker.decompressSize(compressedSize1000)
-    // dummyTracker.registerMapOutput(10, 0, new MapStatus(
-    //   new BlockManagerId("hostA", 1000), Array(compressedSize1000)))
-    val serializedMessage = dummyTracker.getSerializedLocations(10)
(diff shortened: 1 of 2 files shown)


core/src/test/scala/spark/MapOutputTrackerSuite.scala
```diff
@@ -71,6 +77,80 @@ class MapOutputTrackerSuite extends FunSuite {
...
+
+  test("simulatenous fetch fails") {
+    System.clearProperty("spark.master.host")
+    val dummyActorSystem = ActorSystem("testDummy")
+    val dummyTracker = new MapOutputTracker(dummyActorSystem, true)
+    dummyTracker.registerShuffle(10, 1)
+    // val compressedSize1000 = MapOutputTracker.compressSize(1000L)
+    // val size100 = MapOutputTracker.decompressSize(compressedSize1000)
+    // dummyTracker.registerMapOutput(10, 0, new MapStatus(
+    //   new BlockManagerId("hostA", 1000), Array(compressedSize1000)))
```
mateiz: Looks like you left some commented-out code in here

core/src/test/scala/spark/MapOutputTrackerSuite.scala
```diff
@@ -71,6 +77,80 @@ class MapOutputTrackerSuite extends FunSuite {
...
+    val serializedMessage = dummyTracker.getSerializedLocations(10)
+
+    val (actorSystem, boundPort) =
+      AkkaUtils.createActorSystem("test", "localhost", 0)
+    System.setProperty("spark.master.port", boundPort.toString)
```
mateiz: Might be worth explaining in comments how this test works (specifically, that if the intercept in each thread fails, the thread will throw an exception and the firstFailed or secondFailed flag will remain false). Also, what is the purpose of the delayResponseLock -- just to make sure the threads both make a request at a similar time? It seems that Akka won't necessarily guarantee this, unless I'm missing something.

charlesreiss: The goal was to hit the fetching.contains(shuffleId) case -- but looking at that test, it clearly didn't. Guaranteeing that would require adding some instrumentation to MapOutputTracker itself, AFAICT. Removed this second test for now and adding a case of retrieving a cached failure to the previous test.

mateiz:
Great, thanks for the fix.

PR #376, opened 2013-01-15
Python ALS example

MLnick:
This adds a port of the Alternating Least Squares example to Python API. 

I did find it slower than the Scala version, and the RMSE result seems quite different, so let me know if there are any issues.

### DIFF
--- a/python/examples/als.py
+++ b/python/examples/als.py
@@ -1,0 +1,71 @@
+"""
+This example requires numpy (http://www.numpy.org/)
+"""
+from os.path import realpath
+import sys
+
+import numpy as np
+from numpy.random import rand
+from numpy import matrix
+from pyspark import SparkContext
+
+LAMBDA = 0.01   # regularization
+np.random.seed(42)
+
+def rmse(R, ms, us):
+    diff = R - ms * us.T
+    return np.sqrt(np.sum(np.power(diff, 2)) / M * U)
+
+def update(i, vec, mat, ratings):
+    uu = mat.shape[0]
+    ff = mat.shape[1]
+    XtX = matrix(np.zeros((ff, ff)))
+    Xty = np.zeros((ff, 1))
+
+    for j in range(uu):
+        v = mat[j, :]
+        XtX += v.T * v
+        Xty += v.T * ratings[i, j]
+    XtX += np.eye(ff, ff) * LAMBDA * uu
+    return np.linalg.solve(XtX, Xty)
+
+if __name__ == "__main__":
+    if len(sys.argv) < 2:
+        print >> sys.stderr, \
+            "Usage: PythonALS <master> <M> <U> <F> <iters> <slices>"
+        exit(-1)
+    sc = SparkContext(sys.argv[1], "PythonALS", pyFiles=[realpath(__file__)])
+    M = int(sys.argv[2]) if len(sys.argv) > 2 else 100
+    U = int(sys.argv[3]) if len(sys.argv) > 3 else 500
+    F = int(sys.argv[4]) if len(sys.argv) > 4 else 10
+    ITERATIONS = int(sys.argv[5]) if len(sys.argv) > 5 else 5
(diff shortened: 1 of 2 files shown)


python/examples/als.py
```diff
@@ -0,0 +1,71 @@
...
+def update(i, vec, mat, ratings):
+    uu = mat.shape[0]
+    ff = mat.shape[1]
+    XtX = matrix(np.zeros((ff, ff)))
+    Xty = np.zeros((ff, 1))
+
+    for j in range(uu):
+        v = mat[j, :]
+        XtX += v.T * v
+        Xty += v.T * ratings[i, j]
```
JoshRosen, line 28: I don't know if we can replace both `updateUser` and `updateMovie` with a single update function.  Look at how both functions index into the ratings array in the ScalaALS example:

``` scala
def updateMovie(i: Int, ...) {
...
    for (j <- 0 until U) {
    ....
          blas.daxpy(R.get(i, j), u, Xty)
```

and

``` scala
def updateUser(j: Int, ...) {
...
    for (i <- 0 until M) {
    ....
          blas.daxpy(R.get(i, j), m, Xty)
```

In these cases, the position of the loop variable changes in the call to `R.get()`.

In the Python update function, the loop variable always occupies the second position when indexing into `ratings`.  It seems like this will cause range issues.

I'd like to keep the examples consistent across languages, so if you see a way to simplify the ALS example, try implementing the simplifications in the Scala versions first and verify that they do not change that program's output.

python/examples/als.py
```diff
@@ -0,0 +1,71 @@
...
+    XtX += np.eye(ff, ff) * LAMBDA * uu
+    return np.linalg.solve(XtX, Xty)
+
+if __name__ == "__main__":
+    if len(sys.argv) < 2:
+        print >> sys.stderr, \
+            "Usage: PythonALS <master> <M> <U> <F> <iters> <slices>"
```
JoshRosen, line 35: Do you mind modifying ScalaALS so that it has the same argument order and optional arguments?

MLnick, line 35: Will do

python/examples/als.py
MLnick, line 28: The only input which actually needs to be changed is the ratings matrix. So, in the Python version, I passed in the transpose

``` python
us = sc.parallelize(range(U), slices) \
        .map(lambda x: update(x, usb.value[x, :], msb.value, Rb.value.T)) \
```

So indexing into R.T will be correct.

This can be done in the Scala version too - I'll take a look at modifying and checking the result is the same. The downside is of course possibly doubling up on memory for R (unless it's a view - Numpy creates a view into the matrix, not sure about COLT but I'm sure there's a way).

In any case, this approach to ALS is ultimately not scalable since the entire R is broadcast so you're still constrained by single node memory. The approach used in the papers and Mahout is to distribute R twice, once by user and once by item. So in Spark we'd have 2 RDDs of (id, (id, rating)), one grouped by userid, the other grouped by itemid, and cache those.

MLnick:
Results of two runs, first from before commit `a5ba7a9`, the second afterwards.

```
./run spark.examples.SparkALS
Running with M=100, U=500, F=10, iters=5
RMSE = 0.031610830202915136

./run spark.examples.SparkALS
Running with M=100, U=500, F=10, iters=5
RMSE = 0.03131160587463496
```

mateiz:
Have you tested to see whether algebra.transpose is expensive?

MLnick:
Haven't done micro benchmarking, but `time ./run ...` shows fairly negligible difference over a few runs with larger inputs, e.g. `M=1000, U=1000, F=50, iters=5`.

According to the Javadoc, COLT also uses a view into the matrix, so given there's no linear algebra involving R, just indexing into it, this makes sense.

In any case this example uses a dense R where most real-world use cases would be sparse, and the ratings are not optimally distributed, so really it won't scale further than a few thousand users/items, so it's more of an illustrative example.

mateiz:
Alright, thanks for the change. Going to merge this in.

PR #378, opened 2013-01-15
Updated SCALA_VERSION in run2.cmd to match runtime version of Scala

apsaltis:
Changed the SCALA_VERSION to 2.9.2, so that the classes can be found when the classpath is expanded.


mateiz:
Good catch, thanks. Will copy this into branch-0.6 as well.

PR #379, opened 2013-01-15
Add spark.executor.memory to differentiate executor memory from spark-shell

stephenh:
Consider this as a request-for-comment; I'd like to have different heap sizes for my (standalone) cluster's Master, Worker, shell, and Executor processes.

This allows Master, Worker, and spark-shell to use the SPARK_MEM they're given at startup (as currently happens).

But now when running spark-shell, if -Dspark.executor.memory is set, that will be used for that job's executor's heap size on each of the workers.


core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
```diff
@@ -118,8 +118,7 @@ private[spark] class ExecutorRunner(
       for ((key, value) <- jobDesc.command.environment) {
         env.put(key, value)
       }
-      env.put("SPARK_CORES", cores.toString)
```
stephenh, line 116: AFAICT SPARK_CORES isn't read from the environment, but passed in as a command line argument, so I removed this.

core/src/main/scala/spark/scheduler/cluster/SparkDeploySchedulerBackend.scala
```diff
@@ -23,12 +23,11 @@ private[spark] class SparkDeploySchedulerBackend(
...
-      Utils.memoryStringToMb(System.getenv("SPARK_MEM"))
-      // TODO: Might need to add some extra memory for the non-heap parts of the JVM
-    } else {
-      512
-    }
+    // TODO: Might need to add some extra memory for the non-heap parts of the JVM
+    Option(System.getProperty("spark.executor.memory"))
+      .orElse(Option(System.getenv("SPARK_MEM")))
+      .map(Utils.memoryStringToMb)
+      .getOrElse(512)
```
stephenh: I am open to suggestions on the property name--I was trying to see if there was a prefix/namespace common to standalone mode.

As a passing comment, it might eventually be nice to pull all of the System.getProperty/getenv calls out into a separate object--perhaps that is what you had in mind, Matei, when you mentioned a SparkConfig object.

stephenh: Also, instead of a magic system property, I would prefer something more explicit--maybe an optional parameter to the SparkContext constructor, or calling a setter on a SparkConfig that gets passed to SparkContext. 

...huh, I suppose just using the existing SparkContext(environment) parameter (which goes to sc.executorEnvs) would be about the same thing.

stephenh: Thinking some more, it looks like because of the SPARK_MEM/SPARK_MEMORY typo in ExecutorRunner, this executorMemory val was effectively unused, and so the SparkContext.executorEnvs was how SPARK_MEM got shipped over to the executors.

So, the executor can (and was) being sent two ways: one implicitly in the env variables (as SPARK_MEM) and once explicitly in the LaunchExecutor message. Seems like there should be only one way? Which one is better?

I generally prefer more explicit, which is what my patch does (it overwrites any SPARK_ENV environment var with the memoryPerSlave setting).

stephenh: Further musing, but why should any environment variables be sent from the driver program to the executor? Since memoryPerSlave is sent explicitly via the LaunchExecutor message, would you want any other environment variables to really affect the executor?

Seems like the Worker should know its own SPARK_HOME, SPARK_CLASSPATH, etc., such that it would be a bad idea if these driver-side variables (especially if the driver program is spark-shell and so has these variables even if the user hasn't explicitly set them) affects what ExecutorRunner does.

(Note that ExecutorRunner itself taking LaunchExecutor.memory and using that to set SPARK_MEM before calling run still makes sense to me, because SPARK_MEM is like a hidden implementation detail.)

Sorry for the dissertation on what is probably a simple issue.

mateiz: It doesn't make sense to send SPARK_HOME. The other variables were sent because they are meant to be job-specific; for example, SPARK_CLASSPATH can be set to configure different classpaths for each job, SPARK_JAVA_OPTS for JVM options, etc. Ideally these would all be things you can set in a configuration object when you create your SparkContext, but so far, they've been set through environment variables.

Anyway, I want to investigate the SPARK_MEMORY thing because I thought that I'd tested it, and I want to try this patch with Mesos as well. If it works, I think it's good to go.

stephenh: Ah, okay, yeah, per-job classpaths would make sense. I am now intimately familiar with sc.addJar, so had only been thinking of that. And same with the other settings.

Thanks for testing out the patch, I appreciate it.

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -111,8 +111,8 @@ class SparkContext(
 
   // Environment variables to pass to our executors
   private[spark] val executorEnvs = HashMap[String, String]()
-  for (key <- Seq("SPARK_MEM", "SPARK_CLASSPATH", "SPARK_LIBRARY_PATH", "SPARK_JAVA_OPTS",
-       "SPARK_TESTING")) {
+  // Note: SPARK_MEM isn't included because it's set directly in ExecutorRunner
+  for (key <- Seq("SPARK_CLASSPATH", "SPARK_LIBRARY_PATH", "SPARK_JAVA_OPTS", "SPARK_TESTING")) {
```
mateiz: Hey, so it looks like this actually breaks the Mesos mode, because Mesos just launches the spark-executor script directly, and there's no ExecutorRunner involved. However, simply passing SPARK_MEM as in this loop probably isn't right because you added the system property. Instead, just do `executorEnvs("SPARK_MEM") = executorMemory`.

We should also eventually tell users _not_ to set SPARK_MEM in their conf/spark-env.sh. I'll think about whether there's a way to automatically complain if it's set there but it's not super easy because you need to use it sometimes, e.g. when you launch the shell. Maybe the best thing to do is to add a second variable, SPARK_MEMORY, that overrides the value of SPARK_MEM if set; and make this Mesos env code and the ExecutorRunner set that one instead.

stephenh: Hi Matei, thanks for the feedback--let me think about it and I'll work on your suggestion.

mateiz:
Hey Stephen, is this ready to go or are you still working on it? Just saw that you added two new commits (unfortunately I don't get emailed about it unless you comment on the pull request).

stephenh:
Hi Matei, thanks for checking in on this. Technically this branch is working really well for me now, but I want to think about it a bit more. I'll figure something out soon.

stephenh:
Okay, Matei, this is ready to merge. The only change from last time is re-merging master, and then moving the executoryMemory variable up so that it's the same for all of the scheduler backends.

mateiz:
Cool, thanks.

PR #381, opened 2013-01-16
remove unused thread pool

squito:
This threadpool seems unused (unless there is some magic I'm missing somewhere).  My guess is that its been replaced by the threadpool in Executor.


mateiz:
Good catch, thanks.

PR #382, opened 2013-01-18
HttpBroadcast server cache by default in spark.local.dir instead of java.io.tmpdir

fernand:
The HttpBroadcast server cache is by default writing in java.io.tmpdir, which is misleading because you would expect this data to be in the Spark scratch space.


mateiz:
Oh, good catch! This is indeed a problem.

mateiz:
Actually, I realized there's a problem with this pull request: spark.local.dir is allowed to be a comma-separated list of multiple directories (for example, if you have multiple disks), in which case this code won't work. I'll fix it.

mateiz:
Fixed in commit 54c0f9f185576e9b844fa8f81ca410f188daa51c.

PR #383, opened 2013-01-18
Bugfixes to yarn branch

mridulm:
Multiple bugfixes to yarn branch, as mentioned in the commit messages

### DIFF
--- a/core/src/main/scala/spark/util/AkkaUtils.scala
+++ b/core/src/main/scala/spark/util/AkkaUtils.scala
@@ -25,6 +25,8 @@ private[spark] object AkkaUtils {
   def createActorSystem(name: String, host: String, port: Int): (ActorSystem, Int) = {
     val akkaThreads = System.getProperty("spark.akka.threads", "4").toInt
     val akkaBatchSize = System.getProperty("spark.akka.batchSize", "15").toInt
+    val akkaTimeout = System.getProperty("spark.akka.timeout", "20").toInt
+    val akkaFrameSize = System.getProperty("spark.akka.frameSize", "10").toInt
     val akkaConf = ConfigFactory.parseString("""
       akka.daemonic = on
       akka.event-handlers = ["akka.event.slf4j.Slf4jEventHandler"]
@@ -32,14 +34,14 @@ private[spark] object AkkaUtils {
       akka.remote.transport = "akka.remote.netty.NettyRemoteTransport"
       akka.remote.netty.hostname = "%s"
       akka.remote.netty.port = %d
-      akka.remote.netty.connection-timeout = 1s
+      akka.remote.netty.connection-timeout = %ds
+      akka.remote.netty.message-frame-size = %d MiB
       akka.remote.netty.execution-pool-size = %d
       akka.actor.default-dispatcher.throughput = %d
 
-      akka.remote.netty.message-frame-size = 100 MiB
       akka.remote.log-remote-lifecycle-events = on
 
-      """.format(host, port, akkaThreads, akkaBatchSize))
+      """.format(host, port, akkaTimeout, akkaFrameSize, akkaThreads, akkaBatchSize))
 
     val actorSystem = ActorSystem("spark", akkaConf, getClass.getClassLoader)
 
--- a/core/src/main/scala/spark/util/AkkaUtils.scala
+++ b/core/src/main/scala/spark/util/AkkaUtils.scala
@@ -35,6 +35,10 @@ private[spark] object AkkaUtils {
       akka.remote.netty.connection-timeout = 1s
       akka.remote.netty.execution-pool-size = %d
       akka.actor.default-dispatcher.throughput = %d
+
+      akka.remote.netty.message-frame-size = 100 MiB
+      akka.remote.log-remote-lifecycle-events = on
+
       """.format(host, port, akkaThreads, akkaBatchSize))
 
     val actorSystem = ActorSystem("spark", akkaConf, getClass.getClassLoader)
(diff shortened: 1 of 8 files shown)


core/src/main/scala/spark/util/AkkaUtils.scala
```diff
@@ -35,6 +35,10 @@ private[spark] object AkkaUtils {
       akka.remote.netty.connection-timeout = 1s
       akka.remote.netty.execution-pool-size = %d
       akka.actor.default-dispatcher.throughput = %d
+
+      akka.remote.netty.message-frame-size = 100 MiB
+      akka.remote.log-remote-lifecycle-events = on
```
mateiz, line 42: What does turning this on do?

core/src/main/scala/spark/util/AkkaUtils.scala
```diff
@@ -35,6 +35,10 @@ private[spark] object AkkaUtils {
       akka.remote.netty.connection-timeout = 1s
       akka.remote.netty.execution-pool-size = %d
       akka.actor.default-dispatcher.throughput = %d
+
+      akka.remote.netty.message-frame-size = 100 MiB
```
mateiz: We actually added a parameter for this in the code on branch-0.6, called spark.akka.frameSize. Maybe you should copy in the version of the function from there to stay consistent.

core/src/main/scala/spark/util/AkkaUtils.scala
mridulm, line 42: We were getting RemoteClientDisconnected on the master, even though the client(s) were alive.
The problem ended up being that the amount of data slave was sending to master was > 1 Mb (or so) : which caused a netty exception to be thrown, which akka notifies spark via RemoteClientDisconnected.

Ofcourse, getting to this took quite a while since the actual reason was suppressed and not indicated even with TRACE enabled (after fixing log4j handling, which other commits in this set do).

The log-remote-lifecycle-events = on allowed us to finally get to the reason why the RemoteClientDisconnected was getting thrown.

Note that we still have some work to do in this regard : which I will get to later.
When RemoteClientDisconnected is raised, the master does not close the connection : and so slaves continue to run, and no new slave is spawned to replace the one lost - which results in all slaves being lost in a cascade (master thinks they are gone, but yarn knows the slaves are working - so does not reallocate).

core/src/main/scala/spark/util/AkkaUtils.scala
mridulm: Ah ! that sounds like a good idea, let me try to find that and pull it in.
Thanks !

core/src/main/scala/spark/util/AkkaUtils.scala
squito, line 42: On Sat, Jan 19, 2013 at 4:52 AM, Mridul Muralidharan <
notifications@github.com> wrote:

> In core/src/main/scala/spark/util/AkkaUtils.scala:
> 
> > @@ -35,6 +35,10 @@ private[spark] object AkkaUtils {
> >        akka.remote.netty.connection-timeout = 1s
> >        akka.remote.netty.execution-pool-size = %d
> >        akka.actor.default-dispatcher.throughput = %d
> > +
> > -      akka.remote.netty.message-
> >   frame-size = 100 MiB
> > -      akka.remote.log-remote-lifecycle-events = on
> 
> We were getting RemoteClientDisconnected on the master, even though the
> client(s) were alive.
> The problem ended up being that the amount of data slave was sending to
> master was > 1 Mb (or so) : which caused a netty exception to be thrown,
> which akka notifies spark via RemoteClientDisconnected.
> 
> Ofcourse, getting to this took quite a while since the actual reason was
> suppressed and not indicated even with TRACE enabled (after fixing log4j
> handling, which other commits in this set do).
> 
> The log-remote-lifecycle-events = on allowed us to finally get to the
> reason why the RemoteClientDisconnected was getting thrown.
> 
> Note that we still have some work to do in this regard : which I will get
> to later.
> When RemoteClientDisconnected is raised, the master does not close the
> connection : and so slaves continue to run, and no new slave is spawned to
> replace the one lost - which results in all slaves being lost in a cascade
> (master thinks they are gone, but yarn knows the slaves are working - so
> does not reallocate).

Should we also port to the 0.6 & master branches?  We use the standalaone
spark cluster, but it seems like this would help all cases.  Are there
other parts of this commit that are applicable cross-environment?

We've also observed random RemoteClientDisconnecteds, it would be great to
get more information about why that happens.

And, sometimes with the standalone cluster, the StandaloneExecutorBackend
does not get killed when a job dies.  It doesn't happen often, and we
haven't been able to pinpoint the circumstances under which this happens.
It sounds like you may be onto a reason why, which would really help us out.

thanks,
imran

mridulm, line 42: Hi Imran,

  I have been swamped with other issues to resolve (related to our use of spark), but once that is done, I have promised Matei to port our changes (including everything already in yarn branch) to trunk.
That should hopefully address a bunch of issues.

Unfortunately, I do not have a clear timeline on when that would be possible - definitely only after 0.7 is released.
Apologies for the delay !

Regards,
Mridul

squito, line 42: no problem, I understand.  in the meantime I'll just apply

akka.remote.log-remote-lifecycle-events = on

to our own code.
thanks

mridulm, line 42: Please note that (iirc) you need to also have appropriate logging level also set - the message is logged in INFO if I recall right - from some akka package - so that the message can be logged (even if enabled)
Hopefully, i am not misremembering the details ! It was a blur of debugging :-)

mateiz:
Great, thanks!

PR #384, opened 2013-01-18
executor can use a different sparkHome from Worker 

squito:
We'd really like to be able to specify a different sparkHome for an exector, without having to restart the Workers on the cluster.  This would really make it much easier for us to test out small changes for spark -- we'd like to simply leave the Master/Workers running on the cluster, and be able to try out different versions of spark for the executors by setting sparkHome.  We really don't want to just replace the contents of the sparkHome dir, b/c we'd like to have multiple versions around at the same time -- there may be other things using the cluster at the same time as I test out spark versions.

I realize this isn't foolproof -- eventually there may be incompatible versions of the Worker & Executor.  But I'd still like to be able to delay restarting the cluster as much as I can.

I am not sure how this relates to the mesos cluster modes ...

(Without this change, isn't the sparkHome arg to SparkContext completely ignored in standalone clusters?)

### DIFF
--- a/core/src/main/scala/spark/deploy/DeployMessage.scala
+++ b/core/src/main/scala/spark/deploy/DeployMessage.scala
@@ -4,8 +4,6 @@ import spark.deploy.ExecutorState.ExecutorState
 import spark.deploy.master.{WorkerInfo, JobInfo}
 import spark.deploy.worker.ExecutorRunner
 import scala.collection.immutable.List
-import scala.collection.mutable.HashMap
-import java.io.File
 
 
 private[spark] sealed trait DeployMessage extends Serializable
--- a/core/src/main/scala/spark/deploy/DeployMessage.scala
+++ b/core/src/main/scala/spark/deploy/DeployMessage.scala
@@ -5,6 +5,7 @@ import spark.deploy.master.{WorkerInfo, JobInfo}
 import spark.deploy.worker.ExecutorRunner
 import scala.collection.immutable.List
 import scala.collection.mutable.HashMap
+import java.io.File
 
 
 private[spark] sealed trait DeployMessage extends Serializable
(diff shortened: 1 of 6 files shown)


core/src/main/scala/spark/deploy/DeployMessage.scala
```diff
@@ -41,7 +42,8 @@ private[spark] case class LaunchExecutor(
     execId: Int,
     jobDesc: JobDescription,
     cores: Int,
-    memory: Int)
+    memory: Int,
+    sparkHome: File)
```
mateiz: I'd make this a String -- while File is Serializable, it looks a little weird to send it across the network.

mateiz:
Thanks Imran. I'm not sure this is something we'll be able to support in the long run (certainly not across Spark versions for example), but since it's useful to you I think it's fine to include it. I made one comment on the code though.

squito:
I've updated the code to send the dir as a String.

I definitely understand that arbitrary Worker / StandaloneExecutorBackends won't work together.  But I'm just curious -- does the sparkHome arg to SparkContext have any other effect?

thanks

core/src/main/scala/spark/deploy/JobDescription.scala
```diff
@@ -1,10 +1,13 @@
...
 
+import java.io.File
+
 private[spark] class JobDescription(
     val name: String,
     val cores: Int,
     val memoryPerSlave: Int,
-    val command: Command)
+    val command: Command,
+    val sparkHome: File)
```
mateiz: Sorry, this should also be a String for the same reason as above

squito:
oops, sorry I should have caught that.  code updated.

mateiz:
Great, thanks!

mateiz:
I've cherry-picked this into master as well BTW.

PR #385, opened 2013-01-20
Clarifying log directory in EC2 guide

pwendell:


docs/ec2-scripts.md
```diff
@@ -96,7 +96,8 @@ permissions on your private key file, you can run `launch` with the
     `spark-ec2` to attach a persistent EBS volume to each node for
     storing the persistent HDFS.
 -   Finally, if you get errors while running your jobs, look at the slave's logs
-    for that job using the Mesos web UI (`http://<master-hostname>:8080`).
+    for that job inside of the Mesos work directory (/mnt/mesos-work). Mesos errors 
+    can be found using the Mesos web UI (`http://<master-hostname>:8080`).
```
mateiz: The Mesos web UI is not just for Mesos errors; it actually links to the logs of jobs that fail. Maybe you can say something like "you can also view the status of the cluster on the Mesos web UI".

pwendell: Oh I see. The issue I was having is that there was a communications error, so the job was not explicitly failing. So this maybe was why it didn't show up in mesos. Let me change it

mateiz:
Got it, thanks. Yes, the "state != SOME" error from Mesos is confusing.. not sure what to do about it (it's from the C++ code).

pwendell:
One idea would be to add a warning in Spark when the URL is being parsed
using the default parser... something like:

WARN: Unrecognized cluster url: "mesos@foo.bar", attempting to parse as a
"mesos://" URL.

In the programming guide it says to use local, spark://, or mesos://, so I
think it's fair to warn the user if they submit something else.

If you agree, I can submit a patch.

mateiz:
That sounds good. In that case we'll also have to update the AMI though to put mesos:// in the cluster-url file.

pwendell:
Ya it would be ideal to update that file as well (though things will still
work if it is not updated, just with a warning).

PR #386, opened 2013-01-20
Pluggable stream receivers api.

ScrapCodes:


streaming/src/main/scala/spark/streaming/StreamingContext.scala
```diff
@@ -135,6 +140,30 @@ class StreamingContext private (
...
+  }
+
+  /**
+   * Create an input stream with any arbitrary user implemented akka actor receiver.
+   * @param props Props object defining creation of the actor
+   * @param name Name of the actor
+   * @param storageLevel RDD storage level. Defaults to memory-only.
+   */
+  def pluggableActorStream[T: ClassManifest](
+    props: Props, name: String,
```
tdas, line 161: Please name this simply akkaActorStream. That way the name is generic, keeping the API independent of the underlying implementation. Just in case the implementation requires changes in the future.

streaming/src/main/scala/spark/streaming/dstream/KafkaInputDStream.scala
```diff
@@ -31,7 +31,7 @@ case class KafkaDStreamCheckpointData(kafkaRdds: HashMap[Time, Any],
 
 /**
  * Input stream that pulls messages from a Kafka Broker.
- *
```
tdas, line 34: These are changes are all in the white spaces. However, KafkaInputDStream is going to significantly change due to an earlier pending pull request. Once I merge them, I will let you know. And then you will have to merge those changes (mostly accept all the changes to KafkaInputDStream).

streaming/src/main/scala/spark/streaming/receivers/ActorReceiver.scala
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Provides Actors as receivers for receiving stream.
+ *
+ * As Actors can also be used to receive data from almost any stream source.
+ * A nice set of abstraction(s) for actors as receivers is already provided for
+ * a few general cases. It is thus exposed as an API where user may come with
+ * his own Actor to run as receiver for Spark Streaming input source.
+ */
+class ActorReceiver[T: ClassManifest](settings: Settings)
```
tdas, line 52: 1. Please name this AkkaActorReceiver, as this implementation is specific to Akka actors, not any other actors (e.g., someone could implement one for Scala actors).
2. Please explain in the comment how does this supervisor - worker actor model work. It is not clear from the code. You can put appropriate URLs to akka documentation that explain supervisor model and supervisor strategy. 
   http://doc.akka.io/docs/akka/snapshot/scala/fault-tolerance.html
3. Anybody who wants to implement an Actor that links up with this has to know the internals of this receiver, as it has to know the cases Data and PotentiallyHarmful. I dont like that, as it makes the API too tightly tied with the implementation. This is also evident in the WordCount example, the SampleActorReceiver does " context.parent ! Data(msg) "  which is totally non-intuitive unless one understands the internals of the SupervisorActor. Here is an alternate proposition. Implement a trait (say, AkkaActorHelper or something) that has three 3 functions pushData(data: T), pushData(data: Iterator[T]), notifyHiccup/notifyHarmful()  . These will have "context.parent !   ....  " in them. So a person implementing an actor that interfaces with the AkkaActorReceiver, will just have to use this trait and call these 3 interfaces when necessary. The implementation of these 3 functions can change in the future. Does that make sense?

ScrapCodes, line 52: about your first point.
[Scala actors](http://www.scala-lang.org/node/27499) have been deprecated and Akka actors have become scala actors. So in case we change the name now  we would have to do it again when we migrate to scala 2.10. And there is no other [actor framework](http://en.wikipedia.org/wiki/Actor_model#Actor_libraries_and_frameworks) in scala I know off at the moment. 

About supervisor-worker model, its like an actor that starts another actor(worker) who can send data to its parent. And has features for fault tolerance of workers etc.. Is that all I have to mention in comment, Or I missed something?

tdas, line 52: Oops. Sorry I edited the previous post, did not realize that you had already posted a response (did not refresh my page).

tdas, line 52: Okay. Regarding point 1 I accept the argument. ActorReceiver is fine. 
Then, please name the streamingcontext function to actorStream (instead of pluggableActorStream)

Regarding point 2, a very high level idea of the supervisor model and appropriate URLs should be sufficient. So that the akka-supervisor-model-challenged ones (like I am) have enough pointers to understand this.

examples/src/main/scala/spark/streaming/examples/AkkaActorWordCount.scala
```diff
@@ -0,0 +1,80 @@
...
+    val Seq(master, batchDuration, remoteAkkaHost, remoteAkkaPort) = args.toSeq
+
+    // Create the context and set the batch size
+    val ssc = new StreamingContext(master, "AkkaActorWordCount",
+      Seconds(batchDuration.toLong))
+
+    /* 
+     * Following is the use of pluggableActorStream to plug in custom actor as receiver
+     * 
+     * An important point to note:
```
tdas, line 61: Please add this comment about user's responsibility in the scala doc comment of streamingContext.actorStream.

examples/src/main/scala/spark/streaming/examples/FileTextStreamFeeder.scala
```diff
@@ -0,0 +1,63 @@
+package spark.streaming.examples
+
+import java.util.concurrent.CountDownLatch
+
+import scala.collection.mutable.LinkedList
```
tdas, line 5: It would be nice if all the code related to the ActorWordCount example is contained in a single file. So that the examples directory is not cluttered with too many files. In fact a simpler FeederActor may suffice to prove the point - something that generates random text to just one receiver that the wordcount example would create. So it does need any latch or noOfReceivers . That maybe sufficient to show how to write an actor that works with actorStream.

examples/src/main/scala/spark/streaming/examples/AkkaActorWordCount.scala
ScrapCodes, line 61: good point! will do it.

examples/src/main/scala/spark/streaming/examples/FileTextStreamFeeder.scala
ScrapCodes, line 5: This should be fine too!

streaming/src/main/scala/spark/streaming/receivers/ActorReceiver.scala
```diff
@@ -0,0 +1,111 @@
...
+ */
+case class Settings(props: Props,
+  name: String,
+  storageLevel: StorageLevel = StorageLevel.MEMORY_ONLY_SER_2,
+  supervisorStrategy: SupervisorStrategy = ReceiverSupervisorStrategy.defaultStrategy)
+
+/**
+ * Statistcs for querying the supervisor about state of workers
+ */
+case class Statistcs(numberOfMsgs: Int,
```
ryanlecompte, line 36: The name of this class is misspelled.

ScrapCodes, line 36: My bad !, Thanks for pointing out.

tdas:
Shall i close this pull request?

ScrapCodes:
closed

PR #387, opened 2013-01-20
Add accumulators to PySpark

mateiz:
This pull request adds accumulators with the same `value`, `+=` and `AccumulatorParam` interface as the Scala ones to the Python API.


JoshRosen:
Looks good; thanks!

PR #388, opened 2013-01-20
Updated maven build configuration for Scala 2.10

folone:
- updated pom files
- updated 2.10 branch to the latest master state
- fixed the test cases, that were failing before
- left twirl maven plugin at 1.0.2 version, as a version for scala 2.10 is not released yet (or I just did not find it).
  `mvn clean install` now finishes successfully.


mateiz:
Cool, thanks!

PR #389, opened 2013-01-20
Add checkpointing to the Python API

JoshRosen:
This pull request adds checkpointing support to the Python API ([STREAMING-32](https://spark-project.atlassian.net/browse/STREAMING-32)).

### DIFF
--- a/python/pyspark/context.py
+++ b/python/pyspark/context.py
@@ -123,6 +123,10 @@ class SparkContext(object):
         jrdd = self._jsc.textFile(name, minSplits)
         return RDD(jrdd, self)
 
+    def _checkpointFile(self, name):
+        jrdd = self._jsc.checkpointFile(name)
+        return RDD(jrdd, self)
+
     def union(self, rdds):
         """
         Build the union of a list of RDDs.
@@ -145,7 +149,7 @@ class SparkContext(object):
     def accumulator(self, value, accum_param=None):
         """
         Create an C{Accumulator} with the given initial value, using a given
-        AccumulatorParam helper object to define how to add values of the data 
+        AccumulatorParam helper object to define how to add values of the data
         type if provided. Default AccumulatorParams are used for integers and
         floating-point numbers if you do not provide one. For other types, the
         AccumulatorParam must implement two methods:
--- a/python/pyspark/context.py
+++ b/python/pyspark/context.py
@@ -195,3 +195,12 @@ class SparkContext(object):
         filename = path.split("/")[-1]
         os.environ["PYTHONPATH"] = \
             "%s:%s" % (filename, os.environ["PYTHONPATH"])
+
+    def setCheckpointDir(self, dirName, useExisting=False):
+        """
+        Set the directory under which RDDs are going to be checkpointed. This
+        method will create this directory and will throw an exception of the
+        path already exists (to avoid overwriting existing files may be
+        overwritten). The directory will be deleted on exit if indicated.
+        """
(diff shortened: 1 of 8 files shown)


python/pyspark/context.py
```diff
@@ -195,3 +199,12 @@ def addPyFile(self, path):
         filename = path.split("/")[-1]
         os.environ["PYTHONPATH"] = \
             "%s:%s" % (filename, os.environ["PYTHONPATH"])
+
+    def setCheckpointDir(self, dirName, useExisting=False):
+        """
+        Set the directory under which RDDs are going to be checkpointed. This
+        method will create this directory and will throw an exception of the
+        path already exists (to avoid overwriting existing files may be
+        overwritten). The directory will be deleted on exit if indicated.
```
mateiz: Grammar error here? (to avoid overwriting existing files may be overwritten)

python/pyspark/rdd.py
```diff
@@ -49,6 +50,41 @@ def cache(self):
...
+        lineages.  In the current implementation, Spark will save this RDD to
+        a file (using saveAsObjectFile()) after the first job using this RDD is
+        done.  Hence, it is strongly recommended to use checkpoint() on RDDs
+        when
+
+        (i) checkpoint() is called before the any job has been executed on this
+        RDD.
+
+        (ii) This RDD has been made to persist in memory. Otherwise saving it
+        on a file will require recomputation.
```
mateiz, line 60: Will these format correctly as a list in Epydoc?

python/pyspark/context.py
JoshRosen: I copied this description from JavaSparkContext, so I'll fix it there too.

python/pyspark/rdd.py
```diff
@@ -707,6 +744,10 @@ def batched_func(split, iterator):
         self._jrdd_val = python_rdd.asJavaRDD()
         return self._jrdd_val
 
+    @property
```
mateiz: I'd make this a method instead of a property because otherwise it looks at first glance like its value might never change (in code using it)

python/pyspark/tests.py
```diff
@@ -0,0 +1,70 @@
...
+
+        flatMappedRDD.checkpoint()
+        result = flatMappedRDD.collect()
+        time.sleep(1)  # 1 second
+        self.assertTrue(flatMappedRDD.isCheckpointed())
+        self.assertEqual(flatMappedRDD.collect(), result)
+        self.assertEqual(checkpointDir.name,
+                         os.path.dirname(flatMappedRDD.getCheckpointFile()))
+
+        atexit.register(lambda: shutil.rmtree(checkpointDir.name))
```
mateiz: Why is this here instead of right after the directory is created? If the test fails above this, the directory won't be deleted. (Not really the end of the world but kind of weird). In fact it might be good to just do it in setUp / tearDown.

python/pyspark/rdd.py
JoshRosen, line 60: Good point.  I noticed that this description (copied from JavaRDDLike) contains usage suggestions that are missing from the regular RDD.checkpoint().  Should I copy the extended description back to RDD.checkpoint()?

JoshRosen, line 60: Actually, I'll just use RDD's description since it's newer and port it back to JavaRDDLike.

JoshRosen:
I updated the documentation and code; please take a look.

mateiz:
Great, thanks.

PR #390, opened 2013-01-20
Fix PythonPartitioner equality

JoshRosen:
PythonPartitioner did not take the Python-side partitioning function into account when checking for equality; see [SPARK-654](https://spark-project.atlassian.net/browse/SPARK-654)


mateiz:
Looks good, thanks.

PR #391, opened 2013-01-21
Minor cleanup.

stephenh:
Hey, sorry if this is noise, I've been reading the codebase today and just cleaning very minor things up as I see them. If you don't want to be bothered with this small stuff, let me know.

Two questions:

1) It seems like a lot of vars/defs that could be private are left public--is this intentional? When trying to read the codebase, I like as many "private" fields/methods as possible, because it helps me understand the object's true API. Would you like me to add private modifiers as I come across places that can have them?

2) It seems like "splits_", with the underscore last to denote the field, is a common convention, but it makes for odd type declarations because then you have to do "splits_ : Type" instead of just "_splits: Type" (note the extra space because "splits_:" makes the colon a part of the identifier). This probably doesn't matter, but I don't mind doing trivial/boring changes like s/split_/_split/ if it's okay with you guys.

### DIFF
--- a/core/src/main/scala/spark/ParallelCollection.scala
+++ b/core/src/main/scala/spark/ParallelCollection.scala
@@ -23,39 +23,34 @@ private[spark] class ParallelCollectionSplit[T: ClassManifest](
 }
 
 private[spark] class ParallelCollection[T: ClassManifest](
-    @transient sc : SparkContext,
+    @transient sc: SparkContext,
     @transient data: Seq[T],
     numSlices: Int,
-    locationPrefs : Map[Int,Seq[String]])
+    locationPrefs: Map[Int,Seq[String]])
   extends RDD[T](sc, Nil) {
   // TODO: Right now, each split sends along its full data, even if later down the RDD chain it gets
   // cached. It might be worthwhile to write the data to a file in the DFS and read it in the split
   // instead.
   // UPDATE: A parallel collection can be checkpointed to HDFS, which achieves this goal.
 
-  @transient
-  var splits_ : Array[Split] = {
+  @transient var splits_ : Array[Split] = {
     val slices = ParallelCollection.slice(data, numSlices).toArray
     slices.indices.map(i => new ParallelCollectionSplit(id, i, slices(i))).toArray
   }
 
-  override def getSplits = splits_.asInstanceOf[Array[Split]]
+  override def getSplits = splits_
 
   override def compute(s: Split, context: TaskContext) =
     s.asInstanceOf[ParallelCollectionSplit[T]].iterator
 
   override def getPreferredLocations(s: Split): Seq[String] = {
-    locationPrefs.get(s.index) match {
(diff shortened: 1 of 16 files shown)


JoshRosen:
A while back, we had a pull request that made a bunch of classes private: #240 and #244.  Some classes were deliberately left public because Shark depends on them.  So perhaps some of the public fields that you're noticing are required for Shark support (I'd have to see specific examples to say definitively).

mateiz:
Hey Stephen,

Cleanup is appreciated, but I would suggest waiting until after the 0.7 release, because there are some other things we'll need to change in the codebase before that and I don't want to create a lot of conflicts. In terms of the specific changes suggested:
- Underscore in front of names seems fine. So does `@transient`.
- I'd like to look more carefully at the "private" issue and see what other Scala projects do. (For example, is there something on this in the Scala style guide?) Due to Scala's concise nature, it's been very tempting to just write methods with `def`, as in Ruby or Python. However, it may indeed be good to do this for future readability. The one place where we have used `private` on methods has been for user-facing classes, like `SparkContext`, since we don't want internal things to show up in Scaladocs, IDEs, etc. The remaining non-private methods should all be in `private[spark]` classes.

core/src/main/scala/spark/ParallelCollection.scala
```diff
@@ -23,32 +23,28 @@ private[spark] class ParallelCollectionSplit[T: ClassManifest](
...
 
   override def compute(s: Split, context: TaskContext) =
     s.asInstanceOf[ParallelCollectionSplit[T]].iterator
 
   override def getPreferredLocations(s: Split): Seq[String] = {
-    locationPrefs.get(s.index) match {
-      case Some(s) => s
-      case _ => Nil
-    }
+    locationPrefs.get(s.index) getOrElse Nil
```
mateiz: This can just be `locationPrefs.getOrElse(s.index, Nil)`

mateiz:
By the way, the changes in this pull request look good, so I'd like to merge them, modulo one small comment I made. I like the `getOrElse(throw exception)` stuff.. haven't seen it before but it makes a lot of sense.

stephenh:
> Some classes were deliberately left public because Shark depends on them

Ah, okay, I didn't think of that, but that makes sense. Perhaps this is too much in retrospect/backseat-programming, but maybe a "VisibleForShark" type annotation would be interesting to highlight purposefully-left public methods.

> I would suggest waiting until after the 0.7 release

Cool, I'll hold off on anything else.

> I'd like to look more carefully at the "private" issue

Yeah, the brevity is admittedly nice. But as a foreigner in a codebase, differentiating between "used by other classes" and "just a helper method" is really useful. Perhaps the distinction is not as important once you're more familiar with a codebase. Not sure if there is a Scala consensus around this or not.

mateiz:
I've merged this manually due to a small conflict. Thanks Stephen.

PR #392, opened 2013-01-21
Move JavaAPISuite into spark.streaming.

stephenh:
Noticed that on master JavaAPISuite wasn't compiling for me because it seemed to be in the wrong location.


mateiz:
Weird; it's definitely run for me, but maybe it depends on whether you use Maven or SBT. Which one did you use? In any case it's okay to move it.

stephenh:
Eclipse was complaining about it. I didn't try the command line.

I believe being in the root package would be fine, but the JavaAPISuite.java had a "package spark.streaming" at the top of it, which isn't technically Java spec, but perhaps Maven/SBT don't care.

I was considering removing the "package spark.streaming", but it looked like other projects in the repo had their JavaAPISuite in a non-root package, so followed suit.

mateiz:
No, it's good to move it. Thanks for the clarification.

PR #394, opened 2013-01-22
Add SparkFiles.get() API to access files added through addFile().

JoshRosen:
This pull request adds a SparkFiles object that is used to access files added with addFile().  This allows us to break the assumption that these files are downloaded to the current working directory, which caused problems when dependencies were fetched on the master / driver machine: https://groups.google.com/d/topic/spark-developers/BG8a9guHgG4/discussion

This change was originally proposed in #345.  It is backwards-incompatible, but upgrading user code should be pretty simple.

It also fixes [SPARK-662](https://spark-project.atlassian.net/browse/SPARK-662) by adding synchronization to Executor.updateDependencies().

I plan to port #345 and the synchronization fix back to branch-0.6, but I'll do that in a separate pull request.

I also deleted some unused code.

### DIFF
--- a/core/src/main/scala/spark/SparkFiles.java
+++ b/core/src/main/scala/spark/SparkFiles.java
@@ -3,21 +3,21 @@ package spark;
 import java.io.File;
 
 /**
- * Resolves paths to files added through `addFile().
+ * Resolves paths to files added through `SparkContext.addFile()`.
  */
 public class SparkFiles {
 
   private SparkFiles() {}
 
   /**
-   * Get the absolute path of a file added through `addFile()`.
+   * Get the absolute path of a file added through `SparkContext.addFile()`.
    */
   public static String get(String filename) {
     return new File(getRootDirectory(), filename).getAbsolutePath();
   }
 
   /**
-   * Get the root directory that contains files added through `addFile()`.
+   * Get the root directory that contains files added through `SparkContext.addFile()`.
    */
   public static String getRootDirectory() {
     return SparkEnv.get().sparkFilesDir();
--- a/core/src/main/scala/spark/SparkFiles.java
+++ b/core/src/main/scala/spark/SparkFiles.java
@@ -1,0 +1,25 @@
+package spark;
+
+import java.io.File;
+
+/**
+ * Resolves paths to files added through `addFile().
+ */
+public class SparkFiles {
+
+  private SparkFiles() {}
+
+  /**
+   * Get the absolute path of a file added through `addFile()`.
+   */
+  public static String get(String filename) {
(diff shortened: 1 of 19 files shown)


JoshRosen:
I just realized that this introduces a bug in SparkContext.addPyFile: we need to append to the PYTHONPATH on the worker machine because we can no longer assume that these files will be in the current working directory.  Working on a fix now.

JoshRosen:
I fixed the bug and added a test, so this should be ready for review.

core/src/main/scala/spark/SparkFiles.java
```diff
@@ -0,0 +1,25 @@
+package spark;
+
+import java.io.File;
+
+/**
+ * Resolves paths to files added through `addFile().
```
mateiz: Missing a closing tick

core/src/main/scala/spark/SparkFiles.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+public class SparkFiles {
+
+  private SparkFiles() {}
+
+  /**
+   * Get the absolute path of a file added through `addFile()`.
```
mateiz: Change this comment to say SparkContext.addFile to give more context; same with the one below

python/pyspark/files.py
```diff
@@ -0,0 +1,24 @@
+import os
+
+
+class SparkFiles(object):
```
mateiz, line 4: Why is there no getRootDirectory() here?

mateiz:
Made a few small comments.

python/pyspark/files.py
JoshRosen, line 4: I figured that it was only necessary for backwards-compatibility.

mateiz, line 4: I think we should keep it to match the Java/Scala API. The fewer surprises the better. Who knows, people might also want to list the directory and stuff.

JoshRosen:
I just realized that the PySpark version of SparkFiles may not work correctly if it is used inside the user's Python driver code (e.g. outside of closures that run in the Python worker process).  It shouldn't be too hard to support this, though, since I can add code that detects whether we're running on the worker or the driver, allowing PySpark's SparkFiles to directly call its Java counterpart when running in the driver.

python/pyspark/files.py
JoshRosen, line 4: I added this in the latest commit.

JoshRosen:
I added support for using SparkFiles in the Python driver and addressed the documentation typos.

mateiz:
Great, thanks.

PR #395, opened 2013-01-22
Remove unneeded/now-broken saveAsNewAPIHadoopFile overload.

stephenh:
Kudos to @JoshRosen for pointing out the error and fix.


mateiz:
Great, thanks.

PR #396, opened 2013-01-22
Make PySpark AccumulatorParam an abstract base class

JoshRosen:
This commit makes AccumulatorParam an abstract base class, using the [abc](http://docs.python.org/2/library/abc.html) module.  This allows us to use epydoc to document this interface.  I also fixed an epydoc formatting issue in SparkContext.accumulate().

### DIFF
--- a/python/pyspark/accumulators.py
+++ b/python/pyspark/accumulators.py
@@ -25,7 +25,8 @@
 >>> a.value
 13
 
->>> class VectorAccumulatorParam(object):
+>>> from pyspark.accumulators import AccumulatorParam
+>>> class VectorAccumulatorParam(AccumulatorParam):
 ...     def zero(self, value):
 ...         return [0.0] * len(value)
 ...     def addInPlace(self, val1, val2):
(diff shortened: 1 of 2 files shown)


mateiz:
Looks good, but don't you need to change the doctest that creates a VectorAccumulatorParam too?

By the way, it would be nice to add a small section on this in the Python documentation page (maybe with that vector example).

JoshRosen:
Regarding the Python documentation, how much content do you want to duplicate between the Scala / Java / Python programming guides?  The Java programming guide doesn't provide any examples of accumulators or broadcast variables and the Scala one doesn't explain how to implement custom accumulators.

JoshRosen:
Cloudpickle seems to have problems when serializing AccumulatorParam subclasses, so I stopped using ABCMeta.

mateiz:
I would err on the side of including more in the docs, not less. We can add AccumulatorParam in the Scala ones too eventually (but it doesn't have to be in this pull request).

mateiz:
Thanks Josh!

PR #397, opened 2013-01-22
Refactor daemon thread creation

JoshRosen:
Spark has multiple methods for creating daemon thread pools and thread factories, but we can eliminate most of this code using Guava's [ThreadFactoryBuilder](http://docs.guava-libraries.googlecode.com/git-history/release/javadoc/com/google/common/util/concurrent/ThreadFactoryBuilder.html).

It looks like 2d761e3353651049f6707c74bb5ffdd6e86f6f35 might have broken DaemonThreadFactory so that it no longer creates daemon threads.  I wound up removing this class, since it was unnecessary.

Utils had its own `newDaemonThreadFactory` method, which I also removed.


core/src/main/scala/spark/Utils.scala
```diff
@@ -287,29 +288,14 @@ private object Utils extends Logging {
...
-  private def newDaemonThreadFactory: ThreadFactory = {
-    new ThreadFactory {
-      def newThread(r: Runnable): Thread = {
-        var t = Executors.defaultThreadFactory.newThread (r)
-        t.setDaemon (true)
-        return t
-      }
-    }
-  }
+  private[spark] val daemonThreadFactory: ThreadFactory =
```
ryanlecompte, line 291: Might want to set the uncaughtExceptionHandler as well: http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/util/concurrent/ThreadFactoryBuilder.html#setUncaughtExceptionHandler(java.lang.Thread.UncaughtExceptionHandler)

JoshRosen, line 291: @woggling added a default uncaught exception handler in #305 but this only affected worker machines.  This handler ensures that Executors die rather than hanging if they run out of memory or experience other types of failures.  I'm not sure if we want to implement the same "fail fast" behavior in the master or driver, since they might be embedded with other user code that doesn't expect that behavior.

mateiz:
Cool, thanks for the fix.

PR #398, opened 2013-01-22
Actor as receiver.

ScrapCodes:
This pull request provides Actor as stream receivers for spark streaming.

### DIFF
--- a/examples/src/main/scala/spark/streaming/examples/ActorWordCount.scala
+++ b/examples/src/main/scala/spark/streaming/examples/ActorWordCount.scala
@@ -1,0 +1,130 @@
+package spark.streaming.examples
+
+import scala.collection.mutable.LinkedList
+import scala.util.Random
+
+import akka.actor.Actor
+import akka.actor.ActorRef
+import akka.actor.Props
+import akka.actor.actorRef2Scala
+
+import spark.streaming.Seconds
+import spark.streaming.StreamingContext
+import spark.streaming.StreamingContext.toPairDStreamFunctions
+import spark.streaming.receivers.Receiver
+import spark.util.AkkaUtils
+
+case class SubscribeReceiver(receiverActor: ActorRef)
+case class UnsubscribeReceiver(receiverActor: ActorRef)
+
+/**
+ * Sends the random content to every receiver subscribed with 1/2
+ *  second delay.
+ */
+class FeederActor extends Actor {
+
+  val rand = new Random()
+  var receivers: LinkedList[ActorRef] = new LinkedList[ActorRef]()
+
+  val strings: Array[String] = Array("words ", "may ", "count ")
+
+  def makeMessage(): String = {
+    val x = rand.nextInt(3)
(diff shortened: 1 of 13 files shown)


examples/src/main/scala/spark/streaming/examples/ActorWordCount.scala
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  override def postStop() = remotePublisher ! UnsubscribeReceiver(context.self)
+
+}
+
+/**
+ * A sample word count program demonstrating the use of plugging in
+ * Actor as Receiver
+ */
```
tdas, line 121: Very minor point. Can you add some more comments on what the command line parameters should be. Please take a look at the NetworkWordCount example to get an idea.

ScrapCodes, line 121: Alrite! Will fix this.

examples/src/main/scala/spark/streaming/examples/ActorWordCount.scala
```diff
@@ -0,0 +1,136 @@
...
+}
+
+/**
+ * A sample word count program demonstrating the use of plugging in
+ * Actor as Receiver
+ * Usage: ActorWordCount <master> <hostname> <port>
+ *   <master> is the Spark master URL. In local mode, <master> should be 'local[n]' with n > 1.
+ *   <hostname> and <port> describe the AkkaSystem that Spark Sample feeder would work on.
+ * 
+ * and then run the example
```
tdas, line 119: Please remove "and then"

examples/src/main/scala/spark/streaming/examples/ActorWordCount.scala
```diff
@@ -0,0 +1,136 @@
...
+  override def postStop() = remotePublisher ! UnsubscribeReceiver(context.self)
+
+}
+
+/**
+ * A sample word count program demonstrating the use of plugging in
+ * Actor as Receiver
+ * Usage: ActorWordCount <master> <hostname> <port>
+ *   <master> is the Spark master URL. In local mode, <master> should be 'local[n]' with n > 1.
+ *   <hostname> and <port> describe the AkkaSystem that Spark Sample feeder would work on.
```
tdas: Since this example starts the feeder actor locally and actorStream does not connect to any remote feeder actor, the <hostname> is always going to be localhost. So it does not make sense to have hostname as a command line parameter. There are two ways that makes sense - 
(1) either remove the <hostname> as a command-line parameter and have it always be "localhost" in the code.
(2) or, allow the Feeder actor to be start separately from the command line (using another object and main function) and then the ActorWordCount can take the <hostname> + <port> to connect to the feeder. This is much like the nc and NetworkWordCount. And its also more practical usecase that running locally and therefore cooler.

ScrapCodes:
is this planned as part of .7 release?

tdas:
I am closing this pull request. Thank you very much!!

PR #399, opened 2013-01-22
Fix for hanging spark.HttpFileServer on the kind of virtual network

Leemoonsoo:
Some kind of virtual network make spark.HttpFileServer not work.

this is pull request of 

https://groups.google.com/d/topic/shark-users/9x580n5NOfM/discussion


mateiz:
Awesome, thanks for taking the time to send a pull request.

PR #400, opened 2013-01-22
Use default arguments instead of constructor overloads.

stephenh:


mateiz:
Let's keep it as a String for now. I'd ultimately like to replace the way we configure Spark with a SparkConf object that is passed to SparkContext (https://spark-project.atlassian.net/browse/SPARK-544), which would allow a nice way to set multiple parameters (e.g. `new SparkConf.setHome(...).setJars(...)`). It would also get us away from system properties. That's when I'd consider changing the API, although even then, it seems that supporting the current versions with 2, 4 or 5 parameters is not bad.

mateiz:
(By the way, I don't think we've ever broken an API yet in Spark releases -- the worst we've done is change some configuration parameters. It won't be possible to do this forever, but I'd like to try because I think users appreciate it.)

PR #401, opened 2013-01-22
Blockmanager ui

squito:
This is an update to https://github.com/mesos/spark/pull/295

I didn't add any additional functionality beyond just bringing it up to date w/ master; I agree additional changes would be nice, but wanted to get this much merged in at least.

Comments greatly appreciated -- some parts of the merge were a bit ugly, and this section of the code is totally new to me.


core/src/main/scala/spark/RDD.scala
```diff
@@ -94,6 +94,9 @@ abstract class RDD[T: ClassManifest](
   /** How this RDD depends on any parent RDDs. */
   protected def getDependencies(): List[Dependency[_]] = dependencies_
 
+  // A friendly name for this RDD
```
pwendell: Nit: use /*\* */ style comments as in the rest of this file

core/src/main/scala/spark/RDD.scala
```diff
@@ -108,7 +111,13 @@ abstract class RDD[T: ClassManifest](
   /** A unique ID for this RDD (within its SparkContext). */
   val id = sc.newRddId()
 
-  /**
+  /* Assign a name to this RDD */
```
pwendell: Nit: use /*\* */ style comments as in the rest of this file

core/src/main/scala/spark/RDD.scala
```diff
@@ -119,6 +128,8 @@ abstract class RDD[T: ClassManifest](
         "Cannot change storage level of an RDD after it was already assigned a level")
     }
     storageLevel = newLevel
+    // Register the RDD with the SparkContext
+    sc.persistentRdds(id) = this
```
pwendell, line 132: One issue with the introduction of streaming is how to deal with state that accumulates over the course of a long-running job. In this case, it seems like this particular HashMap only gets larger, which will pose a problem. I'm going to defer to TD or Matei here to figure out how to best address this particular case. TD has a handful of MetaDataCleaner implementations which go through and delete things based on a timestamp (checkout DAGScheduler), that may be appropriate to use here.

pwendell, line 132: Just looked closer, and the best way is definitely to use the special HashMap and cleaner that are used in DAGScheduler.

core/src/main/scala/spark/storage/BlockManagerUI.scala
```diff
@@ -0,0 +1,86 @@
...
+        
+        // Calculate macro-level statistics
+        val maxMem = storageStatusList.map(_.maxMem).reduce(_+_)
+        val remainingMem = storageStatusList.map(_.memRemaining).reduce(_+_)
+        val diskSpaceUsed = storageStatusList.flatMap(_.blocks.values.map(_.diskSize))
+          .reduceOption(_+_).getOrElse(0L)
+
+        val rdds = StorageUtils.rddInfoFromStorageStatus(storageStatusList, sc)
+
+        spark.storage.html.index.render(maxMem, remainingMem, diskSpaceUsed, rdds, storageStatusList)
```
pwendell: This line exceeds 100 characters.

core/src/main/scala/spark/storage/BlockManagerUI.scala
```diff
@@ -0,0 +1,86 @@
...
+      }
+    }}} ~
+    get { path("rdd") { parameter("id") { id => { completeWith {
+      val future = master ? GetStorageStatus
+      future.map { status =>
+        val prefix = "rdd_" + id.toString
+
+
+        val storageStatusList = status.asInstanceOf[ArrayBuffer[StorageStatus]].toArray
+        val filteredStorageStatusList = StorageUtils.filterStorageStatusByPrefix(storageStatusList, prefix)
```
pwendell: > 100 characters

pwendell:
Thanks for bringing this up to date! Overall, looks good, with a few small nits. Would be great to get this in as a base for future UI work.

core/src/main/scala/spark/RDD.scala
JoshRosen, line 132: Is simply purging values that are older than a threshold the right approach here?  It seems like most of the space overhead would come from the RDDs rather than the integer HashMap keys.  Purging old values will only benefit us if we can garbage-collect those values. 

Imagine that I have a job that runs hundreds of iterations of an algorithm over a cached data set, which might take a long time to complete.  A naive timeout would purge this cached long-lived RDD from the HashMap even though there are other references keeping it alive, so we unnecessarily lose the ability to display its information.

On the other hand, we don't want to add strong references that would prevent RDDs from being garbage-collected.

Maybe we could use a WeakConcurrentHashMap, which would address this problem.  It looks like Guava's [MapMaker](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/MapMaker.html) can construct these, but I don't have personal experience with it.

This approach doesn't address cases where the cache contains blocks from an RDD that has been garbage-collected.  I suppose that we could use MapMaker to make a map that uses soft references, but I don't know how this performs in practice.

pwendell, line 132: How to deal with cleaning meta-data correctly is a complicated issue and one which is not yet completely addressed in the codebase. The reason this is necessary is streaming jobs which might create thousands of RDD's over a short period of time, so any state really needs to be garbage collected periodically. Otherwise hashmaps like this bloat to the point where they start to impact performance.

For now, TD has introduced these helper data structures which use a timeout based approach. He's added them throughout the codebase, but they are enabled only in streaming mode. In normal batch mode, it acts just like any normal concurrent hashmap. 

Figuring out the correct GC behavior for all types of state is something TD is working on right now, but in the meantime this timeout approach is being employed.

squito:
I fixed the formatting issues, and changed persistentRdds to be a TimeStampedHashMap.  However, its not getting cleaned up anywhere yet -- I wasn't sure if I should move that logic up from DAGScheduler into SparkContext, push the map down to DAGScheduler, or some other refactoring ... thoughts?

Also, as I've played around with this a bit, it seems odd to me that the UI is hosted by the driver, not the master.  (At the very least, the master UI should link to the Storage UI.)  maybe its ok as is for the moment, though ...

squito:
oops, sorry this is not ready yet, some stupid mistakes ...

squito:
fixed the compile errors, but still not sure about the where to clear the persistentRdd values

squito:
I think this is ready to go now.

(The maven build worked, but I had to give it more memory "MAVEN_OPTS=-Xmx1024m mvn -e compile", not sure if that has always been the case, or something from this change ...)

pwendell:
Looks good to me! Thanks for the contribution.

mateiz:
Alright, merged it. Thanks.

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -645,6 +658,12 @@ class SparkContext(
 
   /** Register a new RDD, returning its RDD ID */
   private[spark] def newRddId(): Int = nextRddId.getAndIncrement()
+
+  private[spark] def cleanup(cleanupTime: Long) {
+    var sizeBefore = persistentRdds.size
+    persistentRdds.clearOldValues(cleanupTime)
+    logInfo("idToStage " + sizeBefore + " --> " + persistentRdds.size)
```
stephenh, line 665: idToStage? Copy/paste error?

mateiz, line 665: Hmm yeah good catch.. I'll fix that

PR #402, opened 2013-01-22
Handle slaveLost before slaveIdToHost knows about it.

stephenh:
I was unintentionally being dumb, and causing executors to not get started on the workers, but that led to slaveLost being called before resourceOffers had put the slave in the slaveItToHost map.

This just adds a logError for that case.


core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
```diff
@@ -252,19 +252,24 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
   def slaveLost(slaveId: String, reason: ExecutorLossReason) {
     var failedHost: Option[String] = None
     synchronized {
-      val host = slaveIdToHost(slaveId)
```
stephenh, line 255: This was causing a NoSuchElementException.

mateiz:
Cool, thanks for the fix!

PR #403, opened 2013-01-23
Call removeJob after job failure instead of killing the Master

stephenh:
Again related to my recent incident of being dumb, when executors couldn't be started after 10 times (MAX_NUM_RETRY), the Master threw an exception and the whole cluster died.

This was fairly surprising to me; I was expecting the master/slaves to stay up.

My naive understanding of Master makes it look like just calling removeJob is sensible and the cluster can keep on running.

Does this seem right?

### DIFF
--- a/core/src/main/scala/spark/deploy/worker/Worker.scala
+++ b/core/src/main/scala/spark/deploy/worker/Worker.scala
@@ -143,9 +143,13 @@ private[spark] class Worker(
 
     case KillExecutor(jobId, execId) =>
       val fullId = jobId + "/" + execId
-      val executor = executors(fullId)
-      logInfo("Asked to kill executor " + fullId)
-      executor.kill()
+      executors.get(fullId) match {
+        case Some(executor) =>
+          logInfo("Asked to kill executor " + fullId)
+          executor.kill()
+        case None =>
+          logInfo("Asked to kill non-existent existent " + fullId)
+      }
 
     case Terminated(_) | RemoteClientDisconnected(_, _) | RemoteClientShutdown(_, _) =>
       masterDisconnected()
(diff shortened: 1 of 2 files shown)


core/src/main/scala/spark/deploy/worker/Worker.scala
```diff
@@ -143,9 +143,13 @@ private[spark] class Worker(
 
     case KillExecutor(jobId, execId) =>
       val fullId = jobId + "/" + execId
-      val executor = executors(fullId)
```
stephenh, line 146: With the Master now calling removeJob, the Worker gets a KillExecutor even if the executor is already dead/removed (by ExecutorRunner seeing the non-zero exit code).

JoshRosen:
FYI, this relates to [SPARK-643](https://spark-project.atlassian.net/browse/SPARK-643).

JoshRosen:
When I looked into fixing this, I ran into some issues with propagating the failure back to higher-level components (my early attempt caused hangs): https://spark-project.atlassian.net/browse/SPARK-644

stephenh:
Hi Josh,

When you say higher-level components, do you mean the Cluster/DAGSchedulers in the driver?

True, this simple fix doesn't attempt to tell the driver/job that, hey, go away, but I thought it's at least better to leave the master/slaves running, as the driver/job can always be control-c'd when it's hung and then restarted, without having to recover your whole cluster.

AFAICT that worked for me.

Having the actors handle restarts sounds interesting--at that point, I'm even more out of my area of expertise, so can't really comment.

Now that I'm being less dumb (and can run executors again), this patch isn't critically important to me--I can just close it if you guys would rather do something more robust as apart of the mentioned tickets.

mateiz:
I've merged this manually since it's a pretty important fix. Thanks Stephen. Josh, this doesn't fully solve SPARK-643, right? I'll leave that open.

PR #404, opened 2013-01-23
Add an Avro dependency to REPL to make it compile with Hadoop 2

mbautin:
Without this, I was getting the following compilation error:

```
error: error while loading Path, Missing dependency 'class org.apache.avro.reflect.Stringable', required by /Users/mbautin/.m2/repository/org/apache/hadoop/hadoop-common/2.0.0-cdh4.1.2/hadoop-common-2.0.0-cdh4.1.2.jar(org/apache/hadoop/fs/Path.class)
/wd/spark/repl/src/main/scala/spark/repl/ExecutorClassLoader.scala:37: error: org.apache.hadoop.fs.Path does not have a constructor
          fileSystem.open(new Path(directory, pathInDirectory))
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -542,6 +542,17 @@
             <artifactId>hadoop-client</artifactId>
             <version>2.0.0-mr1-cdh${cdh.version}</version>
           </dependency>
+          <!-- Specify Avro version because Kafka also has it as a dependency -->
+          <dependency>
+            <groupId>org.apache.avro</groupId>
+            <artifactId>avro</artifactId>
+            <version>1.7.1.cloudera.2</version>
+          </dependency>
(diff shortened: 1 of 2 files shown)


mateiz:
Why does this not come from the hadoop2 dependency? I've compiled it with Hadoop 2 before.

mbautin:
I think this is because a transitive dependency on an older version of Avro was introduced through Streaming, which did not include some classes required by `org.apache.hadoop.fs.Path`.

mateiz:
Oh, I see. That does make sense. Going to merge it.

mateiz:
So by the way, does "core" work okay or should we put the dependency there?

PR #405, opened 2013-01-23
Add test for driver hanging on exit (SPARK-530)

JoshRosen:
This adds a test case to detect if a driver cannot exit due to non-daemon threads.  I wrote this as a test case for [SPARK-530](https://spark-project.atlassian.net/browse/SPARK-530); I wasn't able to reproduce that issue, but this should prevent us from introducing bugs in the future.


mateiz:
Looks good, thanks.

PR #406, opened 2013-01-23
Changed StorageLevel and BlockManagerId API to prevent duplication in memory

tdas:
As StorageLevel/BlockManagerId is deserialized, new objects are created that may have exactly the same value as existing level/id objects. This leads to significant memory bloat specially for streaming as many many blocks are created in this case. Ideally, there should be as many unique id objects as there are workers. Same for storagelevel. This fix ensures that. 
For BlockManagerId, the only way to create a new BlockManagerId object is through BlockManagerId.apply(...) which automatically de-dups. I made the constructor private to ensure this. I tried to do the same for StorageLevel. But since StorageLevel is in public API, I kept the main constructor as public so that user programs dont break. However, I have added to the docs saying that the recommended way to create a StorageLevel object is to use StorageLevel.appy(...).

### DIFF
--- a/core/src/main/scala/spark/api/java/StorageLevels.java
+++ b/core/src/main/scala/spark/api/java/StorageLevels.java
@@ -17,4 +17,15 @@ public class StorageLevels {
   public static final StorageLevel MEMORY_AND_DISK_2 = new StorageLevel(true, true, true, 2);
   public static final StorageLevel MEMORY_AND_DISK_SER = new StorageLevel(true, true, false, 1);
   public static final StorageLevel MEMORY_AND_DISK_SER_2 = new StorageLevel(true, true, false, 2);
+
+  /**
+   * Create a new StorageLevel object.
+   * @param useDisk saved to disk, if true
+   * @param useMemory saved to memory, if true
+   * @param deserialized saved as deserialized objects, if true
+   * @param replication replication factor
+   */
+  public static StorageLevel create(boolean useDisk, boolean useMemory, boolean deserialized, int replication) {
+    return StorageLevel.apply(useDisk, useMemory, deserialized, replication);
+  }
 }
--- a/core/src/main/scala/spark/scheduler/MapStatus.scala
+++ b/core/src/main/scala/spark/scheduler/MapStatus.scala
@@ -20,7 +20,7 @@ private[spark] class MapStatus(var address: BlockManagerId, var compressedSizes:
   }
 
   def readExternal(in: ObjectInput) {
-    address = new BlockManagerId(in)
+    address = BlockManagerId(in)
     compressedSizes = new Array[Byte](in.readInt())
     in.readFully(compressedSizes)
   }
--- a/core/src/main/scala/spark/storage/BlockManager.scala
+++ b/core/src/main/scala/spark/storage/BlockManager.scala
@@ -191,7 +191,7 @@ class BlockManager(
         case level =>
           val inMem = level.useMemory && memoryStore.contains(blockId)
           val onDisk = level.useDisk && diskStore.contains(blockId)
-          val storageLevel = new StorageLevel(onDisk, inMem, level.deserialized, level.replication)
+          val storageLevel = StorageLevel(onDisk, inMem, level.deserialized, level.replication)
           val memSize = if (inMem) memoryStore.getSize(blockId) else 0L
           val diskSize = if (onDisk) diskStore.getSize(blockId) else 0L
           (storageLevel, memSize, diskSize, info.tellMaster)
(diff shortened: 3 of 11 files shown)


mateiz:
Have you considered using readResolve for the deserialization stuff? It replaces the deserialized object with another of your choice. Take a look at http://docs.oracle.com/javase/1.5.0/docs/api/java/io/Serializable.html.

tdas:
I am already using that! That has been in place since the streaming branch. However, that works only ObjectInputStream is being used to deserialize the object. However, if it is Externalizable, and you are directly calling readExternal (within some other classes readExternal), then the read resolve thing doesnt work. There were number cases where such was happening. For example, look at line 57 in https://github.com/mesos/spark/blob/master/core/src/main/scala/spark/storage/BlockManagerMessages.scala
This would create a new object as readResolve of BlockManagerId will not be invoked. My fixes takes care of that by making the only way to create BlockManagerId objects is to use the factory methods in the object BlockManagerId.

mateiz:
Ah okay, I remembered seeing that somewhere! Makes sense then.

mateiz:
By the way, for StorageLevel, it doesn't matter a ton for the ones users pass to us, just for the ones we create in Spark worker code, right? Because the one they pass to us will be saved on the master.

I'd be okay changing it to make the constructor private though if you think it will be more consistent. In that case, we should also add a method called build() that is an alias of apply(), since apply() will look a bit weird in Java.

mateiz:
Or actually call it create() instead of build().

tdas:
Yes it doesnt matter. I could add create() as an alias of apply() as well. Should make both Scala and Java folks happy.

mateiz:
Alright, do that then, and maybe update the Jekyll docs too to show the preferred way of creating StorageLevel objects.

tdas:
Where in the docs?

tdas:
I added it to the scala programming guide, after "Which storage level to choose"

mateiz:
Mention it in the Java ones too.

mridulm:
You can ensure this using readResolve.
We do that for our app's.

Regards,
Mridul

tdas:
@mridulm Please refer to the existing conversation with Matei in this pull request. 
@mateiz Made StorageLevel constructor private, and added StorageLevels.create() to the Java API. Updated scala and java programming guides.

mridulm:
I saw the conversation, and imo it is slightly incorrect if I read it correctly.
As documented in the interface, readResolve/writeResolve is invoked
even if it is Externizable (though not read/write Object).

So it can be used to replace the instance - to enforce Singleton
behavior, reduce memory consumption (in our case), etc.

Regards,
Mridul

tdas:
Yes, readResolve is automatically called by ObjectInputStream on Externalizable objects. However, look at the following scenario (line 57 in https://github.com/mesos/spark/blob/master/core/src/main/scala/spark/storage/BlockManagerMessages.scala)

override def readExternal(in: ObjectInput) {
    blockManagerId = new BlockManagerId()
    blockManagerId.readExternal(in)
    blockId = in.readUTF()
    storageLevel = new StorageLevel()
    storageLevel.readExternal(in)
    memSize = in.readInt()
    diskSize = in.readInt()
  }
}

Here the BlockManagerId is being deserialized by creating an empty object and then explicitly called readExternal. This is part of another objects readExternal that contains a blockManagerId field. In this case readResolve will never be called, as the deserialization of blockManagerId is not being managed by ObjectInputStream. So yes, readResolve solves most cases, but there ways to deserialize objects that does not invoke readResolve and therefore create duplicate objects. The changes I propose fixes these scenarios.

mridulm:
From a java point of view, this would be incorrect usage of Extenalizable
interface ... it is not a general purpose interface/method, but to be used
specifically in context of serialization.
Ofcourse, since I am not very familiar with scala, I dont know how it is
used here !

(de)serialization and object management are tightly coupled; and decoupling
them leads to gross and subtle bugs.
At best, you could do something like - static <Type>
readFromStream(ObjectInputStream ois) { ... } and delegate to that from
both readExternal and other methods.
This method would then enforce object management.

Unfortunately, I see your point that this would be wholesale changes to the
code; so probably I should not comment further :-)

Regards,
Mridul

tdas:
Your point is well taken that the example was probably an incorrect use of readExternal. (De)serialization and object management are indeed tough to decouple. What I have done in my changes is pretty much what you suggested. The (default constructor + call to readExternal) is replaced by a call to a static method  
     BlockManagerId.apply( .. input stream .. )  
which returns the de-duped object. Thankfully there werent many places where this problem had occurred, hence I could change use such static functions in all of them. 

Comments are ALWAYS welcome. :)

mateiz:
Great, thanks.

tdas:
Thanks Matei!

PR #407, opened 2013-01-23
Eliminate CacheTracker

charlesreiss:
Replaces DAGScheduler's queries of CacheTracker with BlockManagerMaster
queries. This may have negative performance implications; I have not tested for them.

The remaining CacheTracker functionality of locally coordinating the computation of cachable RDDs is moved to a new class called CacheManager.


mateiz:
Regarding efficiency -- it may be worth optimizing BlockManagerMasterActor.getLocationsMultipleBlockIds for this. When creating the ArrayBuffer to return, give it the locations directly instead of making a (likely size 16) empty one and then calling appendAll. Or even better, it might be possible to return an array. The whole blockLocations data structure is a bit scary in terms of memory usage unfortunately; a bunch of HashSets in there seems like overkill.

mateiz:
One other comment: I think you forgot to git add CacheManager.scala.

charlesreiss:
I indeed forget to add CacheManager.scala; fixed.

mateiz:
Alright, thanks.

PR #408, opened 2013-01-23
Use MapOutputTracker generation for failures in DAGScheduler and test failures more

charlesreiss:
Use MapOutputTracker generation instead of a blacklist of nodes in DAGScheduler to handle failed shuffle results from failed nodes.

Also, increment MapOutputTracker generation for all shuffle tasks, avoiding situations where an map output list with holes is cached. (This fix is suboptimal since it may result in us clearing cached map statuses more often than necessary and more often than the existing code; alternatives welcome.)

Also, add several tests to DistributedSuite which inject failures, including repeated node failures involving shuffles.

### DIFF
--- a/core/src/main/scala/spark/scheduler/DAGScheduler.scala
+++ b/core/src/main/scala/spark/scheduler/DAGScheduler.scala
@@ -76,7 +76,8 @@ class DAGScheduler(taskSched: TaskScheduler) extends TaskSchedulerListener with 
   // sent with every task. When we detect a node failing, we note the current generation number
   // and failed host, increment it for new tasks, and use this to ignore stray ShuffleMapTask
   // results.
-  // TODO: Garbage collect information about failure generations when new stages start.
+  // TODO: Garbage collect information about failure generations when we know there are no more
+  //       stray messages to detect.
   val failedGeneration = new HashMap[String, Long]
 
   val waiting = new HashSet[Stage] // Stages we need to run whose parents aren't done
--- a/core/src/main/scala/spark/scheduler/DAGScheduler.scala
+++ b/core/src/main/scala/spark/scheduler/DAGScheduler.scala
@@ -72,8 +72,12 @@ class DAGScheduler(taskSched: TaskScheduler) extends TaskSchedulerListener with 
   val cacheTracker = env.cacheTracker
   val mapOutputTracker = env.mapOutputTracker
 
-  val deadHosts = new HashSet[String]  // TODO: The code currently assumes these can't come back;
-                                       // that's not going to be a realistic assumption in general
+  // For tracking failed nodes, we use the MapOutputTracker's generation number, which is
+  // sent with every task. When we detect a node failing, we note the current generation number
+  // and failed host, increment it for new tasks, and use this to ignore stray ShuffleMapTask
+  // results.
+  // TODO: Garbage collect information about failure generations when new stages start.
+  val failedGeneration = new HashMap[String, Long]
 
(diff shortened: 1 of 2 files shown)


core/src/test/scala/spark/DistributedSuite.scala
```diff
@@ -188,4 +188,76 @@ class DistributedSuite extends FunSuite with ShouldMatchers with BeforeAndAfter
     val values = sc.parallelize(1 to 2, 2).map(x => System.getenv("TEST_VAR")).collect()
     assert(values.toSeq === Seq("TEST_VALUE", "TEST_VALUE"))
   }
+
+  test("recover from node failures") {
+    import DistributedSuite.{markNodeIfIdentity, failOnMarkedIdentity}
+    DistributedSuite.amMaster = true
+    sc = new SparkContext(clusterUrl, "test")
+    val data = sc.parallelize(Seq(true, true), 2)
+    val singleton = sc.parallelize(Seq(true), 1)
```
mateiz: Why do you create these RDDs? They seem unused.

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -439,9 +445,16 @@ class DAGScheduler(taskSched: TaskScheduler) extends TaskSchedulerListener with
...
+                // recomputation of the map outputs. In that case, some nodes may have cached
+                // locations with holes (from when we detected the error) and will need the
+                // generation incremented to refetch them.
+                // TODO: Only increment the generation number if this is not the first time
+                //       we registered these map outputs.
                 mapOutputTracker.registerMapOutputs(
                   stage.shuffleDep.get.shuffleId,
-                  stage.outputLocs.map(list => if (list.isEmpty) null else list.head).toArray)
+                  stage.outputLocs.map(list => if (list.isEmpty) null else list.head).toArray,
+                  true)
```
mateiz, line 458: Spacing looks weird here

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -72,8 +72,12 @@ class DAGScheduler(taskSched: TaskScheduler) extends TaskSchedulerListener with
   val cacheTracker = env.cacheTracker
   val mapOutputTracker = env.mapOutputTracker
 
-  val deadHosts = new HashSet[String]  // TODO: The code currently assumes these can't come back;
-                                       // that's not going to be a realistic assumption in general
+  // For tracking failed nodes, we use the MapOutputTracker's generation number, which is
+  // sent with every task. When we detect a node failing, we note the current generation number
+  // and failed host, increment it for new tasks, and use this to ignore stray ShuffleMapTask
+  // results.
+  // TODO: Garbage collect information about failure generations when new stages start.
```
mateiz: I'm not sure I understand this; couldn't a node still be down if a new stage starts, and couldn't we want to recompute some old RDD later in a different job?

core/src/test/scala/spark/DistributedSuite.scala
charlesreiss: Removed (and below).

core/src/main/scala/spark/scheduler/DAGScheduler.scala
charlesreiss: Changed TODO to hopefully make it clearer.

mateiz:
I've merged this manually. It conflicted slightly with your previous pull request.

PR #409, opened 2013-01-23
Added pruntSplits method to RDD.

rxin:
One thing I am not entirely sure yet is whether to override getPreferredLocations.


core/src/test/scala/spark/RDDSuite.scala
```diff
@@ -168,4 +168,15 @@ class RDDSuite extends FunSuite with BeforeAndAfter {
       nums.zip(sc.parallelize(1 to 4, 1)).collect()
     }
   }
+
+  test("split pruning") {
+    sc = new SparkContext("local", "test")
+    val data = sc.parallelize(1 to 10, 10)
+    // Note that split number starts from 0, so > 8 means only 10th partition left.
+    val prunedRdd = data.pruneSplits(splitNum => splitNum > 8)
+    assert(prunedRdd.splits.size == 1)
```
JoshRosen: Minor nitpick, but shouldn't you use `===` instead of `==` so that ScalaTest gives better error messages?

rxin: Sure - although in this case the assert is too simple for it to matter, I will change it.

rxin:
Matei - done with my changes. Ready for merge/review.

core/src/main/scala/spark/RDD.scala
```diff
@@ -543,6 +544,15 @@ abstract class RDD[T: ClassManifest](
...
 
+  /**
+   * Prune splits (partitions) so Spark can avoid launching tasks on
+   * all splits. An example use case: If we know the RDD is partitioned by range,
+   * and the execution DAG has a filter on the key, we can avoid launching tasks
+   * on splits that don't have the range covering the key.
+   */
+  def pruneSplits(splitsFilterFunc: Int => Boolean): RDD[T] =
+    new SplitsPruningRDD(this, splitsFilterFunc)
+
```
mateiz: Hey, since this is an advanced feature, it might be better not to add a method like this, and to require people to create a new SplitsPruningRDD manually instead. We might come up with better APIs in the future.

core/src/main/scala/spark/rdd/SplitsPruningRDD.scala
```diff
@@ -0,0 +1,24 @@
...
+
+import spark.{PruneDependency, RDD, SparkEnv, Split, TaskContext}
+
+/**
+ * A RDD used to prune RDD splits so we can avoid launching tasks on
+ * all splits. An example use case: If we know the RDD is partitioned by range,
+ * and the execution DAG has a filter on the key, we can avoid launching tasks
+ * on splits that don't have the range covering the key.
+ */
+class SplitsPruningRDD[T: ClassManifest](
```
mateiz: Sorry, kind of a nitpick, but let's maybe call it PartitionPruningRDD in anticipation of renaming Split -> Partition.

PR #411, opened 2013-01-24
Add LocalSparkContext to manage common sc variable.

stephenh:
I noticed some duplication in the tests around closing the SparkContext and pulled it into a trait to manage.

I had started this awhile ago, but decided to finish it up when I realized all of the clearProperty("spark.master.port") would be spark.driver.port if I start that down that path. So figured doing this first would be good.


mateiz:
This is great! Unfortunately the pull request doesn't merge automatically for some reason -- can you re-merge master into it? Maybe it's based on an older version of the master branch.

stephenh:
Okay, I re-merged master. I wish GitHub worked better with rebasing, because that's what I'd really prefer doing in this scenario. But oh well, I guess.

mateiz:
Great, thanks. This is much cleaner than what we had.

PR #412, opened 2013-01-24
SPARK-541: Adding a warning for invalid Master URL

pwendell:
Right now Spark silently parses master URL's which do not match any
known regex as a Mesos URL. The Mesos error message when an invalid URL gets
passed is really confusing, so this warns the user when the implicit
conversion is happening.


mateiz:
Looks good, thanks.

PR #413, opened 2013-01-25
SPARK-658: Adding logging of stage duration

pwendell:


core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -406,6 +410,15 @@ class DAGScheduler(taskSched: TaskScheduler) extends TaskSchedulerListener with
   def handleTaskCompletion(event: CompletionEvent) {
     val task = event.task
     val stage = idToStage(task.stageId)
+
+    def stageFinished(stage: Stage) = {
```
rxin: rename to markStageAsFinished?

pwendell: Done

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -88,6 +88,7 @@ class DAGScheduler(taskSched: TaskScheduler) extends TaskSchedulerListener with
 
   val activeJobs = new HashSet[ActiveJob]
   val resultStageToJob = new HashMap[Stage, ActiveJob]
+  val stageSubmissionTimes = new HashMap[Stage, Long]
```
mateiz: Instead of this, why not add a submissionTime field on Stage? It would make it easier to garbage-collect data (otherwise we'll have this hashmap that can keep growing).

pwendell: This map won't grow unless a stage never finishes, but I'm fine to have it as a field. Will update.

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -406,6 +409,15 @@ class DAGScheduler(taskSched: TaskScheduler) extends TaskSchedulerListener with
   def handleTaskCompletion(event: CompletionEvent) {
     val task = event.task
     val stage = idToStage(task.stageId)
+
+    def markStageAsFinished(stage: Stage) = {
+      val serviceTime = stage.submissionTime match {
+        case Some(t) => (System.currentTimeMillis() - t).toString
```
mateiz: Sorry, another minor thing, but mind changing this to seconds? Something like "%.03f".format((System.currentTimeMillis() - t) / 1000.0)

pwendell: Good idea. Added new commit with this formatting.

mateiz:
Great, thanks!

PR #415, opened 2013-01-25
Replace old 'master' term with 'driver'.

stephenh:
This is what I have so far; I think it's pretty complete. Let me know if I missed anything obvious.

(I did a "git grep spark.master.port" and "git grep spark.driver.host" and looks like I got all of those strings.)


stephenh:
Huh. Looks like one of the tests is failing--does sbt ever not run all of the tests? I swear I see different output from it between runs.

JoshRosen:
@stephenh Which test is failing?  sbt should run all of the tests, so maybe the test is failing due to some nondeterminism.

stephenh:
@JoshRosen only DriverSuite, but it seems like if I run it directly, it passes:

[info] DriverSuite:
[info] - driver should exit after finishing **\* FAILED ***
[info]   SparkException was thrown during property evaluation. (DriverSuite.scala:14)
[info]   Message: Process List(./run, spark.DriverWithoutCleanup, local) exited with code 1
[info]   Occurred at table row 0 (zero based, not counting headings), which had values (
[info]     master = local
[info]   )

Still poking at it...

JoshRosen:
@stephenh: That's a new test that I added in https://github.com/mesos/spark/pull/405 to guard against non-daemon threads that prevent the driver from exiting.  The test is time-sensitive and I guess I chose a timeout that was too low since the test is failing non-deterministically.  That's not a bug introduced by your pull request, though.

JoshRosen:
Actually, it looks like the driver is failing to launch rather than timing out.  Maybe this test is interfering with other tests because it's using resources that haven't been freed by those tests.

stephenh:
Oh, okay, I didn't have SCALA_HOME set in my shell. That test passes now...still seeing some weirdness. Going to see if it happens on master.

stephenh:
Matei, I just merged master into this commit; I believe it is ready to go.

Although I saw you also did a recent slaves/executors clean up (thank you! I'd been wondering about that), and given this touches a lot of places, I understand if you'd feel more comfortable doing it yourself.

mateiz:
No, I'm fine merging this. Let me look through it a bit more carefully before though.

I have one other change I'll add soon, which simplifies the RDD checkpointing stuff in streaming a little, but it shouldn't conflict with this too much. The executor change I merged earlier was important because it now allows multiple executors to run on the same hostname.

Who knows, maybe we'll also rename splits to partitions!

mateiz:
Merged this, thanks.

PR #416, opened 2013-01-25
Call executeOnCompleteCallbacks in more finally blocks.

stephenh:
Noticed while I was reading the ShuffleMapTask code.


mateiz:
Looks good, thanks.

PR #417, opened 2013-01-27
Fix JavaRDDLike.flatMap(PairFlatMapFunction) (SPARK-668).

JoshRosen:
This is a hack that fixes [SPARK-668](https://spark-project.atlassian.net/browse/SPARK-668), a typechecking bug when using JavaPairRDD.flatMap() with a PairFlatMapFunction.  This seems easier than finding the compiler bug that caused this problem.


mateiz:
Good way to fix it; thanks.

PR #418, opened 2013-01-27
Fix BlockManager reregistration deadlock; do BlockManager reregistration more asynchronously

charlesreiss:
When a block update fails, the BlockManager may try to reregister with a BlockInfo lock held. The reregistration itself also acquires locks on each BlockInfo to send updates for all held blocks. When multiple threads detect they need to reregister at the same time, this can result in a deadlock.

This patch makes reregistration triggered by a failure reply from a block update happen in a separate thread (so no locks are held) and prevents multiple reregistrations as a result of a failed reply from occurring concurrently.


mateiz:
Thanks for the fix.

PR #419, opened 2013-01-27
Detect whether we run on EC2 using ec2-metadata as well

shivaram:
Fixes SPARK-613 when we are using the Spark AMI


mateiz:
Looks good, thanks.

PR #420, opened 2013-01-28
Fix defaultMinSplits to use max.

stephenh:
I am probably just thinking about this wrong, but this:

math.min(defaultParallelism, 2)

Seems like it would always be 2. When is defaultParallelism going to be 0 or 1? (Perhaps if it's unset.) But AFAIK in standalone mode it's number of cores, e.g. much larger than 2.

Given this hint is passed into Hadoop's InputFormat, wouldn't we want to get back ~number of core splits instead of just 2 splits?


mateiz:
I don't think this is a good idea because sometimes, when you operate on small files, you want to have fewer splits, since you might do something like union() a bunch of files or join them or stuff like that. It's a little ridiculous to set it to hundreds in that case. So I'd rather keep this as min and let people choose their own value manually. It is true that the code is a bit weird -- I guess it can be 1 if you run in local mode with just 1 thread, but yeah, most of the time it will be 2.

stephenh:
True, you wouldn't want small files split up, but that looks to be handled by FileInputFormat:

long minSize = Math.max(job.getLong("mapred.min.split.size", 1), minSplitSize);

Unfortunately FileInputFormat itself doesn't have a sane default.

If Spark were to set a sane mapred.min.split.size default in its hadoopConfiguration (granted, defining "small" is hard--1G?), then it seems like math.max would be safe/preferred?

Admittedly, this is not actually affecting me, I just noticed it in the code. So if you'd still rather not, that's fine, I'll close out the pull request.

mateiz:
Would that really be different from the situation now though, where the minimum split size is basically the HDFS block size, unless you ask for more of them?

stephenh:
Min split size is the HDFS block size? How does that happen?

Here is my current analysis of FileInputFormat:

``` java
  public InputSplit[] getSplits(JobConf job, int numSplits)
    throws IOException {
    // numSplits = 2 (basic default of SparkContext.defaultMinSplits)
    FileStatus[] files = listStatus(job);

    // Save the number of input files in the job-conf
    job.setLong(NUM_INPUT_FILES, files.length);
    long totalSize = 0;                           // compute total size
    for (FileStatus file: files) {                // check we have valid files
      if (file.isDir()) {
        throw new IOException("Not a file: "+ file.getPath());
      }
      totalSize += file.getLen();
    }
    // let's say totalSize = 10G

    // goalSize = 5G
    long goalSize = totalSize / (numSplits == 0 ? 1 : numSplits);
    // minSize = 1
    long minSize = Math.max(job.getLong("mapred.min.split.size", 1), minSplitSize);

    // generate splits
    ArrayList<FileSplit> splits = new ArrayList<FileSplit>(numSplits);
    NetworkTopology clusterMap = new NetworkTopology();
    for (FileStatus file: files) {
      Path path = file.getPath();
      FileSystem fs = path.getFileSystem(job);
      long length = file.getLen();
      // Let's say the file is split into 5 blocks of 2G each
      BlockLocation[] blkLocations = fs.getFileBlockLocations(file, 0, length);
      if ((length != 0) && isSplitable(fs, path)) { 
        // blockSize = 2GB
        long blockSize = file.getBlockSize();
        // max(goalSize, min(minSize, blockSize))
        //   == max(5G, min(1, 2G))
        //   == max(5G, 1)
        //   == 5G
        //
        // So our "low" numSplits (2) overrode the file block size
        //
        // Seems like with:
        //
        //     min(minSize, blockSize) we can err on a high minSize, b/c if it's
        //     too high, the actual blockSize will override it
        //
        //     max(goalSize, ...) we can err on a low goalSize (so high
        //     numSplits), b/c if it's too low, then minSize/blockSize will
        //     override it
        //
        // For example:
        // numSplits total  goal min block split comments
        //         2   10G    5G   1    2G    5G // current result, unset min, goal too high
        //       120   10G  .04G   1    2G     1 // "good" numSplits, but unset min, bad result
        //       120   10G  .04G  1G    2G    1G // good numSplits/min, good result
        //         2  100M   50M   1  100M   50M // current result, no min, eh
        //         2   10k    5K   1   10K    5k // current result, no min, bad result
        //       120  100M   .8M  1G  100M  100M // good numSplits/min, good result
        //
        // minSize == don't split unless blockSize is bigger than X
        //
        long splitSize = computeSplitSize(goalSize, minSize, blockSize);

        long bytesRemaining = length;
        // 10G / 5G > 1.1
        while (((double) bytesRemaining)/splitSize > SPLIT_SLOP) {
          // offset = 10G - 5G
          // splitHosts = [a with 2G, b with 2G, c with 2G]
          String[] splitHosts = getSplitHosts(blkLocations, 
              length-bytesRemaining, splitSize, clusterMap);
          splits.add(new FileSplit(path, length-bytesRemaining, splitSize, 
              splitHosts));
          bytesRemaining -= splitSize;
        }

        if (bytesRemaining != 0) {
          splits.add(new FileSplit(path, length-bytesRemaining, bytesRemaining, 
                     blkLocations[blkLocations.length-1].getHosts()));
        }
      } else if (length != 0) {
        String[] splitHosts = getSplitHosts(blkLocations,0,length,clusterMap);
        splits.add(new FileSplit(path, 0, length, splitHosts));
      } else { 
        //Create empty hosts array for zero length files
        splits.add(new FileSplit(path, 0, length, new String[0]));
      }
    }
```

Granted, I am only thinking of Spark standalone mode, where there is no default min split size at this point.

Per that table of numbers in the middle, I believe a low numSplits (2) leads to arbitrarily high split sizes (effectively totalSize / 2), where a split could potentially encompass numerous blocks, that would probably be better off in their own splits to increase parallelism.

But, yes, increasing numSplits would be bad without also setting a sane min size (1G?), which, per the table, I believe leads to good results for both large and small file sizes.

The current settings (low numSplits, no min) I believe is compromising both by not splitting up large files enough (only twice) and still splitting up small files (again twice) even if they don't need to be.

Disclaimer that I haven't actually accessed files this large, but I believe my mental execution of the code, and the figures in the table, are correct.

mateiz:
Hmm, I would actually try that and look into what computeSplitSize does. Unless this changed in Hadoop 2.0, FileInputFormat always created at least one split per block of the file, since each block has different locality preferences (different hosts that it's on). If you run the current Spark EC2 image in a 45 GB file, for example, it will create around 350 splits (given 128 MB block size in HDFS), even though minSplits is 2. I haven't tried this with Hadoop 2 but I imagine that some of our Hadoop 2 users would've noticed if things ran on just 2 tasks.

stephenh:
Oh crap. I mentally inlined "computeSplitSize" but flipped the arguments, it is actually:

  protected long computeSplitSize(long goalSize, long minSize, long blockSize) {
    return Math.max(minSize, Math.min(goalSize, blockSize));
  }

I got goalSize and minSize mixed up. Sheesh. Sorry about that.

mateiz:
Yeah, I guess that's it. I'm going to close this pull request unless you find other reasons for changing this. I think it's good to somewhat match Hadoop anyway (our minSplits setting right now is exactly the same as specifying the # of map tasks in Hadoop).

stephenh:
Cool, that sounds good, thanks.

PR #421, opened 2013-01-28
Switch spark_ec2.py to use the new spark-ec2 scripts. 

shivaram:
This changes spark_ec2.py to use the new cluster initialization scripts from https://github.com/shivaram/spark-ec2. The new scripts preserve existing features, fix a couple of bugs in the standalone setup and also optionally (-g) allow ganglia to be started on the cluster. Users can also fall back to the old script using the --mesos-scripts if required.


ec2/spark_ec2.py
```diff
@@ -84,6 +84,12 @@ def parse_args():
             "maximum price (in dollars)")
   parser.add_option("-c", "--cluster-type", default="mesos",
       help="'mesos' for a mesos cluster, 'standalone' for a standalone spark cluster (default: mesos)")
+  parser.add_option("-g", "--ganglia", action="store_true", default=False,
+      help="Setup ganglia monitoring for the cluster. NOTE: The ganglia " +
+      "monitoring page will be publicly accessible")
```
mateiz, line 89: I think I mentioned this by email, but let's have Ganglia on by default, and just a `--no-ganglia` for turning it off.

mateiz:
This looks pretty good, but I don't want to merge it until we have an official AMI with this that we can point the scripts to by default. I don't want people to see errors if they check out the master branch and run spark-ec2. Do you have an AMI you can make public? If so, I'll add a reference to it in S3 (in our `latest` style files). We will probably also need to change which `latest` file is used (not a big deal, but something to consider).

shivaram:
I turned on ganglia by default and created a new public AMI that works with this script.  The AMI id is ami-bb2ebad2 and has no changes other than a few places where I replaced mesos-ec2 with spark-ec2.  I am not sure I have access to change the `latest` file on S3.

mateiz:
Thanks Shivaram. I've merged this but I will tweak it a little to launch the old AMI for now.

PR #422, opened 2013-01-29
RDDInfo available from SparkContext

squito:
In addition to having info on RDD storage available in the web ui, also add an api so its directly available from a SparkContext.  This gives users more flexibility, eg. logging to a file.

### DIFF
--- a/core/src/main/scala/spark/SparkContext.scala
+++ b/core/src/main/scala/spark/SparkContext.scala
@@ -468,7 +468,7 @@ class SparkContext(
    * Return a map from the slave to the max memory available for caching and the remaining
    * memory available for caching.
    */
-  def getSlavesMemoryStatus: Map[String, (Long, Long)] = {
+  def getExecutorMemoryStatus: Map[String, (Long, Long)] = {
     env.blockManager.master.getMemoryStatus.map { case(blockManagerId, mem) =>
       (blockManagerId.ip + ":" + blockManagerId.port, mem)
     }
@@ -479,13 +479,13 @@ class SparkContext(
    * they take, etc.
    */
   def getRDDStorageInfo : Array[RDDInfo] = {
-    StorageUtils.rddInfoFromStorageStatus(getSlavesStorageStatus, this)
+    StorageUtils.rddInfoFromStorageStatus(getExecutorStorageStatus, this)
   }
 
   /**
    * Return information about blocks stored in all of the slaves
    */
-  def getSlavesStorageStatus : Array[StorageStatus] = {
+  def getExecutorStorageStatus : Array[StorageStatus] = {
     env.blockManager.master.getStorageStatus
   }
 
--- a/core/src/main/scala/spark/SparkContext.scala
+++ b/core/src/main/scala/spark/SparkContext.scala
@@ -46,6 +46,7 @@ import spark.scheduler.cluster.{SparkDeploySchedulerBackend, SchedulerBackend, C
 import spark.scheduler.mesos.{CoarseMesosSchedulerBackend, MesosSchedulerBackend}
 import storage.BlockManagerUI
 import util.{MetadataCleaner, TimeStampedHashMap}
+import storage.{StorageStatus, StorageUtils, RDDInfo}
 
 /**
  * Main entry point for Spark functionality. A SparkContext represents the connection to a Spark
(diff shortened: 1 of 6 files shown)


core/src/main/scala/spark/SparkContext.scala
```diff
@@ -474,6 +475,21 @@ class SparkContext(
...
+   * they take, etc.
+   */
+  def getRDDStorageInfo : Array[RDDInfo] = {
+    StorageUtils.rddInfoFromStorageStatus(getSlavesStorageStatus, this)
+  }
+
+  /**
+   * Return information about blocks stored in all of the slaves
+   */
+  def getSlavesStorageStatus : Array[StorageStatus] = {
```
stephenh: Should this be getExecutorsStorageStatus, given the renaming of slaves->executors?

squito: There is already getSlavesMemoryStatus, which has been part of the public api, which this name mirrors.  Should I rename this method, mark getSlavesMemoryStatus as deprecated, and have it just delegate to a new getExecutorMemoryStatus?

stephenh: I don't know Spark's policy for breaking changes; personally, since it is pre-1.0, I would skip the deprecation step, as it's pretty trivial for end users to update to the new method name. It also seems like it may not be used a lot--it's not a method every Spark program has, like map or filter. But I'll defer to Matei.

mateiz: I'm okay renaming this in this case because getSlavesMemoryStatus wasn't very visibly documented and was really only used by Shark.

squito:
updated method names & brought up to date

core/src/main/scala/spark/storage/StorageUtils.scala
```diff
@@ -22,8 +22,14 @@ case class StorageStatus(blockManagerId: BlockManagerId, maxMem: Long,
...
 
 case class RDDInfo(id: Int, name: String, storageLevel: StorageLevel,
-  numPartitions: Int, memSize: Long, diskSize: Long)
-
+  numPartitions: Int, memSize: Long, diskSize: Long) {
+  override def toString = {
+    import Utils.memoryBytesToString
+    import java.lang.{Integer => JInt}
+    String.format("RDD \"%s\" (%d) Storage: %s; Partitions: %d; MemorySize: %s; DiskSize: %s", name, id.asInstanceOf[JInt],
+      storageLevel.toString, numPartitions.asInstanceOf[JInt], memoryBytesToString(memSize), memoryBytesToString(diskSize))
```
mateiz: Why do you need to cast the integers to JInt? I think you can use Scala's format method instead of Java's and just pass ints. Just do `"RDD %s (%d), ...".format(id, storageLevel, etc)`.

core/src/main/scala/spark/storage/StorageUtils.scala
```diff
@@ -56,9 +62,11 @@ object StorageUtils {
       // Find the id of the RDD, e.g. rdd_1 => 1
       val rddId = rddKey.split("_").last.toInt
       // Get the friendly name for the rdd, if available.
-      val rddName = Option(sc.persistentRdds(rddId).name).getOrElse(rddKey)
-      val rddStorageLevel = sc.persistentRdds(rddId).getStorageLevel
-      
+      val rdd = sc.persistentRdds(rddId)
+      val rddName = Option(rdd.name).getOrElse(rddKey)
+      val rddStorageLevel = rdd.getStorageLevel
+      //TODO get total number of partitions in rdd
```
mateiz: Is this a thing you want to add in? If not, maybe remove the TODO comment. By the way you can just do `rdd.splits.size` to get the number of partitions.

core/src/main/scala/spark/storage/StorageUtils.scala
squito: thanks, I knew that looked wrong but didn't realize it was scala's string format vs. java's

core/src/main/scala/spark/storage/StorageUtils.scala
squito: thanks, I was looking for something to do with partitions ...

squito:
I've updated it to reflect your suggestions, including tracking the total number of partitions.

btw, while reviewing these changes, I noticed something funny from the earlier code to make the info available in the UI.  In StorageUtils.rddInfoFromBlockStatus:

```
// Find all RDD Blocks (ignore broadcast variables)
val rddBlocks = infos.filterKeys(_.startsWith("rdd"))
```

but, actually that variable is never used -- so the final results don't filter for blocks whose name begins with "rdd".  I tried creating a broadcast variable, but it doesn't show up in the block infos at all.  So, should I fix the code to actually use the filter?  Or is the filter unnecessary?  (and if it is unneccessary, how would I get info on blocks used by broadcast variables?)

mateiz:
Hmm, maybe we set the block manager to not report broadcast variables' space to the master. Take a look in the BlockManager class. I think it's fine not to show them on this UI. They could be on some slave-local UI for example. They are also mentioned in the slave's logs.

squito:
I think this is ready to go now.

mateiz:
Looks good, thanks!

PR #424, opened 2013-01-29
Some DEBUG-level log cleanup.

pwendell:
A few changes to make the DEBUG-level logs less noisy and more readable.
- Moved a few very frequent messages to Trace
- Changed some BlockManger log messages to make them more understandable

SPARK-666 #resolve


mateiz:
Thanks for the fixes.

PR #426, opened 2013-01-30
Remember ConnectionManagerId used to initiate SendingConnections

charlesreiss:
This patch remembers the ConectionManagerId used to initiate SendingConnections rather than inferring it from the socket. This prevents ConnectionManager from getting confused if a machine has multiple host names and the one getHostName() finds happens not to be the one that was passed from, e.g., the BlockManagerMaster.

### DIFF
--- a/core/src/main/scala/spark/network/Connection.scala
+++ b/core/src/main/scala/spark/network/Connection.scala
@@ -12,7 +12,14 @@ import java.net._
 
 
 private[spark]
-abstract class Connection(val channel: SocketChannel, val selector: Selector) extends Logging {
+abstract class Connection(val channel: SocketChannel, val selector: Selector,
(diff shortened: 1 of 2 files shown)


mateiz:
Just curious, did you see this causing problems somewhere in practice?

charlesreiss:
Yes; this somehow caused DistributedSuite to fail on watson. A shuffle fetch would be initiated with ConnectionManagerId(169.229.48.63, (some port)) and the connection would be dropped before the block request was sent. ConnectionManager.removeConnection() would then look up MessageStatuses with ConnectionManagerId(watson.millennium.berkeley.edu, (some port)) to fail and find none. As a result the shuffle fetch would hang.

mateiz:
Looks good.

PR #427, opened 2013-01-30
Tests for DAGScheduler

charlesreiss:
This adds a bunch of tests for DAGScheduler, with the actual task execution mocked out.

### DIFF
--- a/core/src/main/scala/spark/scheduler/DAGScheduler.scala
+++ b/core/src/main/scala/spark/scheduler/DAGScheduler.scala
@@ -23,7 +23,14 @@ import util.{MetadataCleaner, TimeStampedHashMap}
  * and to report fetch failures (the submitTasks method, and code to add CompletionEvents).
  */
 private[spark]
-class DAGScheduler(taskSched: TaskScheduler) extends TaskSchedulerListener with Logging {
+class DAGScheduler(taskSched: TaskScheduler, 
+                   mapOutputTracker: MapOutputTracker,
+                   blockManagerMaster: BlockManagerMaster,
+                   env: SparkEnv)
+                      extends TaskSchedulerListener with Logging {
+  def this(taskSched: TaskScheduler) {
+    this(taskSched, SparkEnv.get.mapOutputTracker, SparkEnv.get.blockManager.master, SparkEnv.get)
+  }
   taskSched.setListener(this)
 
   // Called by TaskScheduler to report task completions or failures.
@@ -66,10 +73,6 @@ class DAGScheduler(taskSched: TaskScheduler) extends TaskSchedulerListener with 
 
   var cacheLocs = new HashMap[Int, Array[List[String]]]
 
-  val env = SparkEnv.get
-  val mapOutputTracker = env.mapOutputTracker
-  val blockManagerMaster = env.blockManager.master
-
   // For tracking failed nodes, we use the MapOutputTracker's generation number, which is
   // sent with every task. When we detect a node failing, we note the current generation number
   // and failed executor, increment it for new tasks, and use this to ignore stray ShuffleMapTask
@@ -90,12 +93,14 @@ class DAGScheduler(taskSched: TaskScheduler) extends TaskSchedulerListener with 
   val metadataCleaner = new MetadataCleaner("DAGScheduler", this.cleanup)
 
   // Start a thread to run the DAGScheduler event loop
-  new Thread("DAGScheduler") {
-    setDaemon(true)
-    override def run() {
-      DAGScheduler.this.run()
-    }
-  }.start()
+  def start() {
+    new Thread("DAGScheduler") {
+      setDaemon(true)
+      override def run() {
+        DAGScheduler.this.run()
+      }
+    }.start()
+  }
 
   def getCacheLocs(rdd: RDD[_]): Array[List[String]] = {
     if (!cacheLocs.contains(rdd.id)) {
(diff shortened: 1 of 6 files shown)


core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -23,7 +23,14 @@ import util.{MetadataCleaner, TimeStampedHashMap}
  * and to report fetch failures (the submitTasks method, and code to add CompletionEvents).
  */
 private[spark]
-class DAGScheduler(taskSched: TaskScheduler) extends TaskSchedulerListener with Logging {
+class DAGScheduler(taskSched: TaskScheduler, 
+                   mapOutputTracker: MapOutputTracker,
+                   blockManagerMaster: BlockManagerMaster,
+                   env: SparkEnv)
+                      extends TaskSchedulerListener with Logging {
```
mateiz: This formatting is weird (doesn't match the rest of the code). It should look like this:

```
class DAGScheduler(
    taskSched: TaskScheduler,
    mapOutputTracker: MapOutputTracker,
    blockManagerMaster: BlockManagerMaster,
    env: SparkEnv)
  extends TaskSchedulerListener with Logging {
...
}
```

mateiz:
Hey Charles,

This looks really useful, but since it's doing some nontrivial stuff with ScalaTest and EasyMock, it would really benefit from some documentation. There should be a brief overview of how the setup works at the top, and Scaladoc comments on methods like resetExpecting, expectStage, expectGetLocations, etc.

If it's not too hard, it may also be useful to factor this into a trait like LocalSparkContext, so that we can mix it into multiple test suites. I can see a lot more suites wanting to use this.

core/src/test/scala/spark/scheduler/DAGSchedulerSuite.scala
```diff
@@ -0,0 +1,541 @@
...
+class DAGSchedulerSuite extends FunSuite
+    with BeforeAndAfter with EasyMockSugar with TimeLimitedTests
+    with AsyncAssertions with spark.Logging {
+
+  // If we crash the DAGScheduler thread, our test will probably hang.
+  override val timeLimit = Span(5, Seconds)
+
+  val sc: SparkContext = new SparkContext("local", "DAGSchedulerSuite")
+  var scheduler: DAGScheduler = null
+  var w: Waiter = null
```
mateiz: This probably needs to be called `waiter`, since it's really unclear what it does

core/src/test/scala/spark/scheduler/DAGSchedulerSuite.scala
```diff
@@ -0,0 +1,541 @@
...
+    mapOutputTracker = new MapOutputTracker(actorSystem, true)
+    resetExpecting {
+      taskScheduler.setListener(anyObject())
+    }
+    whenExecuting {
+      scheduler = new DAGScheduler(taskScheduler, mapOutputTracker, blockManagerMaster, null)
+    }
+    w = new Waiter
+    schedulerException = null
+    schedulerThread = new Thread("DAGScheduler under test") {
```
stephenh: Just my initial impression is that this has a lot of waiting/mocking magic that is more akin to an integration test than a unit test.

Maybe instead of assuming we have to use the thread/event queue, we cheat and just refactor DAGScheduler to be easier to test?

E.g. keep DAGScheduler as having an event queue, and all of its public methods post messages to the event queue, but then move the internal implementation into a separate class that could be easily tested. E.g.:

``` scala
class DAGScheduler {
  val eventQueue = ...
  val impl = new DAGSchedulerImpl(...)
  def foo() = { eventQueue.adafterSubmitd(new FooCommand) }
    private def run() {
      while (true) {
        val m = eventQueue.poll
        m match {
          case FooCommand => impl.foo()
        }
      }
   }
}
class DAGSchedulerImpl(...) {
   def foo() = { ... }
}
```

Then this DAGSchedulerTest can poke DAGSchedulerImpl directly and not have to worry about waiting around for the command to flush through the event queue and cause affects that can be asserted against.

charlesreiss:
Based on stephenh's point that it wouldn't be hard to test DAGScheduler without the DAGScheduler thread, I did a rather severe refactoring that runs the test in the same thread. This makes the tests rely on less ScalaTest wizardry and be somewhat less confusing. I added some comments that try to explain what's going on though I'm not sure if it's enough.

With this design, it would probably be straightforward to change the TaskScheduler from a simple object that stores TaskSets it is submitted, though I'm not sure that would make things much clearer (and we're still left with BlockManagerMaster, which is not straightforward to replace this way).

core/src/test/scala/spark/scheduler/DAGSchedulerSuite.scala
```diff
@@ -0,0 +1,624 @@
...
+import spark.SparkContext
+import spark.SparkException
+import spark.Split
+import spark.TaskContext
+import spark.TaskEndReason
+
+import spark.{FetchFailed, Success}
+
+/**
+ * Tests for DAGScheduler. These tests directly call the event processing functinos in DAGScheduler
```
mateiz: Typo: functinos

core/src/test/scala/spark/scheduler/DAGSchedulerSuite.scala
```diff
@@ -0,0 +1,624 @@
...
+
+  after {
+    assert(scheduler.processEvent(StopDAGScheduler))
+    resetExpecting {
+      taskScheduler.stop()
+    }
+    whenExecuting {
+      scheduler.stop()
+    }
+    System.clearProperty("spark.master.port")
```
mateiz, line 129: Shouldn't you do `sc.stop()` too?

mateiz:
Hey Charles,

This looks good design-wise, but a few style comments:
- In multi-line doc comments, don't start text on the first line after the `/**`. Start it on the one below
- `resubmitFailedStages` and `submitWaitingStages` should be called with parentheses since they have side-effects

Also, I recently made a small change to DAGScheduler that I thought wouldn't affect this, but it does, mostly because `JobSucceeded` no longer holds a result now so you can't check job results as easily. So please merge that in. I should've merged this first but I wanted to do the change when I remembered that problem.

charlesreiss:
Style issues and missing sc.stop() should be fixed.

Merged in DAGScheduler changes from trunk; this was a bit non-trivial, so please take a look.

mateiz:
Thanks Charles. I've merged it, but there were still some unfixed formatting issues in DAGScheduler.scala, so I'll fix those manually.

PR #429, opened 2013-01-30
Include message and exitStatus if availalbe.

stephenh:
Really small change, but I'm seeing executors just up and disappear (probably due to partitions being too large) and so am curious what the exit code was.


mateiz:
Cool, thanks.

PR #431, opened 2013-01-31
Remove activation of profiles by default

mbautin:
See the discussion at https://github.com/mesos/spark/pull/355 for why default profile activation is a problem.


mateiz:
Thanks Mikhail.

PR #432, opened 2013-01-31
Add more private declarations.

stephenh:
These are "private" declarations I'd added locally while trying to understand things, and figure I would submit them, if anything because I kept getting trivial conflicts each time I'd pull in master.

I was going to see if Shark used any of these methods, but sbt is being uncooperative--I don't really know how the Shark/Spark integration works, but admittedly I'm touching DAGScheduler and friends here, which is pretty core, so I suppose I would not be surprised if some of it is used.

Perhaps if this does break Shark, the offending methods could be explicitly documented as "used by shark", just as a hint to otherwise-unaware readers like me that it really is a public contract.

### DIFF
--- a/core/src/main/scala/spark/MapOutputTracker.scala
+++ b/core/src/main/scala/spark/MapOutputTracker.scala
@@ -170,7 +170,7 @@ private[spark] class MapOutputTracker(actorSystem: ActorSystem, isDriver: Boolea
     }
   }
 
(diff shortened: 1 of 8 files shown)


rxin:
This should be fine with Shark.

mateiz:
Looks good, thanks.

stephenh:
Cool, thanks, Matei.

PR #433, opened 2013-02-01
Changed PartitionPruningRDD's split to make sure it returns the correct split index.

rxin:


core/src/main/scala/spark/Dependency.scala
```diff
@@ -72,6 +72,14 @@ class PruneDependency[T](rdd: RDD[T], @transient partitionFilterFunc: Int => Boo
...
   @transient
   val partitions: Array[Split] = rdd.splits.filter(s => partitionFilterFunc(s.index))
+    .zipWithIndex
+    .map { case(split, idx) => new PruneDependency.PartitionPruningRDDSplit(idx, split) : Split }
 
   override def getParents(partitionId: Int) = List(partitions(partitionId).index)
 }
+
+object PruneDependency {
+  class PartitionPruningRDDSplit(idx: Int, val parentSplit: Split) extends Split {
```
mateiz: Why did you make this an inner class instead of a top-level one? Also, it should probably be private[spark].

core/src/main/scala/spark/rdd/PartitionPruningRDD.scala
```diff
@@ -2,20 +2,20 @@ package spark.rdd
...
+
 /**
  * A RDD used to prune RDD partitions/splits so we can avoid launching tasks on
  * all partitions. An example use case: If we know the RDD is partitioned by range,
  * and the execution DAG has a filter on the key, we can avoid launching tasks
  * on partitions that don't have the range covering the key.
- *
- * TODO: This currently doesn't give partition IDs properly!
  */
 class PartitionPruningRDD[T: ClassManifest](
```
mateiz, line 32: You should also remove getPartitioner here since the partitioner won't really be compatible with the parent (not ideal unfortunately, but that's the way it is now)

mateiz:
Cool, thanks.

PR #434, opened 2013-02-01
SPARK-673: Capture and re-throw Python exceptions

pwendell:
This patch alters the Python <-> executor protocol to pass on
exception data when they occur in user Python code.

### DIFF
--- a/core/src/main/scala/spark/api/python/PythonRDD.scala
+++ b/core/src/main/scala/spark/api/python/PythonRDD.scala
@@ -103,21 +103,30 @@ private[spark] class PythonRDD[T: ClassManifest](
 
       private def read(): Array[Byte] = {
         try {
-          val length = stream.readInt()
-          if (length != -1) {
-            val obj = new Array[Byte](length)
-            stream.readFully(obj)
-            obj
-          } else {
-            // We've finished the data section of the output, but we can still read some
-            // accumulator updates; let's do that, breaking when we get EOFException
-            while (true) {
-              val len2 = stream.readInt()
-              val update = new Array[Byte](len2)
-              stream.readFully(update)
-              accumulator += Collections.singletonList(update)
+          stream.readInt() match {
+            case length if length > 0 => {
+              val obj = new Array[Byte](length)
+              stream.readFully(obj)
+              obj
             }
-            new Array[Byte](0)
+            case -2 => {
+              // Signals that an exception has been thrown in python
+              val exLength = stream.readInt()
+              val obj = new Array[Byte](exLength)
+              stream.readFully(obj)
+              throw new PythonException(new String(obj))
+            }
+            case -1 => {
+              // We've finished the data section of the output, but we can still read some
+              // accumulator updates; let's do that, breaking when we get EOFException
+              while (true) {
+                val len2 = stream.readInt()
+                val update = new Array[Byte](len2)
+                stream.readFully(update)
+                accumulator += Collections.singletonList(update)
+              }
+              new Array[Byte](0)
+          }
           }
         } catch {
           case eof: EOFException => {
@@ -139,6 +148,9 @@ private[spark] class PythonRDD[T: ClassManifest](
 
   val asJavaRDD : JavaRDD[Array[Byte]] = JavaRDD.fromRDD(this)
 }
+
+/** Thrown for exceptions in user Python code. */
+private class PythonException(msg: String) extends Exception(msg)
 
 /**
  * Form an RDD[(Array[Byte], Array[Byte])] from key-value pairs returned from Python.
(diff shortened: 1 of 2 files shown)


pwendell:
You can test this pretty easily as follows:

`sc.parallelize(range(100), 10).map(lambda x: x/0).count()`

core/src/main/scala/spark/api/python/PythonRDD.scala
```diff
@@ -103,21 +103,30 @@ private[spark] class PythonRDD[T: ClassManifest](
...
+              // We've finished the data section of the output, but we can still read some
+              // accumulator updates; let's do that, breaking when we get EOFException
+              while (true) {
+                val len2 = stream.readInt()
+                val update = new Array[Byte](len2)
+                stream.readFully(update)
+                accumulator += Collections.singletonList(update)
+              }
+              new Array[Byte](0)
+          }
```
mateiz, line 127: Indentation here seems to be off. Also, I'm not sure whether we put curly braces around cases in the rest of the code.. might be worth checking that and seeing whether this matches it style.

mateiz:
Cool! What does the output look like on the Python side?

pwendell:
Here is what the full log message looks like, incl. Java and Python stacks

13/01/31 21:44:29 INFO TaskSetManager: Loss was due to spark.api.python.PythonException: Traceback (most recent call last):
  File "/home/patrick/Documents/spark/python/pyspark/worker.py", line 45, in main
    for obj in func(split_index, iterator):
  File "/home/patrick/Documents/spark/python/pyspark/rdd.py", line 727, in batched_func
    return batched(oldfunc(split, iterator), batchSize)
  File "/home/patrick/Documents/spark/python/pyspark/rdd.py", line 702, in pipeline_func
    return func(split, prev_func(split, iterator))
  File "/home/patrick/Documents/spark/python/pyspark/rdd.py", line 702, in pipeline_func
    return func(split, prev_func(split, iterator))
  File "/home/patrick/Documents/spark/python/pyspark/rdd.py", line 702, in pipeline_func
    return func(split, prev_func(split, iterator))
  File "/home/patrick/Documents/spark/python/pyspark/rdd.py", line 114, in func
    def func(s, iterator): return f(iterator)
  File "/home/patrick/Documents/spark/python/pyspark/rdd.py", line 343, in <lambda>
    return self.mapPartitions(lambda i: [sum(1 for _ in i)]).sum()
  File "/home/patrick/Documents/spark/python/pyspark/rdd.py", line 343, in <genexpr>
    return self.mapPartitions(lambda i: [sum(1 for _ in i)]).sum()
  File "<stdin>", line 1, in <lambda>
ZeroDivisionError: integer division or modulo by zero

at spark.api.python.PythonRDD$$anon$1.read(PythonRDD.scala:117)
at spark.api.python.PythonRDD$$anon$1.<init>(PythonRDD.scala:143)
at spark.api.python.PythonRDD.compute(PythonRDD.scala:97)
at spark.RDD.computeOrReadCheckpoint(RDD.scala:203)
at spark.RDD.iterator(RDD.scala:192)
at spark.scheduler.ResultTask.run(ResultTask.scala:76)
at spark.executor.Executor$TaskRunner.run(Executor.scala:96)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1110)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:603)
at java.lang.Thread.run(Thread.java:722)

mateiz:
Ah cool, that's not too bad. It's unfortunate that you still see some Java stack trace elements in there but I guess it's going to be required because that's Spark's logging. Anyway if you fix the code style stuff I think this is good to go.

pwendell:
K style stuff is cleaned up. Rest of the code does not use braces in case statements.

mateiz:
Cool, thanks!

python/pyspark/worker.py
```diff
@@ -40,8 +41,13 @@ def main():
     else:
         dumps = dump_pickle
     iterator = read_from_pickle_file(sys.stdin)
-    for obj in func(split_index, iterator):
-        write_with_length(dumps(obj), old_stdout)
+    try:
+        for obj in func(split_index, iterator):
+           write_with_length(dumps(obj), old_stdout)
+    except Exception as e:
```
JoshRosen, line 47: Do you think that we should also log the inspection on the worker?  It might help during debugging.

pwendell, line 47: When the PythonException is thrown inside the Java process on the worker,
won't that get logged? Isn't that just like any other exception that occurs
inside of an RDD compute(). ?

JoshRosen, line 47: Ah, you're right.

PR #436, opened 2013-02-01
Once we find a split with no block, we don't have to look for more.

stephenh:
As far as I can tell, once we find a locs(p) == Nil, and that rdd's dependencies as stages, we don't need to continue the loop looking for more locs(p)s because we'll just end up adding the same stages (which has no effect since we're adding them to a hash set that they've already been added too).

Admittedly, this won't have a huge affect on runtime, because even with a large number of splits, this should be a fast loop, but nonetheless it made me think slightly when I saw it.


stephenh:
I am technically still running the tests for this, so please hold off on merging.

mateiz:
OK, let me know if it helps. By the way the check would be slightly more concise as `locs.contains(Nil)`.

mateiz:
Looks good, thanks.

PR #437, opened 2013-02-01
Stop BlockManagers metadataCleaner.

stephenh:
Otherwise while running the tests it leaves a lot of threads/tasks running.


mateiz:
Thanks for the fix.

PR #438, opened 2013-02-01
Do not launch JavaGateways on workers (SPARK-674).

JoshRosen:
This fixes [SPARK-674](https://spark-project.atlassian.net/browse/SPARK-674), which describes how Py4J JavaGateway processes were accidentally being launched on remote worker machines.  This had a huge performance overhead.

After fixing this, the unit tests run 3-4x faster on my machine.


mateiz:
Woah, good catch.

PR #441, opened 2013-02-02
Reduce the amount of duplicate logging Akka does to stdout.

stephenh:
Given we have Akka logging go through SLF4j to log4j, we don't need
all the extra noise of Akka's stdout logger that is supposedly only
used during Akka init time but seems to continue logging lots of
noisy network events that we either don't care about or are in the
log4j logs anyway.

See:

http://doc.akka.io/docs/akka/2.0/general/configuration.html

```
# Log level for the very basic logger activated during AkkaApplication startup
# Options: ERROR, WARNING, INFO, DEBUG
# stdout-loglevel = "WARNING"
```

This restored my sanity while running the tests and trying to watch the sbt/scalatest console output.

### DIFF
--- a/core/src/main/scala/spark/util/AkkaUtils.scala
+++ b/core/src/main/scala/spark/util/AkkaUtils.scala
@@ -30,6 +30,7 @@ private[spark] object AkkaUtils {
     val akkaConf = ConfigFactory.parseString("""
       akka.daemonic = on
       akka.event-handlers = ["akka.event.slf4j.Slf4jEventHandler"]
+      akka.stdout-loglevel = "ERROR"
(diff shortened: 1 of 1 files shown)


mateiz:
Oh wow, amazing. Didn't know you could do this. I would even consider switching it to ERROR later.

stephenh:
Wtf...the log messages came back. What a day.

stephenh:
This is ridiculous. After the ActorSystem shuts down, they revert back to the StandardOutLogger, in a state that may/may not be correct.

I swear I saw it work, but perhaps I was delusional.

I've followed up with akka-users to see if they have any ideas.

PR #444, opened 2013-02-03
SPARK-532 Log something when no resources have been offered in the cluster

pwendell:
This solves a long-standing usability issue, that jobs seem to just "hang" when they are submitted to a cluster without enough resources.

This patch adds two warning messages that address two variants of this problem:

(a) A job is submitted to the cluster scheduler, but there aren't any slaves registered yet with the scheduler. Therefore, no offers are made.
(b) Slaves are registered with the Standalone scheduler, but none have enough memory to launch executors on.

This is a preliminary patch which handles the case where these errors happen on cluster startup. Some types of "starvation" are not covered, such as a job arriving in an already full cluster that needs to wait until other jobs finished (it's much more subjective what to log and when in that situation).

### DIFF
--- a/core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
+++ b/core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
@@ -11,6 +11,7 @@ import spark.TaskState.TaskState
 import spark.scheduler._
 import java.nio.ByteBuffer
 import java.util.concurrent.atomic.AtomicLong
+import java.util.{TimerTask, Timer}
 
 /**
  * The main TaskScheduler implementation, for running tasks on a cluster. Clients should first call
@@ -22,8 +23,8 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
 
   // How often to check for speculative tasks
   val SPECULATION_INTERVAL = System.getProperty("spark.speculation.interval", "100").toLong
-  // How often to check for starved TaskSets
-  val STARVATION_CHECK_INTERVAL = System.getProperty("spark.starvation_check.interval", "5000").toLong
+  // Threshold above which we warn user initial TaskSet may be starved
+  val STARVATION_TIMEOUT = System.getProperty("spark.starvation.timeout", "5000").toLong
 
   val activeTaskSets = new HashMap[String, TaskSetManager]
   var activeTaskSetsQueue = new ArrayBuffer[TaskSetManager]
@@ -31,6 +32,10 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
   val taskIdToTaskSetId = new HashMap[Long, String]
   val taskIdToExecutorId = new HashMap[Long, String]
   val taskSetTaskIds = new HashMap[String, HashSet[Long]]
+
+  var hasReceivedTask = false
+  var hasLaunchedTask = false
+  val starvationTimer = new Timer(true)
 
   // Incrementing Mesos task IDs
   val nextTaskId = new AtomicLong(0)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
```diff
@@ -249,6 +266,20 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
     }
   }
 
+  // Find and resource-starved TaskSets and alert the user
+  def detectStarvedTaskSets() {
+    val noOfferThresholdSeconds = 5
```
JoshRosen: Shouldn't you use `STARVATION_CHECK_INTERVAL` here, so that it's configurable?

pwendell: This is actually a different interval. STARVATION_CHECK_INTERVAL determines how often to run the check. This determines how long  of a period without offers qualifies as too long. Someone could want to run the check really frequently but want the timeout to be long, for instance. I could simplify buy just enforcing that they are the same value, since I'm guessing this isn't something anyone will care to configure.

JoshRosen: Why not use a [Timer](http://docs.oracle.com/javase/1.4.2/docs/api/java/util/Timer.html) and schedule a check to be performed `noOfferThresholdSeconds` after the job is submitted?

mateiz: Agree with Josh on making this a timer. Also, the approach of iterating over all task sets seems a little brute-force. Can't we just have a boolean that starts as false and that we set to true the first time we actually launch a task?

pwendell: Timer is nifty - I hadn't used that before. Updated to use a Timer and a boolean variable.

core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
```diff
@@ -94,6 +101,18 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
       activeTaskSets(taskSet.id) = manager
       activeTaskSetsQueue += manager
       taskSetTaskIds(taskSet.id) = new HashSet[Long]()
+
+      if (hasReceivedTask == false) {
+        starvationTimer.scheduleAtFixedRate(new TimerTask() {
```
JoshRosen, line 106: Why is this scheduled periodically rather than only once (e.g. why did you use `Timer.scheduleAtFixedRate()` instead of `Timer.schedule()`)?

pwendell, line 106: What this does is that it periodically prints a warning message until a
slave joins the cluster, then they stop. I'd be fine to just have it print
once, but I feel like it has more of an effect as a periodic message,
because other messages may be printing out at the same time. For instance,
if a bunch of jobs are being submitted, this would keep pinging with the
warning until the user resolved the situation.

pwendell, line 106: Don't have strong feelings either way though, happy to just log once.

JoshRosen, line 106: I don't really have a preference about whether we should log more than once.  I suppose that the periodic logging lets the user know that the system as a whole hasn't hung.

You can prevent the check from being rescheduled after the task has launched by calling `cancel()` inside the TimerTask: http://docs.oracle.com/javase/1.4.2/docs/api/java/util/TimerTask.html#cancel()

mateiz:
Thanks Patrick, I've merged this manually since I had changed a few other things in the standalone master.

PR #445, opened 2013-02-04
Fix exit status in PySpark unit tests; fix/optimize PySpark's RDD.take()

JoshRosen:
This pull request fixes three problems that I ran into while working on the Python tutorial:
- My custom doctest `__main__` was not using the number of failures to determine the exit status, which could lead to doctest failures going undetected unless you inspected the logs.
- `PythonRDD.takePartition` crashed when consuming the output of `textFile()` because it returned the iterator passed to the `runJob` callback function rather than reading it into an array.  This caused problems because the HDFS iterator was closed in an `onComplete` callback.
- PySpark's `RDD.take()` and `RDD.first()` were slow because they transferred more objects from Java to Python than were strictly necessary.

### DIFF
--- a/python/pyspark/rdd.py
+++ b/python/pyspark/rdd.py
@@ -372,6 +372,10 @@ class RDD(object):
         items = []
         for partition in range(self._jrdd.splits().size()):
             iterator = self.ctx._takePartition(self._jrdd.rdd(), partition)
+            # Each item in the iterator is a string, Python object, batch of
+            # Python objects.  Regardless, it is sufficient to take `num`
+            # of these objects in order to collect `num` Python objects:
+            iterator = iterator.take(num)
             items.extend(self._collect_iterator_through_file(iterator))
             if len(items) >= num:
                 break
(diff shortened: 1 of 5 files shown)


python/pyspark/rdd.py
```diff
@@ -748,8 +752,10 @@ def _test():
     # The small batch size here ensures that we see multiple batches,
     # even in these small test examples:
     globs['sc'] = SparkContext('local[4]', 'PythonTest', batchSize=2)
-    doctest.testmod(globs=globs)
+    (failure_count, test_count) = doctest.testmod(globs=globs)
```
mateiz, line 755: Do you need to make this same fix to the accumulator.py doctest?

JoshRosen, line 755: I looked at `run-tests` and noticed that only `rdd.py` and `context.py` are directly executed as scripts.  `accumulator.py` and `broadcast.py` have doctest `__main__` methods, but these aren't used because those tests are run using `python -m doctest` since they don't rely on custom fixtures.  I removed the unnecessary `__main__` methods.

mateiz:
OK, makes sense. Thanks.

PR #446, opened 2013-02-04
SPARK-678: Adding an example with an OLAP roll-up

pwendell:


examples/src/main/scala/spark/examples/OLAPQuery.scala
```diff
@@ -0,0 +1,66 @@
...
+      apache_log_regex findFirstIn line match {
+        case Some(apache_log_regex(ip, _, user, dateTime, query, status, bytes, referer, ua)) =>
+          if (user != "\"-\"") (ip, user, query)
+          else (null, null, null)
+        case _ => (null, null, null)
+      }
+    }
+
+    def extractStats(line: String): Stats = {
+      apache_log_regex findFirstIn line match {
```
mateiz: Try not to use the infix syntax for methods that aren't operators -- it's in our style guide (http://spark-project.org/docs/latest/contributing-to-spark.html)

examples/src/main/scala/spark/examples/OLAPQuery.scala
```diff
@@ -0,0 +1,66 @@
...
+      System.err.println("Usage: OLAPQuery <master> [logFile]")
+      System.exit(1)
+    }
+    val sc = new SparkContext(args(0), "OLAP Query")
+
+    val dataSet =
+      if (args.length == 2) sc.textFile(args(1))
+      else sc.parallelize(exampleApacheLogs)
+
+    val apache_log_regex =
```
mateiz: I would call this apacheLogRegex, and same with the other snake_case names

examples/src/main/scala/spark/examples/OLAPQuery.scala
```diff
@@ -0,0 +1,66 @@
+package spark.examples
+
+import spark.SparkContext
+import spark.SparkContext._
+/**
+ * Executes a roll up-style query against Apache logs.
+ */
+object OLAPQuery {
```
mateiz: Would be nice to call this something like LogQuery for people who don't know what OLAP means (to make it sound less daunting)

pwendell:
Thanks Matei, some code was borrowed from a user example and I forgot to fix the style.

mateiz:
Cool, thanks.

PR #447, opened 2013-02-05
Streaming constructor which takes JavaSparkContext

pwendell:
It's sometimes helpful to directly pass a JavaSparkContext,
and take advantage of the various constructors available for that.


mateiz:
Looks good.

PR #448, opened 2013-02-05
add as many fetch requests as we can, subject to maxBytesInFlight

squito:


mateiz:
So looking at this more carefully, it seems that this will make a difference when there is very little data per node (so each request is smaller than maxBytesInFlight/5). In that case, the current code limits itself to 5 parallel fetches, while this version won't. On the other hand, the current code will launch more than 5 requests at the start anyway. Have you tested this and seen it make a difference? I will include it but I'd like to do some scale tests to check that it doesn't hurt things.

squito:
No, I haven't done any tests at all, just saw this and thought it was odd, which is why I asked about it.  Its easy enough to change to limit it to 5 open requests, in addition to the maxBytesInFlight limit.

However, I don't actually think the old code limited to 5 parallel fetches.  It sort of did, by trying to size fetches to be around maxBytesInFlight / 5.  But, there is no guarantee it actually did that -- fetches could be much smaller, which would lead to more open fetches. (Not that I've observed it actually happening, just that it could.)  With my change as is, we'd have the same 5-ish cap on parallel fetches.  But, I could update this to always be a hard limit of 5 fetches.

actually, I'm working in another branch on instrumenting fetching to find out the time taken during these various stages, which is why I noticed this, but don't have any results from that yet.

squito:
another thought on the limits to bytesInFlight & parallel fetches -- do you want these limits per task, or per jvm?  Right now they are per task (I think), which means that on machines w/ 32 cores, you get 160 open parallel fetches.  Seems like you might want to throttle these at the jvm level

mateiz:
Yeah, good point. I'll think about that but for now I've merged this version.

PR #449, opened 2013-02-06
Increase DriverSuite timeout.

stephenh:
This was timing out for me, plus fails if SPARK_HOME isn't set, so added an explicit assert for that.


mateiz:
Good idea, thanks.

PR #450, opened 2013-02-06
Inline mergePair to look more like the narrow dep branch.

stephenh:
No functionality changes, I think this is just more consistent
given mergePair isn't called multiple times/recursive.

Also added a comment to explain the usual case of having two parent RDDs.


mateiz:
Good catch.

PR #451, opened 2013-02-06
Handle Terminated to avoid endless DeathPactExceptions.

stephenh:
Credit to Roland Kuhn, Akka's tech lead, for pointing out this
various obvious fix, but StandaloneExecutorBackend.preStart's
catch block would never (ever) get hit, because all of the
operation's in preStart are async.

So, the System.exit in the catch block was skipped, and instead
Akka was sending Terminated messages which, since we didn't
handle, it turned into DeathPactException, which started
a postRestart/preStart infinite loop.

### DIFF
--- a/core/src/main/scala/spark/deploy/worker/Worker.scala
+++ b/core/src/main/scala/spark/deploy/worker/Worker.scala
@@ -1,19 +1,16 @@
 package spark.deploy.worker
 
 import scala.collection.mutable.{ArrayBuffer, HashMap}
-import akka.actor.{ActorRef, Props, Actor}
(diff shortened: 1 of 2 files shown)


stephenh:
I will make no comments about how the tests may/may not be working better with this change, lest I jinx it.

mateiz:
Got it, that makes sense. Thanks for the fix!

mateiz:
Actually this seems to not be in sync with master -- can you rebase it? Or want me to just merge it manually?

stephenh:
Sure, I'll merge.

stephenh:
Cool, fixed, just some import conflicts.

mateiz:
Alright, thanks.

PR #452, opened 2013-02-06
Add RDD.coalesce, clean up some RDDs, other misc.

stephenh:
Sorry, I put 4 commits into this pull request, but if you review each commit separately, it should be fairly self-contained/explanatory.

### DIFF
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -656,7 +656,6 @@ abstract class RDD[T: ClassManifest](
    */
   private[spark] def markCheckpointed(checkpointRDD: RDD[_]) {
     clearDependencies()
-    dependencies_ = null
     splits_ = null
     deps = null    // Forget the constructor argument for dependencies too
   }
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -20,6 +20,7 @@ import spark.partial.BoundedDouble
 import spark.partial.CountEvaluator
 import spark.partial.GroupedCountEvaluator
 import spark.partial.PartialResult
+import spark.rdd.CoalescedRDD
 import spark.rdd.CartesianRDD
 import spark.rdd.FilteredRDD
 import spark.rdd.FlatMappedRDD
@@ -230,6 +231,12 @@ abstract class RDD[T: ClassManifest](
     map(x => (x, null)).reduceByKey((x, y) => x, numSplits).map(_._1)
 
   def distinct(): RDD[T] = distinct(splits.size)
+
+  /**
+   * Return a new RDD that is reduced into `numSplits` partitions.
+   */
+  def coalesce(numSplits: Int = sc.defaultParallelism): RDD[T] =
+    new CoalescedRDD(this, numSplits)
 
   /**
    * Return a sampled subset of this RDD.
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -235,8 +235,7 @@ abstract class RDD[T: ClassManifest](
   /**
    * Return a new RDD that is reduced into `numSplits` partitions.
    */
-  def coalesce(numSplits: Int = sc.defaultParallelism): RDD[T] =
-    new CoalescedRDD(this, numSplits)
+  def coalesce(numSplits: Int): RDD[T] = new CoalescedRDD(this, numSplits)
 
   /**
    * Return a sampled subset of this RDD.
--- a/core/src/main/scala/spark/api/java/JavaRDD.scala
+++ b/core/src/main/scala/spark/api/java/JavaRDD.scala
@@ -39,6 +39,11 @@ JavaRDDLike[T, JavaRDD[T]] {
     wrapRDD(rdd.filter((x => f(x).booleanValue())))
 
   /**
+   * Return a new RDD that is reduced into `numSplits` partitions.
+   */
+  def coalesce(numSplits: Int): JavaRDD[T] = rdd.coalesce(numSplits)
+
+  /**
    * Return a sampled subset of this RDD.
    */
   def sample(withReplacement: Boolean, fraction: Double, seed: Int): JavaRDD[T] =
--- a/core/src/main/scala/spark/api/java/JavaRDDLike.scala
+++ b/core/src/main/scala/spark/api/java/JavaRDDLike.scala
@@ -133,7 +133,7 @@ trait JavaRDDLike[T, This <: JavaRDDLike[T, This]] extends PairFlatMapWorkaround
   /**
    * Return a new RDD that is reduced into `numSplits` partitions.
    */
-  def coalesce(numSplits: Int): RDD[T] = rdd.coalesce(numSplits)
+  def coalesce(numSplits: Int): JavaRDD[T] = rdd.coalesce(numSplits)
 
   /**
    * Return an RDD of grouped elements. Each group consists of a key and a sequence of elements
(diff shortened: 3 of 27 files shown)


core/src/main/scala/spark/RDD.scala
```diff
@@ -649,7 +656,6 @@ abstract class RDD[T: ClassManifest](
    */
   private[spark] def markCheckpointed(checkpointRDD: RDD[_]) {
     clearDependencies()
-    dependencies_ = null
```
stephenh, line 652: This is done by clearDependencies().

Also, just for my own curiosity, why do we null out splits and deps here in markCheckpointed, but not in clearDependencies?

core/src/main/scala/spark/rdd/BlockRDD.scala
```diff
@@ -11,10 +11,6 @@ private[spark]
 class BlockRDD[T: ClassManifest](sc: SparkContext, @transient blockIds: Array[String])
   extends RDD[T](sc, Nil) {
 
-  @transient var splits_ : Array[Split] = (0 until blockIds.size).map(i => {
-    new BlockRDDSplit(blockIds(i), i).asInstanceOf[Split]
-  }).toArray
```
stephenh, line 16: I went through the RDDs looking for places where the old splits_ could be changed to just use getSplits.

core/src/main/scala/spark/rdd/CartesianRDD.scala
```diff
@@ -66,6 +66,7 @@ class CartesianRDD[T: ClassManifest, U:ClassManifest](
   )
 
   override def clearDependencies() {
+    super.clearDependencies()
```
stephenh, line 69: AFAICT it seems like calling super.clearDependencies() is needed to unset the dependencies_ field in RDD.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -31,7 +31,7 @@ class CoalescedRDD[T: ClassManifest](
     maxPartitions: Int)
   extends RDD[T](prev.context, Nil) {  // Nil since we implement getDependencies
 
-  override def getSplits: Array[Split] = {
+  override def getSplits = {
```
stephenh: It made it consistent so that the override getSplits/getDependencies don't declare the return type.

stephenh:
Hi Matei; just a quick ping about this, I know I got lazy making just 1 pull request for these 4 commits; would you rather I submit them as 4 separate pull requests?

mateiz:
No, I just haven't had the time to look through it yet; will do so soon.

core/src/main/scala/spark/RDD.scala
```diff
@@ -232,6 +233,12 @@ abstract class RDD[T: ClassManifest](
   def distinct(): RDD[T] = distinct(splits.size)
 
   /**
+   * Return a new RDD that is reduced into `numSplits` partitions.
+   */
+  def coalesce(numSplits: Int = sc.defaultParallelism): RDD[T] =
```
mateiz: Let's make this not take a default value for now, since people may be confused when they just see `coalesce()`

core/src/main/scala/spark/api/java/JavaRDDLike.scala
```diff
@@ -131,6 +131,16 @@ trait JavaRDDLike[T, This <: JavaRDDLike[T, This]] extends PairFlatMapWorkaround
       other.classManifest)
 
   /**
+   * Return a new RDD that is reduced into the default number of partitions.
+   */
+  def coalesce(): RDD[T] = coalesce(rdd.context.defaultParallelism)
```
mateiz: Same here, remove this version

core/src/main/scala/spark/rdd/UnionRDD.scala
```diff
@@ -38,7 +38,7 @@ class UnionRDD[T: ClassManifest](
     array
   }
 
-  override def getDependencies: Seq[Dependency[_]] = {
```
mateiz, line 41: So I wouldn't remove types like here and in `getSplits`; I think it's better documentation to leave the types in. I realize that the code is a bit inconsistent on this, especially for things like `toString`.

mateiz:
Looks good overall, but I made a few comments on the style and on the coalesce() API. Thanks for doing all this cleanup!

core/src/main/scala/spark/rdd/UnionRDD.scala
stephenh, line 41: Makes sense, yeah, I just didn't know which way to go. Added them all back.

core/src/main/scala/spark/api/java/JavaRDDLike.scala
```diff
@@ -131,6 +131,11 @@ trait JavaRDDLike[T, This <: JavaRDDLike[T, This]] extends PairFlatMapWorkaround
       other.classManifest)
 
   /**
+   * Return a new RDD that is reduced into `numSplits` partitions.
+   */
+  def coalesce(numSplits: Int): RDD[T] = rdd.coalesce(numSplits)
```
mateiz: Actually, one more thing: I believe this should return a JavaRDD / JavaPairRDD / etc based on what it was called on, not an RDD. Look at the `filter` method for example. In order to deal with this, it was implemented separately in JavaRDD.scala, JavaPairRDD.scala, and JavaDoubleRDD.scala.

stephenh:
Good catch on the JavaRDD return type; fixed.

mateiz:
It's actually a bit more complicated -- the problem is that for RDDs of key-value pairs and Doubles, there are separate JavaRDDLike subclasses, called JavaPairRDD and JavaDoubleRDD. Since coalesce() doesn't change an RDD's type, it should be defined separately in each one to return the corresponding type. Take a look at how filter() is implemented for example (it has the same property of not changing type). There are basically 3 different implementations of it.

JoshRosen:
In principle, we should have been able to implement something similar to Scala's [TraversableLike](http://docs.scala-lang.org/overviews/core/architecture-of-scala-collections.html) to factor out these common operations into a `JavaRDDLike` class (in fact, some vestiges of an early attempt at this are still in the code).  Due to some Scala compiler bugs (e.g. [SI-6050](https://issues.scala-lang.org/browse/SI-6050)), this doesn't actually work from Java, hence the duplication of code across the `Java*RDD` classes.

If we can get the "implementation trait" technique to work reliably, I'd love to use it since it would result in much less code duplication.  In the meantime, you'll have to add `coalesce()` to each class.

stephenh:
Ah, okay, I completely missed that. Thanks for pointing it out.

core/src/main/scala/spark/api/java/JavaRDD.scala
```diff
@@ -39,6 +39,11 @@ JavaRDDLike[T, JavaRDD[T]] {
     wrapRDD(rdd.filter((x => f(x).booleanValue())))
 
   /**
+   * Return a new RDD that is reduced into `numSplits` partitions.
+   */
+  def coalesce(numSplits: Int): JavaRDD[T] = rdd.coalesce(numSplits)
```
mateiz, line 44: Does this actually compile? Looks like the rdd.coalesce should be wrapped to turn it into a JavaRDD (sorry if I'm missing something)

stephenh, line 44: Yeah, fromRDD is implicit in this class (although not the others, so I agree it stands out).

mateiz, line 44: Oh, weird. Okay, thanks!

JoshRosen, line 44: Yeah, there's an implicit conversion from RDD[T] to JavaRDD[T].  The `wrapRDD` calls on the other methods are a leftover from when I tried to factor these methods into a common implementation trait.

PR #453, opened 2013-02-06
Change docs on 'reduce' since the merging of local reduces no longer pre...

markhamstra:
...serves

ordering, so the reduce function must also be commutative.


mateiz:
Hey Mark, this seems to have a merge conflict now. Mind re-merging master into your branch?

mateiz:
Great, thanks!

PR #454, opened 2013-02-07
SPARK-685 Adding IPYTHON environment variable support for launching pyspark using ...

MLnick:
...ipython shell


mateiz:
Awesome, thanks!

PR #456, opened 2013-02-08
Utility function to get a setup & cleanup function for mapping each partition

squito:
Often when mapping some RDD, you want to do a bit of setup before processing each partition, followed by cleanup at the end of the partition; this adds utility functions to make that easier.

I felt that this is worth including because its a little tricky to get right -- I needed to add a "CleanupIterator", and I have an example in the unit tests of how this fails w/out it.  OTOH, I wasn't sure if this necessarily belongs in the spark api itself (eg., do we add a version of foreach with this also?)

We find it a useful addition, and so thought others might also ...


core/src/test/scala/spark/RDDSuite.scala
```diff
@@ -4,6 +4,8 @@ import scala.collection.mutable.HashMap
 import org.scalatest.FunSuite
 import spark.SparkContext._
 import spark.rdd.{CoalescedRDD, PartitionPruningRDD}
+import spark.RDD.PartitionMapper
+import collection._
```
JoshRosen: This style of import doesn't match the convention used in the rest of the codebase.  It should probably be replaced with an import of `scala.collection.mutable.Set`, since that appears to be the only class that it's importing.

core/src/test/scala/spark/RDDSuite.scala
```diff
@@ -173,4 +175,48 @@ class RDDSuite extends FunSuite with LocalSparkContext {
...
+        val values = mutable.Set[Int]()
+        val mItr = itr.map{i => values += i; i * 2}
+        //you haven't actually put anything into values yet, b/c itr.map defines another
+        // iterator, which is lazily computed.  so the Set is empty
+        acc2.localValue += (partition -> Set[Int](values.toSeq: _*))
+        mItr
+    }.collect
+
+    assert(m2.toSet === (1 to 100).map{_ * 2}.toSet)
+    assert(acc2.value.keySet == (0 to 3).toSet)
```
JoshRosen: This should be a triple-equals.

core/src/test/scala/spark/RDDSuite.scala
```diff
@@ -173,4 +175,48 @@ class RDDSuite extends FunSuite with LocalSparkContext {
...
+    //this condition will fail
+//    acc2.value.foreach { case(partition, values) =>
```
JoshRosen, line 222: If this is supposed to fail, should we wrap it in an `intercept` block and check that an appropriate exception is actually thrown?

core/src/main/scala/spark/RDD.scala
```diff
@@ -34,6 +31,8 @@ import spark.rdd.ZippedRDD
 import spark.storage.StorageLevel
 
 import SparkContext._
+import spark.RDD.PartitionMapper
+import util.CleanupIterator
```
mateiz: Please don't use relative import names for packages (see http://spark-project.org/docs/latest/contributing-to-spark.html)

core/src/test/scala/spark/RDDSuite.scala
squito, line 222: well, its not really _supposed_ to fail -- there is no exception, it just doesn't give the "expected" result.  that second part isn't really a unit test at all, its just documentation of why this method is needed.  probably doesn't belong here at all, I just wanted it as part of the pull request to demonstrate why the method was needed.

I guess I can just write it up somewhere else (seems too long to put in the spark docs also, at least in the current layout ...)

squito:
I cleaned up the style issues, sorry about that.
I left that "failing" test in there ... happy to remove it if you want me to, just wanted to be clear if you want me to document the trouble w/ mapPartitions somewhere

stephenh:
Hi Imran,

Have you seen the TaskContext and addOnCompleteCallback? That is what HadoopRDD uses to close the FileStream after all of the lines in a Hadoop file have been read.

You might be able to do what you're doing with just a custom RDD that did something like:

```
override def compute(s: Split, context: TaskContext): Iterator[(T, U)] = {
   setupDbConnection()
   context.addOnCompleteCallback { tearDownDbConnection()
   // call parent rdd or do own compute stuff
```

I believe this will achieve the same thing, as compute will be called on each partition, and you'll have start/stop hooks around the execution on each partition.

mateiz:
I agree with Stephen here. The addOnCompleteCallback mechanism also makes sure to call your handler if the task throws an exception, which is important.

Also, can you add a similar method in the Java API? I guess you would need to create an interface in Java for the setup/close stuff too.

squito:
good point, I definitely hadn't thought about ensuring cleanup w/ exceptions.

I've updated it to use onCompleteCallback.  I also added it to the java api -- I added separate classes for PairRDDs & DoubleRDDs, dunno if there is a better way to do that.

stephenh:
I wonder if this could be done with something more like decoration:

``` scala
val rdd = sc.textFile(...).setupPartitions(dbSetupLogic).mapPartitions(...).cleanupPartitions(dbCloseLogic)
```

So there would be two new RDDs, PartitionSetupRDD that first invoked its setup function once/partition, then called firstParent.compute, and then PartitionCleanupRDD, that setup the complete callback for its cleanup function.

Not sure if the decoupling would lead to unintended/nonsensical use cases. But, just musing, then perhaps they could be used separately, if you only need one or the other, or without the map, which PartitionMapper currently forces you do to.

Also, I just like that this would use plain functions and not a new "PartitionMapper" interface--for some reason that doesn't feel quite right, but I can't think of a better name.

I see what you're trying to do though.

core/src/main/scala/spark/api/java/JavaRDDLike.scala
```diff
@@ -117,6 +119,41 @@ trait JavaRDDLike[T, This <: JavaRDDLike[T, This]] extends PairFlatMapWorkaround
...
+  def mapWithSetupAndCleanup[U](m: PartitionMapper[T,U]): JavaRDD[U] = {
+    JavaRDD.fromRDD(rdd.mapWithSetupAndCleanup(m)(fakeManifest[U]))(fakeManifest[U])
+  }
+
+  /**
+   * Return a new RDD by applying a function to each element of the RDD, with an additional
+   * setup & cleanup that happens before & after computing each partition
+   */
+  def mapWithSetupAndCleanup[K,V](m: JavaPairPartitionMapper[T,K,V]): JavaPairRDD[K,V] = {
+    val scalaMapper = new PartitionMapper[T,(K,V)] {
```
JoshRosen: Can`JavaPairPartitionMapper<T, K, V>` be an abstract class that extends or implements `PartitionMapper<T, Tuple2<K, V>>`?  If you can do that, then you wouldn't have to wrap the the Java PartitionMapper to convert it into its Scala counterpart.

core/src/main/scala/spark/api/java/ManifestHelper.java
```diff
@@ -0,0 +1,11 @@
+package spark.api.java;
+
+import scala.reflect.ClassManifest;
+import scala.reflect.ClassManifest$;
+
+class ManifestHelper {
```
JoshRosen, line 6: This ManifestHelper class is a good idea.  We could also use it to create the fake manifests in the Java `Function*` classes.

mateiz:
Regarding Stephen's comment -- I think it's better to keep PartitionMapper a single object instead of doing functions, in case you need to share state among the setup, map and clean methods (e.g. you open some external resource, use it in your map, then close it).

stephenh:
Ah, good point. That makes sense.

squito:
I updated JavaPairPartitionMapper, per Josh's suggestion.  (We lose the ability for map to throw an exception, but that is already the case for the basic PartitionMapper.)

I tried doing the same thing for JavaDoubleRDD, but somehow I got stuck with weird manifest errors.  First it complained:

[error]  found   : ClassManifest[scala.Double]
[error]  required: ClassManifest[java.lang.Double]

Then when I switched to explicitly using a java.lang.Double manifest, it reversed:

[error]  found   : spark.RDD[java.lang.Double]
[error]  required: spark.RDD[scala.Double]

so I just left it as is

squito:
OK I think this is ready to go now.  I got rid of the need for the helper object for JavaDoubleRDD, by just casting from java.lang.Double to scala.Double, and it seems happy.  Also I put a throws Exception declaration on map in PartitionMapper, for the java api.

squito:
Just curious what the status is on this -- waiting for some additional changes here, decided against merging it, or just haven't gotten to it yet.

core/src/test/scala/spark/JavaAPISuite.java
```diff
@@ -401,6 +398,54 @@ public void mapPartitions() {
   }
 
   @Test
+  public void mapPartitionsWithSetupAndCleanup() {
```
mateiz, line 401: Please use two spaces for the indentation in this file (it looks like maybe it's tabs, or multiple spaces)

core/src/main/scala/spark/RDD.scala
```diff
@@ -340,6 +339,18 @@ abstract class RDD[T: ClassManifest](
     new MapPartitionsWithSplitRDD(this, sc.clean(f), preservesPartitioning)
 
   /**
+   * Return a new RDD by applying a function to every element in this RDD, with extra setup & cleanup
+   * at the beginning & end of processing every partition.
+   *
+   * This might be useful if you need to setup some resources per task & cleanup them up at the end, eg.
+   * a db connection
+   */
+  def mapWithSetupAndCleanup[U: ClassManifest](
```
mateiz: I would rename the instances of WithSetupAndCleanup to just WithCleanup for simplicity. Having cleanup will also imply that there's something to be cleaned up.

mateiz:
Sorry, just hadn't had a chance to look at it. It looks good but I made two small comments.

squito:
thanks!  I've updated to take those comments into account.

squito:
wow, somhow I totally missed committing changes to one file before ... hope you didn't waste time looking at it before, now its actually all there

PR #458, opened 2013-02-09
Use stubs instead of mocks for DAGSchedulerSuite.

stephenh:
@woggling @mateiz This is still in "request for comment state", but I took a crack at changing the DAGSchedulerSuite over from mocks to stubs.

I admittedly have a somewhat obsessive fascination with stubs, in that I think they can lead to simpler behavior than mocks and so make the tests easier to read.

This is because with stubs the state is being passed around much more directly, instead of juggling with the mock framework's expect abstractions.

Still outstanding is a few assertions I need to put back, and then stubbing out DAGScheduler's use of "new Thread" in it's run local mode, either by introducing a "LocalRunner" abstraction, or, more likely, just overriding DAGScheduler.runLocally to invoke job.func immediately (so the test doesn't have to wait for the thread to do it).

This would remove the last bit of "waiting on threads" behavior (the Thread.sleep hack I have now) and mean everything happens (for the unit test) synchronously/immediately.

I can continue to clean this up if you want, but thought since I have all of the existing tests ported over and passing, it would be a good place to check in.


core/src/main/scala/spark/MapOutputTracker.scala
```diff
@@ -38,9 +38,10 @@ private[spark] class MapOutputTrackerActor(tracker: MapOutputTracker) extends Ac
   }
 }
 
-private[spark] class MapOutputTracker(actorSystem: ActorSystem, isDriver: Boolean) extends Logging {
```
stephenh, line 41: These changes to MapOutputTracker and BlockManagerMaster are somewhat unrelated to the DAGSchedulerSuite change, but basically pull the common actor lookup code out of MapOutputTracker/BlockManagerMaster and do it in SparkEnv, so that these two classes just need their actor, and that's it.

mateiz:
Hey Stephen,

Sorry for taking some time to get to this -- I thought I had replied earlier. I think using stubs here is great. Regarding DAGScheduler.runLocally though, that needs to happen in a separate thread to avoid blocking the DAGScheduler event loop when there are multiple concurrent jobs. If you want, you can add an option for running it in the same thread in tests.

stephenh:
Hi Matei. Yeah, I know the production runLocally has to keep using the thread, I'll just try and trick it for the tests. Since you okay the approach, I'll take a look at this again, merge in the latest master, and let you know when it's ready for re-review.

stephenh:
Okay, took me awhile to get to it, but I've merged in master and override runLocally so that the test does not need a Thread.sleep anymore.

stephenh:
Hm. Actually, hold on, looking into a test failing; not sure if it's because of this or not.

stephenh:
Okay, now the tests pass, and it should be ready to merge.

mateiz:
Alright, I'll merge it soon (was out of town last week).

mateiz:
I've merged this manually due to some small conflicts with the task metrics stuff. Thanks Stephen!

stephenh:
Awesome, thanks for getting to this--I'd be meaning to come back and merge in master. Thanks for taking care of it.

PR #459, opened 2013-02-10
Change defaultPartitioner to use upstream split size.

stephenh:
Previously it used the SparkContext.defaultParallelism, which occassionally
ended up being a very bad guess. Looking at upstream RDDs seems to make
better use of the context.

Also sorted the upstream RDDs by partition size first, as if we have
a hugely-partitioned RDD and tiny-partitioned RDD, it is unlikely
we want the resulting RDD to be tiny-partitioned.

AFAIK these changes make sense--let me know if I'm missing something.

Currently I have an explicit split size littering my code, and I think this would basically get rid of it.


stephenh:
Just as an illustration, this snippet:

``` scala
    val sc = new SparkContext("local", "test", "/home/stephen/other/spark")
    val a = sc.makeRDD(Array(1, 2, 3, 4), 2000)
    val b = a.map(a => (a, (a * 2).toString))
    val c = b.groupByKey()
    println(c.toDebugString)
```

Started out with 2000 splits, but in the groupByKey was dropped down to 1:

```
MapPartitionsRDD[4] at main at <unknown>:0 (1 splits)
  ShuffledRDD[3] at main at <unknown>:0 (1 splits)
    MapPartitionsRDD[2] at main at <unknown>:0 (2000 splits)
      MappedRDD[1] at main at <unknown>:0 (2000 splits)
        ParallelCollection[0] at main at <unknown>:0 (2000 splits)
```

Obviously this is an extreme example (since it uses local mode).

With this pull request, the toDebugOutput changes to:

```
MapPartitionsRDD[4] at main at <unknown>:0 (2000 splits)
  ShuffledRDD[3] at main at <unknown>:0 (2000 splits)
    MapPartitionsRDD[2] at main at <unknown>:0 (2000 splits)
      MappedRDD[1] at main at <unknown>:0 (2000 splits)
        ParallelCollection[0] at main at <unknown>:0 (2000 splits)
```

Which I believe is much less surprising behavior for someone to starts out specifying a large of number of partitions.

pwendell:
Hey Stephen,

The behavior of using defaultParallelism is mentioned a few times in the docs (at least once in the Scala programming guide, in the ScalaDoc for combineByKey, e.g.). If this ends up being merged, those will need to be updated along with the patch. It would be best to look through the Scala, Java, and Python API docs and the programming guides and update accordingly.

stephenh:
Hi Patrick, that's a good point--if this change is given a preliminary okay, I'll go through and update the docs.

stephenh:
Just fyi, @mateiz I am working on flushing this out, updating the docs, etc. If you think it's a horribly bad idea, now would be a good time. :-)

(Basically, defaultParallelism is going to be used a lot less--the max parent split size just seems like a saner default, as it keeps the same upstream split size, instead of reverting back when the user may not expect it.)

stephenh:
Okay, I believe this is ready to merge, assuming it's given the okay.

stephenh:
...er, right, found more docs. Hold on.

stephenh:
Okay, Patrick was right, that was mentioned a lot. I believe I have got all of them.

mateiz:
Hey Stephen, so I think this is good, but I'm a bit uncomfortable changing the meaning of spark.default.paralellism for user programs that set it. How about calculating the default this way only if spark.default.paralellism is not set? Otherwise, if that has a value, we would use that.

stephenh:
Hi Matei, okay, I suppose that makes sense. With this change, spark.default.parallelism was going to become fairly unused, so using it as the "old behavior on/off" switch makes sense. I'll work on that.

JoshRosen:
@mateiz: How about retaining the `spark.default.parallelism` setting, but removing it from the documentation and raising deprecation warnings when it's used?

mateiz:
I'd rather not deprecate spark.default.parallelism right now because I'm not convinced that it's a bad idea. It seems more convenient than relying only on the upstream size in some cases. So let's see how things go with this change, and we might remove it later.

mridulm:
We are also looking at how to infer the right number of reduce tasks
required .. This looks like an interesting approach : if it was possible to
allow for hints as part of flatmap, filter, etc (in terms of expected
record change) that would additionally help in sizing the number of splits
and reducers.

For example 2000 mappers (by input split) which is heavily filtered.
A global default is not optimal - but inferring wrong number of reducers in
this case would be very expensive ... If user could give a hint, that would
be great ...
Regards
Mridul

stephenh:
Okay, I changed defaultPartitioner to use defaultParallelism if the System property is set.

mateiz:
Great, thanks!

PR #462, opened 2013-02-11
Track assorted metrics for each task, report summaries to user at stage completion

squito:
Three major parts here:
1) each task sends a TaskMetrics object back to the driver, to track some of the stats that used to just get printed out (eg. run time)
2) add some stats related to shuffles
3) create a SparkListener to get notification of various events (for now w/ just a StageCompletion event).

This is not ready to get merged in right now (missing docs & tests, at least), but I think its ready for some general feedback.  In particular, I dunno if there is a better way to get all of the metrics back to the driver.  I had to some coupling between components that used to be more independent.  (eg., a TaskContext now has a reference to a Task).

I probably don't have the right set of metrics at the moment, but with this setup, we can add a bunch more and see what sticks.  If possible, I'd like to just get it merged into master first, rather than fret over the exact metrics right now.

With the metrics I have so far, & using the provided SparkListener, after each stage the user gets output like:

```
Finished stage: MappedRDD[7] at map at <console>:18
13/02/10 09:49:22 INFO scheduler.StatsReportListener: task runtime:(count: 64, mean: 144.562500, stdev: 83.793249)
13/02/10 09:49:22 INFO scheduler.StatsReportListener:   0%  5%  10% 25% 50% 75% 90% 95% 100%
13/02/10 09:49:22 INFO scheduler.StatsReportListener:    102     104     107     112     131     148     170     174     610
13/02/10 09:49:17 INFO scheduler.StatsReportListener: shuffle bytes written:(count: 64, mean: 265079.562500, stdev: 3616.055045)
13/02/10 09:49:17 INFO scheduler.StatsReportListener:   0%  5%  10% 25% 50% 75% 90% 95% 100%
13/02/10 09:49:17 INFO scheduler.StatsReportListener:   253.8 KB    254.9 KB    255.7 KB    256.8 KB    258.6 KB    259.6 KB    261.5 KB    262.8 KB    278.6 KB
13/02/10 09:49:22 INFO scheduler.StatsReportListener: fetch wait time:(count: 64, mean: 0.765625, stdev: 1.608009)
13/02/10 09:49:22 INFO scheduler.StatsReportListener:   0%  5%  10% 25% 50% 75% 90% 95% 100%
13/02/10 09:49:22 INFO scheduler.StatsReportListener:      0       0       0       0       0       1       2       3       9
13/02/10 09:49:22 INFO scheduler.StatsReportListener: remote bytes read:(count: 64, mean: 4498011.000000, stdev: 14247.297252)
13/02/10 09:49:22 INFO scheduler.StatsReportListener:   0%  5%  10% 25% 50% 75% 90% 95% 100%
13/02/10 09:49:22 INFO scheduler.StatsReportListener:   4.2 MB  4.3 MB  4.3 MB  4.3 MB  4.3 MB  4.3 MB  4.3 MB  4.3 MB  4.3 MB
```


core/src/main/scala/spark/rdd/CheckpointRDD.scala
```diff
@@ -104,7 +104,7 @@ private[spark] object CheckpointRDD extends Logging {
     val deserializeStream = serializer.deserializeStream(fileInputStream)
 
     // Register an on-task-completion callback to close the input stream.
-    context.addOnCompleteCallback(() => deserializeStream.close())
+    context.addOnCompleteCallback(_ => deserializeStream.close())
```
stephenh: Does this behave differently from what was there before?

stephenh: Oh, sorry, I see you changed on addOnCompleteCallback works.

core/src/main/scala/spark/BlockStoreShuffleFetcher.scala
```diff
@@ -45,6 +46,9 @@ private[spark] class BlockStoreShuffleFetcher extends ShuffleFetcher with Loggin
         }
       }
     }
-    blockManager.getMultiple(blocksByAddress).flatMap(unpackBlock)
+    val blockFetcherItr = blockManager.getMultiple(blocksByAddress)
+    val itr = new TimedIterator(blockFetcherItr.flatMap(unpackBlock)) with DelegateBlockFetchTracker
+    itr.setDelegate(blockFetcherItr)
```
ryanlecompte, line 53: Any reason you don't pass in the delegate to TimedIterator's constructor as opposed to a separate setter?

core/src/main/scala/spark/executor/TaskMetrics.scala
```diff
@@ -0,0 +1,19 @@
+package spark.executor
+
+import spark.scheduler.Task
+
+case class TaskMetrics(
+  val totalBlocksFetched : Option[Int],
```
ryanlecompte: "val" here is redundant for case classes.

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -73,6 +74,10 @@ class DAGScheduler(
 
   val shuffleToMapStage = new TimeStampedHashMap[Int, Stage]
 
+  private[spark] val stageToInfos = new TimeStampedHashMap[Stage, StageInfo]
+
+  private[spark] var sparkListeners = ArrayBuffer[SparkListener]()
```
ryanlecompte: Should this be a val instead of a var?

core/src/main/scala/spark/rdd/CheckpointRDD.scala
ryanlecompte: This looks awesome, @squito!

On Sun, Feb 10, 2013 at 10:11 PM, Stephen Haberman <notifications@github.com

stephenh:
This looks pretty awesome; I'd love to have these stats.

I wonder if eventually we could store them on the Stage object, because, for my jobs anyway (which are big enough to have lots of console output, but have only ~5-10 or so stages), it'd be nice to have a summary at the very end when the job is done like:

```
Stage 0 -- stats
   Stage 1 -- stats
   Stage 2 -- stats
```

Just because I'm lazy and don't want to hunt through all of the logs.

ryanlecompte:
I thought about this a bit more last night, and I think it might make sense to introduce a proven metrics collection library within Spark, such as Twitter's [ostrich](https://github.com/twitter/ostrich) library or Coda Hale's [metrics](https://github.com/codahale/metrics) library. These libraries give us gauges, timers, histograms, and counters. They already handle things like tracking the mean, standard deviation, quartiles, etc. Also you get the added benefit of having the standard JVM metrics being tracked also. The metrics are exposed as JSON too, which means we can expose them in the Spark dashboard, dump them in the log file, or pump them into graphite or librato for graphing. I'd encourage us to use Ostrich or the Metrics library here.

core/src/main/scala/spark/rdd/CheckpointRDD.scala
squito: but looks like I didn't have any reason to change the behavior.  Must have been from some intermediate version I was trying.  good catch, I undid those changes.

core/src/main/scala/spark/BlockStoreShuffleFetcher.scala
squito, line 53: setDelegate is actually a method of the _trait_ DelegateBlockFetchTracker, which is why it can't be set in a constructor.  But maybe that trait is overkill ... I agree that it really ought to be in the constructor

squito:
@stephenh -- absolutely, I don't want to troll through the logs either, so I do plan to add the ability to just dump out all the stageInfos at once, along w/ the rdd storage info.  we can add that later.  For now, you can always do this by creating a listener that just accumulates all the stageInfos, and then display them yourself whenever you want.

core/src/main/scala/spark/TaskContext.scala
```diff
@@ -1,9 +1,11 @@
 package spark
 
 import scala.collection.mutable.ArrayBuffer
+import spark.scheduler.Task
 
-
-class TaskContext(val stageId: Int, val splitId: Int, val attemptId: Long) extends Serializable {
+class TaskContext(val stageId: Int, val splitId: Int, val attemptId: Long, val task: Task[_]) extends Serializable {
+  //by adding Task here, I'm destroying the separation between Task & TaskContext ... not sure why they need to
+  // be separate
```
mateiz: This is because Task was an internal API that we might want to change, while TaskContext was a more external one that anyone subclassing RDD might use. Which fields of Task do you need to use in metrics?

mateiz:
Hey, so this approach definitely looks good overall. I made one small comment on TaskContext. I can make more detailed ones, but let me know when you think this is ready for a close look.

core/src/main/scala/spark/TaskContext.scala
squito: actually, I only need access to fields of Task that I added.  That way, as the metrics are collected, I can stuff them directly into the Task, and then they are easy to send back along w/ the task results.  I'll highlight the places in the code where its used.

The TaskContext is not available in the Executor when its sending task results back -- would it be better if I put the metrics into the TaskContext, made a Task store a reference to the TaskContext, and then extract the metrics that way at the end of the Task?

core/src/main/scala/spark/rdd/CoGroupedRDD.scala
```diff
@@ -102,9 +102,17 @@ class CoGroupedRDD[K](@transient var rdds: Seq[RDD[(_, _)]], part: Partitioner)
...
+        for ((k, vs) <- fetchItr) {
           getSeq(k)(depNum) ++= vs
         }
+        context.task.shuffleReadMillis = Some(fetchItr.getNetMillis)
+        context.task.remoteFetchTime = Some(fetchItr.remoteFetchTime)
+        context.task.remoteFetchWaitTime = Some(fetchItr.remoteFetchWaitTime)
+        context.task.remoteReadBytes = Some(fetchItr.remoteBytesRead)
+        context.task.totalBlocksFetched = Some(fetchItr.totalBlocks)
+        context.task.localBlocksFetched = Some(fetchItr.numLocalBlocks)
+        context.task.remoteBlocksFetched = Some(fetchItr.numRemoteBlocks)
```
squito: Here's where I put the metrics from the shuffle fetches directly into the Task -- I can only do this b/c I made TaskContext keep a reference to the Task.

core/src/main/scala/spark/executor/Executor.scala
```diff
@@ -95,7 +95,7 @@ private[spark] class Executor extends Logging {
         env.mapOutputTracker.updateGeneration(task.generation)
         val value = task.run(taskId.toInt)
         val accumUpdates = Accumulators.values
-        val result = new TaskResult(value, accumUpdates)
+        val result = new TaskResult(value, accumUpdates, TaskMetrics(task))
```
squito: Here the metrics are read out of the task, and sent back to the driver along w/ the TaskResults.  (TaskContext is not available here)

core/src/main/scala/spark/TaskContext.scala
mateiz: I think the right way is to either add these values onto TaskContext, or reference a different object, such as a MetricsReporter of some kind. The latter is cleaner from a design point of view.

squito:
I think this is ready to merge now.  TaskContext no longer has a reference to Task, instead it has a reference to TaskMetrics, which is shared w/ the Task.  I though it was better to make TaskMetrics a member of TaskContext, then to change the signature of RDD.compute() to include another TaskMetrics argument, since that would effect anybody that made their own RDD.

I've also expanded the set of metrics somewhat -- if you like, I can trim this down to just a few of them for the moment, so that this can get merged into master, and keep experiments with other metrics separate.

core/src/main/scala/spark/executor/TaskMetrics.scala
```diff
@@ -0,0 +1,34 @@
...
+ * @param localBlocksFetched local blocks fetched in a shuffle
+ * @param shuffleReadMillis total time to read shuffle data
+ * @param remoteFetchWaitTime total time that is spent blocked waiting for shuffle to fetch remote data
+ * @param remoteFetchTime the total amount of time for all the shuffle fetches.  This adds up time from overlapping
+ *                        shuffles, so can be longer than task time
+ * @param remoteBytesRead total number of remote bytes read from a shuffle
+ * @param shuffleBytesWritten number of bytes written for a shuffle
+ * @param executorDeserializeTime time taken on the executor to deserialize this task
+ * @param executorRunTime time the executor spends actually running the task (including fetching shuffle data)
+ * @param resultSize the number of bytes this task transmitted back to the driver as the TaskResult
```
squito: Any suggestions on where to put these docs?  I'd really like them to be attached to the fields, so eg. in Intellij they show up when you hover on `metrics.totalBlocksFetched` -- however, if I put them on the fields, Intellij ignores them.  At least this way, they show up as args in the constructor.

squito:
ahh, whoops, actually hold off.  forgot that I hadn't got to ShuffleRDD yet.  I think I'll do a little refactoring for that ...

squito:
Ok I added support for ShuffledRDD also, so I think its ready now.

mateiz:
Thanks Imran! I'll probably wait until after the 0.7 release to merge this because it's a bigger change; we can include it in 0.7.1.

squito:
cool, I understand.  Would appreciate getting it merged somewhere soon, if possible ... its a bit of headache to keep it up to date, as there are constantly conflicts in BlockManager & ShuffleFetcher.  As you're OK with this approach in general, I might keep going w/ adding more metrics

core/src/main/scala/spark/executor/TaskMetrics.scala
mateiz: It would probably work if you did it like this:

```
class TaskMetrics {
  /** Total blocks fetched */
  var totalBlocksFetched: Option[Int] = None

  /** Reote blocks fetched */
  var remoteBlocksFetched: Option[Int] = None

  ...
}
```

core/src/main/scala/spark/executor/TaskMetrics.scala
```diff
@@ -0,0 +1,34 @@
...
+ */
+case class TaskMetrics(
```
mateiz: Any reason why this is a case class instead of a class? I don't think people will want to pattern-match on it.

core/src/main/scala/spark/storage/BlockManager.scala
```diff
@@ -986,3 +842,198 @@ object BlockManager extends Logging {
     }
   }
 }
+
+
+private[spark] trait BlockFetchTracker {
```
mateiz: Place this in a different source file

core/src/main/scala/spark/storage/BlockManager.scala
```diff
@@ -986,3 +842,198 @@ object BlockManager extends Logging {
...
+  def totalBlocks : Int
+  def numLocalBlocks: Int
+  def numRemoteBlocks: Int
+  def remoteFetchTime : Long
+  def remoteFetchWaitTime: Long
+  def remoteBytesRead : Long
+}
+
+private[spark] trait DelegateBlockFetchTracker extends BlockFetchTracker {
```
mateiz: Same with this (different source file)

core/src/main/scala/spark/util/CleanupIterator.scala
```diff
@@ -0,0 +1,25 @@
...
+  def next = sub.next
+  def hasNext = {
+    val r = sub.hasNext
+    if (!r) {
+      cleanup
+    }
+    r
+  }
+
+  def cleanup
```
mateiz: This should have parens `()` since it's an action, not a field. Also, is it guaranteed to be called if there's an exception? Or I guess we only get these metrics for succeeded tasks?

core/src/main/scala/spark/util/Distribution.scala
```diff
@@ -0,0 +1,65 @@
+package spark.util
+
+import java.io.PrintStream
+
+/**
+ * util for getting some stats from a small sample of numeric values, with some handy summary functions
```
mateiz: Start sentences in comments with capital letters to match the rest of the code.

mateiz:
Hey, so I took another look at it and made a few more comments. One other thing: why are the fields of TaskMetrics options instead of just being initialized to zero? It seems to complicate consumer code.

squito:
All of the metrics aren't always applicable, eg. if there is no shuffle, those metrics aren't relevant.  I guess you could say that if there is no shuffle, then shuffleBytesWritten = 0.  But I don't actually think that makes consumer code any cleaner -- before displaying a metric, you'd want to check if its zero, as opposed to checking if its None.  That seemed more scala-like to me.

An alternative might be to wrap up those metrics in another class, like `ShuffleReadMetrics` and `ShuffleWriteMetrics`, which just have longs, and then TaskMetrics would have added fields `var shuffleReadMetrics: Option[ShuffleReadMetrics]` & `var shuffleWriteMetrics: Option[ShuffleWriteMetrics]`.  Or perhaps have subclasses of TaskMetrics w/ the added fields.

Happy to go whichever direction you prefer

mateiz:
I see, I guess that makes sense. I prefer doing that then -- it will be clearer which combinations can be present together.

core/src/main/scala/spark/util/CleanupIterator.scala
squito: yeah, we only get metrics for tasks that succeed, so its OK if `cleanup` doesn't get called.  I think I'll rename the class & method, though -- `cleanup` does sort of imply a behavior like `finally`

squito:
thanks for feedback, I updated it again based on your comments.

mateiz:
Cool, thanks! I've merged it -- hopefully it will make your life easier.

A couple of minor things on style for the future, by the way:
- If a method or variable's doc comment fits in one line, you can also write `/** Whatever */`
- Try to start sentences with capitals in our comments (we've done that in the rest of the code)

I'll fix these manually if I still see some instances. Anyway, thanks for adding this! Looking forward to seeing some sort of UI for it too.

MLnick:
@squito quick question, how does one actually use this? I've done `sbt/sbt publish-local` after pulling in the latest master changes, and run a simple Spark job but I don't see the summary information for stages being printed out. Is there another step to activate this?

squito:
You need to register a `SparkListener` to process the metrics after each stage is completed.  `StatsReportListener` is provided as a way to log a few of the measures.  Eg.

```
val l = new spark.scheduler.StatsReportListener
sc.addSparkListener(l)
```

MLnick:
Ah got it, thanks!

MLnick:
@squito running this off latest master gives a NPE

```
val sc = new SparkContext("local[2]", "test")
val l = new StatsReportListener()
sc.addSparkListener(l)

val result = sc.parallelize(1 to 10).map(x => x^2).collect()
println(result.mkString(","))
```

```
950  [DAGScheduler] DEBUG spark.scheduler.DAGScheduler  - Got event of type spark.scheduler.CompletionEvent
951  [DAGScheduler] INFO  spark.scheduler.DAGScheduler  - Completed ResultTask(0, 1)
952  [DAGScheduler] DEBUG spark.scheduler.DAGScheduler  - Got event of type spark.scheduler.CompletionEvent
952  [DAGScheduler] INFO  spark.scheduler.DAGScheduler  - Completed ResultTask(0, 0)
952  [DAGScheduler] INFO  spark.scheduler.DAGScheduler  - Stage 0 (map at StatsTest.scala:21) finished in 0.074 s
954  [DAGScheduler] INFO  spark.scheduler.StatsReportListener  - Finished stage: MappedRDD[1] at map at StatsTest.scala:21
962  [DAGScheduler] INFO  spark.scheduler.StatsReportListener  - task runtime:(count: 2, mean: 70.000000, stdev: 0.000000)
964  [DAGScheduler] INFO  spark.scheduler.StatsReportListener  -    0%  5%  10% 25% 50% 75% 90% 95% 100%
965  [DAGScheduler] INFO  spark.scheduler.StatsReportListener  -    70.0 ms 70.0 ms 70.0 ms 70.0 ms 70.0 ms 70.0 ms 70.0 ms 70.0 ms 70.0 ms
Exception in thread "DAGScheduler" java.lang.NullPointerException
    at spark.scheduler.StatsReportListener$$anonfun$onStageCompleted$3.apply(SparkListener.scala:31)
    at spark.scheduler.StatsReportListener$$anonfun$onStageCompleted$3.apply(SparkListener.scala:31)
    at spark.scheduler.StatsReportListener$$anonfun$extractLongDistribution$1.apply(SparkListener.scala:63)
    at spark.scheduler.StatsReportListener$$anonfun$extractLongDistribution$1.apply(SparkListener.scala:63)
    at spark.scheduler.StatsReportListener$$anonfun$extractDoubleDistribution$1.apply(SparkListener.scala:58)
    at spark.scheduler.StatsReportListener$$anonfun$extractDoubleDistribution$1.apply(SparkListener.scala:57)
    at scala.collection.TraversableLike$$anonfun$flatMap$1.apply(TraversableLike.scala:239)
    at scala.collection.TraversableLike$$anonfun$flatMap$1.apply(TraversableLike.scala:239)
    at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:60)
    at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:47)
    at scala.collection.TraversableLike$class.flatMap(TraversableLike.scala:239)
    at scala.collection.mutable.ArrayBuffer.flatMap(ArrayBuffer.scala:47)
    at spark.scheduler.StatsReportListener.onStageCompleted(SparkListener.scala:31)
    at spark.scheduler.DAGScheduler$$anonfun$markStageAsFinished$1$2.apply(DAGScheduler.scala:482)
    at spark.scheduler.DAGScheduler$$anonfun$markStageAsFinished$1$2.apply(DAGScheduler.scala:482)
    at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:60)
    at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:47)
    at spark.scheduler.DAGScheduler.markStageAsFinished$1(DAGScheduler.scala:482)
    at spark.scheduler.DAGScheduler.handleTaskCompletion(DAGScheduler.scala:504)
    at spark.scheduler.DAGScheduler.processEvent(DAGScheduler.scala:300)
    at spark.scheduler.DAGScheduler.spark$scheduler$DAGScheduler$$run(DAGScheduler.scala:364)
    at spark.scheduler.DAGScheduler$$anon$1.run(DAGScheduler.scala:107)
```

squito:
shoot, this is totally broken in local mode, sorry.  I meant to come back to it and completely forgot.

I'll submit another pull request w/ a fix but it will be a few days ... until then its broken in local mode.  Other than task runtime, I'm not sure how much more we can get in local mode.

squito:
@MLnick I think this should be fixed by https://github.com/mesos/spark/pull/516

rxin:
FYI

Some recent experiments we have been conducting saw significant performance degradation in workloads with large-shuffles between the master branch and the original 0.7 release. Initial investigation points to TimedIterator as the culprit.

We will investigate further and report findings here - but my suspicion is that per-record timing is too costly to do.

squito:
Hmm, yeah I can see how that would be the case.  I kind of just threw a bunch of different metrics up there, and I don't think that is even very useful in the end.  I did do a really quick check on performance myself, but (a) that was before all of the other recent improvements and (b) I may have only done one big shuffle, where it might not have mattered.

I've found the most useful metric to be fetchWaitTime, which is tracked in `BlockFetchIterator.next()`.  That timing is per-block, so I think the cost shouldn't be too high, but that is worth double checking under your loads.

Incidentally, I think I _meant_ to put the `TimedIterator` in `BlockStoreShuffleFetcher` around fetching each _block_, not around each _record_,  but now I realized I didn't do that.  Maybe the performance hit wouldn't be so bad if we changed it to time per-block.  But then again, it doesn't add much over the other metrics, so feel free to just rip it out.

pwendell:
I just committed a pull request which makes it possible to disable this
tracking.

I would be fine to just disable it - Imran which situation did you have in
mind where this would be useful to track? It seems like it's tracking the
time to call next() on an iterator that just reads a block which has
already been fetched I'd assume that time will be pretty much constant
under any workload.

pwendell:
Correction: "I would be fine to entirely remove it"

squito:
yeah, there is probably no need for it at all.  Really I just stuck profiling around lots of different things, I think I might have put that it in before putting better tracking down in `BlockFetchIterator`.

I vote to just get rid of it.

PR #465, opened 2013-02-12
SPARK-696: sortByKey should use 'ascending' parameter

pwendell:


mateiz:
Woah, kind of important!

PR #466, opened 2013-02-12
STREAMING-50: Support transform workaround in JavaPairDStream

pwendell:
This ports a useful workaround (the `transform` function) to
JavaPairDStream. It is necessary to do things like sorting which
are not supported yet in the core streaming API.


tdas:
Looks good! Thanks!

PR #467, opened 2013-02-13
include jobid in Executor commandline args

squito:
For whatever reasons, we've observed that sometimes the StandaloneExecutorBackend process doesn't die when a spark context finishes.  Obviously we'd like to eliminate that from happening in the first place, but in the meantime we just need an automated way to kill these rogue processes.

If the command line args include the jobid for each StandaloneExecutorBackend, then we can use `jps` to find the jobid of all the executors, and filter out those that the master doesn't know about.

### DIFF
--- a/core/src/main/scala/spark/executor/StandaloneExecutorBackend.scala
+++ b/core/src/main/scala/spark/executor/StandaloneExecutorBackend.scala
@@ -68,8 +68,9 @@ private[spark] object StandaloneExecutorBackend {
   }
 
   def main(args: Array[String]) {
-    if (args.length != 4) {
-      System.err.println("Usage: StandaloneExecutorBackend <driverUrl> <executorId> <hostname> <cores>")
+    if (!(args.length >= 4)) {
+      //the reason we allow the last frameworkId argument is to make it easy to kill rogue executors
+      System.err.println("Usage: StandaloneExecutorBackend <driverUrl> <executorId> <hostname> <cores> [<frameworkid>]")
       System.exit(1)
     }
     run(args(0), args(1), args(2), args(3).toInt)
--- a/core/src/main/scala/spark/executor/StandaloneExecutorBackend.scala
+++ b/core/src/main/scala/spark/executor/StandaloneExecutorBackend.scala
@@ -68,9 +68,9 @@ private[spark] object StandaloneExecutorBackend {
   }
 
   def main(args: Array[String]) {
-    if (!(args.length >= 4)) {
+    if (args.length < 4) {
       //the reason we allow the last frameworkId argument is to make it easy to kill rogue executors
-      System.err.println("Usage: StandaloneExecutorBackend <driverUrl> <executorId> <hostname> <cores> [<frameworkid>]")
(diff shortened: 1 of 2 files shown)


stephenh:
Out of curiosity, have you seen this on master as of last week-ish? Specifically https://github.com/mesos/spark/pull/451?

I was seeing this behavior locally, but that fixed it. (I know you had commented on that PR, just double checking that you tried it to see if it cleared up this issue.)

squito:
I haven't tried that out yet, and I'm definitely hopeful that it fixes things ... but I'd still like to have this as a backup, in case there are problems in the future.

core/src/main/scala/spark/executor/StandaloneExecutorBackend.scala
```diff
@@ -68,8 +68,9 @@ private[spark] object StandaloneExecutorBackend {
   }
 
   def main(args: Array[String]) {
-    if (args.length != 4) {
-      System.err.println("Usage: StandaloneExecutorBackend <driverUrl> <executorId> <hostname> <cores>")
+    if (!(args.length >= 4)) {
+      //the reason we allow the last frameworkId argument is to make it easy to kill rogue executors
+      System.err.println("Usage: StandaloneExecutorBackend <driverUrl> <executorId> <hostname> <cores> [<frameworkid>]")
```
stephenh: Why frameworkid instead of jobId?

squito: I thought there was some discussion of disambiguating the uses of "job", by calling a connection to Standalone cluster a "framework", and a group of computations on rdds a "job" ... but maybe I'm remembering wrong / this is pre-emptive

mateiz: Yeah this is a good question. I would actually call it App ID and eventually replace other stuff to call it that as well.

mateiz: BTW one minor nitpick, do args.length < 4 instead of !(args.length > 4)

stephenh: Interesting...so what's now being called AppID? Sorry, I'm not following...

mateiz: Well, so currently we have this:
- In spark.deploy: an application (SparkContext instance) is called a job
- In SparkContext itself: one parallel query (runJob call) is called a job
- In Mesos: an application (SparkContext instance) is called a framework

I want to change the spark.deploy "job" to an "app", which better reflects its nature, and leave "job" as "a set of parallel tasks computing one goal" as it is in SparkContext / DAGScheduler. And for Mesos, maybe we will rename frameworks to apps at some point too.

stephenh: Cool, thanks for the explanation.

squito:
updated.  (thanks for catching the stupid condition I had)

mateiz:
Great, thanks for adding this.

PR #470, opened 2013-02-16
Make CoGroupedRDDs explicitly have the same key type.

stephenh:
Pretty minor, especially given most parameters to CoGroupedRDD are casts anyway.


mateiz:
Looks good, thanks.

PR #471, opened 2013-02-16
Move ParallelCollection into spark.rdd package.

stephenh:
I had missed ParallelCollection when I was cleaning up the RDDs because it wasn't in the spark.rdd package. This change moves it there, and renames it to ParallelCollectionRDD, as that's how the other RDDs are named.


mateiz:
Looks good, thanks.

PR #472, opened 2013-02-16
Add clearDependencies for FlatMappedValuesRDD and MappedValuesRDD.

stephenh:
Some more RDD cleanup; I am not an expert at how/when/what clearDependencies is for, but I can cargo cult with the best of them.


mateiz:
So I don't think this is really needed, because the "prev" kept to the constructor should not be kept as a field if we only pass it to the superclass constructor. A lot of other RDD subclasses (e.g. MappedRDD) work the same way.

stephenh:
Ah, okay, I was actually wondering if Scala would be that smart, but didn't know for sure--good to know that it is.

PR #473, opened 2013-02-16
Add RDD.subtract.

stephenh:
Instead of reusing the cogroup primitive, this adds a SubtractedRDD that knows it only needs to keep rdd1's values (per split) in memory.


stephenh:
The tests for this pass, but this was trickier than I expected because it has to deal with both non-tupled and already-tupled RDDs. I believe it is working for all cases, narrow/shuffle/etc., as shown by the tests.

But the "part2" partitioner thing is odd enough that I would appreciate extra scrutiny.

stephenh:
Okay, I ran this on our job and saw a nice speed up.

I am subtracting an RDD of ~600 partitions (last N days of data) from an RDD of ~60 partitions (today's data), and previously because cogroup would first build a combined map of all keys, and then keep only today's uniques, I had cogroup use 600 partitions. On the current data set, it took ~160 seconds.

With this patch, using subtract, I was able to use 60 partitions, and on the same data set it completed in ~20 seconds.

Granted, both of these numbers are small, and won't make a different to our report, but I think the optimization in general is still worthwhile.

mateiz:
Cool, thanks! Mind adding this to the Java API too though?

stephenh:
Sure; I'll get to it here pretty soon.

stephenh:
Okay, I merged in master, and added subtract to JavaRDD, JavaDoubleRDD, and JavaPairRDD.

mateiz:
Alright, thanks. The key-value pair stuff is indeed tricky -- I just spent 15 minutes looking at it again. I wonder if we should eventually just use ClassManifest to decide whether an RDD contains key-value pairs. It will be trickier to call from Java, so it might be best not to; but on the other hand, a `match` statement in the inner loop doesn't look that great either (it's probably not _that_ expensive, but it is basically an `instanceof`).

PR #475, opened 2013-02-18
Remove hack workaround for SPARK-668

JoshRosen:
While fixing [SPARK-694](https://spark-project.atlassian.net/browse/SPARK-694), Patrick discovered that [SPARK-668](https://spark-project.atlassian.net/browse/SPARK-668) can be fixed by simply renaming type parameters.

I tried renaming the type parameters before I submitted my original fix in #417; it failed then, but it turns out that you have to run `sbt/sbt clean` in order for the parameter renaming to have its full effect.

This commit undoes my hack workaround and uses renamed type parameters instead.  I might want to add some additional tests later, but I'm opening this PR now so that this change makes it into 0.7.


mateiz:
Cool, thanks!

PR #477, opened 2013-02-19
Ganglia port change

shivaram:
We might need to back port this to 0.6 if we want to support launches using new-scripts from 0.6.


mateiz:
So regarding the printing at the end, I actually changed setup.sh earlier to print the master's web UI URL. Can you do that in spark_ec2.py too? Then we can remove it from setup.sh. It should be done for Mesos as well as standalone mode (in both cases, the web UI port is 8080).

shivaram:
Done

mateiz:
Great, thanks.

PR #479, opened 2013-02-19
Zeromq streaming

ScrapCodes:


project/SparkBuild.scala
```diff
@@ -134,6 +134,7 @@ object SparkBuild extends Build {
       "com.typesafe.akka" % "akka-actor" % "2.0.3",
       "com.typesafe.akka" % "akka-remote" % "2.0.3",
       "com.typesafe.akka" % "akka-slf4j" % "2.0.3",
+      "com.typesafe.akka" % "akka-zeromq" % "2.0.3",
```
tdas: Can you add akka-zeromq to the streaming project settings and not the shared one? We dont want the Spark core JAR to get bloated with too many libraries. Since only Spark Streaming project needs the zeroMQ, it should get included on the streaming project settings.

docs/zeroMQ-intro.md
```diff
@@ -0,0 +1,59 @@
+---
+layout: global
+title: ZeroMQ Stream setup guide
+---
+
+## Install ZeroMQ (using JNA)
```
tdas: Instead of having a separate markdown doc which people may not know about, can add the gist of this (basically installation instructions) in the comments of ZeroMQWordCount example? In that way they will not have to hunt around for more instruction, the zeroMQ word count will be self explanatory.

pwendell: But what if the code changes - do we keep adding new gist's? I sort of
prefer the markdown option since it will evolve correctly with the code and
be versioned like all the rest of our documentation. What about including a
reference to the md file in the docstring?

tdas: Umm, sorry if it was confusing. I didnt mean Github Gist. I meant "gist" as in "key idea" of this md doc, that should be in the comments of the ZeroMQWordCount example. :)

pwendell: Haha that makes much more sense.

ScrapCodes: @tdas Should we mention, the JNA and JNI difference and all in the method comments? Or leave them for user to self explore those things. Then method comments would just point to links to zeromq installation guide.

tdas: I think the line "To work with akka-zeromq, .... " in the md has all the links and high level information about JNA and JNI. Just putting that line should be fine. Also should have a line about installing the zeromq core libraries that is necessary for running this example.

streaming/src/main/scala/spark/streaming/StreamingContext.scala
```diff
@@ -175,6 +177,26 @@ class StreamingContext private (
   }
 
   /**
+   * ZeroMQ stream receiver
```
tdas: Please complete the sentence to make it consistent with other input streams. See the other input streams.

tdas:
Thank you very much! 
However, when I am trying to compile this, I get unresolved dependencies for zero-mq2.0.3. Taking a look at that.

haitaoyao:
Same as @tdas , I got the same error. 
Can somebody fix this?
Thanks.

ScrapCodes:
@haitaoyao and @tdas I am sorry for overlooking your comment for a while. Actually this is because in sbt file dependencies are not shared between core and streaming. And since akka repo is defined in core it has to be again defined for streaming. 

A temporary fix would be to move the Akka zeromq dependency back to core. 

I am trying to properly fix it, will send a pull request in a few hours perhaps.

haitaoyao:
Thanks very much. 

Haitao Yao
yao.erix@gmail.com
weibo: @haitao_yao
Skype:  haitao.yao.final

在 2013-2-25，下午3:58，Prashant Sharma notifications@github.com 写道：

tdas:
This is a mistake I made the at the last moment. I had fixed it by adding the Akka repo to the streaming project, but moved it to core at the last moment. The correct thing to do is to add it to the resolvers of sharedSettings. We should not add zero-mq to the core project as the core project does not need it and should not get bloated with unnecessary things.

Also, may be the pom.xml for streaming needs to be fixed as well.

ScrapCodes:
https://github.com/mesos/spark/pull/496 here is the fix. I think i guessed your intentions this time. It does not have fix for pom though.

haitaoyao:
@ScrapCodes I can not even find the akka-zeromq#2.0.3 in the main maven repo and in the akka offical download site. Can you provide the jar and pom in some place that I can donwload? 
I think no matter where you put the dependency, once others must need it if want to build  from the source.
Thanks.

tdas:
That was fast! Thanks Prashant. Though I have asked you in the pull request to issue a different one directly to master. If you get a chance, can you also compile with maven ( mvn -Phadoop1 -DSkipTests package ) after clearing the maven caches to see whether it compiles. It may be that pom.xml does not need to be fixed.

tdas:
@haitaoyao http://repo.akka.io/releases/com/typesafe/akka/akka-zeromq/2.0.3/

ScrapCodes:
@tdas Looking at pom.xml i have similar feeling. It does not need to be fixed. Unfortunately I am on leave today trying to shift my house sitting on the laptop where maven is not properly working. If you can wait for sometime, then I think i can look at that. :)

tdas:
Thats okay, I am testing out the maven one myself right now.

Happy shifting! ;)

tdas:
I checked. pom is fine.

tdas:
@haitaoyao If you pull from the master again, your problem should be resolved.

haitaoyao:
@tdas I see. Thanks. It works very well.

PR #480, opened 2013-02-19
[Streaming] Examples using Twitter's Algebird library

MLnick:
This PR adds two examples for streaming that use monoids from Twitter's Algebird library:
- HyperLogLog for approximate distinct object counting with low memory overhead
- CountMinSketch for approximating object frequency in a stream as well as TopK or "heavy hitters" estimation

See https://groups.google.com/forum/?fromgroups=#!topic/spark-users/4ht9ndVaZQY


examples/src/main/scala/spark/streaming/examples/TwitterAlgebirdCMS.scala
```diff
@@ -0,0 +1,78 @@
...
+    val TOPK = 10
+
+    val Array(master, username, password) = args.slice(0, 3)
+    val filters = args.slice(3, args.length)
+
+    val ssc = new StreamingContext(master, "TwitterAlgebirdCMS", Seconds(10))
+    val stream = ssc.twitterStream(username, password, filters,
+      StorageLevel.MEMORY_ONLY_SER)
+
+    val users = stream.map(status => status.getUser.getId)
```
MLnick, line 49: A note about this: currently Algebird CMS only supports Long inputs. Since it uses hashing under the hood it should be possible to have any hashable input as with HyperLogLog, but not currently.

So for now this example works on user ids, so running it over relatively small durations will not result in very heavily-skewed data (which is where the sketch will be most useful). If we could take String inputs then it would be more interesting as we could do TopK on hashtags (for example) which is likely to be a lot more skewed.

streaming/pom.xml
```diff
@@ -47,6 +47,16 @@
       <artifactId>zkclient</artifactId>
       <version>0.1</version>
     </dependency>
+    <dependency>
+      <groupId>org.twitter4j</groupId>
+      <artifactId>twitter4j-stream</artifactId>
```
tdas, line 52: Aah, twitter4j was missing in the streaming pom.xml! Thanks for fixing this!

pyspark
```diff
@@ -36,4 +36,9 @@ if [[ "$SPARK_LAUNCH_WITH_SCALA" != "0" ]] ; then
     export SPARK_LAUNCH_WITH_SCALA=1
 fi
 
-exec "$PYSPARK_PYTHON" "$@"
+if [[ "$IPYTHON" = "1" ]] ; then
+    export PYSPARK_PYTHON="ipython"
```
tdas, line 40: Why is this change necessary for the example?

examples/src/main/scala/spark/streaming/examples/TwitterAlgebirdCMS.scala
```diff
@@ -0,0 +1,78 @@
...
+
+import spark.streaming.{Seconds, StreamingContext}
+import spark.storage.StorageLevel
+import com.twitter.algebird._
+import spark.streaming.StreamingContext._
+import spark.SparkContext._
+
+/**
+ * Example of using CountMinSketch monoid from Twitter's Algebird together with Spark Streaming's
+ * TwitterInputDStream
```
tdas: Can you please add a bit more verbose comment about what this program does? That is, what is CountMinSketch, what it does, and what output to expect. This would help those who are new to Algebird library.

examples/src/main/scala/spark/streaming/examples/TwitterAlgebirdCMS.scala
tdas, line 49: This maybe an important point that may confused people. Can you added a line to the comment at the top?

examples/src/main/scala/spark/streaming/examples/TwitterAlgebirdHLL.scala
```diff
@@ -0,0 +1,62 @@
...
+    val ssc = new StreamingContext(master, "TwitterAlgebirdHLL", Seconds(5))
+    val stream = ssc.twitterStream(username, password, filters, StorageLevel.MEMORY_ONLY_SER)
+
+    val users = stream.map(status => status.getUser.getId)
+
+    var globalHll = new HyperLogLogMonoid(BIT_SIZE).zero
+    var userSet: Set[Long] = Set()
+
+    val approxUsers = users.mapPartitions(ids => {
+      val hll = new HyperLogLogMonoid(BIT_SIZE)
```
tdas: Few words of comments explaining the steps of how partial HLLs are used to update global HLLs would be nice! For those who are not familiar with HLLs.

pyspark
MLnick, line 40: hmm, it is pulling in a commit from another PR: https://github.com/mesos/spark/pull/454

I guess somehow this was not merged master <-> streaming branch.

tdas, line 40: Aah, I can try to fix that by merging the master branch again to streaming.

examples/src/main/scala/spark/streaming/examples/TwitterAlgebirdCMS.scala
```diff
@@ -0,0 +1,78 @@
...
+      StorageLevel.MEMORY_ONLY_SER)
+
+    val users = stream.map(status => status.getUser.getId)
+
+    var globalCMS = new CountMinSketchMonoid(DELTA, EPS, SEED, PERC).zero
+    var globalExact = Map[Long, Int]()
+    val mm = new MapMonoid[Long, Int]()
+
+    val approxTopUsers = users.mapPartitions(ids => {
+      val cms = new CountMinSketchMonoid(DELTA, EPS, SEED, PERC)
```
johnynek: you should be able to move this above mpaPartitions, I think. The cms doesn't have any internal mutable state.

examples/src/main/scala/spark/streaming/examples/TwitterAlgebirdHLL.scala
johnynek: you can reuse these monoids. They are immutable. So, if it is clearer at all, you can move hll above, and use the zero from that on in the globalHll also.

pyspark
tdas, line 40: I merged all the latest stuff from master. But still this PR diff shows this change. This is weird. I checked the pyspark file from your branch https://github.com/MLnick/spark/blob/streaming-eg-algebird/pyspark and the mesos spark branch https://github.com/mesos/spark/blob/streaming/pyspark . Both seem to have the commit. Not sure why this is still showing up. Can you update your branch with all the changes of mesos/spark/streaming branch, just to be sure.

johnynek:
Glad to see this pull req. Hope this helps CMS and HLL make more impacts.

I agree that the CMS interface is suboptimal now. We are going to update it to support the same approach as HLL (probably in algebird 0.2.0). Let us know if there are any algorithms to add.  I'd love to collaborate and share this code in Algebird (which we extracted from scalding).

examples/pom.xml
```diff
@@ -20,11 +20,10 @@
       <artifactId>jetty-server</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.twitter4j</groupId>
-      <artifactId>twitter4j-stream</artifactId>
-      <version>3.0.3</version>
+      <groupId>com.twitter</groupId>
+      <artifactId>algebird-core_2.9.2</artifactId>
+      <version>0.1.8</version>
```
sritchie: 0.1.9 is out!

MLnick:
@johnynek thanks for the comments! Look forward to 0.2.0 in that case since CMS with any hashable inputs will be neat. Also if I find some time I'd be happy to try a scalding version of the example.

tdas:
Thank you very much. This is a great addition.

PR #481, opened 2013-02-19
STREAMING-51: Add RDD type as a type parameter in JavaDStreamLike Edit (streaming/ version) 

pwendell:
This patch adds the JavaRDD type as a type parameter to JavaDStreamLike. That change simplifies some repeated code between JavaDStream and JavaPairDStream. The patch adds tests to verify new functionality. There is some more detail in individual commits.


tdas:
Looks good, I am accepting this and closing the other pull request. Thanks Patrick!

PR #482, opened 2013-02-19
Don't call System.exit over uncaught exceptions from shutdown hooks

charlesreiss:
Currently, spark.executor.Executor's uncaught exception handler always tries to call System.exit(). Unfortunately, calling System.exit() from a shutdown hook deadlocks. This patch modifies the uncaught exception handler to avoid calling System.exit() when a shutdown is in progress.

This also modifies the delete local directories shutdown hook to catch and log exceptions, which is how I noticed the deadlock.

### DIFF
--- a/core/src/main/scala/spark/executor/Executor.scala
+++ b/core/src/main/scala/spark/executor/Executor.scala
@@ -52,20 +52,8 @@ private[spark] class Executor extends Logging {
             logError("Uncaught exception in thread " + thread, exception)
             
             // We may have been called from a shutdown hook. If so, we must not call System.exit().
-            // (If we do, we will deadlock.) Runtime#addShutdownHook should fail if we are shutting
-            // down, which would either occur if we were called from a shutdown hook or if
-            // a System.exit() occured concurrently.
-            var shuttingDown = false
-            try {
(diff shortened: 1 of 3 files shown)


core/src/main/scala/spark/executor/Executor.scala
```diff
@@ -50,14 +50,31 @@ private[spark] class Executor extends Logging {
...
+            var shuttingDown = false
+            try {
+              val hook = new Thread {
+                override def run() {}
+              }
+              Runtime.getRuntime.addShutdownHook(hook)
+              Runtime.getRuntime.removeShutdownHook(hook)
+            } catch {
+              case ise: IllegalStateException => shuttingDown = true
+            }
```
mateiz: Is this really the only way to detect whether you're shutting down? If so, factor it into a different method since it's a little hacky.

charlesreiss: Pulled out into utility function.

I didn't see any other way to do it. If we only cared about shutdown hooks we registered, we could keep a list of the threads and check it, but this wouldn't work for anything registered by external libraries.

mateiz:
Alright, thanks.

PR #488, opened 2013-02-20
Fix SPARK-698. From ExecutorRunner, launch java directly vs via the run scripts

cgrothaus:
Fixes https://spark-project.atlassian.net/browse/SPARK-698

Currently, doesn't handle "%FWDIR%conf\java-opts.cmd", as I found no docs about that.

### DIFF
--- a/core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
+++ b/core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
@@ -75,9 +75,45 @@ private[spark] class ExecutorRunner(
 
   def buildCommandSeq(): Seq[String] = {
     val command = appDesc.command
-    val script = if (System.getProperty("os.name").startsWith("Windows")) "run.cmd" else "run"
-    val runScript = new File(sparkHome, script).getCanonicalPath
-    Seq(runScript, command.mainClass) ++ (command.arguments ++ Seq(appId)).map(substituteVariables)
+    val runner = if (getEnvOrEmpty("JAVA_HOME") == "") {
+      "java"
+    } else {
+      getEnvOrEmpty("JAVA_HOME") + "/bin/java"
+    }
+    // SPARK-698: do not call the run.cmd script, as process.destroy()
+    // fails to kill a process tree on Windows
+    Seq(runner) ++ buildJavaOpts() ++ Seq(command.mainClass) ++
+      command.arguments.map(substituteVariables)
+  }
+  
+  /*
+   * Attention: this must always be aligned with the environment variables in the run scripts and the
+   * way the JAVA_OPTS are assembled there.
+   */
+  def buildJavaOpts(): Seq[String] = {
+    val _javaLibPath = if (getEnvOrEmpty("SPARK_LIBRARY_PATH") == "") {
+      ""
+    } else {
+      "-Djava.library.path=" + getEnvOrEmpty("SPARK_LIBRARY_PATH")
+    }
+    
+    Seq("-cp", 
+        getEnvOrEmpty("CLASSPATH"), 
+        // SPARK_JAVA_OPTS is overwritten with SPARK_DAEMON_JAVA_OPTS for running the worker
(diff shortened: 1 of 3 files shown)


core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
```diff
@@ -75,9 +75,45 @@ private[spark] class ExecutorRunner(
...
+    }
+    
+    Seq("-cp", 
+        getEnvOrEmpty("CLASSPATH"), 
+        // SPARK_JAVA_OPTS is overwritten with SPARK_DAEMON_JAVA_OPTS for running the worker
+        getEnvOrEmpty("SPARK_NONDAEMON_JAVA_OPTS"), 
+        _javaLibPath,
+        "-Xms" + memory.toString + "M",
+        "-Xmx" + memory.toString + "M")
+    .filter(_ != "")
```
mateiz: Why not just use `System.getenv` instead of `getEnvOrEmpty`, and then `filter(_ != null)`?

core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
```diff
@@ -75,9 +75,45 @@ private[spark] class ExecutorRunner(
...
   def buildCommandSeq(): Seq[String] = {
     val command = appDesc.command
-    val script = if (System.getProperty("os.name").startsWith("Windows")) "run.cmd" else "run"
-    val runScript = new File(sparkHome, script).getCanonicalPath
-    Seq(runScript, command.mainClass) ++ (command.arguments ++ Seq(appId)).map(substituteVariables)
+    val runner = if (getEnvOrEmpty("JAVA_HOME") == "") {
+      "java"
+    } else {
+      getEnvOrEmpty("JAVA_HOME") + "/bin/java"
+    }
```
mateiz, line 82: Here also you could just use `System.getenv` instead of `getEnvOrEmpty`

run
```diff
@@ -22,6 +22,8 @@ fi
 # values for that; it doesn't need a lot
 if [ "$1" = "spark.deploy.master.Master" -o "$1" = "spark.deploy.worker.Worker" ]; then
   SPARK_MEM=${SPARK_DAEMON_MEMORY:-512m}
+  # Backup current SPARK_JAVA_OPTS for use in ExecutorRunner.scala
+  SPARK_NONDAEMON_JAVA_OPTS=$SPARK_JAVA_OPTS
   SPARK_JAVA_OPTS=$SPARK_DAEMON_JAVA_OPTS   # Empty by default
```
mateiz: Instead of saving SPARK_NONDAEMON_JAVA_OPTS, I think it would be better to modify the scripts to use a different variable name for the JAVA_OPTS they do eventually use. For example:

```
OUR_JAVA_OPTS="$SPARK_JAVA_OPTS"
if [ "$1" = "spark.deploy.master.Master ]; then
    OUR_JAVA_OPTS="$SPARK_DAEMON_JAVA_OPTS"
fi
```

This will make the scripts clearer to understand.

mateiz:
Thanks a lot for the fix! I've made a few comments on the code but it looks good overall. Just update the pull request by pushing a new commit to your branch when you fix them.

cgrothaus:
Hello Matei, thanks for your feedback! I just updated the pull request.

mateiz:
Hey, sorry it's taken a while to merge this -- I wanted to test it to make sure the environment variables get passed the right way. I'm going to look at both it and #540 soon.

mateiz:
Thanks for submitting this. I've now merged it into the master branch. I did one change, which is to read the classpath by calling an external script, because we can't be guaranteed that the ExecutorRunner thread will be launched by something that had the CLASSPATH environment variable set (specifically, in unit tests, when we use the "local-cluster" mode, that won't be true). I also made it split SPARK_JAVA_OPTS into separate arguments. See commits 15b00914c53f1f4f00a3313968f68a8f032e7cb7 and https://github.com/mesos/spark/commit/6c8d1b2ca618a1a17566ede46821c0807a1b11f5.

PR #489, opened 2013-02-22
SPARK-702: Replace Function --> JFunction in JavaAPI Suite.

pwendell:
In a few places the Scala (rather than Java) function class is used.


mateiz:
Looks good, thanks.

PR #490, opened 2013-02-23
Detect when SendingConnections disconnect even if we aren't sending to them

charlesreiss:
This patch modifies SendingConnection to always select for reading; when the conneciton is readable, it reads a byte to detect EOF or an I/O Error and trigger connection shutdown, allowing sent but unack'd messages to fail instead of hanging.

This also adds a test to DistributedSuite which can trigger the issue by doing memory-to-memory replications operations around failures.

This addresses issue SPARK-704.

### DIFF
--- a/core/src/main/scala/spark/network/Connection.scala
+++ b/core/src/main/scala/spark/network/Connection.scala
@@ -198,7 +198,7 @@ extends Connection(SocketChannel.open, selector_, remoteId_) {
     outbox.synchronized {
       outbox.addMessage(message)
       if (channel.isConnected) {
-        changeConnectionKeyInterest(SelectionKey.OP_WRITE)
+        changeConnectionKeyInterest(SelectionKey.OP_WRITE | SelectionKey.OP_READ)
       }
     }
   }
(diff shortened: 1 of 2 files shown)


mridulm:
If the socket is closed on remote side, there is no gaurantee that a
read event will be triggered.
- Mridul

charlesreiss:
I think we will eventually get a read event, because we set all connections keep alive and read events are documented to occur for end-of-stream or error conditions. But TCP keep-alive is way too slow (like hours!) by default to be useful to us, so, for most users, this patch only fixes the case where the remote OS stays alive enough to signal EOF/connection reset or where the remote machine disappears while it has data pending. (I think these are important enough cases to handle.)

Some options for a more complete fix:
- timeout ConnectionManager operations (tricky since they can involve sending lots of data);
- periodically send probes on idle SendingConnections;
- get notified from the BlockManagerMaster when a node heartbeat fails;

mateiz:
Thanks for the pull request. I think we'll have to add heartbeats on the connections for a complete solution, but this is a good start.

PR #491, opened 2013-02-23
Adding Java-accessible methods to Vector.scala

pwendell:
This is needed for the Strata machine learning tutorial (and also is generally helpful).


core/src/main/scala/spark/util/Vector.scala
```diff
@@ -10,12 +10,14 @@ class Vector(val elements: Array[Double]) extends Serializable {
       throw new IllegalArgumentException("Vectors of different length")
     return Vector(length, i => this(i) + other(i))
   }
+  def add(other: Vector) = +(other)
```
mateiz: Looks good but please add blank lines before these; for clarity I would also change these to `this + other`, `this += other`, etc.

pwendell:
Okay fixed.

mateiz:
Great, thanks!

PR #492, opened 2013-02-24
Change numSplits to numPartitions in PySpark

JoshRosen:
This renames `numSplits` to `numPartitions` in PySpark for consistency with the Scala / Java APIs.


mateiz:
Good catch.

PR #494, opened 2013-02-24
Final set of Spark Streaming changes

tdas:
Here the major changes at a high-level
- Major changes to the streaming scheduler and checkpointing modules to handle driver-failure recovery correctly with file input stream.
- Added information on checkpointing, fault-tolerance and Java API to the streaming programming guide 
- Added Twitter input stream to StreamingContext
- Changed names of few functions (countByKeyAndWindow => CountByValueAndWindow, _network_Stream => _socket_Stream)
- Fixed missing functionalities in the Java API to make them consistent with the Scala API
- Removed unnecessary stuff from Kafka input stream as well as other fixes (contributed by Sean McNamara)
- Added actor and zeroMQ input streams along with support for arbitrary receivers (contributed by Prashant Sharma)
- Added Twitter + Algebird examples (contributed by Nick Pentreath)


project/SparkBuild.scala
```diff
@@ -155,18 +155,21 @@ object SparkBuild extends Build {
...
+    libraryDependencies ++= Seq("com.twitter" % "algebird-core_2.9.2" % "0.1.8")
   )
 
   def bagelSettings = sharedSettings ++ Seq(name := "spark-bagel")
 
   def streamingSettings = sharedSettings ++ Seq(
     name := "spark-streaming",
+    resolvers ++= Seq(
+      "Akka Repository" at "http://repo.akka.io/releases"
+    ),
```
mateiz: Why isn't the typesafe repo we had for the projects above enough? If it's not, change that one to be the Akka one

streaming/pom.xml
```diff
@@ -20,6 +20,17 @@
...
+      <id>akka-repo</id>
+      <name>Akka Repository</name>
+      <url>http://repo.akka.io/releases</url>
+      <releases>
+        <enabled>true</enabled>
+      </releases>
+      <snapshots>
+        <enabled>false</enabled>
+      </snapshots>
+    </repository>
```
mateiz: Ditto here on Akka repository vs Typesafe

run
```diff
@@ -116,6 +116,9 @@ CLASSPATH+=":$CORE_DIR/src/main/resources"
 CLASSPATH+=":$REPL_DIR/target/scala-$SCALA_VERSION/classes"
 CLASSPATH+=":$EXAMPLES_DIR/target/scala-$SCALA_VERSION/classes"
 CLASSPATH+=":$STREAMING_DIR/target/scala-$SCALA_VERSION/classes"
+for jar in `find "$STREAMING_DIR/lib" -name '*jar'`; do
+  CLASSPATH+=":$jar"
+done
```
mateiz: This kind of change will also need to be done in `run2.cmd` for Windows. Also, in Java 1.6 and above, you can avoid putting a lot of JARs by just adding a path that ends with `*`, such as `lib/*`; I believe we did that for the other projects that used to have JARs, so try it here too.

project/SparkBuild.scala
tdas: It wasnt fetching akka-zeromq from the Typesage repo, even though akka-actor and akka-remote was working. I can update the main resolvers.

mateiz: Okay, got it. Then just replace the Typesafe repo with the Akka one. We also did it in 0.6.2.

docs/plugin-custom-receiver.md
```diff
@@ -0,0 +1,101 @@
+---
+layout: global
+title: Tutorial - Spark streaming, Plugging in a custom receiver.
+---
+
+A "Spark streaming" receiver can be a simple network stream, streams of messages from a message queue, files etc. A receiver can also assume roles more than just receiving data like filtering, preprocessing, to name a few of the possibilities. The api to plug-in any user defined custom receiver is thus provided to encourage development of receivers which may be well suited to ones specific need.
```
mateiz: As a minor thing, are you going to call it "Spark streaming" or "Spark Streaming"? I think capital S is better, and will be more consistent with the other docs. Also, maybe this doc should be renamed to `custom-streaming-receiver.md` to make it a bit easier to find.

docs/streaming-programming-guide.md
```diff
@@ -292,22 +337,182 @@ For a Spark Streaming application running on a cluster to be stable, the process
...
+
+Since all data is modeled as RDDs with their lineage of deterministic operations, any recomputation always leads to the same result. As a result, all DStream transformations are guaranteed to have _exactly-once_ semantics. That is, the final transformed result will be same even if there were was a worker node failure. However, output operations (like `foreach`) have _at-least once_ semantics, that is, the transformed data may get written to an external entity more than once in the event of a worker failure. While this is acceptable for saving to HDFS using the `saveAs*Files` operations (as the file will simply get over-written by the same data), additional transactions-like mechanisms may be necessary to achieve exactly-once semantics for output operations.
+
+## Failure of a Driver Node
```
mateiz: This should say failure of the driver, not a driver.

docs/streaming-programming-guide.md
```diff
@@ -292,22 +337,182 @@ For a Spark Streaming application running on a cluster to be stable, the process
...
+A system that is required to operate 24/7 needs to be able tolerate the failure of the drive node as well. Spark Streaming does this by saving the state of the DStream computation periodically to a HDFS file, that can be used to restart the streaming computation in the event of a failure of the driver node. To elaborate, the following state is periodically saved to a file.
```
mateiz: drive -> driver

docs/plugin-custom-receiver.md
tdas: Done.

docs/streaming-programming-guide.md
mateiz: Also, is checkpointing on by default or does ti have to be configured? Not super clear from here.

docs/streaming-programming-guide.md
tdas: Done.

docs/streaming-programming-guide.md
tdas: Done.

tdas: Some of the information has been given in the earlier checkpoint section. However, to make the link more explicitly clear, I have updated it to this. 

A system that is required to operate 24/7 needs to be able tolerate the failure of the driver node as well. Spark Streaming does this by saving the state of the DStream computation periodically to     a HDFS file, that can be used to restart the streaming computation in the event of a failure of the driver node. This checkpointing is enabled by setting a HDFS directory for checkpointing using `ssc.checkpoint(<checkpoint directory>)` as described [earlier](#rdd-checkpointing-within-dstreams) .

run
tdas: Similar for loop exists in the run script, for pyspark for example. Also, this method does not search for subdirectories recursively (http://docs.oracle.com/javase/6/docs/technotes/tools/windows/classpath.html). So even though it may generate a long jar list, this will catch all JAR files in lib without missing anything AND without specifying each directory manually.

In this particular case, since there is just one Jar file (kafka), its fine to leave it as it is. Its unlikely that any other jar is going to be added. In the event it does happen, this format will automatically catch that JAR as well without any other changes.

mateiz: I know it's not recursive, but we changed the previous ones in the run script for a reason -- the path was getting really long, and making it hard to identify processes from the command line. I'd prefer to use \* there.

tdas: Done. Updated run2.cmd as well.

tdas:
@mateiz I think I have done all the fixes that you suggested. Once the pull request goes through, then I will able to test the strata stuff properly with Shivaram's AMI.

run
```diff
@@ -111,11 +111,13 @@ CLASSPATH+=":$FWDIR/conf"
 CLASSPATH+=":$CORE_DIR/target/scala-$SCALA_VERSION/classes"
 if [ -n "$SPARK_TESTING" ] ; then
   CLASSPATH+=":$CORE_DIR/target/scala-$SCALA_VERSION/test-classes"
+  CLASSPATH+=":$STREAMING_DIR/target/scala-$SCALA_VERSION/test-classes"
 fi
 CLASSPATH+=":$CORE_DIR/src/main/resources"
 CLASSPATH+=":$REPL_DIR/target/scala-$SCALA_VERSION/classes"
 CLASSPATH+=":$EXAMPLES_DIR/target/scala-$SCALA_VERSION/classes"
 CLASSPATH+=":$STREAMING_DIR/target/scala-$SCALA_VERSION/classes"
+CLASSPATH+=":$STREAMING_DIR/lib/org/apache/kafka/kafka/0.7.2-spark/*" # <-- our in-project Kafka Jar
```
mateiz, line 120: Ah, I see now, the problem was that Kafka was in this weird directory structure for our Maven repository hack to include it. I guess we'll leave it like this until we publish that Kafka JAR somewhere.

mateiz:
Alright, thanks TD!

run
tdas, line 120: That is precisely why i did not want to include customize it like that. Makes it a little brittle. Especially because there is no testcase for KafkaStream as it requires Zookeeper. So any changes in the location of the JAR would break stuff without noticing. In fact things were actually broken until Sean McNamara tested it and fixed it. Something to keep in mind in the future if these locations are changed.

PR #496, opened 2013-02-25
Moving akka dependency resolver to shared.

ScrapCodes:


tdas:
Prashant, can you send this pull request against mesos/master directly, so that it comes to Matei's attention directly.

ScrapCodes:
Moved to master

PR #498, opened 2013-02-25
Disable remote lifecycle logging from Akka.

pwendell:
This changes the default setting to `off` for remote lifecycle events. When this is on, it is very chatty at the INFO level. It also prints out several ERROR messages sometimes when sc.stop() is called.


mateiz:
Cool, thanks.

JoshRosen:
What was the original reason for turning this on in c0b9ceb8c3d56c6d6f6f6b5925c87abad06be646?

mateiz:
It was to help debug some user problems, but I agree that it's too verbose. Maybe I can add a config option to turn it on though (spark.akka.logLifecycleEvents).

Matei

pwendell:
There might be an equivalent to log4j.properties we can just put on the
classpath that akka will detect. Not sure how their logging works, but they
must have something like this right?
- Patrick

PR #506, opened 2013-02-28
Fixed SPARK-706: Failures in block manager put leads to read task hanging.

rxin:


rxin:
Thanks @shivaram for pair coding.

mateiz:
Good catch, thanks.

PR #508, opened 2013-03-01
Avoid bind failure in InputStreamsSuite

markhamstra:
Trying to use a fixed port (testPort = 9999) for multiple instances of TestServer fails because the port is already bound and remains so immediately after a TestServer is stopped.  Letting the OS choose an available port avoids the reuse problem.


mateiz:
Good idea, thanks.

PR #509, opened 2013-03-04
Fix SPARK-663: Implement Fair scheduler within ClusterScheduler

rold50:
The SparkContext now has a DynamicVariable Properties object that threads can set using addLocalProperties(). initLocalProperties() is used when you are sharing threads (e.g., processing requests from a thread pool).
SparkContext also can be configured to choose the type of scheduler. It uses the spark.cluster.taskscheduler system property. By default it will use the ClusterScheduler. To use the fair scheduler we can set it to spark.scheduler.cluster.fair.FairClusterScheduler.

Changes to DagScheduler, DagSchedulerEvent, TaskSet, Stage are for propagating meta information down to the TaskScheduler.

FairClusterScheduler and FairTaskSetManager contain the implementation of the fair scheduler. It is similar to the Hadoop fair scheduler but simplified. Within a pool, we just use FIFO. Pools are statically defined through an allocation XML, which we can set the path to through the system property: spark.fairscheduler.allocation.file. If that is not specified, it will just create 1 pool named default. The current implementation will always create a default pool. Also, the current implementation only supports "weight". Currently, it does not support minShares, etc.

The following is a sample allocation XML:
`<allocations>
  <pool name="pool1">
     <weight>25</weight>
  </pool>
  <pool name="pool2">
     <weight>25</weight>
  </pool>
  <pool name="pool3">
     <weight>25</weight>
  </pool>
  <pool name="default">
     <weight>25</weight>
  </pool>
</allocations>`

To assign a job to a specific pool, in the spark program, the user can set the LocalProperty:
sc.addLocalProperties("spark.scheduler.cluster.fair.pool","some pool name")
If the pool name does not exist (in the Scheduler) or the user did not specify any pool for their job, the scheduler will just assign the job to the default pool.


mateiz:
Thanks for putting this together! I haven't had time to look at it in too much detail, but here's one quick question -- could we have a simpler pluggable scheduler interface that just sorts the queue of pending TaskSets instead of having to replicate so much of ClusterScheduler in each subclass? (I'm talking about the replicated bookkeeping code in resourceOffers, statusUpdate, etc).

rold50:
One issue is that it's not just about sorting the queue of pending TaskSets. Because in the fair scheduler implementation, I am using a different data structure: It has a queue of pools and within a pool there is a queue of tasksets, so in the resourceOffer, It's actually a 2-level sorting (sort the pools and then sort the TaskSet queue in the pool). Also in the ClusterScheduler, it only sort once and then make offers to the TaskSetManager. In my fair scheduler implementation, I re-sort after each time we have found a task for an offer.

We can probably reduce the replicated bookkeeping code, if some of the functions in ClusterScheduler are not set to private. For example, I had to override statusUpdate but it's exactly the same as the code in ClusterScheduler because I wanted to modify what happens in private def removeExecutor (to remove executor in our queue of pool, instead of the queue of tasksets in ClusterScheduler). If we change the visibility of removeExecutor to public in ClusterScheduler, then we don't have to override statusUpdate. 

For the most part,  most of the code that I override are to add additional bookkeeping code for pool datastructure. 

FairTaskSetManager is essentially the same as TaskSetManager. I only added code to also notify the scheduler when a task is lost or finished (rather than just calling the DagScheduler's TaskEnded). I needed this call so that I can keep track of the currently number of running tasks, because we use that information when we sort the pools.
We can probably do away with this class altogether if we modified TaskSetManager, to also call a TaskEnded function in ClusterScheduler, and we just override this in FairClusterScheduler.

mridulm:
As the size of the cluster increases, we have found that both
TaskSetManager and ClusterScheduler become extremely critical for
scaling and performance.
I am not sure how much trunk has diverged from 0.6 (yarn branch has
already diverged a bit from it) in this respect : but some of our
local changes have resulted in very nice speedup while having minimal
impact when number of workers is low.

So changes to both should also consider the overall system performance
and scaling also IMHO.

Regards,
Mridul

mateiz:
Rold -- instead of viewing it as sorting TaskSets, we could also have an abstraction that manages a task set queue and can be asked to send an offer to one of them. But I'd prefer to have a special class for that if possible. In general, it's better to use composition than inheritance when designing things like this -- ClusterScheduler is a rather complex class and we don't want changes in it to mess up all these subclasses. This is the same reason that things like SchedulerBackend are separate classes instead of subclasses of Scheduler.

Mridul -- definitely share what you guys had to optimize to make it run faster when you get a chance.

rold50:
Hi Matei,

I updated the code to use composition instead of inheritance. The Fair Task Scheduler does not inherit ClusterScheduler anymore. Rather, ClusterScheduler is now the entry point for everything. It uses TaskSetQueuesManager. 
FIFOTaskSetQueuesManager contains the original comparison logic in ClusterScheduler (single ActiveTaskSet queue, etc).
FairTaskSetQueuesManager contains the fair scheduling logic (pools of ActiveTaskSet queue, etc)

By default ClusterScheduler will use the FIFOTaskSetQueuesManager.
We can set the system property spark.cluster.taskscheduler=spark.scheduler.cluster.FairTaskSetQueuesManager to use the fair policy.

core/src/main/scala/spark/scheduler/cluster/FairTaskSetQueuesManager.scala
```diff
@@ -0,0 +1,183 @@
...
+    val file = new File(schedulerAllocFile)
+    if(!file.exists()) {
+    //if file does not exist, we just create 1 pool, default
+      val pool = new Pool("default",100)
+      pools += pool
+      poolNameToPool("default") = pool
+      logInfo("Created a default pool with weight = 100")
+    }
+    else {
+      val xml = XML.loadFile(file)
```
shivaram, line 41: Can we use JSON or YAML instead of XML ? Scala has a JSON parser at scala.util.parsing.json

mateiz, line 41: Hmm, I think we should actually decide this more carefully. I found XML more human-readable than JSON (mostly because I don't like writing curly braces). Have you seen systems with readable config in JSON?

As an example of XML, look at the Hadoop Fair Scheduler config: http://hadoop.apache.org/docs/r1.1.1/fair_scheduler.html#Allocation+File+%28fair-scheduler.xml%29. (There's a sample lower down on that.) I don't think it's that bad despite being XML.

pwendell, line 41: In the early days of Avro you could only use JSON to write out schema's and it was incredibly cumbersome to write by hand. JSON is useful as a verbose serialization format (which is now basically the role it plays in Avro - they added a more concise IDL), but for hand-crafting I'd much prefer XML.

shivaram, line 41: My main concerns with XML are that it gets very verbose as  you add more options and things like commenting are hard / ugly.  I tried to write the same hadoop example is YAML (https://gist.github.com/shivaram/5123598) and JSON. https://gist.github.com/shivaram/5123599 

The YAML one looks more nicer and also has elegant comments that can be a part of the config. However YAML parsing isn't available in the scala library and we'll need another dependency to parse this.

Another option to consider is the typesafe config library. https://github.com/typesafehub/config . This is Scala friendly and is used for akka configuration --It looks like a modified version of JSON which is easier to write.

rxin, line 41: I would like to echo the sentiment that XML is more human-readable/writable. In particular, I think JSON is too error prone for humans to write.

xiajunluan:
Rold- After reviewing your code, it amazed me  that you completed fair scheduler very quickly, and I think there may be two points that need refactoring
1. There is no problem to use DynamicVariable to get pool name for each job, but it looks like somewhat  redundancy to transfer local properies in recursion construction of stages. we could just hold properies in ActiveJob, when we construct "TaskSet" in submitMissingTasks function ,we could find properties by stage->jobid(priority in stage)->ActiveJob(need to add HashSet(JobId, ActiveJob))->properties, that is to say, we will consider properties as the member element of job not stage.
2. I assume that requirement of fair scheduler we will implement in Spark is similar with it in Hadoop. For fair schedule in Spark, the schedulable entities are Pools and TaskSetManagers, we will firstly choose one pool using fair scheduler algorithm between pools, and then choose one TaskSetManager within pool using FIFO or FS. So taking FairTaskSetQueuesManager for example, after choosing pool in receiveOffers function, we should not expose how we get a proper TaksSetManager(Here is FIFO) within one pool to FairTaskSetQueuesManager, we could abstract this section into pool class, also we should abstract one interface of FIFO and FS algorithm for different schedulable entities. I think we might refer to the design of fair scheduler in Hadoop and form below diagram.
   Fair Scheduler another version

Schedulable class: contain scheduler related memeber elements and functios(e.g. minishare and weight for FS, priority and starttime for FIFO)
Comparator<Schedulable>:Sort the Pools or TaskSetManagers with certain algorithem
TaskSetManager: Do a minor changes to make TaskSetManager schedulable

rold50:
Hi Andrew,

Thanks. I'll refactor the code based on your comments, when I get back from
Spring break.

>    1.
> 
>    I assume that requirement of fair scheduler we will implement in Spark
>    is similar with it in Hadoop. For fair schedule in Spark, the schedulable
>    entities are Pools and TaskSetManagers, we will firstly choose one pool
>    using fair scheduler algorithm between pools, and then choose one
>    TaskSetManager within pool using FIFO or FS. So taking
>    FairTaskSetQueuesManager for example, after choosing pool in receiveOffers
>    function, we should not expose how we get a proper TaksSetManager(Here is
>    FIFO) within one pool to FairTaskSetQueuesManager, we could abstract this
>    section into pool class, also we should abstract one interface of FIFO and
>    FS algorithm for different schedulable entities. I think we might refer to
>    the design of fair scheduler in Hadoop and form below diagram.
>    [image: Fair Scheduler another version]https://f.cloud.github.com/assets/1168928/240843/c83f6b3a-8953-11e2-80e4-7875e58c4f74.jpg
> 
> Schedulable class: contain scheduler related elements(e.g. minishare and
> weight for FS, priority and starttime for FIFO)
> Comparator:Sort the Pools or TaskSetManagers with certain algorithem
> 
> Right. When I implemented the fair scheduler, I was only interested in the
> most basic case of fair scheduler (wrt my own research project). But, I
> agree, I think it makes sense to have it similar to the Hadoop case and
> what you mention makes sense (i.e., have both Pools and TaskSetManager
>  implement a Schedulable class)

-Harold

xiajunluan:
Hi Harold

   It happened that we would like to add fair scheduler feature to spark in short time, so do you mind that we collaborate to develop this feature together or I continue to refactor points I mentioned above based on your codes?

rold50:
Hi Andrew,

Either way is fine for me. It seems that the points you mentioned shouldn't
take that long to implement. Since I won't be back until next week (March
18ish), maybe you can go ahead and continue to refactor it based on my code
(Not really familiar yet with github, do you now fork my forked spark repo
so that you can use my code, then make a pull request of the changes to my
repo?)? If when I'm back and it's not yet done, then we can collaborate  to
develop this feature together.

-Harold

mateiz:
Andrew, you should be able to just pull Harold's changes into your own repo (by checking out his master branch) and then send a pull request to us. That sounds like a good plan.

xiajunluan:
Hi Harold
    That is fine, I will go head and look forward for your return soon.

Hi Mateiz
    I will continue to develop this feature following your guide.

core/src/main/scala/spark/scheduler/cluster/FairTaskSetQueuesManager.scala
jason-dai, line 41: An objective we have is to have the FairScheduler in Spark to reuse the FairScheduler config files in Hadoop, which will make it easy for people to deploy Spark alongside Hadoop. Any comments?

rold50, line 41: Right, the format of the config file should be the same (or similar). I
currently only parse the weight and pool name info because the current
implementation only uses weight for it's fair comparison (it doesn't have
the other bells and whistles such as min/max slots, etc). However, I think
the Hadoop fair allocation config file distinguished between map slots and
reduce slots, I think in our case we won't have that but instead maybe just
have minSlots, maxSlots, etc.

rold50:
I think this is now folded into: https://github.com/mesos/spark/pull/529

rxin:
This is now in #529

PR #511, opened 2013-03-06
Use scala found in PATH by default if it's the right version

jey:
This commit allows for a smoother out-of-the-box experience by not requiring the user to create a `conf/spark-env.sh` file if they already have an apropriate `scala` in their `PATH`. 

This is done by setting `SPARK_LAUNCH_WITH_SCALA` if `SCALA_HOME` is unset and a suitable `scala` is found.


mateiz:
Hi Jey,

Unfortunately this is not a good idea because the reason we sometimes launch with java instead of scala is to make sure that we can kill the child process. When our standalone cluster Worker spawns a child process, if the shell script execs scala instead of java, you get the scala process itself (also a shell script) and a child of that which is the JVM. When we call Process.kill() on the scala process, it doesn't kill the JVM. So we have to run with Java. Perhaps a better fix would be to ask the users to do sbt assembly or mvn package and then assume that the final built JAR includes Scala.

rxin:
A side note - it might be good to write that into the run script as a block comment.

mateiz:
Well, it might get fixed in this: https://github.com/mesos/spark/pull/488. But the problem is that this will also require SCALA_HOME.

jey:
Makes sense, I'll look into it further.

PR #512, opened 2013-03-08
Fix reference bug in Kryo serializer, add test, update version

patelh:
Fix reference bug in kryo serializer, add test, and bump version.

https://groups.google.com/forum/#!msg/kryo-users/Eu5V4bxCfws/k-8UQ22y59AJ


mateiz:
Thanks for the fix.

PR #513, opened 2013-03-09
Adds choice of persistence level to Bagel.

MLnick:
This PR adds support for selecting the `StorageLevel` to use when running a Bagel computation. The default remains `MEMORY_ONLY` and should be fully backward-compatible in terms of `run` method calls.

Note that due to the way Scala handles default arguments it was not possible to simply add the `storageLevel` parameter to each `run` method with a default. Hence I did so for the "main" `run` method and then added a second method for each of the remaining ones, that takes care of the default argument (and whose signature matches the old versions).

See https://groups.google.com/forum/?fromgroups=#!topic/spark-developers/F_UOsKM8ZEM for the discussion.

I also added some documentation and a (probably superfluous) test to the `BagelSuite` to test use of the non-default `StorageLevel` parameter.

### DIFF
--- a/bagel/src/main/scala/spark/bagel/Bagel.scala
+++ b/bagel/src/main/scala/spark/bagel/Bagel.scala
@@ -34,7 +34,7 @@ object Bagel extends Logging {
    * @tparam M message type
    * @tparam C combiner
    * @tparam A aggregator
-   * @return a set of (K, V) pairs representing the graph after completion of the program
+   * @return an RDD of (K, V) pairs representing the graph after completion of the program
    */
   def run[K: Manifest, V <: Vertex : Manifest, M <: Message[K] : Manifest,
           C: Manifest, A: Manifest](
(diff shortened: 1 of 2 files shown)


bagel/src/main/scala/spark/bagel/Bagel.scala
```diff
@@ -73,13 +120,27 @@ object Bagel extends Logging {
...
   ): RDD[(K, V)] = {
     val part = new HashPartitioner(numPartitions)
     run[K, V, M, C, Nothing](
-      sc, vertices, messages, combiner, None, part, numPartitions)(
+      sc, vertices, messages, combiner, None, part, numPartitions, storageLevel)(
       addAggregatorArg[K, V, M, C](compute))
   }
 
+  /** Runs a Bagel program with no [[spark.bagel.Aggregator]], default [[spark.HashPartitioner]],
+    * [[spark.bagel.DefaultCombiner]] and the default storage level */
```
mateiz: Looks good, but make the comment formatting consistent with the rest of the codebase; it should look like this:

```
/**
 * Blah blah
 */
```

Or, for single-line comments, this is also okay:

```
/** Blah */
```

MLnick:
@mateiz made the suggested doc style changes.

mateiz:
Cool, thanks!

PR #514, opened 2013-03-09
Prevent DAGSchedulerSuite from corrupting driver.port.

charlesreiss:
Use the LocalSparkContext abstraction to properly manage clearing
spark.driver.port.

This will probably conflict with pull request #458 (but that pull request does not fix this issue).


mateiz:
Merged this manually; thanks.

PR #515, opened 2013-03-09
Notify standalone deploy client of application death.

charlesreiss:
Send a message to deploy clients when their application is removed.

Usually, this isn't necessary since the application will be removed as a result of the deploy client disconnecting, but occassionally, the standalone deploy master removes an application otherwise.

Also mark applications as FAILED instead of FINISHED when they are killed as a result of their executors failing too many times.

I kept the marking applications as FINISHED for the Akka-detected disconnection cases, since I don't know which of those, if any, we believe indicate failures.


mateiz:
Good idea, thanks!

PR #516, opened 2013-03-10
Fix local metrics

squito:
TaskMetrics actually work in local mode now, with a unit test.

Also, I renamed `remoteFetchWaitMillis` to `fetchWaitMillis`, b/c it also includes time spent waiting on the local fetch.

### DIFF
--- a/core/src/test/scala/spark/scheduler/SparkListenerSuite.scala
+++ b/core/src/test/scala/spark/scheduler/SparkListenerSuite.scala
@@ -20,7 +20,7 @@ class SparkListenerSuite extends FunSuite with LocalSparkContext with ShouldMatc
     //just to make sure some of the tasks take a noticeable amount of time
     val w = {i:Int =>
       if (i == 0)
-        Thread.sleep(25)
+        Thread.sleep(100)
       i
     }
 
--- a/core/src/test/scala/spark/scheduler/SparkListenerSuite.scala
+++ b/core/src/test/scala/spark/scheduler/SparkListenerSuite.scala
@@ -1,0 +1,80 @@
+package spark.scheduler
+
+import org.scalatest.FunSuite
+import spark.{SparkContext, LocalSparkContext}
+import scala.collection.mutable
+import org.scalatest.matchers.ShouldMatchers
+import spark.SparkContext._
+
+/**
+ *
+ */
+
+class SparkListenerSuite extends FunSuite with LocalSparkContext with ShouldMatchers {
+
+  test("local metrics") {
+    sc = new SparkContext("local[4]", "test")
+    val listener = new SaveStageInfo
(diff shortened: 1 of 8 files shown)


core/src/test/scala/spark/scheduler/SparkListenerSuite.scala
```diff
@@ -0,0 +1,80 @@
...
+    }
+  }
+
+  def checkNonZeroAvg(m: Traversable[Long], msg: String) {
+    assert(m.sum / m.size.toDouble > 0.0, msg)
+  }
+
+  def isStage(stageInfo: StageInfo, rddNames: Set[String], excludedNames: Set[String]) = {
+    val names = Set(stageInfo.stage.rdd.name) ++ stageInfo.stage.rdd.dependencies.map{_.rdd.name}
+    !names.intersect(rddNames).isEmpty && names.intersect(excludedNames).isEmpty
```
squito, line 76: I know this is really weird, but I'm not sure of a better way to figure out what is actually in a shuffle stage.  `stage.rdd` isn't the "user" rdd, so you have to check the dependencies.  There probably should be a better way of tying this stage back to user-defined rdds, b/c I'm guessing this will confuse users also.

core/src/test/scala/spark/scheduler/SparkListenerSuite.scala
```diff
@@ -0,0 +1,80 @@
...
+    val d3 = d.map{i => i -> (0 to (i % 5))}.setName("shuffle input 2")
+
+    val d4 = d2.cogroup(d3, 64).map{case(k,(v1,v2)) => k -> (v1.size, v2.size)}
+    d4.setName("A Cogroup")
+
+    d4.collectAsMap
+
+    listener.stageInfos.size should be (4)
+    listener.stageInfos.foreach {stageInfo =>
+      //small test, so some tasks might take less than 1 millisecond, but average should be greater than 1 ms
```
mateiz, line 42: Is this guaranteed? It might be good to add a Thread.sleep(100) in the tasks just in case some people's computers run them too quickly, or their clock is low resolution. For example, on Windows, I think System.currentTimeMillis only advances every 10 or so milliseconds.

squito:
good point, sorry that was lazy of me.  Any idea how I can easily add in a small wait for the fetch time?  Otherwise that test for non-zero average has the same problem.  Maybe I should just remove that check.

mateiz:
Yeah, maybe you can remove that for fetch time (or just change it to >= 0). Not sure how to easily add it.

mateiz:
Actually I'll just implement that myself I guess. Thanks for the fix.

squito:
thanks! sorry I hadn't gotten to it yet

PR #518, opened 2013-03-11
Send block sizes as longs in BlockManager updates

charlesreiss:
Currently, UpdateBlockMessage stores the sizes of blocks (in memory and on disk) as longs but serializes them as ints, silently truncating the values. This means that blocks which are larger (or estimated larger) than 2G are accounted for incorrectly, which can result in the master thinking a block manager has more memory free than its capacity. 

This patch serializes UpdateBlockMessage sizes as longs.

I assume that the sizes were serialized as ints for some reason, in which case this may create a regression. (I suspect that there are other problems with blocks >2G, e.g. it may not be possible to persist them at MEMORY_ONLY_SER, but these should at least not be silent problems.)

### DIFF
--- a/core/src/main/scala/spark/storage/BlockManagerMessages.scala
+++ b/core/src/main/scala/spark/storage/BlockManagerMessages.scala
@@ -49,16 +49,16 @@ class UpdateBlockInfo(
     blockManagerId.writeExternal(out)
     out.writeUTF(blockId)
     storageLevel.writeExternal(out)
-    out.writeInt(memSize.toInt)
-    out.writeInt(diskSize.toInt)
+    out.writeLong(memSize)
+    out.writeLong(diskSize)
   }
 
   override def readExternal(in: ObjectInput) {
     blockManagerId = BlockManagerId(in)
     blockId = in.readUTF()
     storageLevel = StorageLevel(in)
-    memSize = in.readInt()
-    diskSize = in.readInt()
+    memSize = in.readLong()
+    diskSize = in.readLong()
   }
 }
 


rxin:
Charles,

Do you expect block sizes to be greater than 2G? Increasing the range from ints to longs would double the size of the messages sent from workers to the master during shuffle (and caching). It could become a scalability problem as the number of nodes increases.

Another option is to use the lowest bit to indicate whether we should multiply the number by 1 or 1000.

charlesreiss:
Although I would not recommend a configuration with block sizes larger than 2G, I don't think they're unreasonable on machines which have >>4G RAM per core, which we have some of in the AMP Lab, and people are likely to run in to them otherwise (because they have such machines or because of a reducer of death) and we should do something unsurprising with BlockManager sizes.

This does not affect messages sent during shuffle. Shuffle blocks are not reported to the master through BlockManager (false flag on the BlockManager#put() here: https://github.com/mesos/spark/blob/master/core/src/main/scala/spark/scheduler/ShuffleMapTask.scala#L142 ); they are collected at the master through ShuffleMapTask results and MapOutputTracker, which already sends the sizes in a compressed/approximate way.

If we are worried about the scalability issue, it would be reasonable to use something like ZigZag encoding or to use an approximate encoding like in MapOutputTracker (assuming that there's no problem with BlockManagerMaster only reporting approximate memory free sizes). But, given that shuffles should not use this code path and that we are already passing the rather bulky BlockManagerId with each UpdateBlockInfo, I don't think a scalability issue is likely.

rxin:
Ah ok, as long as it doesn't affect shuffle, it is much less of a concern.

mateiz:
Good catch, thanks.

PR #521, opened 2013-03-13
Close the reader in HadoopRDD as soon as iteration end.

stephenh:
This fixes a bug where HadoopRDD's reader was only closed in the TaskContext completion context.

This is normally fine, but when coalescing a large number of S3 files, the task may not complete until too many readers have been left open, and the S3 connection pool will fill up (resulting in very misleading "connection timeout" exceptions).

This changes the HadoopRDD to call close() both at the end of iteration (when on errors) and also in the TaskContext completion (if there are failures, but will call close again if there are no errors--calling it twice should be fine).

There were several "gotNext"-based iterator implementations in the codebase, each slightly different, so along with fixing this bug, I pulled the common code out into a "NextIterator" helper class, so that each location only needs to implement "getNext".

### DIFF
--- a/core/src/main/scala/spark/rdd/HadoopRDD.scala
+++ b/core/src/main/scala/spark/rdd/HadoopRDD.scala
@@ -72,7 +72,7 @@ class HadoopRDD[K, V](
     reader = fmt.getRecordReader(split.inputSplit.value, conf, Reporter.NULL)
 
     // Register an on-task-completion callback to close the input stream.
-    context.addOnCompleteCallback{ () => close() }
+    context.addOnCompleteCallback{ () => closeIfNeeded() }
 
     val key: K = reader.createKey()
     val value: V = reader.createValue()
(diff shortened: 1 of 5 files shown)


mateiz:
Hey Stephen,

This is good, but I think there were some problems before with close() being called twice on some streams, which caused some InputFormats to throw an exception. Could you add a guard around the call to close() perhaps? Or maybe this has to be in each subclass.

stephenh:
Hi Matei; okay, I added a "closeIfNeeded" check in the NextIterator. That should handle HadoopRDD. The other classes were not calling close directly (and Serializer was using ByteArrayInputStreams that don't need closing anyway), so I think it should be fine.

stephenh:
Hm, just a sec, let me check one thing...

stephenh:
Okay, everything was fine, but I wanted to sanity check some boundary cases, so I added a test for NextIterator.

mateiz:
Cool, thanks!

PR #522, opened 2013-03-13
Shuffle Performance fix: Use netty embeded OIO file server instead of Co...

shane-huang:
...nnectionManager

This patch provides an alternative way to do shuffle. Basically it uses an embeded netty OIO file server to serve blocks to BlockManager instead of the current message exchanging ConnectionManager.  We provide an option for user to choose between the two alternatives. Set system property "spark.shuffle.use.netty=true" to enable this feature. Default is disabled. 
In our tests using netty can speedup jobs from 2x~5x. 
BTW, I have to use a lot of java implementations when using netty because scala 2.9.2 has a bug (please refer to https://github.com/netty/netty/issues/781). They can be reimplemented with scala if we upgrade scala to 2.10x

### DIFF
--- a/core/src/main/scala/spark/netty/ShuffleCopier.scala
+++ b/core/src/main/scala/spark/netty/ShuffleCopier.scala
@@ -1,0 +1,95 @@
+package spark.netty
+
+import io.netty.buffer.ByteBuf
+import io.netty.channel.ChannelHandlerContext
+import io.netty.channel.ChannelInboundByteHandlerAdapter
+import io.netty.util.CharsetUtil
+
+import java.util.concurrent.atomic.AtomicInteger
+import java.util.logging.Logger
+import spark.Logging
+import spark.network.ConnectionManagerId
+import java.util.concurrent.Executors
+
+private[spark] class ShuffleCopier extends Logging {
+
+  def getBlock(cmId: ConnectionManagerId,
+    blockId: String,
+    size: Long,
+    resultCollectCallback: (String, Long, ByteBuf) => Unit) = {
+
+    val handler = new ShuffleClientHandler(resultCollectCallback)
+    val fc = new FileClient(handler)
+    fc.init();
+    fc.connect(cmId.host, cmId.port)
+    fc.sendRequest(blockId)
+    fc.waitForClose();
+    fc.close();
+  }
+  
+  def getBlocks(cmId: ConnectionManagerId,
+    blocks: Seq[(String, Long)],
+    resultCollectCallback: (String, Long, ByteBuf) => Unit) = {
+
+    blocks.map { 
+      case (blockId, size) => {
+        getBlock(cmId,blockId,size,resultCollectCallback)
+      }
+    }
+  }
+
+}
+
+private[spark] class ShuffleClientHandler(val resultCollectCallBack: (String, Long, ByteBuf) => Unit ) extends FileClientHandler with Logging {
+
+  def handle(ctx: ChannelHandlerContext, in: ByteBuf, header: FileHeader) {
+
+    logInfo("Received Block: " + header.blockId);
+    //in.readBytes(header.fileLen)
+    resultCollectCallBack(header.blockId, header.fileLen.toLong, in.readBytes(header.fileLen))
+    closeContext(ctx);
+  }
+
+}
+
+private[spark] object ShuffleCopier extends Logging {
+
+  def echoResultCollectCallBack(blockId: String, size: Long, content: ByteBuf) = {
+    logInfo("File: " + blockId + " content is : \" "
+      + content.toString(CharsetUtil.UTF_8) + "\"")
+  }
+
+  def runGetBlock(host:String, port:Int, file:String){
+    val handler = new ShuffleClientHandler(echoResultCollectCallBack)
+    val fc = new FileClient(handler)
+    fc.init();
+    fc.connect(host, port)
+    fc.sendRequest(file)
+    //Thread.sleep(2000)
+    fc.waitForClose();
+    fc.close()
+  }
+
+  def main(args: Array[String]) {
+    if (args.length < 3) {
+      System.err.println("Usage: ShuffleCopier <host> <port> <shuffle_block_id> <threads>")
+      System.exit(1)
+    }
+    val host = args(0)
+    val port = args(1).toInt
+    val file = args(2)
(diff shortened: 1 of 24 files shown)


core/src/main/scala/spark/netty/ShuffleCopier.scala
```diff
@@ -0,0 +1,95 @@
...
+      System.err.println("Usage: ShuffleCopier <host> <port> <shuffle_block_id> <threads>")
+      System.exit(1)
+    }
+    val host = args(0)
+    val port = args(1).toInt
+    val file = args(2)
+    val threads = if (args.length>3) args(3).toInt else 10
+
+    val copiers = Executors.newFixedThreadPool(80)
+    for (i <- Range(1,threads)){
```
shane-huang: forgot to change this, should be Range (0, threads)

shane-huang:
Commit again to address all review comments till now.

mridulm:
This looks very promising !
We have had to hack on Connection and ConnectionManager a bit due to
performance issues and bugs - this change should hopefully replace all
that with a more robust codebase.

Thanks !
Mridul

shane-huang:
Please do not submit this patch right now. I'm working on a few optimizations and will check-in soon.

shane-huang:
Close this pull request temporarily. Will submit another optimized version later.

shane-huang:
The latest optimization could further speedup some of our workload by 1~2x.

core/src/main/scala/spark/storage/BlockManager.scala
```diff
@@ -474,52 +641,59 @@ class BlockManager(val master: BlockManagerMaster, val serializer: Serializer, m
...
+        val iterator = blockInfos.iterator
+        var curRequestSize = 0L
+        var curBlocks = new ArrayBuffer[(String, Long)]
+        while (iterator.hasNext) {
+          val (blockId, size) = iterator.next()
+          if (size > 0) {
+            curBlocks += ((blockId, size))
+            curRequestSize += size
+          } else if (size == 0){
+            totalBlocks -= 1
```
jason-dai, line 678: We can apply this to local blocks as well, yes?

shivaram:
Thanks for sending this out. I have a couple of high level comments/clarifications

a. The netty file handler and server code are right now in java/spark/network/netty. Since all of Spark's core components are in Scala, I think it'll be better to translate these files to scala (say scala/spark/network/netty).

b. Right now, it looks like netty handlers read from the filesystem directly bypassing spark. While this works for shuffles where the intermediate data is stored on disk, it would be better to also support in memory shuffles. Could we just read from the BlockManager instead of the filesystem in the handler to do this ? Are there any performance implications when using the BlockManager vs. the filesystem ?

shane-huang:
> a. The netty file handler and server code are right now in java/spark/network/netty. Since all of Spark's core components are in Scala, I think it'll be better to translate these files to scala (say scala/spark/network/netty).
>    As I noted in the opening comment,  I have to use java implementations (in java/spark/network/netty) because scala 2.9.2 compiler has a bug (please refer to netty/netty#781). They can be reimplemented with scala if we upgrade scala to 2.10x
> b. Right now, it looks like netty handlers read from the filesystem directly bypassing spark. While this works for shuffles where the intermediate data is stored on disk, it would be better to also support in memory shuffles. Could we just read from the BlockManager instead of the filesystem in the handler to do this ? Are there any performance implications when using the BlockManager vs. the filesystem ?
>    We make it a file server mostly for simpleness. It looks like we're only serving files in shuffle in current implementation, right? And it seems Matei also thinks a simple file server will do (see discussion in shark user group https://groups.google.com/d/msg/shark-users/IHOb2u5HXSk/huTWyosI1n4J).

rxin:
Hi Shane,

I think it'd be better to isolate the Scala bug into a single Java file, and move the rest Java files into Scala. This way, when we move to Scala 2.10, it is easy to just re-implement the single file.

Re: using block manager

I talked to Shivaram and for now it might be better to stick to the current way (not using block manager) because we want to be able to leverage netty's zero copy strategies.

In the long run (beyond this patch), we should look into:
1. Performance of netty nio vs communication layer - ideally it'd be nice to use netty for both nio and oio.
2. Refactoring block manager to support shuffling both disk store data (files) and in memory data. We found in the Shark experiments that when shuffle data was put in memory, it decreased the tail latency of individual tasks and thus improved overall query execution performance. But it is hard to manage memory when you put shuffle data in memory - that's why that hasn't made its way into master branch yet.

shane-huang:
Ok. I'll move java implementation to scala as much as possible and leave the rest in one file. .

Sigehere:
Hi Friends,
I have made changes as you told in " https://github.com/mesos/spark/pull/522/files#r3413134 " Link.
Then i have build spark with 
$ sbt/sbt package
Now, am I ready to use spark with new changes?
Or anything else is required over hear?

shane-huang:
You have to enable it by setting system property "spark.shuffle.use.netty=true" in spark-env.sh. Format like follows:
SPARK_JAVA_OPTS+="-Dspark.shuffle.use.netty=true "
export SPARK_JAVA_OPTS

shane-huang:
@rxin @shivaram When is scala-2.10 support available? If it is available in master soon, I think I can wait till then and move all java code into scala instead of leaving a java snippet around.

shivaram:
I am not sure when support for Scala 2.10 will be in place.  @mateiz might be able to add more about this.

mateiz:
My guess is that it might take around 2 weeks, mostly because I'm traveling a bunch. If it's not much trouble, it might be better to make it work now with 2.9, by keeping some classes in Java. That would also allow us to release this fix in 0.7.1 for existing 2.9 users.

shane-huang:
@mateiz This patch is currently based on branch-0.6. And I have to backport a few code from master branch in file core/src/main/scala/spark/MapOutputTracker.scala (about compress/decompress block size) to make it correct. So if user disable the netty option, the original code path is NOT remain unchanged. I'm not sure whether this is an issue. I'm going to rebase this patch into master and do all tests again so that I don't have to do the modifications in MapOutputTracker. Later when I have time I can backport it to branch-0.6

shane-huang:
Close this pull request and make it against master branch (0.7) instead. The new pull request is https://github.com/mesos/spark/pull/557

PR #526, opened 2013-03-14
Add foldByKey

markhamstra:
Motivation: https://groups.google.com/forum/?fromgroups=#!topic/spark-users/Uo2KCj5HJH4


mateiz:
Hey Mark,

It's a little bit unfortunate that we have to do groupByKey instead of reduceByKey here, since it won't do local combining on each node. In Scala's parallel collections' fold, is the zero value assumed to really act as a zero? That is, can they fold separate ranges with it, and then fold the results? If so, we should assume that too and write this in terms of combineByKey instead.

markhamstra:
Yes, Matei, your comments are consistent with how the zero value works in the standard library's fold over collections.  Doc comments updated to mirror the standard library and foldByKey refactored to use combineByKey.

mateiz:
Great, thanks!

PR #527, opened 2013-03-15
KafkaInputDStream fixes and improvements

mcnamaras:
Improvements:
- fixes critical memory leak.
- makes available all the kafka consumer config properties (ie- autooffset.reset, see http://kafka.apache.org/configuration.html).
- removes dependency on private kafka class ZookeeperConsumerConnector.
- cleans up kafkaStream interface by removing initialOffets/setOffsets logic (now handled by kafkaParams /w autooffset.reset)

Example creating kafkaStream:

``` scala
val kafkaParams = Map(
    "zk.connect" -> "zoo01,zoo02,zoo03",
    "groupid" -> "my-consumer-group",
    "autooffset.reset" -> "smallest" // start at topic beginning
)

val inputStream = ssc.kafkaStream(kafkaParams, topics)
```

### DIFF
--- a/streaming/src/main/scala/spark/streaming/StreamingContext.scala
+++ b/streaming/src/main/scala/spark/streaming/StreamingContext.scala
@@ -204,8 +204,6 @@ class StreamingContext private (
    * @param groupId The group id for this consumer.
    * @param topics Map of (topic_name -> numPartitions) to consume. Each partition is consumed
    * in its own thread.
-   * @param initialOffsets Optional initial offsets for each of the partitions to consume.
-   * By default the value is pulled from zookeper.
    * @param storageLevel  Storage level to use for storing the received objects
    *                      (default: StorageLevel.MEMORY_AND_DISK_SER_2)
    */
@@ -213,11 +211,10 @@ class StreamingContext private (
       zkQuorum: String,
       groupId: String,
       topics: Map[String, Int],
-      initialOffsets: Map[KafkaPartitionKey, Long] = Map[KafkaPartitionKey, Long](),
       storageLevel: StorageLevel = StorageLevel.MEMORY_ONLY_SER_2
     ): DStream[T] = {
     val kafkaParams = Map[String, String]("zk.connect" -> zkQuorum, "groupid" -> groupId, "zk.connectiontimeout.ms" -> "10000");
-    kafkaStream[T](kafkaParams, topics, initialOffsets, storageLevel)
+    kafkaStream[T](kafkaParams, topics, storageLevel)
   }
 
   /**
@@ -225,16 +222,14 @@ class StreamingContext private (
    * @param kafkaParams Map of kafka configuration paramaters. See: http://kafka.apache.org/configuration.html
    * @param topics Map of (topic_name -> numPartitions) to consume. Each partition is consumed
    * in its own thread.
-   * @param initialOffsets Optional initial offsets for each of the partitions to consume.
    * @param storageLevel  Storage level to use for storing the received objects
    */
   def kafkaStream[T: ClassManifest](
       kafkaParams: Map[String, String],
       topics: Map[String, Int],
-      initialOffsets: Map[KafkaPartitionKey, Long],
       storageLevel: StorageLevel
     ): DStream[T] = {
-    val inputStream = new KafkaInputDStream[T](this, kafkaParams, topics, initialOffsets, storageLevel)
+    val inputStream = new KafkaInputDStream[T](this, kafkaParams, topics, storageLevel)
     registerInputStream(inputStream)
     inputStream
   }
--- a/streaming/src/main/scala/spark/streaming/StreamingContext.scala
+++ b/streaming/src/main/scala/spark/streaming/StreamingContext.scala
@@ -199,7 +199,7 @@ class StreamingContext private (
   }
 
   /**
-   * Create an input stream that pulls messages form a Kafka Broker.
+   * Create an input stream that pulls messages from a Kafka Broker.
    * @param zkQuorum Zookeper quorum (hostname:port,hostname:port,..).
    * @param groupId The group id for this consumer.
    * @param topics Map of (topic_name -> numPartitions) to consume. Each partition is consumed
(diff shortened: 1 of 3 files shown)


streaming/src/main/scala/spark/streaming/StreamingContext.scala
```diff
@@ -213,10 +211,25 @@ class StreamingContext private (
       zkQuorum: String,
       groupId: String,
       topics: Map[String, Int],
-      initialOffsets: Map[KafkaPartitionKey, Long] = Map[KafkaPartitionKey, Long](),
```
tdas, line 216: Cant the initialOffsets be supported in this new code? I would like to keep the existing API as intact as possible.

streaming/src/main/scala/spark/streaming/api/java/JavaStreamingContext.scala
```diff
@@ -125,7 +98,6 @@ class JavaStreamingContext(val ssc: StreamingContext) {
       zkQuorum,
       groupId,
       Map(topics.mapValues(_.intValue()).toSeq: _*),
-      Map(initialOffsets.mapValues(_.longValue()).toSeq: _*),
       storageLevel)
   }
```
tdas, line 101: The new kafkaStream function in the StreamingContext (with the kafkaParams map) should also be added in the JavaStreamingContext. Can you do that? And it should take java map and not scala map, and inside the function that conversion from java to scala map should be made.

streaming/src/main/scala/spark/streaming/dstream/KafkaInputDStream.scala
```diff
@@ -97,29 +88,29 @@ class KafkaReceiver(zkQuorum: String, groupId: String,
...
         blockGenerator += msgAndMetadata.message
-        // Keep on handling messages
-
-        true
       }
     }
   }
+
+  // Handles cleanup of consumer group znode. Lifted with love from Kafka's
+  // ConsumerConsole.scala tryCleanupZookeeper()
```
tdas: Can  you just add one more line to explain what this code effectively does? And a link to the documentation/code of ConsumerConsole.scala (which is from Kafka I guess) would be great. These are just for future ease, so the people who are not that experienced with Kafka (sadly, me) can understand whats going on.

streaming/src/main/scala/spark/streaming/StreamingContext.scala
mcnamaras, line 216: Yes, absolutely could be and this was something I debated over.  In the end I felt changing it was the best option. My reasoning for doing so is-
1. cfa8e76 adds the functionality that people will mostly be looking for with kafka (which isn't easy to do with the existing interface).  Specifically it makes it easy to:
   - consume from start of topic (great for testing or seeding data)
   - pick up where consumer group left off
   - start consuming new events only (it was easy for us to overwhelm the block generator if a driver has been down for maintainance for an extended period of time. Though this was partly caused by the mem leak in the MessageHander)
2. The current implementation has a dependency on a private kafka class (ZookeeperConsumerConnector).
3. With cfa8e76, I don't see a real need to keep initialOffsets around from the user perspective.

That is my 2 cents at least =)  I will let you have the final say.  I am happy to tweak this further of course.

streaming/src/main/scala/spark/streaming/api/java/JavaStreamingContext.scala
mcnamaras, line 101: Good catch, will do!

streaming/src/main/scala/spark/streaming/StreamingContext.scala
tdas, line 216: I think I understand. Users are going either start from the beginning (which can be done using the autoffset.reset=smallest) and continue with new stuff (which is the default). The use cases where one has to set arbitrary initial offsets are rare. Am I correct in understanding this?

Assuming yes, can you give an example on how users who may have used the arbitrary initial offset can migrate their code to the new API? Is it even possible?

mcnamaras, line 216: Correct.  The default is to continue where you last left off.  There is also autoffset.reset=largest which lets you skip passed events that you haven't read (ie- when taking a consumer down for maintenance and bringing it back online, there could be hours worth of data queued up that you may not be interested in).

Here is some background on setting offsets using the high level consumer (which spark and most other people use):
https://groups.google.com/forum/?fromgroups=#!topic/kafka-dev/7n_l-O70mk0
http://grokbase.com/t/kafka/users/12b9bmsy7k/question-about-resetting-offsets-and-the-high-level-consumer

Long story story short, setting arbitrary offsets using the high level api isn't supported (which is why kafka had to be modified to use ZookeeperConsumerConnector in Spark).  In the second link Jun mentions that exposing that functionality would be post 0.8.

> can you give an example on how users who may have used the arbitrary initial offset can migrate their code to the new API? Is it even possible?

Since the high level consumer doesn't support setting arbitrary offsets, it shouldn't have been exposed in the first place.  Worst case, if someone still needed this functionality in spark, they should be able to copy the old code that was in spark and use that as a template.  Ie- have their spark driver connect as a consumer, call setOffets, and then close before consuming.  After that, they'd call kafkaStream and it would pick up where they told it to.

The "proper" way to allow arbitrary offsets in kafka currently is to use the lower level kafka API (called the SimpleConsumer, which is a misnomer because it is significantly more complex to use).  While the SimpleConsumer lets you manage offsets, there is a great deal that it doesn't do for you.  Integrating with zookeeper, manage saving/loading offsets, etc.. is all left up to you.

In my book the best bet is to stick with the high level consumer.  When arbitrary offsets are exposed beyond kafka 0.8 they should be able to be passed in as kafkaParams, so we won't have to worry about making any more API changes down the road.

That is probably WAY more than you ever wanted to know about Kafka =)

tdas, line 216: I guess you have convinced me. :)
I do know a bit about the high level and low level consumers though only conceptually. So I wanted to get a complete idea. Thanks for providing that.

tdas:
Just update the java streaming context and the comments and we are ready to go.

mcnamaras:
Should all be in there now.  Thanks TD!

mateiz:
Thanks for the fix! I've merged it in.

PR #530, opened 2013-03-15
Add a log4j compile dependency to fix build in IntelliJ

mbautin:
Also rename parent project to spark-parent (otherwise it shows up as
"parent" in IntelliJ, which is very confusing).


mateiz:
Thanks for the fix, though it's worked fine for me in IntelliJ -- I guess something changed in a new version?

PR #533, opened 2013-03-18
Fix performance issues and bugs in YARN branch.

mridulm:
Individual commits have more details on what is fixed in each of them (hopefully, they are logically grouped). These were the fixes we did to fix some severe bugs, make slaves more responsive as cluster increases (and there are more than 8 reducers !), reduce latency of computation, improve rack awareness, etc.

For some of our ML jobs, the time came down from 20+ mins to under 5 mins without any user code change (YMMV ofcourse !).

More conservatively, after all changes (bug fixes, etc) applied, the scheduling changes itself gave us a 2x-2.5x improvement on a small 40 container test (which increases as number of nodes increase) - by default, the scheduling changes are disabled - and are explicitly enabled by flags.

I am probably missing out other details, please query in case there are any issues, comments or notes !

### DIFF
--- a/core/src/main/scala/spark/deploy/yarn/Client.scala
+++ b/core/src/main/scala/spark/deploy/yarn/Client.scala
@@ -222,16 +222,12 @@ class Client(conf: Configuration, args: ClientArguments) extends Logging {
     // Add Xmx for am memory
     JAVA_OPTS += "-Xmx" + amMemory + "m "
 
-    if (env.isDefinedAt("SPARK_JAVA_OPTS")) {
-      JAVA_OPTS += env("SPARK_JAVA_OPTS") + " "
-    }
     // Commenting it out for now - so that people can refer to the properties if required. Remove it once cpuset version is pushed out.
     // The context is, default gc for server class machines end up using all cores to do gc - hence if there are multiple containers in same
     // node, spark gc effects all other containers performance (which can also be other spark containers)
     // Instead of using this, rely on cpusets by YARN to enforce spark behaves 'properly' in multi-tenant environments. Not sure how default java gc behaves if it is
     // limited to subset of cores on a node.
-/*
-    else {
+    if (env.isDefinedAt("SPARK_USE_CONC_INCR_GC") && java.lang.Boolean.parseBoolean(env("SPARK_USE_CONC_INCR_GC"))) {
       // In our expts, using (default) throughput collector has severe perf ramnifications in multi-tenant machines
       JAVA_OPTS += " -XX:+UseConcMarkSweepGC "
       JAVA_OPTS += " -XX:+CMSIncrementalMode "
@@ -239,7 +235,9 @@ class Client(conf: Configuration, args: ClientArguments) extends Logging {
       JAVA_OPTS += " -XX:CMSIncrementalDutyCycleMin=0 "
       JAVA_OPTS += " -XX:CMSIncrementalDutyCycle=10 "
     }
-*/
+    if (env.isDefinedAt("SPARK_JAVA_OPTS")) {
+      JAVA_OPTS += env("SPARK_JAVA_OPTS") + " "
+    }
 
     // Command for the ApplicationMaster
     val commands = List[String]("java " +
--- a/core/src/main/scala/spark/deploy/yarn/Client.scala
+++ b/core/src/main/scala/spark/deploy/yarn/Client.scala
@@ -132,8 +132,9 @@ class Client(conf: Configuration, args: ClientArguments) extends Logging {
     // Add them as local resources to the AM
     val fs = FileSystem.get(conf)
     Map("spark.jar" -> System.getenv("SPARK_JAR"), "app.jar" -> args.userJar, "log4j.properties" -> System.getenv("SPARK_LOG4J_CONF"))
-    .foreach { case(destName, localPath) =>
-      if (null != localPath) {
+    .foreach { case(destName, _localPath) =>
+      val localPath: String = if (null != _localPath) _localPath.trim() else ""
+      if (! localPath.isEmpty()) {
         val src = new Path(localPath)
         val pathSuffix = appName + "/" + appId.getId() + destName
         val dst = new Path(fs.getHomeDirectory(), pathSuffix)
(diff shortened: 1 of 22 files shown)


core/src/main/scala/spark/util/AkkaUtils.scala
```diff
@@ -27,6 +27,9 @@ private[spark] object AkkaUtils {
     val akkaBatchSize = System.getProperty("spark.akka.batchSize", "15").toInt
     val akkaTimeout = System.getProperty("spark.akka.timeout", "20").toInt
     val akkaFrameSize = System.getProperty("spark.akka.frameSize", "10").toInt
+    // 10 seconds is the default akka timeout, but in a cluster, we need higher by default.
+    val akkaWriteTimeout = System.getProperty("spark.akka.writeTimeout", if (Utils.isYarnMode()) "30" else "10").toInt
```
mateiz: Why not make this 30 for everything? Issues might happen on schedulers other than YARN too.

core/src/main/scala/spark/deploy/yarn/Client.scala
```diff
@@ -222,9 +225,25 @@ class Client(conf: Configuration, args: ClientArguments) extends Logging {
...
+/*
+    else {
+      // In our expts, using (default) throughput collector has severe perf ramnifications in multi-tenant machines
+      JAVA_OPTS += " -XX:+UseConcMarkSweepGC "
+      JAVA_OPTS += " -XX:+CMSIncrementalMode "
+      JAVA_OPTS += " -XX:+CMSIncrementalPacing "
+      JAVA_OPTS += " -XX:CMSIncrementalDutyCycleMin=0 "
+      JAVA_OPTS += " -XX:CMSIncrementalDutyCycle=10 "
+    }
+*/
```
mateiz: Instead of leaving this commented out, it might be good to add an option for it and an if statement. By the way, I believe it's also possible to control the number of threads for the throughput GC with -XX:ParallelGCThreads=N. Might be worth testing out.

mateiz:
Hey Mridul,

Apologies for taking a while to get to this, but it looks quite good overall. I'm curious to see whether your ConnectionManager fixes help in the master branch too (the patch at https://github.com/mesos/spark/pull/522 still requires the use of threaded IO for the best performance, and has problems with NIO; I'd really prefer to use NIO if possible, unless there's a Linux/JVM bug causing the problem).

One other question -- is the rack resolver something we could call on non-YARN backends too, or is it an API provided by YARN? For example, does it use a config file with a rack->host mapping script, similar to Hadoop?

core/src/main/scala/spark/util/AkkaUtils.scala
mridulm: One of our main goals was to ensure that there is minimal side-effects of our changes : since we are unable to test it in all modes unfortunately : mesos, windows, etc.
Bumping from 10 to 30 means it takes more time to recover in case the slave is "hung".

The 10 seconds write timeout was causing issues particularly before the ConnectionManager and Connection changes. I really do not like fudging with timeouts since they are (usually) just procrastinating the problem instead of fixing it - but in a yarn cluster, given mulch-tennency, 10 was a bit strict at times (particularly with gc pauses when worker is in heavy load - proc or IO).

So to answer your query, we can make it 30 for all - if you think that is fine in other modes too.

mateiz: I agree, but I guess I've found this timeout small myself due to garbage collection, so that's why I'd increase it by default. It's confusing to have a special case for YARN. Mesos and even standalone clusters will also have multitenancy.

core/src/main/scala/spark/deploy/yarn/Client.scala
mridulm: I was planning on using ParallelGCThreads after yarn starts using cpusets - otherwise I was not sure of the side effects (performance, correctness) : of parallel gc not occupying all threads I mean.

I asked around hadoop team, and they seem to be ok with not doing anything for gc specifically - assuming things even out in the end for the cluster I guess - and since yarn was adding cpusets anyway, other than perf hit of more gc threads than core (which we can control with ParallelGCThreads), there wont be any other issue

Since this was, I assumed, not a long term change - I left it for documentation : we can ofcourse remove it !

mridulm:
Hi Matei,

  The changes to Connection/ConnectionManager addresses some logical bugs, and makes it MT (as part of which, it makes the ConnectionManager more responsive).
As such, I think it is fairly isolated - so it might be possible to simply replace the two files in trunk with these two - if you want to test it : I have not tried it though :-)

I did look at 522, since I was interested in it for the potential performance improvements promised !
The changes to Connection/ConnectionManager does improve spark performance in our tests quiet drastically (particularly jobs which are shuffle bound) - but I am not sure how well it compares against the netty server - offhand, I dont expect it to beat netty; though imo the performance should be pretty close.

On other hand, the netty server route means we close the door to non-file based shuffle forever, which is a future optimization direction we were was hoping to explore.

The current/default implementation of rack resolver in yarn invokes a shell script which returns the rack given a hostname. Though this can be overridden in conf to something else. Take a look at init() in 
org.apache.hadoop.yarn.util.RackResolver [1], etc.
It is part of yarn common package and as such I dont think it has any other ties to hadoop/yarn and should be fairly self-contained.
IIRC even dfs uses it for block placement ...

So without having investigated deeply, I think it should be fine to use it elsewhere too : assuming it is an environment where Configuration can be initialized properly ofcourse.

Regards,
Mridul

[1] http://grepcode.com/file/repository.cloudera.com/content/repositories/releases/org.apache.hadoop/hadoop-yarn-common/0.23.1-cdh4.0.0b2/org/apache/hadoop/yarn/util/RackResolver.java

mateiz:
The problem with 522 though was that their NIO performance was _similar_ to our old ConnectionManager; they had to go to synchronous IO with one thread per connection to go faster. So if your change makes things better while remaining asynchronous, that could be very nice. We'll have to compare with what they saw though, because I believe they saw more than 2x on their hardware.

Is there an equivalent of the RackResolver in Hadoop 1? I guess not in the public API... but we might be able to use the internal class that the JobTracker uses.

core/src/main/scala/spark/util/AkkaUtils.scala
mridulm: Sounds good, I will remove the "if (Utils.isYarnMode())" switch then, and make it 30 for all cases !

mridulm:
To give more details about our test:
It was part of a logistic regression execution run on 30 containers, where reduce was using all 30 containers.
As the number of reducers went up, for non trivial amounts of data, we saw a few things which lead us to investigate Connection/ConnectionManager :

a) Accept timeout - sockets connections were not being accepted fast enough : which lead to socket timeout - resulting in other failures.

b) Data was not being sent even when it was available.
This was due to not registering event with selector - the selector was not being woken up for all interest changes.

c) finishConnect need not complete even if socket is connectable.

d) multiple outbound connections can (and were) getting created to same slave due to MT bug.

Also, the performance of read, write could be better since it was all being attempted in the main selector thread (though low, there was still performance hit to actually fetch and send the data/read and fire events for read data).

In addition, there was a bug I introduced as part of previous yarn change - where we were incorrectly not identifying the paired connection as part of closing a SendingConnection (no measurable performance hit, but still a bug nonetheless !).

The changes to Connection/ConnectionManager resolves all of these - and as part of that, we saw a nice speed bump too : the initial aim was not to work on the performance issues actually !
So for us the changes were important enough even disregarding the performance impact.

About RackResolver in hadoop 1 ... it should exist - since both mapreduce job submission and dfs block placement rely on it. Unfortunately, I am not very familiar with the codebase to point you to it unfortunately.
I am not sure if hadoop1 is just relying on yarn commons or they have it replicated.

mridulm:
Hi Matei,

  Please let me know in case there are any further changes to be made !

Thanks,
Mridul

mateiz:
Looks good, thanks. Now the key thing is to port this to master. Let me know how that goes and if there's any way we can help with part of it.

PR #534, opened 2013-03-18
Remove try/catch block that can't be hit.

stephenh:
Minor change, but like the code from the StandaloneExecutorBackend.preStart that was fixed a few weeks ago, all of these method calls are async, and so will not cause connection/etc. exceptions.

Granted, this is not causing any bugs, but it can lure the programmer into thinking the calls are sync (it did me anyway) and so lead to seemingly-surprising behavior.

It seems simpler to just take it out, so it mirrors what StandaloneExecutorBackend.preStart looks like now.


mateiz:
Good idea, thanks.

PR #535, opened 2013-03-18
porting of repl to scala-2.10

ScrapCodes:


mateiz:
Woah, cool! Let me take a close look at this but this is awesome.

repl/src/main/scala/spark/repl/SparkExprTyper.scala
```diff
@@ -0,0 +1,109 @@
...
+
+import scala.tools.nsc._
+import scala.tools.nsc.interpreter._
+
+import scala.reflect.internal.util.BatchSourceFile
+import scala.tools.nsc.ast.parser.Tokens.EOF
+
+import spark.Logging
+
+trait SparkExprTyper extends Logging {
```
stephenh, line 16: What are the changes to this file vs. the ExprTyper that (I assume) this was copy/pasted from?

Seems like there should be a few "// NOTE: Changed for Spark because of X or Y" comments in here that explain to future maintainers how this is exactly different from Scala's ExprTyper, and why.

Right now the only line that looks remotely related to Spark (as in it mentions a Spark type) is the "repl: SparkIMain"...was that really the only change? I have no idea...just can't tell without looking at the Scala source what has changed.

ScrapCodes, line 16: The file names are congruent with the names given at scala 2.10.x branch with only difference being the prefix Spark. It is always possible to see diff, so, felt writing NOTE, unless there is a specific reason for doing something that other maintainer shall be careful about, is redundant. There are comments in the code mentioned otherwise. Please take a closer look and let me know if you find something specific that can be improved. 

About above file, you are right. Only reason, I had to copy paste it is to solve a type mismatch.

emchristiansen:
Can we do this without copy-pasting a bunch of code from the Scala REPL? My sense is copy-pasting now will lead to more pain later when we upgrade to 2.11.

mateiz:
Unfortunately there's no way, because the Scala REPL doesn't provide a plugin architecture. I've been trying to get Scala folks to change this, and it might happen eventually, or we might be able to use macros to avoid having to modify the interpreter.

repl/src/main/scala/spark/repl/SparkImports.scala
```diff
@@ -144,10 +147,10 @@ trait SparkImports {
     // add code for a new object to hold some imports
     def addWrapper() {
       val impname = nme.INTERPRETER_IMPORT_WRAPPER
-      code append "class %sC extends Serializable {\n".format(impname)
-      trailingBraces append "}\nval " + impname + " = new " + impname + "C;\n"
+      code append "object %s {\n".format(impname)
+      trailingBraces append "}\n"
```
mateiz: Hey, so this change was actually important (see the MATEI: comment below)

repl/src/main/scala/spark/repl/SparkImports.scala
```diff
@@ -159,36 +162,25 @@ trait SparkImports {
...
-            /*if (currentImps contains imv)*/
-            val imvName = imv.toString
-            if (currentImps exists (_.toString == imvName)) addWrapper()
-        
-            val objName = req.lineRep.readPath
-            val valName = "$VAL" + newValId();
-            code.append("val " + valName + " = " + objName + ".INSTANCE;\n")
-            code.append("import " + valName + req.accessPath + ".`" + imv + "`;\n")
-
-            //code append ("import %s\n" format (req fullPath imv))
```
mateiz, line 191: This change was also important. The problem with Scala's default "wrapper" object for creating one singleton object per line is that the initialization code for that variable needs to run on the workers. For example, say that you type this:

```
scala> var x = Console.readLine()

scala> val data = sc.parallelize(1 to 10).map(_ + x)

scala> data.collect()
```

What Scala will do by default is compile one object for each line, like this:

```
object Line1 { var x = Console.readLine() }
```

And then call it inside subsequent lines:

```
object Line3 {
  import Line1.x
  val data = sc.parallelize(1 to 10).map(_ + x)
}
```

Unfortunately, since Line1 is a global singleton object, this is a static method call. This means that on the worker node running the map(), Scala will try to run Line1's initializer again, which tries to do Console.readLine. This is bad! Try the current patch against a cluster instead of local mode and you'll see this problem.

My fix in these two methods was to change object Line1 to something like this:

```
class Line1 { var x = Console.readLine() }; object Line1 { val INSTANCE = new Line1 }
```

And then to reference the instance of type Line1 directly inside further lines that use it, instead of going through a static call (code in this method). Luckily, it looks like this aspect of the REPL hasn't changed much in 2.10, so hopefully these changes, plus a few of the others that had MATEI: comments, can be added back.

mateiz:
So I tried this out and added a comment about an important change that was missing. It doesn't affect local mode unfortunately, but does affect clusters (we should really add a test for it with the "local standalone cluster" mode we recently added). Apart from that, one other thing I noticed is that the scala> prompt comes up before the SparkContext is initialized, leading to somewhat confusing output. Can you just join that initialization thread before printing the prompt? I believe I had to change this in 2.9 too.

Anyway, thanks for taking a look at this!

repl/src/main/scala/spark/repl/SparkIMain.scala
```diff
@@ -608,115 +735,170 @@ class SparkIMain(val settings: Settings, protected val out: PrintWriter) extends
...
+     * append to objectName to access anything bound by request.
+     */
+    val SparkComputedImports(importsPreamble, importsTrailer, accessPath) =
       importsCode(referencedNames.toSet)
 
     /** Code to access a variable with the specified name */
     def fullPath(vname: String) = (
-      //lineRep.readPath + accessPath + ".`%s`".format(vname)
-      lineRep.readPath + ".INSTANCE" + accessPath + ".`%s`".format(vname)
+      lineRep.readPath + accessPath + ".`%s`".format(vname)
```
mateiz: This change was also related to the class/object stuff. Basically, look at the diff of this file on GitHub -- a lot of the changes from 2.10 seem to be due to my edits and not due to the 2.9 -> 2.10 upgrade.

repl/src/main/scala/spark/repl/SparkImports.scala
mateiz, line 191: By the way, I should also add that if you have any other suggestion to solve this problem, that would be great. For example, I don't know whether this would work, but it might be simpler to leave `Line1` as an `object` as above, but do the importing like this instead:

```
val Line1instance = Line1
import LIne1instance.x
```

The point is that we want the call to `x` in the closure in Line3 to reference the local _instance_ of Line1, rather than doing a static method call which will go to the worker node's instance of that object. I don't know whether this particular proposal works though.

ScrapCodes:
@mateiz  Looks like the jenkins is not a bot.And I am sorry for a long silence. Anyways was not able to work at this, was caught up in something else. Will be able to see it this week, hopefully. Hope we have some time before we release 0.7.1?

mateiz:
No, Jenkins actually is a bot. It's kind of funny because you can reply certain things to it and it will do stuff. Anyway, if you do have time to fix it, that would be great. I think it's pretty close to done.

ScrapCodes:
@mateiz Incorporated your feedback, It was a great help. Thanks for detailed feedback. To solve that problem, I have simply copied your approach. I have one question regarding. [SparkIMain:872](https://github.com/mesos/spark/blob/master/repl/src/main/scala/spark/repl/SparkIMain.scala#L872) and [SparkIMain:884](https://github.com/mesos/spark/blob/master/repl/src/main/scala/spark/repl/SparkIMain.scala#L884)'s comment. Why were they needed, If they are still needed. I mean code concerning them have changed a bit. Have to take a deeper look though.  

Everything as far as I tested worked fine on a stand alone cluster. Take a look at changes.

mateiz:
It's true that those two might not be needed. If the code works on a cluster, that's good. Do you think it's ready to go or are you still working on it? (I was wondering whether you want to change those back to the 2.10 version).

SparkIMain:884 is probably the more important one by the way; it was there because the REPL used to run each command in a separate thread, and we relied on having a thread-local variable set to do stuff.

repl/src/main/scala/spark/repl/SparkILoopInit.scala
```diff
@@ -0,0 +1,143 @@
...
+        asyncMessage("[info] total init time: " + elapsed() + " s.")
+        withLock(initLoopCondition.signal())
+      }
+    }
+  }
+
+ def initializeSpark() {
+    intp.beQuietDuring {
+      command("""
+         @transient lazy val sc = spark.repl.Main.interp.createSparkContext();
```
mateiz: Any reason why this is a lazy val? It makes the interpreter print "Spark context available as sc" before it actually starts Spark.

ScrapCodes: It can speed up loading of repl. `sc` will only materialize the first time it is used. It made sense to do so because as repl loads it spawns many threads and deferred initialization of spark context will only speed up loading of repl. That was the only intention, if it hurts in some other way perhaps I am not aware - other than printing of that message (which I suppose, we can work around !). Your thoughts @mateiz ? Should we make it a normal val ?

ScrapCodes:
It is ready except :
There was a mention about a test using a standalone cluster, which is missing at the moment. So was wondering if there is already one written somewhere which I can probably adapt. Writing a new one would be fun too.

ScrapCodes:
It is ready to go, I suppose.

repl/src/main/scala/spark/repl/SparkILoopInit.scala
mateiz: I don't like this approach because it means that if you have any error in how you configured the shell (e.g. you gave the wrong master URL), you won't see it until the first time you use sc. I'd prefer if we tried to initialize it before gave you a shell, like in the current version.

ScrapCodes:
@mateiz Did you get a chance to look at this?

mateiz:
Looks great; I've merged it in. Thanks a lot for contributing this! This means we'll have Scala 2.10 support in Spark 0.8.

PR #536, opened 2013-03-19
CoalescedRDD for many partitions

sasurfer:
... original and the new coalesced RDD.

The limitation was in the range that Scala.Int can represent.


mateiz:
Ah, so how many partitions did you have? I guess this is fine but you won't be able to have more than 2 billion partitions in total today. (I guess you had something like sqrt(2 billion) though?)

mateiz:
Actually never mind, just saw your mailing list post. That's a lot of partitions! Thanks for the fix though.

sasurfer:
Sorry if I write you now but the time zone...
I was trying to do a wordcount on data bigger and bigger to see which parameters I have to adjust in my cluster for big data analysis with spark. So I tried the classic wordcount on 12 TB of books divided in 97 files. The number of map/task generated are 394790 but I wanted to compare with hadoop in which I set 13580 mappers. In the CoalescedRDD when it computes the range there is a multiplication which leads to a number greater than 2^31-1 so I needed temporarily a Long there.

PR #538, opened 2013-03-20
Added mapSideCombine flag to CoGroupedRDD. Added unit test for CoGroupedRDD.

rxin:


mateiz:
Looks good, but why not always have mapSideCombine on? Is there any advantage to creating one ArrayBuffer per element?

rxin:
When mapSideCombine is off, each value is simply the value object iself (not the array buffer). This would be helpful when the number of keys is large to avoid an extra pass of creating a lot of arraybuffers.

mateiz:
Ah, got it; that does make sense.

PR #539, opened 2013-03-22
Bugfix: WorkerWebUI must respect workDirPath from Worker

cgrothaus:
... or else serving the stdout and stderr log files from worker web ui is broken.


mateiz:
Good catch, thanks.

cgrothaus:
You're welcome.

Am 30.03.2013 um 06:16 schrieb Matei Zaharia notifications@github.com:

PR #540, opened 2013-03-22
Shutdown StandaloneExecutorBackend properly in non-error case by sending...

cgrothaus:
... a termination message instead of calling process.destroy.

This gives the executor process the chance to properly clean up after itself. E.g. spark.Utils.createTempDir() registers a JVM shutdown hook per created dir. These get never executed if we terminate the executor via process.destroy, so I think sending a termination message is the nicer way, at least in the good case. In an error case - e.g. the executor not reacting anymore - we can still stick with the destroy method.

### DIFF
--- a/core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
+++ b/core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
@@ -52,7 +52,7 @@ private[spark] class ExecutorRunner(
 
   /** Stop this executor runner, including killing the process it launched */
   def kill() {
-    if (workerThread != null) {
+    if (workerThread != null && workerThread.isAlive) {
       workerThread.interrupt()
       workerThread = null
       if (process != null) {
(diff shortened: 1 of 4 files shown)


mridulm:
Process.destroy results in a SIGTERM - which will cause all shutdown
hooks to be executed on the child process.
What exactly is the issue you are observing ? Since we are seeing all
temp directories cleaned up, might be something else ?

Regards,
Mridul

cgrothaus:
Ahh, now I know where our difference is: Linux vs. Windows.
Your point regarding may be true on Linux, I don't know how Process.destroy is implemented there. But on Windows, there are no process signals like in Linux. Process.destroy simply kills the process immediatly (like SIGKILL or kill -9 in Linux, I think). Because of that I already opened #488 (launch executor from worker directly with "java" and without the "run" script). I don't know another way than the here proposed termination message to cleanly shutdown the executor process on Windows systems.

Regards,
Christoph

mateiz:
Hey Christoph,

Sorry again for the long time getting to this, but I haven't forgotten! I've merged this fix into branch-0.7, so it will be in the 0.7.3. release. (I added some small changes on top to send the StopExecutor message from the Worker instead of the Driver: 05d377397ec5d48945e20b05b720ebbbae91e4d5, but it's the same mechanism). In addition, I want to merge #488 into the master branch and Spark 0.8. The reason I don't want to do it for 0.7 is because it's a bigger change that might break some peoples' setup if they depend on spark-env.sh being read on each worker machine when new executors are spawned. Thanks for contributing these!

PR #541, opened 2013-03-22
Add a shuffle parameter to coalesce.

stephenh:
This is useful for when you want just 1 output file (part-00000) but
still up the upstream RDD to be computed in parallel.

### DIFF
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -240,7 +240,7 @@ abstract class RDD[T: ClassManifest](
    */
   def coalesce(numPartitions: Int, shuffle: Boolean = false): RDD[T] = {
     if (shuffle) {
-      // include a shuffle step so that are upstream tasks are still distributed
+      // include a shuffle step so that our upstream tasks are still distributed
       new CoalescedRDD(new ShuffledRDD(map(x => (x, null)), new HashPartitioner(numPartitions)), numPartitions).keys
     } else {
       new CoalescedRDD(this, numPartitions)
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -31,6 +31,7 @@ import spark.rdd.MapPartitionsRDD
 import spark.rdd.MapPartitionsWithIndexRDD
 import spark.rdd.PipedRDD
 import spark.rdd.SampledRDD
+import spark.rdd.ShuffledRDD
 import spark.rdd.SubtractedRDD
 import spark.rdd.UnionRDD
 import spark.rdd.ZippedRDD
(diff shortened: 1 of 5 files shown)


core/src/main/scala/spark/RDD.scala
```diff
@@ -237,7 +238,14 @@ abstract class RDD[T: ClassManifest](
   /**
    * Return a new RDD that is reduced into `numPartitions` partitions.
    */
-  def coalesce(numPartitions: Int): RDD[T] = new CoalescedRDD(this, numPartitions)
+  def coalesce(numPartitions: Int, shuffle: Boolean = false): RDD[T] = {
+    if (shuffle) {
+      // include a shuffle step so that are upstream tasks are still distributed
```
markhamstra: ...so that our upstream tasks...

stephenh: Thanks, fixed.

core/src/main/scala/spark/api/java/JavaDoubleRDD.scala
```diff
@@ -58,6 +58,12 @@ class JavaDoubleRDD(val srdd: RDD[scala.Double]) extends JavaRDDLike[Double, Jav
   def coalesce(numPartitions: Int): JavaDoubleRDD = fromRDD(srdd.coalesce(numPartitions))
 
   /**
+   * Return a new RDD that is reduced into `numPartitions` partitions.
+   */
+  def coalesce(numPartitions: Int, shuffle: java.lang.Boolean): JavaDoubleRDD =
```
mateiz: Why not make the shuffle parameter just a Bool? It will translate into a primitive boolean in Java.

core/src/test/scala/spark/RDDSuite.scala
```diff
@@ -154,6 +154,10 @@ class RDDSuite extends FunSuite with LocalSparkContext {
     assert(coalesced4.collect().toList === (1 to 10).toList)
     assert(coalesced4.glom().collect().map(_.toList).toList ===
       (1 to 10).map(x => List(x)).toList)
+
+    // we can optionally shuffle to keep the upstream parallel
+    val coalesced5 = data.coalesce(1, shuffle = true)
+    assert(coalesced5.dependencies.head.rdd.dependencies.head.rdd.asInstanceOf[ShuffledRDD[_, _]] ne null)
```
mateiz: Here I would use != instead of ne because I think it will mean the same thing

mateiz:
Hey, this is a good idea, but unfortunately the PR doesn't merge cleanly -- is it based on an older branch? I don't think we changed too much in this code recently.

core/src/main/scala/spark/api/java/JavaDoubleRDD.scala
stephenh: Ah, cool, didn't know that--I'll change to Boolean.

stephenh:
Just an import conflict in RDDSuite; I'll merge in master.

mateiz:
Thanks for the fix.

PR #542, opened 2013-03-24
Surfacing decoders on KafkaInputDStream

mcnamaras:
This pull request builds off of https://github.com/mesos/spark/pull/527

The existing kafkaStream interface allows type parameterizing, but currently only works when the type is a String.  This pull request fixes it so that types other than String may be used.

There are two kafkaStream methods in StreamingContext:
- One assumes a DStream[String] and is easy to get up and running with (it's used in the KafkaWordCount example)
- The other kafkaStream method remains the same but adds on a parameterized type for the kafka serializer.  ie-  `ssc.kafkaStream[String, kafka.serializer.StringDecoder](kafkaParams, topics, storageLevel)`

I was unable to find a clever way to expose it using a decoder as a single parameterized type.

So, there you have it.  Input is always welcome!


mcnamaras:
I am going to rebase this commit and re-submit under a new pull request.

PR #543, opened 2013-03-26
Re-enable deprecation warnings and fix deprecated warning.

holdenk:


mateiz:
Good idea, thanks. I think there were a bunch of them in our Hadoop API usage that maybe go away if you use Hadoop 2.

PR #544, opened 2013-03-27
Support for spawning multiple workers on every slave

kalpit:
- Support for spawning multiple worker instances/processes on every slave machine
- upgraded sbt version, sbt plugins and some library dependencies to latest stable version

The bash scripts are not backward compatible in the sense that a cluster that was started with older version of scripts will have to be brought down with the same version of scripts.


mateiz:
Thanks for the fixes. I've merged this manually to deal with some conflicts.

PR #545, opened 2013-03-28
Don't use deprecated Application in example

ash211:
As of 2.9.0 extending from Application is not recommended

http://www.scala-lang.org/api/2.9.3/index.html#scala.Application


mateiz:
Thanks for the fix.

PR #547, opened 2013-03-28
Move streaming test initialization into 'before' blocks

jey:
This fix is needed for the `spark-streaming` tests to pass under Maven.


mateiz:
Thanks for the fix.

PR #548, opened 2013-03-31
Fixed broken filter in getWritableClass[T]

markhamstra:
Here's one that we were getting away with.  When filtering based on m.getReturnType().toString, the value that we need to test for is actually "class java.lang.Object", which means that the first part of the compound predicate wasn't doing anything.  Neither does the second part usually, but I left that as is.  The result was that the filter accomplished nothing, both apply methods were present in the resulting Array, and we were just lucky that the correct apply method was always at the zero index.  Or at least that was the case.  In Scala 2.10, the result of getDeclaredMethods is ordered differently than in 2.9.x, which means that getWritableClass was returning Object from the else clause instead of an appropriate Writable.


mateiz:
Good catch, thanks!

PR #549, opened 2013-04-01
Bump development version to 0.8.0

jey:
This patch bumps the version number of the master branch to "0.8.0-SNAPSHOT" so that bugfix releases of the 0.7.x branch can be made, and to comply with the [Semantic Versioning](http://semver.org/) convention.

### DIFF
--- a/bagel/pom.xml
+++ b/bagel/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.spark-project</groupId>
     <artifactId>spark-parent</artifactId>
-    <version>0.7.1-SNAPSHOT</version>
+    <version>0.8.0-SNAPSHOT</version>
     <relativePath>../pom.xml</relativePath>
   </parent>
 
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.spark-project</groupId>
     <artifactId>spark-parent</artifactId>
-    <version>0.7.1-SNAPSHOT</version>
+    <version>0.8.0-SNAPSHOT</version>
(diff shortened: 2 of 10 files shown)


pwendell:
Thanks for catching this - I think maybe it was just a typo when Matei
bumped versions to 0.7.1 in the master branch. If you look at the last
major release, we bumped the master version to

0.7.0-SNAPSHOT

Just after we cut 0.6.0

https://github.com/mesos/spark/commit/0967e71a004d2b723d1e66e3dbc62a12726bbb94

mateiz:
As discussed on the mailing list, I don't think we need to do this yet. Eventually we will need to create a branch-0.7 which contains fixes to 0.7.x and change master to 0.8, but right now the fixes in master are also good for 0.7.1.

markhamstra:
My only problem with your approach, Matei, is with new functionality added to 0.7.1-SNAPSHOT leaking out into an eventual 0.7.1 release before it is fully mature.  For example, in catching up with https://github.com/mesos/spark/pull/456, it strikes me that there is enough overlap with the recently merged https://github.com/mesos/spark/pull/510 (mapWith, filterWith, etc.) that I'll likely do a refactoring/combining/extending of these efforts later this week.  That's fine as long as there haven't been any severe bugs requiring a 0.7.1 release yet, but we probably should try to segregate new API functionality from purely bugfix changes so that a bugfix release can easily be made without leaking new public API that we may still want to rework.

massie:
Jenkins, test this please.

mateiz:
I've merged this in and created a branch-0.7 for further development of 0.7. Thanks Jey!

PR #551, opened 2013-04-02
Python bugfixes

jey:
A couple of fixes in PySpark -- one is a failing test which implicitly assumed RDD.parallelize preserved ordering, the other is a missing type coercion of a command-line argument from string to int.


mateiz:
Thanks, good catch on both!

PR #552, opened 2013-04-03
Bumping version for Twitter Algebird to latest

MLnick:
Originally when I included Algebird in the examples project, the latest version didn't target JDK 6, hence we went back to 0.1.8.

In https://github.com/twitter/algebird/pull/137 this is fixed so we can bump the version up.

This is nice since the latest Algebird has some other useful stuff like TopK counting via PriorityQueues, for example.

### DIFF
--- a/project/SparkBuild.scala
+++ b/project/SparkBuild.scala
@@ -157,7 +157,7 @@ object SparkBuild extends Build {
 
   def examplesSettings = sharedSettings ++ Seq(
     name := "spark-examples",
-    libraryDependencies ++= Seq("com.twitter" % "algebird-core_2.9.2" % "0.1.8")
+    libraryDependencies ++= Seq("com.twitter" % "algebird-core_2.9.2" % "0.1.11")
   )
 
   def bagelSettings = sharedSettings ++ Seq(name := "spark-bagel")
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -22,7 +22,7 @@
     <dependency>
       <groupId>com.twitter</groupId>
       <artifactId>algebird-core_2.9.2</artifactId>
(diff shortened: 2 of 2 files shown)


MLnick:
@tdas 
I've tested locally that this works by compiling using JDK6:

```
$ JAVA_HOME=/Library/Java/JavaVirtualMachines/1.6.0_29-b11-402.jdk/Contents/Home sbt clean package
...
[success] Total time: 197 s, completed Apr 3, 2013 1:08:22 PM
```

And then running again using JDK6 `JAVA_HOME=/Library/Java/JavaVirtualMachines/1.6.0_29-b11402.jdk/Contents/Home ./run spark.streaming.examples.TwitterAlgebirdCMS local[2] XXX YYY`. Previously this failed with Major/Minor version exceptions.

project/SparkBuild.scala
```diff
@@ -157,7 +157,7 @@ object SparkBuild extends Build {
 
   def examplesSettings = sharedSettings ++ Seq(
     name := "spark-examples",
-    libraryDependencies ++= Seq("com.twitter" % "algebird-core_2.9.2" % "0.1.8")
+    libraryDependencies ++= Seq("com.twitter" % "algebird-core_2.9.2" % "0.1.11")
```
markhamstra, line 160: Prefer "com.twitter" %% "algebird-core" % "0.1.11" so that the correct version is used based on the Scala version -- e.g. algebird-core_2.10 will be used when building with Scala 2.10.x.

mateiz:
I'll fix the Scala version thing manually. Thanks for the patch.

MLnick:
@mateiz sorry, been swamped and didn't get a chance to make that change! Thanks for that

PR #553, opened 2013-04-03
SPARK-724 - Have Akka logging enabled by default for standalone daemons

pwendell:
See the JIRA for more details.

I was only able to test the bash version (don't have Windows)
so maybe check over that the syntax is correct there.

### DIFF
--- a/run
+++ b/run
@@ -22,7 +22,7 @@ fi
 # values for that; it doesn't need a lot
 if [ "$1" = "spark.deploy.master.Master" -o "$1" = "spark.deploy.worker.Worker" ]; then
   SPARK_MEM=${SPARK_DAEMON_MEMORY:-512m}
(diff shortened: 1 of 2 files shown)


run
```diff
@@ -22,6 +22,7 @@ fi
 # values for that; it doesn't need a lot
 if [ "$1" = "spark.deploy.master.Master" -o "$1" = "spark.deploy.worker.Worker" ]; then
   SPARK_MEM=${SPARK_DAEMON_MEMORY:-512m}
+  SPARK_DAEMON_JAVA_OPTS+="-Dspark.akka.logLifecycleEvents=true"
```
mateiz: This will break if they have set SPARK_DAEMON_JAVA_OPTS already because there's no space before the -D

pwendell: Updated, good catch,

mateiz:
Thanks Patrick.

PR #554, opened 2013-04-04
Fixes SPARK-723 - Update build to Scala 2.9.3

andyk:
This builds fine for me.

Though the test suite (i.e., running `sbt/sbt test`) doesn't seem to complete on my Macbook Air, though it doesn't report failed tests (as described in this bug https://spark-project.atlassian.net/browse/SPARK-716)

Fixes https://spark-project.atlassian.net/browse/SPARK-723

### DIFF
--- a/project/SparkBuild.scala
+++ b/project/SparkBuild.scala
@@ -36,7 +36,7 @@ object SparkBuild extends Build {
   def sharedSettings = Defaults.defaultSettings ++ Seq(
     organization := "org.spark-project",
     version := "0.7.1-SNAPSHOT",
-    scalaVersion := "2.9.2",
+    scalaVersion := "2.9.3",
     scalacOptions := Seq("-unchecked", "-optimize", "-deprecation"),
     unmanagedJars in Compile <<= baseDirectory map { base => (base / "lib" ** "*.jar").classpath },
     retrieveManaged := true,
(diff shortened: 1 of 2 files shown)


mateiz:
Did the test suite work before? It shouldn't be hanging. The one small thing to watch out for is that you need to configure conf/spark-env.sh for it to run, since it tries to spawn processes using that.

Note that SBT itself sometimes hangs if you run a lot of commands in it. It's best to do sbt package, then exit it, then do sbt test.

massie:
Jenkins, test this please and add to whitelist.

massie:
Jenkins, test this please.

andyk:
@mateiz I figured out what was causing my tests to hang (and eventually to silently crash), I had this line in my spark-env.sh file:

export SPARK_MEM=3G

After removing that line the tests pass without a problem.

project/SparkBuild.scala
```diff
@@ -140,7 +140,7 @@ object SparkBuild extends Build {
       "colt" % "colt" % "1.2.0",
       "cc.spray" % "spray-can" % "1.0-M2.1",
       "cc.spray" % "spray-server" % "1.0-M2.1",
-      "cc.spray" %%  "spray-json" % "1.1.1",
+      "cc.spray" % "spray-json_2.9.2" % "1.1.1",
```
mateiz, line 143: So they didn't have this for 2.9.3? Maybe they have a later version of spray-json.

andyk, line 143: From their readme: "The latest release is 1.2.3 and is built against Scala 2.9.2 as well as Scala 2.10.0." So I guess we could upgrade it but I don't think it'll help us in this case with the 2.9.2 issue.

mateiz:
Great, thanks a lot. I think that with this, we should do a 0.7.1 release soon.

PR #555, opened 2013-04-06
Retry failed ssh commands in the ec2 python script.

holdenk:
This is especially useful when starting up when the hosts may not have yet come on-line but can be useful at other points for people with flakey connections


mateiz:
This looks like a good thing to fix, but I'd prefer doing it just inside the "ssh" method instead of using a decorator, since we won't use this in a lot of other places, and it will make the code easier to follow. Just put a loop in there.

holdenk:
Sounds good, I'll do that.

holdenk:
Cool I've switched it to just be a simple loop in the script and removed the retry decorator code.

mateiz:
Thanks, that looks great.

PR #559, opened 2013-04-10
Uniform whitespace across scala examples

ash211:


mateiz:
Cool, thanks for the fix!

PR #562, opened 2013-04-11
Add table for system properties

ash211:
Still needs information about what each property does, but having all the properties in one place is a start.


JoshRosen:
We already have a table of system properties in https://github.com/mesos/spark/blob/master/docs/configuration.md

PR #563, opened 2013-04-11
Optimize PySpark worker invocation

jey:
This is an initial patch to reduce the startup time of PySpark workers by preforking the workers. Code review would be appreciated. Currently passes all PySpark tests, and additional tests will be added.

### DIFF
--- a/core/src/main/scala/spark/api/python/PythonRDD.scala
+++ b/core/src/main/scala/spark/api/python/PythonRDD.scala
@@ -47,6 +47,7 @@ private[spark] class PythonRDD[T: ClassManifest](
       currentEnvVars.put(variable, value)
     }
 
+    val startTime = System.currentTimeMillis
     val proc = pb.start()
     val env = SparkEnv.get
 
@@ -108,6 +109,17 @@ private[spark] class PythonRDD[T: ClassManifest](
               val obj = new Array[Byte](length)
               stream.readFully(obj)
               obj
+            case -3 =>
+              // Timing data from child
+              val bootTime = stream.readLong()
+              val initTime = stream.readLong()
+              val finishTime = stream.readLong()
+              val boot = bootTime - startTime
+              val init = initTime - bootTime
+              val finish = finishTime - initTime
+              val total = finishTime - startTime
+              logInfo("Times: total = %s, boot = %s, init = %s, finish = %s".format(total, boot, init, finish))
+              read
             case -2 =>
               // Signals that an exception has been thrown in python
               val exLength = stream.readInt()
--- a/core/src/main/scala/spark/api/python/PythonRDD.scala
+++ b/core/src/main/scala/spark/api/python/PythonRDD.scala
@@ -38,8 +38,8 @@ private[spark] class PythonRDD[T: ClassManifest](
 
   override def compute(split: Partition, context: TaskContext): Iterator[Array[Byte]] = {
     val startTime = System.currentTimeMillis
-    val worker = SparkEnv.get.getPythonWorker(pythonExec, envVars.toMap).create
     val env = SparkEnv.get
+    val worker = env.createPythonWorker(pythonExec, envVars.toMap)
 
     // Start a thread to feed the process input from our parent's iterator
     new Thread("stdin writer for " + pythonExec) {
--- a/core/src/main/scala/spark/api/python/PythonRDD.scala
+++ b/core/src/main/scala/spark/api/python/PythonRDD.scala
@@ -2,10 +2,9 @@ package spark.api.python
 
 import java.io._
 import java.net._
-import java.util.{List => JList, ArrayList => JArrayList, Collections}
+import java.util.{List => JList, ArrayList => JArrayList, Map => JMap, Collections}
 
 import scala.collection.JavaConversions._
-import scala.io.Source
 
 import spark.api.java.{JavaSparkContext, JavaPairRDD, JavaRDD}
 import spark.broadcast.Broadcast
@@ -16,7 +15,7 @@ import spark.rdd.PipedRDD
 private[spark] class PythonRDD[T: ClassManifest](
     parent: RDD[T],
     command: Seq[String],
-    envVars: java.util.Map[String, String],
+    envVars: JMap[String, String],
     preservePartitoning: Boolean,
     pythonExec: String,
     broadcastVars: JList[Broadcast[Array[Byte]]],
@@ -25,7 +24,7 @@ private[spark] class PythonRDD[T: ClassManifest](
 
   // Similar to Runtime.exec(), if we are given a single string, split it into words
   // using a standard StringTokenizer (i.e. by spaces)
-  def this(parent: RDD[T], command: String, envVars: java.util.Map[String, String],
+  def this(parent: RDD[T], command: String, envVars: JMap[String, String],
       preservePartitoning: Boolean, pythonExec: String,
       broadcastVars: JList[Broadcast[Array[Byte]]],
       accumulator: Accumulator[JList[Array[Byte]]]) =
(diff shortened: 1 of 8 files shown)


core/src/main/scala/spark/api/python/PythonRDD.scala
```diff
@@ -88,16 +70,16 @@ private[spark] class PythonRDD[T: ClassManifest](
...
     }.start()
 
     // Return an iterator that read lines from the process's stdout
-    val stream = new DataInputStream(proc.getInputStream)
+    val stream = new DataInputStream(worker.getInputStream)
     return new Iterator[Array[Byte]] {
       def next(): Array[Byte] = {
         val obj = _nextObj
-        _nextObj = read()
+        if(hasNext) _nextObj = read()
```
mateiz: Change this to

```
if (hasNext) {
 _nextObj = read()
}
```

to match the project code style (in general it's good to have if blocks delimited with braces in case we expand them)

core/src/main/scala/spark/api/python/PythonRDD.scala
```diff
@@ -307,3 +287,82 @@ class PythonAccumulatorParam(@transient serverHost: String, serverPort: Int)
     }
   }
 }
+
+// NOTE: There is only one worker daemon per JVM
+private object PythonWorker extends Logging {
```
mateiz: Unfortunately this will make it tricky to use PySpark in unit tests that share the same JVM (though I guess our unit tests for it currently all spawn new JVMs because they're in Python). Could you make this a class and add a member of it in SparkEnv instead? Then you can do SparkEnv.get.pythonWorker in your RDD.

JoshRosen:
Disclaimer: I haven't hacked on preforking servers before, so my question might be naive.

I think that our goal here is to avoid fork()ing Python workers from a JVM that has a large heap ([SPARK-671](https://spark-project.atlassian.net/browse/SPARK-671)).  I like this strategy of lazily launching a persistent Python process per worker and delegating forks() to it.

The top-level daemon (one per worker) forks `POOLSIZE` workers, which share a listening socket.  When PythonRDD wants to spawn a worker Python process, it opens a connection to this socket.  This connection is handled by one of the workers (whichever one grab's the OS's internal mutex that guards accepts() on this socket).  Rather than directly servicing this request (e.g. running `worker_main`) and re-entering the idle worker pool, this worker process forks a child to perform the work and re-enters the pool immediately.

What does this second level of indirection buy us compared to having the daemon manage the listening socket and fork all of the workers itself?  It seems like we don't benefit from having multiple processes competing for the right to spawn workers, since everything is serialized at the accept() mutex anyways?

I could see benefit from a pool model where the daemon preforks worker processes that run `worker_main` directly.  Workers would notify the daemon when they exited the pool, allowing the daemon to know when it should begin accepting() on the listening socket and forking new workers.

core/src/main/scala/spark/api/python/PythonRDD.scala
jey: Sounds good, will do

jey:
In this case, preforking is just a rather small optimization over accept-and-fork. When N requests arrive nearly all at once in the accept-and-fork scenario, all the requests would have to be serially handled one-by-one in the manager, which has to accept, fork, accept, fork, etc. N times. By preforking the workers and having them all sit in an accept() call, we avoid the serial bottleneck in the manager with all of the context switches, system calls, and fork()s that involves. You're absolutely right that we're losing some of this performance gain since the children call fork() themselves at just about the worst possible time: when a client has arrived! This is just going to thrash the kernel's scheduler and memory bus further as extra processes get created and pages get copied for a process that'll do nothing but enter accept().

core/src/main/scala/spark/api/python/PythonRDD.scala
```diff
@@ -307,3 +287,82 @@ class PythonAccumulatorParam(@transient serverHost: String, serverPort: Int)
...
+
+      // Important: don't close child's stdin (daemon.getOutputStream) so it can correctly
+      // detect our disappearance.
+    }
+  }
+
+  private def stopDaemon() {
+    synchronized {
+      // Request shutdown of existing daemon by sending SIGTERM
+      if(daemon != null) daemon.destroy
```
mateiz: Here's another place to put the if body in its own block

JoshRosen:
What's the status of this pull request?  Does the code need to be cleaned up or does the preforking need to be re-worked so that children don't immediately call fork() and thrash the scheduler?

Were you able to reproduce the performance issues that I noticed when running the AMP Camp exercises?  Does this patch offer a significant performance improvement for those exercises?

jey:
Hey Josh, my plan was to just submit it with a few cleanups, since I think the `fork()` right after `accept()` shouldn't be too big of an issue right now. I've pushed my current code to the `python-optimization` branch with most (or all?) of the suggested changes applied, but I haven't yet reviewed it closely to make sure there aren't any other changes I should make before considering it a final submission.

JoshRosen:
Does Jenkins run the PySpark tests?

jey:
Don't think so; we should probably make it so that both SBT and Maven will
run the PySpark tests during Spark unit test execution

jey:
I've updated my `python-optimization` branch and it should be ready to be merged. The one caveat is that I haven't tested this under Windows, and I doubt it'll work well under Windows as-is -- we might want to do some kind of fallback to the old scheme if running on a non-POSIX system?

mateiz:
Jey, did you run the PySpark tests for this?

jey:
Yes, and they've been updated to include additional tests for the new daemon

jey:
Jenkins, test this please

core/src/main/scala/spark/api/python/PythonWorkerFactory.scala
```diff
@@ -0,0 +1,95 @@
...
+private[spark] class PythonWorkerFactory(pythonExec: String, envVars: Map[String, String])
+    extends Logging {
+  var daemon: Process = null
+  val daemonHost = InetAddress.getByAddress(Array(127, 0, 0, 1))
+  var daemonPort: Int = 0
+
+  def create(): Socket = {
+    synchronized {
+      // Start the daemon if it hasn't been started
+      startDaemon
```
mateiz: Small code style things: for calls to methods that have side-effects, use parentheses; so for example `startDaemon()`.

core/src/main/scala/spark/api/python/PythonWorkerFactory.scala
```diff
@@ -0,0 +1,95 @@
...
+          startDaemon
+          new Socket(daemonHost, daemonPort)
+        }
+        case e => throw e
+      }
+    }
+  }
+
+  def stop() {
+    stopDaemon
```
mateiz: Same thing, use stopDaemon()

mateiz:
Hey Jey,

One issue I notice in testing this is that Python exceptions are no longer reported to the master properly. There is still a line that writes them to stdout in worker.py, but it seems that the "worker process crashed" caused by the raise after that is noticed first, and that's what propagates back to the master. Try it out -- this is the code I typed into pyspark:

```
data = sc.parallelize(range(1, 10))
data.map(lambda x: x/2).collect()
data.map(lambda x: x/0).collect()
```

core/src/main/scala/spark/api/python/PythonRDD.scala
```diff
@@ -108,6 +95,17 @@ private[spark] class PythonRDD[T: ClassManifest](
...
+              // Timing data from worker
+              val bootTime = stream.readLong()
+              val initTime = stream.readLong()
+              val finishTime = stream.readLong()
+              val boot = bootTime - startTime
+              val init = initTime - bootTime
+              val finish = finishTime - initTime
+              val total = finishTime - startTime
+              logInfo("Times: total = %s, boot = %s, init = %s, finish = %s".format(total, boot, init, finish))
+              read
```
mateiz, line 108: read()

core/src/main/scala/spark/api/python/PythonRDD.scala
```diff
@@ -115,23 +113,21 @@ private[spark] class PythonRDD[T: ClassManifest](
...
               throw new PythonException(new String(obj))
             case -1 =>
-              // We've finished the data section of the output, but we can still read some
-              // accumulator updates; let's do that, breaking when we get EOFException
-              while (true) {
-                val len2 = stream.readInt()
+              // We've finished the data section of the output, but we can still
+              // read some accumulator updates; let's do that, breaking when we
+              // get a negative length record.
+              var len2 = stream.readInt
```
mateiz: readInt()

jey:
Good catch, I'll look into that

jey:
Hi Matei, I've fixed the issues you pointed out. Exceptions are once again being correctly reported to the master, and I added parens for function calls with side-effects.

mateiz:
Thanks Jey, this looks great! I've merged it in.

mateiz:
FYI, I've also merged this into branch 0.7.

lucktroy:
As shown in https://issues.apache.org/jira/browse/SPARK-2764, at default, The per-java-worker daemon.py process launches a **numCores-sized** pool of subprocesses. 

For example, it will lauch 16 subprocesses if there are 16 cores at a worker. However, in our application, more subprocesses are expected. How can I set the number? Like 32 in each worker. 

Thanks.

JoshRosen:
@lucktroy In https://github.com/apache/spark/pull/1680, linked from that pull request that you mentioned, we removed the intermediate, numCores-sized layer of subprocesses, since the only thing that the intermediate pool of processes did was to process requests to `fork()` the processes that actually performed work.  Now, the `fork()` is performed directly from `daemon.py`.  This removes some parallelism from the actual forking stage, but the impact of should have been mitigated by other performance improvements, such as https://github.com/apache/spark/pull/2259 to re-use the Python worker processes across tasks.

The number of actual computation-performing processes is controlled at a higher level: there's one PySpark `worker.py` process per executor thread.  Do you want to run more Python threads than cores?  If so, you should configure your executors / workers to use more threads.

lucktroy:
@JoshRosen Thanks for the detailed info. In our application, we'd like to more python threads than cores. However, in the current application, more executors / workers may not be helpful. The dataset is not very large. To accelerate it, we propose a hybrid method which combined model-parallelism and data-parallelism. To implement it, we also use python multi-threads. So it can run many jobs at the same time. 

For example, there are 8 nodes (16 cores/node), we set spark.default.parallelism=8 because it is enough. When we used data-parallelism (1 job each time) only, each node would launches 1 pyspark daemon. 

We also tested with the hybrid method, if 8 jobs are launched at the same time, each node launches 8 pyspark daemons. 

As same, if 64 jobs are launched at the same time, it is expected that there are 64 pyspark daemons at each node. However, it is only up to 16 pyspark daemons. When viewed with `top`, we found that  there are many idle CPU resources.

PR #565, opened 2013-04-12
Update wording of section on RDD operations in quick start guide in docs

andyk:
This minor change to the heading matches the contents of the section a bit better IMO. Noticed as I work on the Screencast that includes these instructions.


mateiz:
Looks good.

PR #566, opened 2013-04-13
Kafka decoders on KafkaInputDStream

mcnamaras:
This pull request allows setting the kafka decoder on kafkaStreams, so that types other than String may be supported.

Using a type other than String causes a ClassCastException. Here is some background on the issue:
https://groups.google.com/forum/?fromgroups=#!searchin/spark-users/vanderborgh/spark-users/pOZ9tOa5yRk/9ofOdkboNqUJ

There are now two kafkaStream methods in StreamingContext:
1. One assumes a DStream[String] and is easy to get up and running with (it's used in the KafkaWordCount example)
2. The other kafkaStream method remains the same but adds a method parameter for the kafka serializer. ie- `ssc.kafkaStream[String](kafkaParams, topics, new StringDecoder, storageLevel)`


tdas:
Thanks for implementing this! However, I have a high level question about the technique used to implement this. From the way the StringDecoder object is used ...

val topicMessageStreams = consumerConnector.createMessageStreams(topics, new StringDecoder())

... it seems, all that is necessary is a decoder object. Why cant this object passed as a parameter in the KafkaInputDStream's constructor and ssc.kafkaStream . The advantage of this method maybe is that for both types of kafkaStream methods, the decoder parameter can have "new StringDecoder()" as the default. What do you think?

mcnamaras:
It could absolutely be passed in as a param.  The "new StringDecoder()" won't be able to be set as a default param on the kafkaStream that accepts kafkaParams though.  This is because the decoder is actually a Decoder[T].  I used a <: Decoder[_] because it did not like me using T before the class params.  So the other downside with my implementation is less type safety.

After thinking about it some, I agree.  It will be best to make the decoder a param. It will still need to be specified on one of the kafkaStreams, but in the end it will have better type-safety.

This was a great call out, I will have a fix pushed up soon. Thank you TD!

mcnamaras:
Ok, fixed now.  Tomorrow morning I will get this running on our standalone cluster for good measure.

streaming/src/main/scala/spark/streaming/StreamingContext.scala
```diff
@@ -204,17 +205,18 @@ class StreamingContext private (
    * @param groupId The group id for this consumer.
    * @param topics Map of (topic_name -> numPartitions) to consume. Each partition is consumed
    * in its own thread.
+   * @param decoder Kafka message decoder.
```
tdas, line 208: The decoder param does not exist in this method.

mcnamaras, line 208: Good catch, I will fix that

mcnamaras:
I am rebasing this contribution so that it is cleaner.  I will open a new pull request for it.

PR #567, opened 2013-04-15
Fixing count() in Spark Streaming

mcnamaras:
Previously when there was no data count() would not emit anything downstream.  This pull request fixes it so that under this condition 0 will be emitted.


mcnamaras:
One question I just had about my implementation- Will the ConstantInputDStream RDD stick around, or will it get removed by the cleaner?

tdas:
The ConstantInputDStream RDD will get removed. So it should not be used for anything other than simple testing.

tdas:
I think its better to implement this fix using transform operation rather than using constantinputdstream. In the transform operation, each mapped RDD can be unified with an special RDD created using makeRDD(...) . Note that the special RDD must be created every time and not reused, other wise it will have the same problem as the constantinputdstream.

mcnamaras:
That makes sense. I will look into reformulating it with a transform.  Thanks!

tdas:
Any update on this?

mcnamaras:
Hey TD!  So sorry, I have been sidetracked with travel and other things.  I will try to get to this very soon.

mcnamaras:
Please let me know how this looks.  Thanks!

pwendell:
Jenkins, test this please

PR #569, opened 2013-04-16
Fix committing problem when saving as text file on hdfs 

viirya:
Tested on hadoop 2.0.3-alpha.

When saving RDD as text file on hdfs, file output will be cleaned after committing. It is needed to call commitJob after commitTask to fix this problem.

Also updated building file to hadoop 2.0.3-alpha.

### DIFF
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -480,7 +480,9 @@ abstract class RDD[T: ClassManifest](@transient sc: SparkContext) extends Serial
    */
   def saveAsTextFile(path: String) {
     this.map(x => (NullWritable.get(), new Text(x.toString)))
-      .saveAsNewAPIHadoopFile[NewTextOutputFormat[NullWritable, Text]](path)
+      .saveAsHadoopFile[TextOutputFormat[NullWritable, Text]](path)
+      //or call new API with:
+      //.saveAsNewAPIHadoopFile[NewTextOutputFormat[NullWritable, Text]](path)
   }
 
   /**
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -25,6 +25,8 @@ import org.apache.hadoop.mapred.OutputFormat
 import org.apache.hadoop.mapred.SequenceFileOutputFormat
 import org.apache.hadoop.mapred.TextOutputFormat
 
+import org.apache.hadoop.mapreduce.lib.output.{TextOutputFormat => NewTextOutputFormat} 
+
 import it.unimi.dsi.fastutil.objects.{Object2LongOpenHashMap => OLMap}
 
 import spark.partial.BoundedDouble
@@ -478,7 +480,7 @@ abstract class RDD[T: ClassManifest](@transient sc: SparkContext) extends Serial
    */
   def saveAsTextFile(path: String) {
     this.map(x => (NullWritable.get(), new Text(x.toString)))
-      .saveAsHadoopFile[TextOutputFormat[NullWritable, Text]](path)
+      .saveAsNewAPIHadoopFile[NewTextOutputFormat[NullWritable, Text]](path)
   }
 
   /**
--- a/examples/src/main/scala/spark/examples/SparkPi.scala
+++ b/examples/src/main/scala/spark/examples/SparkPi.scala
@@ -4,6 +4,9 @@ import scala.math.random
 import spark._
 import SparkContext._
 
+import org.apache.hadoop.conf.Configuration
+
+
 object SparkPi {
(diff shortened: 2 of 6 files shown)


examples/src/main/scala/spark/examples/SparkPi.scala
```diff
@@ -19,6 +19,7 @@ object SparkPi {
       if (x*x + y*y < 1) 1 else 0
     }.reduce(_ + _)
     println("Pi is roughly " + 4.0 * count / n)
+    spark.makeRDD(List("Pi is roughly " + 4.0 * count / n), 1).saveAsTextFile("hdfs://localhost/tmp/pi")
```
rxin: Does the addition of this line require the user to have a local hdfs working?

viirya: Yes. It is inappropriate to hard coding the location in the code. I just committed a change for that to obtain namenode address from hadoop configuration. Thanks.

mridulm:
Note that on changing to 2.0.3, you lose ability to work with cloudera clusters : due to a protocol change.
But the basic change of commitJob is valid even without change of version (and something we just ported to trunk as part of merging yarn branch to trunk).

Did not realize others were using yarn branch already :-)

core/src/main/scala/spark/RDD.scala
```diff
@@ -478,7 +480,7 @@ abstract class RDD[T: ClassManifest](@transient sc: SparkContext) extends Serial
    */
   def saveAsTextFile(path: String) {
     this.map(x => (NullWritable.get(), new Text(x.toString)))
-      .saveAsHadoopFile[TextOutputFormat[NullWritable, Text]](path)
+      .saveAsNewAPIHadoopFile[NewTextOutputFormat[NullWritable, Text]](path)
```
mridulm: A similar commitJob api exists even for mapred api - addition of that would be sufficient (instead of moving to new hadoop api) : it would also mean fixing that spark impl

viirya:
Since this is a yarn branch and it is building against hadoop version 2.0.1-alpha in latest SparkBuild.scala, I suppose that using hadoop new api should be ok if you cannot run it on non-yarn hadoop cluster?

Using old mapred api is fine. I noticed that there is a commitJob api too, but actually it is only a wrapper of new hadoop api. For you concerns, I committed another change that uses old mapred api but adds commitJob calling too. That fixes same problem but with less change in api level.

examples/src/main/scala/spark/examples/SparkPi.scala
```diff
@@ -19,6 +22,8 @@ object SparkPi {
       if (x*x + y*y < 1) 1 else 0
     }.reduce(_ + _)
     println("Pi is roughly " + 4.0 * count / n)
+    val namenode =  new Configuration().get("fs.default.name")
+    spark.makeRDD(List("Pi is roughly " + 4.0 * count / n), 1).saveAsTextFile(namenode + "/tmp/pi")
```
mateiz: Please don't change the SparkPi example to test this. SparkPi is meant to be a small example to show people how to use the API. You can add another example that saves to a file, or WordCount might already do it.

viirya: Undo the change to SparkPi and add another example.

I do not find any examples using saveAsTextFile to save files on hdfs. So as I ran SparkPi example, I can not see any output to indicate it works or not. Using println to print on stdout will require users to examine the output across the machines of a cluster. A simple result file on hdfs will be better to see if it works.

mateiz:
That println in SparkPi was just in the main program, so you'll see the result printed in the shell where you run that program (./run spark.examples.SparkPi whatever). Same shell that says "connecting to master" and things like that.

mridulm:
Hi Matei,

  The problem is that the stdout/stderr, when run in yarn mode, goes to
application master's (AM) stdout/stderr files : and is not available at
./run script's stdout/stderr.
This is different from all other modes of execution I think.

Regards,
Mridul

viirya:
OK. I think the problem is the difference between running the example on yarn and other settings. By running the example on yarn-based hadoop cluster (./run spark.deploy.yarn.Client), SparkPi will be submitted to run across hadoop machines. Its stdout/stderr will be the stdout/stderr files of those machines.

mridulm:
Please note that yarn branch has been finally merged with master manually ... we should probably remove the yarn branch.

mateiz:
Thanks for contributing this, but it's actually been implemented in master as part of the YARN pull request (#575), so I'm going to close this request. Let me know if you think the implementation there is missing something (you can click "Files" on the pull request to see the changes made).

viirya:
Hi Mateiz,

Great to hear that and thanks for informing.

PR #570, opened 2013-04-16
Increase ReservedCodeCacheSize for sbt

jey:
Increasing the CodeCache size so that an `sbt/sbt clean package test` reliably finishes with all tests passing (none timing out) and finishes quickly in 10m instead of lingering for >20m.

Please add to `branch-0.7`.

### DIFF
--- a/sbt/sbt
+++ b/sbt/sbt
@@ -5,4 +5,4 @@ if [ "$MESOS_HOME" != "" ]; then
 fi
 export SPARK_HOME=$(cd "$(dirname $0)/.."; pwd)
 export SPARK_TESTING=1  # To put test classes on classpath
-java -Xmx1200M -XX:MaxPermSize=250m $EXTRA_ARGS -jar $SPARK_HOME/sbt/sbt-launch-*.jar "$@"
+java -Xmx1200m -XX:MaxPermSize=250m -XX:ReservedCodeCacheSize=128m $EXTRA_ARGS -jar $SPARK_HOME/sbt/sbt-launch-*.jar "$@"


mateiz:
Cool; does this remove the need for the Runtime.halt change?

jey:
Not as far as I know, but it's conceivable that the sporadic deadlock becomes harder to reproduce after this change. Also, `Runtime.halt` seems like the right way to simulate a crash, instead of cleanly exiting via `System.exit` (but the presence of a deadlock is generally worrying too).

mateiz:
I see; was that deadlock in the forked JVMs or in the main one that SBT runs? I just don't like using Runtime.halt if it can be avoided, since we do use shutdown hooks to clean up some temporary files. For example, another fix might be to try a newer Akka version, if its cleanup was the problem.

jey:
The deadlock is in the forked JVMs. (BTW, I can send you full stacktraces of all threads of all four JVMs.) I agree that cleaning up temporary files etc is important, but we probably also want to have testing for unclean crashes, so maybe we should write the temporary files to some temporary directory, then have the main process (the SBT one) delete the whole temp dir after running the test that uncleanly creates temp files?

markhamstra:
After this PR, 'sbt/sbt clean test' consistently fails for me with exhausted PermGen.  At least for Unix, the manual install instructions for SBT (http://www.scala-sbt.org/release/docs/Getting-Started/Setup.html#id1) specify a larger -XX:MaxPermSize=384M -- and this is sufficient for me to successfully 'clean test'.  Additionally, homebrew has for quite a while supported a ~/.sbtconfig file (http://davethecodemonkey.blogspot.com/2010/12/homebrew-now-has-improved-sbt-launcher.html).  Perhaps we need to make some effort not to clobber SBT options that have already been set elsewhere in a system -- although such system-wide configuration doesn't appear to be standardized....

jey:
I'm running into similar problems too. I've mostly been using `-Xmx4096m -XX:MaxPermSize=512m -XX:ReservedCodeCacheSize=128m` for my builds, but did have a successful full rebuild & test with the patch submitted in this PR. Maybe we can change the configuration in `sbt/sbt` to something like `-Xmx2048m -XX:MaxPermSize=512m -XX:ReservedCodeCacheSize=128m`?

mateiz:
Why did you submit the PR then? I'll fix it later today.

We can't require 4 GB for the build. I'd rather turn on forking in tests to avoid this problem, or tell people to run package and test separately.

jey:
Because it was NOT passing `clean package test` with the old parameters of `-Xmx1200M -XX:MaxPermSize=250m`. In the PR I submitted the minimal change that allowed `clean package test` to work on my machine, even though I have been using the overkill parameters of `-Xmx4096m -XX:MaxPermSize=512m -XX:ReservedCodeCacheSize=128m`. 

(Yesterday I ran into the same problems again when using the parameters as submitted in the PR.)

jey:
Also, I think heap size of 2G is sufficient (don't need 4G). I'm also kind of confused about why Java doesn't have cleverer heuristics about when to run the GC and instead requires such careful knob-twiddling.

mateiz:
Ah OK, so this is nondeterministic. I think the solution is to just run tests in a new JVM -- look up sbt's fork option if you want to try it. Those sizes are for permanent fixed memory space for things like classes, which Java assumes you won't have a lot of but sbt does.

PR #571, opened 2013-04-16
Surfacing decoders on KafkaInputDStream

mcnamaras:
This pull request allows setting the kafka decoder on kafkaStreams, so that types other than String may be supported.  It is a rebasing of https://github.com/mesos/spark/pull/566 that has been tidied up. 

Using a type other than String causes a ClassCastException. Here is some background on the issue:
https://groups.google.com/forum/?fromgroups=#!searchin/spark-users/vanderborgh/spark-users/pOZ9tOa5yRk/9ofOdkboNqUJ

There are now two kafkaStream methods in StreamingContext:
1. One assumes a DStream[String] and is easy to get up and running with (it's used in the KafkaWordCount example)
2. The other kafkaStream method remains the same but adds a method parameter for the kafka serializer. ie- `ssc.kafkaStream[String, StringDecoder](kafkaParams, topics, storageLevel)`

### DIFF
--- a/streaming/src/main/scala/spark/streaming/api/java/JavaStreamingContext.scala
+++ b/streaming/src/main/scala/spark/streaming/api/java/JavaStreamingContext.scala
@@ -105,7 +105,7 @@ class JavaStreamingContext(val ssc: StreamingContext) {
    * @param storageLevel RDD storage level. Defaults to memory-only
    */
   def kafkaStream[T, D <: kafka.serializer.Decoder[_]: Manifest](
-	kafkaParams: JMap[String, String],
+    kafkaParams: JMap[String, String],
     topics: JMap[String, JInt],
     storageLevel: StorageLevel)
   : JavaDStream[T] = {
--- a/streaming/src/main/scala/spark/streaming/api/java/JavaStreamingContext.scala
+++ b/streaming/src/main/scala/spark/streaming/api/java/JavaStreamingContext.scala
@@ -68,33 +68,50 @@ class JavaStreamingContext(val ssc: StreamingContext) {
    * @param topics Map of (topic_name -> numPartitions) to consume. Each partition is consumed
    * in its own thread.
    */
-  def kafkaStream[T](
+  def kafkaStream(
     zkQuorum: String,
     groupId: String,
     topics: JMap[String, JInt])
-  : JavaDStream[T] = {
-    implicit val cmt: ClassManifest[T] =
-      implicitly[ClassManifest[AnyRef]].asInstanceOf[ClassManifest[T]]
-    ssc.kafkaStream[T](zkQuorum, groupId, Map(topics.mapValues(_.intValue()).toSeq: _*))
+  : JavaDStream[String] = {
+    implicit val cmt: ClassManifest[String] =
+      implicitly[ClassManifest[AnyRef]].asInstanceOf[ClassManifest[String]]
+    ssc.kafkaStream(zkQuorum, groupId, Map(topics.mapValues(_.intValue()).toSeq: _*), StorageLevel.MEMORY_ONLY_SER_2)
+  }
+
+  /**
+   * Create an input stream that pulls messages form a Kafka Broker.
+   * @param zkQuorum Zookeper quorum (hostname:port,hostname:port,..).
+   * @param groupId The group id for this consumer.
+   * @param topics Map of (topic_name -> numPartitions) to consume. Each partition is consumed
+   * @param storageLevel RDD storage level. Defaults to memory-only
+   * in its own thread.
+   */
+  def kafkaStream(
+    zkQuorum: String,
+    groupId: String,
+    topics: JMap[String, JInt],
+    storageLevel: StorageLevel)
+  : JavaDStream[String] = {
+    implicit val cmt: ClassManifest[String] =
+      implicitly[ClassManifest[AnyRef]].asInstanceOf[ClassManifest[String]]
+    ssc.kafkaStream(zkQuorum, groupId, Map(topics.mapValues(_.intValue()).toSeq: _*), storageLevel)
   }
 
   /**
    * Create an input stream that pulls messages form a Kafka Broker.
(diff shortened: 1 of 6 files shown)


mcnamaras:
I should add that we have been using this on our standalone spark cluster all day and it's running great.  It also fixes a java test case in the streaming branch.

So, barring any further tweaks that anyone may have, I'd say it is ready.

mcnamaras:
Good catch, yes I will fix that and get the API tests slipped in. Thanks!

tdas:
Any update on this (and the other pull requests)?

mcnamaras:
Yes, I got stuck implementing the java api test case.  Maybe it is just my limited knowledge of java generics, but Java does not seem to like type parameters on methods and takes issue with:

`JavaDStream test3 = ssc.kafkaStream<String, StringDecoder>(kafkaParams, topics, StorageLevel.MEMORY_AND_DISK());`

I will finish up the test cases (minus test3 as show above) and we can take it from there.

tdas:
Can we please close this soon? This has been open for too long and causing problems in our automated testing. 
Same for the other pending pull requests to the streaming branch.

rxin:
@tdas  I don't think this is ready for merge, is it?

mcnamaras:
@rxin  We've been testing this code in our openstack environment for a few weeks now and it's working great.  I will try to get the api tests put in today, sorry I have been stretched on time recently.  Aside from the java API issue I mentioned above it should be good to go.

mcnamaras:
Ok, tests added!

Again, I couldn't figure out why the java api doesn't like the kafkaStream method where you specify a decoder on.  If someone smarter than I would like to take a crack at it, this is the test I tried in JavaAPISuite.java, and the error I would see: http://pastie.org/7827809

JoshRosen:
@Reinvigorate The [Java syntax](http://stackoverflow.com/questions/3012781/java-syntax-for-explicitly-specifiying-generic-arguments-in-method-calls) for calling a generic method with type parameters is a little weird.  The correct syntax is

``` java
ssc.<String, StringDecoder>kafkaStream(kafkaParams, topics, StorageLevel.MEMORY_AND_DISK());
```

with the type parameters appearing before the method's name.  This is what's causing the confusing `javac` error message.

However, the real reason why your test case isn't working is because of how Scala "context bounds" work.

It looks like

``` java
ssc.kafkaStream(kafkaParams, topics, StorageLevel.MEMORY_AND_DISK());
```

is intended to call the method

``` scala
  def kafkaStream[T, D <: kafka.serializer.Decoder[_]: Manifest](
    kafkaParams: JMap[String, String],
    topics: JMap[String, JInt],
    storageLevel: StorageLevel)
  : JavaDStream[T]
```

When this method is compiled into Java, the resulting method has the type signature

```
public spark.streaming.api.java.JavaDStream kafkaStream(java.util.Map, java.util.Map, spark.storage.StorageLevel, scala.reflect.Manifest);
```

(To see this, I ran  `javap -s -classpath streaming/target/scala-2.9.2/classes/  spark.streaming.api.java.JavaStreamingContext`).

What's happening is that the `: Manifest` part of `kafkaStream[T, D <: kafka.serializer.Decoder[_]: Manifest]` is specifying that there's an implicit parameter of type `Manifest[D]` (see [this answer on StackOverflow](http://stackoverflow.com/a/2982293/590203)).  So the above definition is of `kafkaStream` is actually equivalent to

``` scala
def kafkaStream[T, D <: kafka.serializer.Decoder[_]](
      kafkaParams: JMap[String, String],
    topics: JMap[String, JInt],
    storageLevel: StorageLevel)(implicit m: Manifest[D])
  : JavaDStream[T]
```

Java doesn't support implicit parameters, so the manifest must be explicitly passed when calling the Java method.

I'm not familiar enough with Spark Streaming to say what's the best way to fix this.  Hopefully we can eliminate the implicit parameter / context bound and have the Java API wrapper automatically supply the appropriate manifest to the Scala code.

streaming/src/main/scala/spark/streaming/api/java/JavaStreamingContext.scala
```diff
@@ -68,33 +68,50 @@ class JavaStreamingContext(val ssc: StreamingContext) {
...
+  /**
+   * Create an input stream that pulls messages form a Kafka Broker.
+   * @param kafkaParams Map of kafka configuration paramaters. See: http://kafka.apache.org/configuration.html
+   * @param topics Map of (topic_name -> numPartitions) to consume. Each partition is consumed
    * in its own thread.
    * @param storageLevel RDD storage level. Defaults to memory-only
    */
-  def kafkaStream[T](
+  def kafkaStream[T, D <: kafka.serializer.Decoder[_]: Manifest](
 	kafkaParams: JMap[String, String],
```
JoshRosen: When you remove the context bound, could you also fix this mis-indented line from the original code?

mcnamaras: Sure, no problem!

tdas:
Thanks Josh. This makes sense. And the Java syntax of putting the generic type earlier is very weird. A better way to define the Java function is to what we have done for JavaSparkContext.hadoopFile() 

https://github.com/mesos/spark/blob/master/core/src/main/scala/spark/api/java/JavaSparkContext.scala
Line 188

Basically, take the classes for data type (i.e. String) and the decoder type (i.e. StringDecoder) as parameters (just like JavaSparkContext.hadoopFile() takes the classes for key, value and inputformat.

mcnamaras:
@JoshRosen Thank you! This is much more clear to me and makes sense now.

@tdas Perfect! I will implement it in a similar fashion to hadoopFile()

mcnamaras:
The java API is now happily working and passing tests.  Please review and let me know if you spot anything else.  Otherwise, this should be ready to rock and roll.

mcnamaras:
Just a friendly reminder that this pull request along with https://github.com/mesos/spark/pull/572 https://github.com/mesos/spark/pull/567 are still floating out here.  I am happy to make further changes if needed. I would really like to see these make it in time for 0.8, as it fixes + greatly improves kafkaStream.  Thanks! =)

PR #573, opened 2013-04-18
add JobLogger to Spark

shimingfei:
add a new class named JobLogger
1.each SparkContext has one JobLogger, one folder is created for every JobLogger
2.JobLogger manages all history files of activeJobs running in that SparkContext,the file name is jobID
job history include:
1.additional information from outside
For example: queryplan from Shark
2.RDD Graph for each Job
The RDD graph is printed using a top-down approach
3.stage information and tasks' start/stop & shuffle information
gotten from TaskMetrics and DAGScheduler


rxin:
I will take a look at this today. 

Jenkins: this is ok to test

rxin:
Thanks, Mingfei.

Below are some high level comments:
1. Instead of adding JobLogger specific calls to DAGScheduler, why don't you extend SparkListener to add more callback APIs (for every recordTaskMetrics call in DAGScheduler, create a call back API in SparkListener)
2. With 1, JobLogger becomes an implementation of the SparkListener
3. JobLogger itself should also expose the information it collects so we can build a web UI for the JobLogger.

I have some detailed comments on the code that I will also post inline.

core/src/main/scala/spark/JobLogger.scala
```diff
@@ -0,0 +1,228 @@
...
+import java.util.Date
+import java.text.SimpleDateFormat
+import java.io.PrintWriter
+import java.io.File
+import java.io.FileNotFoundException
+import scala.collection.mutable.Map
+import scala.collection.mutable.HashMap
+import scala.collection.mutable.ListBuffer
+import spark.scheduler.Stage
+import scala.io.Source
```
rxin: For imports, sort them in the following order:
1. java packages
2. scala packages
3. everything else in alphabetical order.

Add a blank line for classes in different domain. 

Do this for other files - Spark code didn't strictly follow this but we are enforcing that now ...

core/src/main/scala/spark/JobLogger.scala
```diff
@@ -0,0 +1,228 @@
...
+import spark.executor.TaskMetrics
+import spark.scheduler.cluster.TaskInfo
+
+//it is used to record runtime information for each job, including RDD graph tasks start/stop and shuffle information 
+// and query plan information if there is any
+
+sealed trait JobLogger extends Logging {
```
rxin: With my proposed change (extend SparkListener and make JobLogger a specific implementation of SparkListener), you only need a single JobLogger implementation (no need to have JobLoggerOff).

core/src/main/scala/spark/JobLogger.scala
```diff
@@ -0,0 +1,228 @@
...
+  
+  def createLogWriter(jobID: Int): Unit
+  
+  def addStageIDToJobID(stages: List[Stage], jobID: Int): Unit
```
rxin: Try to combine addStageIDToJobID with addJobIDToStageIDs. They convey the same information and having two APIs make it more error-prone.

core/src/main/scala/spark/JobLogger.scala
```diff
@@ -0,0 +1,228 @@
...
+import scala.io.Source
+import spark.executor.TaskMetrics
+import spark.scheduler.cluster.TaskInfo
+
+//it is used to record runtime information for each job, including RDD graph tasks start/stop and shuffle information 
+// and query plan information if there is any
+
+sealed trait JobLogger extends Logging {
+  
+  def createLogWriter(jobID: Int): Unit
```
rxin: For functions that don't return anything, don't define the type as Unit. e.g. Just write

def createLogWriter(jobID: Int) {
 ...
}

core/src/main/scala/spark/RDD.scala
```diff
@@ -114,6 +114,14 @@ abstract class RDD[T: ClassManifest](
     name = _name
     this
   }
+  
+  /**generator of this RDD*/
+  var generator = Utils.getRddGenerator
```
rxin: Can we reuse "origin" for this? I noticed there are some minor differences in generator vs origin - but it would be great if we can merge the two since they are too similar.

core/src/main/scala/spark/RDD.scala
```diff
@@ -114,6 +114,14 @@ abstract class RDD[T: ClassManifest](
...
+  
+  /**reset generator*/
```
rxin, line 121: I don't see this used anywhere else. What is the purpose of this?

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -65,6 +65,10 @@ class SparkContext(
   // Ensure logging is initialized before we spawn any threads
   initLogging()
 
+  val jobLogger = JobLogger.init
+  
+  val addInfo = new DynamicVariable[String]("")
```
rxin: Don't see this set anywhere. What does it do?

core/src/main/scala/spark/Utils.scala
```diff
@@ -475,4 +475,21 @@ private object Utils extends Logging {
     }
     return false
   }
+  
+  def getRddGenerator = {//first class name out of Spark
```
rxin: As mentioned earlier, try to use getSparkCallsite for this... If you need to modify getSparkCallsite, that is fine.

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -507,10 +524,15 @@ class DAGScheduler(
                     activeJobs -= job
                     resultStageToJob -= stage
                     markStageAsFinished(stage)
+                    stageStatus = "RESULTSTAGE SUCCESS STAGE_ID:" + stage.id
```
rxin: This triggers only a single job logger log when the whole result task stage is completed. Is this intentional?

rxin:
I am done with my detailed comments.

Overall this looks like a great first step - we really appreciate you doing this!

There is one more comment about the SparkListener. Right now, DAGScheduler accepts a list of SparkListener's, and use a for loop to invoke the listeners. I think that can become expensive in large clusters. It would be great to have the DAGScheduler only accept one SparkListener, and if multiple SparkListener's are needed, we can create a composed SparkListener implementation that simply delegates to multiple SparkListener's. 

Again, thanks a lot for doing this.

core/src/main/scala/spark/JobLogger.scala
```diff
@@ -0,0 +1,228 @@
...
+  }
+  
+  //clean stages related to one job in stageIDToJobID
+  def cleanStageIDToJobID(jobID: Int): Unit = {
+    jobIDToStageIDs.get(jobID) match{
+      case Some(stageIDList) => for(stageid <- stageIDList) stageIDToJobID -= stageid 
+      case None =>
+    }
+  }
+
```
markhamstra: Avoid pattern matching on Option -- especially when you are doing nothing with the None case;

jobIDToStageIDs.get(jobID).map(_.foreach(stageid => stageIDToJobID -= stageid))

core/src/main/scala/spark/JobLogger.scala
```diff
@@ -0,0 +1,228 @@
...
+  
+  def recordTaskMetrics(stageID: Int, status: String, taskInfo: TaskInfo, taskMetrics: TaskMetrics): Unit = { }
+}
+
+class JobLoggerOn(val contextDirName: String) extends JobLogger {
+  private val logDir = { if (System.getenv("SPARK_LOG_DIR") != null)   System.getenv("SPARK_LOG_DIR")
+                         else   "/tmp/spark"
+                       } //get log directory setting default is /tmp/spark 
+  private var jobIDToPrintWriter = new HashMap[Int, PrintWriter] 
+  private var stageIDToJobID = new HashMap[Int, Int]
```
markhamstra: This is broken.  A stage can be associated with more than one active job.  See discussion of https://github.com/mesos/spark/pull/414

core/src/main/scala/spark/scheduler/TaskScheduler.scala
```diff
@@ -8,6 +8,9 @@ package spark.scheduler
  * the TaskSchedulerListener interface.
  */
 private[spark] trait TaskScheduler {
+  
+  val sc: SparkContext
+
```
markhamstra: Why?  Making the SparkContext part of the TaskScheduler seems to run contrary to the layered design of Spark.

markhamstra:
@rxin [3. JobLogger itself should also expose the information it collects so we can build a web UI for the JobLogger.](https://github.com/mesos/spark/pull/573#issuecomment-16611137)

That's a bit trickier.  Not only do you need to construct one or more data structures to which you must add the appropriate information when tasks/stages/jobs start/complete/fail/resubmit, but you also need to figure out when that information is no longer needed so that you can remove it from the appropriate data structure, avoiding that structure's unbounded growth.

I'm working on a substantial portion of that right now.

core/src/main/scala/spark/JobLogger.scala
shimingfei: The Map here only records the jobID which really "creates" the stage, the addStageIDToJobID function is called when the Stage is created. it means that the stage is executed in the activeJob that creates it.

core/src/main/scala/spark/scheduler/TaskScheduler.scala
shimingfei: it is used to make DAGScheduler be able to access jobLogger which is initialized  in SparkContext
that may be not necessary if modify the JobLogger according to Reynold's suggestion.

core/src/main/scala/spark/SparkContext.scala
shimingfei: this is used to pass some additional information from outside, for example query plan from Shark.
it can be replaced by localProperties in SparkContext adding by FairScheduler.

core/src/main/scala/spark/scheduler/DAGScheduler.scala
shimingfei: yes, it is used to identify the end of result stage which also means the end of job

core/src/main/scala/spark/RDD.scala
shimingfei, line 121: just like setName above, this is used to set generator intentionally

shimingfei:
Reynold and  markhamstra,
Thanks for your comments, I will modify the code according to the comments

pwendell:
Hi There,

I see that this still has jobLogger-specific calls in the DAGScheduler. As @rxin was saying earlier, it would be better to extend the SparkListener interface to have the necessary callbacks and have jobLogger be an instance of a SparkListener. 

This will be great to add since some other things we want to include (like a UI which reports job progress) will depend on these callbacks as well.

Were you planning to do this?
- Patrick

shimingfei:
Hi Patrick,

I am planning to do this, but now there seems to be more than one SparkListeners in Spark, and there is an interface in SparkContext to add an SparkListener to the ArrayBuffer which stores SparkListers. while JobLogger is designed to be one for each SparkContext and record job logs in that SparkContext.
I would like to know more details about the design of SparkListener interface, what is it planed to used for, and the right way to implement it?
Thanks
Mingfei

pwendell:
Hi Mighfei,

The listener in general is a way to expose state about Spark scheduling and workflow to other components, A SparkContext may have one or more SparkListeners. So the JobLogger would be a listener and there may both other listeners as well. Each JobLogger listens to exactly one SparkContext.

Our suggestion was that JobLogger implements SparkListener and we extend the SparkListener interface to have any methods that are currently missing.

This also means that JobLoggers should be a class rather than a Java object, so one can create multiple JobLoggers within a JVM - each tied to a single SparkContext. We didn't say that in our earlier review but it's a natural consequence of the model. Does that make sense? Since job information is not shared amongst multiple SparkContexts, this shouldn't be a problem.
- Patrick

shimingfei:
Hi Patrick
I  rewrite JobLogger and Implement SparkListener according to your suggestions

Mingfei

core/src/main/scala/spark/scheduler/JobLogger.scala
```diff
@@ -0,0 +1,209 @@
...
+import scala.collection.mutable.{Map, HashMap, ListBuffer}
+import scala.io.Source
+import spark._
+import spark.executor.TaskMetrics
+import spark.scheduler.cluster.TaskInfo
+
+//it is used to record runtime information for each job, including RDD graph tasks start/stop shuffle information 
+// and query plan information if there is any
+class JobLogger(val contextDirName: String) extends SparkListener with Logging{
+  private val logDir =  if (System.getenv("SPARK_LOG_DIR") != null)   System.getenv("SPARK_LOG_DIR")
```
pwendell: This should be broken into multiple lines (no braces is okay). We use the scala styleguide: http://docs.scala-lang.org/style/control-structures.html

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -476,7 +485,7 @@ class DAGScheduler(
   private def handleTaskCompletion(event: CompletionEvent) {
     val task = event.task
     val stage = idToStage(task.stageId)
-
+
```
pwendell: Maybe this is just from github, but if these are actually extra spaces (here and line 608) would be good to remove them.

core/src/main/scala/spark/scheduler/JobLogger.scala
```diff
@@ -0,0 +1,209 @@
...
+        case shufDep: ShuffleDependency[_,_] => val depInfo = "SHUFFLE_ID:" + shufDep.shuffleId
+                                                jobLogInfo(jobID, indentString(indent + 1) + depInfo, false)
+        case _ => recordRddInStageGraph(jobID, dep.rdd, indent + 1)
+      }
+    }
+  }
+  
+  private def recordStageDepGraph(jobID: Int, stage: Stage, indent: Int = 0) {
+    var stageInfo: String = ""
+    if(stage.isShuffleMap){
```
pwendell: Include spaces around control structures. This applies to lines 135-146.

if(state.isShuffleMap){
should be:
if (state.isShuffleMap) {

core/src/main/scala/spark/scheduler/JobLogger.scala
```diff
@@ -0,0 +1,209 @@
...
+    var writeInfo = info
+    if (withTime) {
+      val date = new Date(System.currentTimeMillis())
+      writeInfo = DATE_FORMAT.format(date) + ": " +info
+    }
+    jobIDToPrintWriter.get(jobID).foreach(_.println(writeInfo))
+  }
+  
+  //write log information to log file by stageID, withTime parameter controls whether to recored time stamp for the information
+  private def stageLogInfo(stageID: Int, info: String, withTime: Boolean) = stageIDToJobID.get(stageID).foreach(jobID => jobLogInfo(jobID, info, withTime))
```
pwendell: This line is well over 100 characters

core/src/main/scala/spark/scheduler/JobLogger.scala
```diff
@@ -0,0 +1,209 @@
...
+    }else
+      jobLogInfo(jobID, indentString(indent) + stageInfo + " JOB_ID=" + stage.priority, false)
+  }
+  
+  //record task metrics into job log files
+  private def recordTaskMetrics(stageID: Int, status: String, taskInfo: TaskInfo, taskMetrics: TaskMetrics) {
+    val info = " TID=" + taskInfo.taskId + " STAGE_ID=" + stageID + " START_TIME=" + taskInfo.launchTime + " FINISH_TIME=" + taskInfo.finishTime + 
+               " DURATION=" + taskInfo.duration + " EXECUTOR_ID=" + taskInfo.executorId + " HOST=" + taskInfo.host
+    val executorRunTime = " EXECUTOR_RUN_TIME=" + taskMetrics.executorRunTime
+    val readMetrics = { taskMetrics.shuffleReadMetrics match{
```
pwendell: Several lines here over 100 characters

pwendell: Also, the style here is a bit wonky. It should be like this:

val readMetrics = 
  taskMetrics.shuffleReadMetrics match {

(with second line indented)

pwendell:
Hi Mingfei,

Thanks for updating this - the new approach is exactly what Reynold and I were suggesting.

I've made several comments about style for your latest patch. We adhere to the Scala style guide with a few changes. Lines must be at most 100 characters. If you could review this patch (not only where I pointed out nits, but the entire thing) for style, that would be great. Here is our style guide:
http://spark-project.org/docs/latest/contributing-to-spark.html

From a code perspective, there are a few remaining things:
- It would be great if you could add some tests for this to verify the logic inside of JobLogger.scala
- I'm still a bit confused why you can't sue getCallSite instead of the generator class. Could you fill us in about the thought there?

Thanks again for taking time to update this.
- Patrick

shimingfei:
Hi Patrick
It is impossible to get the RDD's generator  from callsite information some time, because some classes may call into the same code where RDD is generated (for example in Shark, many operators call executeProcessPartition in Operator class to generate mapPartitionsWithIndexRDD), it is better to let user's code be able to set who generate the RDD. and I also merged getRddGenerator function to getSparkCallSite 
I added it only for analyzing, if it is not suitable to add it into RDD class, I will remove it from the code and use callsite information instead.
Besides I am adding test case for JobLogger now.

Thanks
Mingfei

shimingfei:
Hi Patrick

I have modified code according to your comments and added test case for joblogger

Thanks
Mingfei

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -575,10 +578,10 @@ class SparkContext(
       partitions: Seq[Int],
       allowLocal: Boolean,
       resultHandler: (Int, U) => Unit) {
-    val callSite = Utils.getSparkCallSite
+    val callSite = Utils.getSparkCallSite()
     logInfo("Starting job: " + callSite)
     val start = System.nanoTime
-    val result = dagScheduler.runJob(rdd, func, partitions, callSite, allowLocal, resultHandler)
+    val result = dagScheduler.runJob(rdd, func, partitions, callSite + "|" + addInfo.value.toString, allowLocal, resultHandler)
```
pwendell: This line is over 100 characters

core/src/main/scala/spark/Utils.scala
```diff
@@ -427,11 +428,15 @@ private object Utils extends Logging {
         else {
           firstUserLine = el.getLineNumber
           firstUserFile = el.getFileName
+          firstUserClass = el.getClassName
           finished = true
         }
       }
     }
-    "%s at %s:%s".format(lastSparkMethod, firstUserFile, firstUserLine)
+    if(getfirstUserClass)
```
pwendell: put a space before first paran:
if (getFirstUserClass)

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -65,6 +65,8 @@ class SparkContext(
   // Ensure logging is initialized before we spawn any threads
   initLogging()
 
+  val addInfo = new DynamicVariable[String]("")
```
pwendell: It would be good to have a docstring for this and also to maybe give it a different name. How about this:

/\* Allows higher layer frameworks to describe the context of a job. */
val annotation = new DynamicVariable[String]("")

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -465,6 +467,7 @@ class SparkContext(
     dagScheduler.sparkListeners += listener
   }
 
+//  SparkListener(this)
```
pwendell: Remove this (?)

core/src/main/scala/spark/scheduler/SparkListener.scala
```diff
@@ -2,26 +2,49 @@ package spark.scheduler
...
+  
+  def onTaskEnd(event: CompletionEvent) { }
+  
+  def onJobStart(job: ActiveJob, info: String = "") { }
+  
+  def onJobEnd(job: ActiveJob, event: SparkListenerEvents) { }
+  
+}
+
+object SparkListener {
```
pwendell: Could you remove this SparkListener object? Let's keep both the JobLogger and the StatsReportListener as optional add-on for now, rather than have a configuration option.

I think pretty soon we make listeners configurable by options, but I'd like to keep that separate from this pull request.

core/src/main/scala/spark/Utils.scala
```diff
@@ -401,7 +401,7 @@ private object Utils extends Logging {
    * (outside the spark package) that called into Spark, as well as which Spark method they called.
    * This is used, for example, to tell users where in their code each RDD got created.
    */
-  def getSparkCallSite: String = {
+  def getSparkCallSite(getfirstUserClass: Boolean = false): String = {
```
pwendell: A better way to refactor this would be to have two functions:
getCallSiteInfo() => returns (lastSparkMethod, firstUserFile, firstUserLine, firstUserClass)

formatSparkCallSite() => calls getCallSiteInfo then returns the formatted string

you can replace the earlier references to getSparkCallSite with formatSparkCallSite, then for the new cases you added here, you can just call getCallSiteInfo and you only use the firstUserClass and ignore the other stuff.

core/src/main/scala/spark/RDD.scala
```diff
@@ -114,6 +114,14 @@ abstract class RDD[T: ClassManifest](
     name = _name
     this
   }
+  
+  /**generator of this RDD*/
```
pwendell: Change the comment to:

/*\* User-defined generator of this RDD. */

core/src/main/scala/spark/scheduler/JobLogger.scala
```diff
@@ -0,0 +1,236 @@
...
+import scala.collection.mutable.{Map, HashMap, ListBuffer}
+import scala.io.Source
+import spark._
+import spark.executor.TaskMetrics
+import spark.scheduler.cluster.TaskInfo
+
+//it is used to record runtime information for each job, including RDD graph 
+//tasks' start/stop shuffle information and information from outside if there is any
+class JobLogger(val contextDirName: String) extends SparkListener with Logging{
+  private val logDir =  if (System.getenv("SPARK_LOG_DIR") != null)
```
pwendell: put a linebreak here before the if:

   private val logDir = 
     if (XXX)...

pwendell: also indent the line with "if" (comment doesn't show that)

core/src/main/scala/spark/scheduler/JobLogger.scala
```diff
@@ -0,0 +1,236 @@
...
+  private var jobIDToPrintWriter = new HashMap[Int, PrintWriter] 
+  private var stageIDToJobID = new HashMap[Int, Int]
+  private var jobIDToStages = new HashMap[Int, ListBuffer[Stage]]
+  private val DATE_FORMAT = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss")
+  
+  createContextDir()
+  
+  def this() = this(String.valueOf(System.currentTimeMillis()))
+  
+  //create a folder for each SparkContext, the folder's name is the creation time of the jobLogger
```
pwendell: Every comment in this file should have a space after the //

core/src/main/scala/spark/scheduler/JobLogger.scala
```diff
@@ -0,0 +1,236 @@
...
+  
+  //close log file for one job, and clean the stages related to the job in stageIDToJobID 
+  private def closeLogWriter(jobID: Int) = jobIDToPrintWriter.get(jobID).foreach { fileWriter => 
+    fileWriter.close()
+    jobIDToStages.get(jobID).foreach(_.foreach(stage => stageIDToJobID -= stage.id))
+    jobIDToPrintWriter -= jobID
+    jobIDToStages -= jobID
+  }
+  
+  //write log information to log file by JobID, withTime parameter controls whether to recored time stamp for the information
```
pwendell: > 100 characters

core/src/main/scala/spark/scheduler/JobLogger.scala
```diff
@@ -0,0 +1,236 @@
...
+  private def jobLogInfo(jobID: Int, info: String, withTime: Boolean) {
+    var writeInfo = info
+    if (withTime) {
+      val date = new Date(System.currentTimeMillis())
+      writeInfo = DATE_FORMAT.format(date) + ": " +info
+    }
+    jobIDToPrintWriter.get(jobID).foreach(_.println(writeInfo))
+  }
+  
+  //write log information to log file by stageID, withTime parameter controls whether to recored time stamp for the information
```
pwendell: > 100 characters

pwendell:
Hi Mighfei,

I added a handful of comments on this latest submission.
1. For the callsite stuff, I proposed a slightly different refactoring than the one you give here. In  general though I support re-factoring this to use the same code path.
2. I would prefer not to have the JobLogger based on a Spark configuration option just yet. I'd rather have it be something that people need to explicitly add. Also, you changed the code to add a StatsListener by default which wasn't the case before. My proposal was just to remove the entire StatsListener object which would fix both of these issues.
3. There were some more style issues which I commented on.
4. For the unit test - ideally you want to write a test in a way that actually verifies some of the logic in the JobLogger class. For instance, you could try to verify that the hashmaps you maintain are being setup correctly. There are a few ways to do this. One is to mock out calls to the various handler functions then to look directly at the state of the JobLogger. Another way is to add a constructor option that avoids cleaning the state of the maps inside of JobLogger (default false) and in the tests you set this option to true. Then you could just run a job and verify the contents of the hashmaps.

Thanks,
Patrick

shimingfei:
Hi Patrick

I have modified the code and add unit test according to your suggestion,

Thanks
Mingfei

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -65,6 +65,9 @@ class SparkContext(
   // Ensure logging is initialized before we spawn any threads
   initLogging()
 
+  // Allows higher layer frameworks to describe the context of a job
+  val annotation = new DynamicVariable[String]("")
```
pwendell, line 69: rather than having this be a possibly blank string. How about:

new DynamicVariable[Option[String]]("")

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -575,10 +578,11 @@ class SparkContext(
       partitions: Seq[Int],
       allowLocal: Boolean,
       resultHandler: (Int, U) => Unit) {
-    val callSite = Utils.getSparkCallSite
+    val callSite = Utils.formatSparkCallSite
     logInfo("Starting job: " + callSite)
     val start = System.nanoTime
-    val result = dagScheduler.runJob(rdd, func, partitions, callSite, allowLocal, resultHandler)
+    val result = dagScheduler.runJob(rdd, func, partitions, callSite + "|" +
```
pwendell, line 584: Rather than piggyback on callsite to add the annotation. Could you instead add another argument to runJob, called annotation, which accepts an Option[String]?

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -275,11 +275,16 @@ class DAGScheduler(
   private[scheduler] def processEvent(event: DAGSchedulerEvent): Boolean = {
     event match {
       case JobSubmitted(finalRDD, func, partitions, allowLocal, callSite, listener) =>
```
pwendell, line 277: It would be good to add an annotation field to JobSubmitted as well, (of type Option[String]) rather than using string parsing for this.

core/src/main/scala/spark/Utils.scala
```diff
@@ -427,13 +428,18 @@ private object Utils extends Logging {
...
         }
       }
     }
-    "%s at %s:%s".format(lastSparkMethod, firstUserFile, firstUserLine)
+    (lastSparkMethod, firstUserFile, firstUserLine, firstUserClass)
   }
 
+  def formatSparkCallSite = {
+    val callSiteInfo = getCallSiteInfo
+    "%s_%s_%s_%s".format(callSiteInfo._1, callSiteInfo._2, callSiteInfo._3, callSiteInfo._4)
```
pwendell, line 441: Could you make this formatting match what was there before, rather than changing it to this one with underscores? It's fine to leave out the fourth entry, just make it matches what we used to have. The reason why I proposed this refactoring was so that you could leave the existing log messages unchanged.

pwendell:
Hi Mingfei,

A few notes about the annotation field - it would be good to not use string parsing as a way to pass additional arguments. Instead, you can use an optional value.

Also, I notice that now you've changed the default format for the callsite info - it would be good to actually leave that unchanged.

Overall, this is looking in good shape. Once you fix these I'll ask Matei or Reynold to do a pass.

Jenkins, test this please

shimingfei:
Hi Patric,
I would like to use "localProperties" variable in latest SparkContext to pass job annotation information, so that interface of "runJob" and "submitJob" will not be changed.
and I also intend to use a daemon thread to do joblogger's work, since joblogger has lots of disk access.
But current spark I forked don't have "localProperties" variable in SparkContext, need I close the current pull request and create a new one?

pwendell:
Okay I think we can close this now. I'm going to take a look at the new PR.

PR #574, opened 2013-04-18
Remove spark-repl's extraneous dependency on spark-streaming

jey:
Per Reynold's request, this patch removes the extraneous dependency `spark-repl` has on `spark-streaming`, which was causing problems in Shark.

This is for `branch-0.7`.

### DIFF
--- a/project/SparkBuild.scala
+++ b/project/SparkBuild.scala
@@ -21,7 +21,7 @@ object SparkBuild extends Build {
 
   lazy val core = Project("core", file("core"), settings = coreSettings)
 
-  lazy val repl = Project("repl", file("repl"), settings = replSettings) dependsOn (core) dependsOn (streaming)
+  lazy val repl = Project("repl", file("repl"), settings = replSettings) dependsOn (core)
 
   lazy val examples = Project("examples", file("examples"), settings = examplesSettings) dependsOn (core) dependsOn (streaming)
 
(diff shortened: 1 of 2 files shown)


rxin:
Thanks, Jey.

To add some context:

spark-streaming introduces dependency on a version of zookeeper that is different from the version of zookeeper Hive uses. It leads to problems in Shark.

The repl binary itself doesn't really need to depend on the streaming project. spark-shell should, but that's simply a classpath setup in the run script.

jey:
BTW -- I'm ready to release 0.7.1 once this is approved/merged.

mateiz:
Please don't release anything or post a final version to Maven until I can take a look. We also need to fix that SBT issue.

tdas:
Also, I have a few streaming related pull requests that have been accepted
in the streaming branch, but is yet to be pulled into the master branch. I
think they should be in the 0.7.1 release.

mateiz:
TD, we can also include them in 0.7.2. I guess you missed the discussions on 0.7.1 when you were away. If they are small though please send a pull request to branch-0.7.

tdas:
Aah, okay. Then I will probably hold them off. They are not super
important. They just make the kafkaStream slightly better.

TD

mateiz:
I plan to do 0.7.2 soon after 0.7.1 with some perf fixes, so it shouldn't be a long wait.

mateiz:
Thanks for the patch, Jey. Committed it.

PR #575, opened 2013-04-18
Manual merge of yarn branch to trunk

mridulm:
Hi,

This is a manual merge of yarn branch to trunk.
All the tests have passed, and we have tested the resulting jar on a yarn cluster - and the performance is comparable to the yarn branch.

Note that there are a few todo's - which have been left for later.

a) Enhancement of yarn support - monitoring, management, hadoop security, and so on.
These can be seen as feature additions/enhancements - this patch (and yarn branch) just provides basic support.

b) Add support for mirror 'assembly' functionality from sbt.
I am not very comfortable with maven - so this is left as a todo for later.
Essentially, we need to mirror whatever is done as part of assembly for core while building the hadoop2-yarn maven profile.

c) Using multiple cores per container.
Currently, a container is restricted to a single core - but the allocation of the container might allow for multiple cores to be used.
This has been done due to an MT bug in hadoop dfs code (HADOOP-8406) - which has been resolved. Unfortunately, not 2.0.x versions seem to support this (particularly cloudera versions which have cherry picked the patches to apply).
We can either wait for GA of hadoop 2.0 before modifying the code to add support for this.

Nothing else comes to mind off hand. No new changes have been introduced compared to yarn branch iirc - but the codebase there was significantly different : so a closer look as part of review would be great !

We have also NOT tested this for streaming (other than what is in maven tests) and mesos.
Any help in these would be greatly appreciated.

Thanks,
Mridul


project/SparkBuild.scala
```diff
@@ -40,9 +47,10 @@ object SparkBuild extends Build {
     scalacOptions := Seq("-unchecked", "-optimize", "-deprecation"),
     unmanagedJars in Compile <<= baseDirectory map { base => (base / "lib" ** "*.jar").classpath },
     retrieveManaged := true,
-    retrievePattern := "[type]s/[artifact](-[revision])(-[classifier]).[ext]",
+    // retrievePattern := "[type]s/[artifact](-[revision])(-[classifier]).[ext]",
     transitiveClassifiers in Scope.GlobalScope := Seq("sources"),
-    testListeners <<= target.map(t => Seq(new eu.henkelmann.sbt.JUnitXmlTestsListener(t.getAbsolutePath))),
+    // For some reason this fails on some nodes and works on others - not yet debugged why
+    // testListeners <<= target.map(t => Seq(new eu.henkelmann.sbt.JUnitXmlTestsListener(t.getAbsolutePath))),
```
mateiz: Don't leave this commented out in the pull request to us

project/SparkBuild.scala
```diff
@@ -40,9 +47,10 @@ object SparkBuild extends Build {
     scalacOptions := Seq("-unchecked", "-optimize", "-deprecation"),
     unmanagedJars in Compile <<= baseDirectory map { base => (base / "lib" ** "*.jar").classpath },
     retrieveManaged := true,
-    retrievePattern := "[type]s/[artifact](-[revision])(-[classifier]).[ext]",
+    // retrievePattern := "[type]s/[artifact](-[revision])(-[classifier]).[ext]",
```
mateiz: You can just delete this line

project/SparkBuild.scala
```diff
@@ -142,8 +152,26 @@ object SparkBuild extends Build {
...
+          Seq(
+            "org.apache.hadoop" % "hadoop-core" % HADOOP_VERSION,
+            "org.apache.hadoop" % "hadoop-client" % HADOOP_VERSION
+          )
+        }
+      } else {
+        Seq("org.apache.hadoop" % "hadoop-core" % HADOOP_VERSION)
+      }),
+    unmanagedSourceDirectories in Compile <+= baseDirectory{ _ /
+      ( if (HADOOP_YARN && HADOOP_MAJOR_VERSION == "2") "src/hadoop2-yarn/scala" else "src/hadoop" + HADOOP_MAJOR_VERSION + "/scala" )
```
mateiz: Break the else onto a new line

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -99,7 +102,7 @@ class SparkContext(
 
 
   // Add each JAR given through the constructor
-  jars.foreach { addJar(_) }
+  if (jars != null) jars.foreach { addJar(_) }
```
mateiz: Put this in a block rather than on the same line as the if; so like this:

```
if (jars != null) {
  jars.foreach(...)
}
```

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -111,7 +114,7 @@ class SparkContext(
       executorEnvs(key) = value
     }
   }
-  executorEnvs ++= environment
+  if (environment != null) executorEnvs ++= environment
```
mateiz: Same here, put if body in a block

core/src/main/scala/spark/Utils.scala
```diff
@@ -273,7 +317,98 @@ private object Utils extends Logging {
    * Get the local machine's hostname.
    */
   def localHostName(): String = {
-    customHostname.getOrElse(InetAddress.getLocalHost.getHostName)
+    // customHostname.getOrElse(InetAddress.getLocalHost.getHostName)
```
mateiz: Remove the commented-out line

core/src/main/scala/spark/Utils.scala
```diff
@@ -273,7 +317,98 @@ private object Utils extends Logging {
...
+  def checkHost(host: String, message: String = "") {
+    // Currently catches only ipv4 pattern, this is just a debugging tool - not rigourous !
+    // if (host.matches("^[0-9]+(\\.[0-9]+)*$")) {
+    if (ipPattern.matcher(host).matches()) {
+      Utils.logErrorWithStack("Unexpected to have host " + host + " which matches IP pattern. Message " + message)
+    }
+    if (Utils.parseHostPort(host)._2 != 0){
+      Utils.logErrorWithStack("Unexpected to have host " + host + " which has port in it. Message " + message)
+    }
+  }
```
mateiz, line 350: We'll want to leave these out of the final version; do you mind replacing them by the blank versions below? My guess is that some people might have nodes with their hostname matching their IP.

core/src/main/scala/spark/storage/BlockManager.scala
```diff
@@ -37,17 +37,27 @@ class BlockManager(
...
+      this.initThread = Thread.currentThread()
+    }
 
     /**
      * Wait for this BlockInfo to be marked as ready (i.e. block is finished writing).
      * Return true if the block is available, false otherwise.
      */
     def waitForReady(): Boolean = {
-      if (pending) {
+      if (initThread != Thread.currentThread() && pending) {
```
mateiz, line 60: Is the goal of this to avoid a lock if we're being called in the initializer thread? That's fine, but one bad aspect is that the initThread variable keeps pointing to that thread and might prevent GC. We should set initThread = null when we're done.

mateiz, line 60: Actually, never mind, I see you are setting it to null. Still wondering why this was needed though -- did it make a difference in performance?

mateiz:
Hey Mridul, I made a few small comments, but this looks good overall. Thanks for doing the merge! Let me know when you've gone through them.

mateiz:
By the way, in Maven, the equivalent of sbt assembly is done by package. It creates a big single JAR in the repl directory.

mateiz:
Also, HADOOP-8406 seems to have been fixed in Hadoop 2.0.2-alpha, or so they say. Is it still broken?

project/SparkBuild.scala
mridulm: Done

project/SparkBuild.scala
mridulm: Done

project/SparkBuild.scala
mridulm: Done, please check

core/src/main/scala/spark/SparkContext.scala
mridulm: Done

core/src/main/scala/spark/SparkContext.scala
mridulm: Done

core/src/main/scala/spark/Utils.scala
mridulm: Done

core/src/main/scala/spark/Utils.scala
mridulm, line 350: Done.
But please do note that while testing streaming and mesos, please do use the other version for your deployment - I want to ensure that there are no bugs which have crept in along code paths we have not tested !

Thanks.

core/src/main/scala/spark/storage/BlockManager.scala
mridulm, line 60: In some codepaths (I dont recall which), particularly after waitForReady was added uniformly to all usage of info, there can be a deadlock if we do not check if it is reenterent.

mridulm:
Hi Matei,

  I have updated the pull request with fixes for your comments.
I have also added a new api to SparkHadoopUtil to create a new Configuration (in yarn mode, for example, it must be YarnConfiguration so that underlying subsystems are appropriately initialized in some cases).
Additionally, I have modified an example to make use of SplitInfo to influence the node allocation as part of creating SparkContext (for illustration of how to use this api).

Please do review, thanks for all your comments and feedback !

Regards,
Mridul

mridulm:
Regarding HADOOP-8406, Cloudera seems to be picking and choosing which patches to apply - so I am unsure if this one made it into their distro or not !
On other hand, within a month or more, 2.0.5 RC should be out (I am told) - and everyone will move to it, so we can always re-enable it then.

The loss of functionality is minimal (mostly folks just allocate one mappers memory per container anyway), but the impact if the bug hits is very bad (job fails intermittently).

Thanks,
Mridul

mridulm:
Hi Matei,

  We are done with our changes.
Unless there are any required based on review comments, the patch is good to go - finally !

Thanks,
Mridul

mateiz:
Thanks for the updates -- they look good. However, the pull request doesn't seem to merge automatically anymore. Mind updating it? I think it may be due to small changes we did in the Maven build.

mridulm:
Hi Matei,

  I have updated project/SparkBuild.scala based on what is in master.
Is there any other conflict when trying to merge ?

Thanks,
Mridul

mridulm:
scratch that, there are still some issues - let me try to fix those.

mridulm:
Hi Matei,

  I have hopefully merged with trunk properly this time around. Can you please give it a shot ?

Thanks,
Mridul

mateiz:
Thanks Mridul! I've merged this in now.

mridulm:
Brilliant !
Thanks a lot Matei !!

Regards, 
Mridul

mateiz:
Hey Mridul, just FYI, the removal of this line from SparkBuild.scala:

```
retrievePattern := "[type]s/[artifact](-[revision])(-[classifier]).[ext]"
```

broke sbt test, because it broke the "run" script by downloading files into subdirectories of lib_managed. I've fixed it now -- I guess you probably ran your tests with Maven. It's unfortunate that we didn't catch this earlier, but we're working on setting up automated test infrastructure that will catch such things.

mridulm:
Hi Matei,
  We had tested only on maven ... though we do run sbt assembly.
That is why we had to remove that line ... sbt picks up wrong netty as part
of assembly even though required netty is explicitly mentioned as first
dependency .. not sure from which dependency older netty is getting picked
up - we will need to exclude older netty somehow.

Regards,
Mridul

mateiz:
I see, that is a problem. We might need to add "exclude" rules for the things that pull in other versions of Netty.

mateiz:
By the way, now that I think about it, I'm not sure why this would cause sbt assembly to fail. That setting is only used to pull files into "lib_managed" from the Ivy cache. See whether it's still a problem with SBT 0.12 now that we're using that.

The right solution might actually be to rely on sbt assebly to build a "good" complete JAR of the project, and use that on the classpath of "run". I think SBT it self doesn't put the multiple Netty versions on the run or build classpath; it just happens to download the JARs into that folder, and our run script picks them up.

mridulm:
Hi Matei,

  The assembly does not fail - it just picks wrong netty jar to include in
the assembled jar  : causing akka to fail when spark comes up in the
application master.

Note that unlike local mode, etc we need to submit the spark assembled jar
and user jar to yarn to launch the job : so we use the assembled jar to
consolidate all depenencies (so run classpath is not relevant anymore).

The correct solution in my opinion, given this, is what you proposed
initially - add appropriate exclude to dependencies which are indirectly
pulling in older netty.
Thankfully, maven does the right thing as part of package !

I will try to resolve this tomorrow - though I suspect it might be getting
pulled indirectly from hadoop and spray ...

Regards,
Mridul

mateiz:
So I've added some "exclude" rules to remove Netty, but it's not too pretty (1b169f190c5c5210d088faced86dee1007295ac8). We should find another solution -- both an assembly-like plugin that pulls in only the right versions, and a way to to the equivalent of "retrieve" (create a runtime classpath) that has the right JARs. For the latter we could just use the result of "assembly".

One weird thing for Netty was that its organization name changed from org.jboss.netty to io.netty. I think that might've confused SBT too.

mridulm:
That is great Matei, this fixes the issue properly : tested the sbt
assembled jar, and things are working great.
I did not realize the org name change - that explains the problems we saw
better !

Thanks for resolving this.
Regards,
Mridul

mateiz:
Hi Mridul,

I noticed that unfortunately this patch seems to lead to some sporadic test failures, in spark.streaming.FailureSuite. It happens even with the version directly on your branch (commit 3b594a4e3b94de49a09dc679a30d857e3f41df69), not just the version in mesos/spark master. This is a test that basically relies on saving some RDDs to files with checkpoint() and resuming from an old checkpoint when we randomly kill the SparkContext. It seems to read or compute some incorrect values with the patch. I'm trying to track it down, but if you have time to take a look, it would be appreciated. Just do "project streaming" and then "test-only spark.streaming.FailureSuite" in SBT. Feel free to modify the test to keep some files around for debugging. The problem happens about 1/2 of the time in my experience so far.

mateiz:
By the way, the actual test that fails is "multiple failures with updateStateByKey". This tests essentially adds the stream of values 1, 2, 3, 4, etc to the count of a given key, so it's supposed to get values 1, 3, 6, 10, etc ("triangular numbers") as a result. However, sometimes it gets a non-triangular number, meaning it's somehow added the wrong values or read the checkpoint file incorrectly. It's pretty weird but maybe you have some ideas on what changes could've caused it.

mridulm:
That is strange, the only test failures I saw were the hdfs test failures
and the distributed suite failures - but that was due to some api issues :
which I added to resolve it.
Never saw streaming failures !

Ofcourse, I always tested as part of install in maven - not from sbt.
Let me give this a shot : will let you know in case I am able to reproduce
and resolve it.
I am re-enabling the checkHost and checkHostPort code in Utils - just in
case that is the issue (can you try with that too ?).

Btw, is there anything specific about the environment of the test machine I
should try to replicate ?

Thanks,
Mridul

mateiz:
Not that I know of; I tried the same test from an old revision on the same machine and it works. But who knows, maybe there's something wrong with the hostname configuration.

mridulm:
Hi Matei,

  Unfortunately, I am unable to reproduce the issue inspite of some 30 odd
tries.
I even tried after an explicit "./sbt/sbt clean assembly" of the entire
source, and cleaning up /tmp just in case there was some other issue.

If you have some log files I can take a look at, that would be great.
In meantime, let me try on a few other machines.

Thanks,
Mridul

rxin:
Did this pull request change the logging level? After merging this in, I see almost no logging messages printed when running spark-shell ...

core/src/main/scala/spark/scheduler/ShuffleMapTask.scala
```diff
@@ -77,13 +77,21 @@ private[spark] class ShuffleMapTask(
...
     var partition: Int,
-    @transient var locs: Seq[String])
+    @transient private var locs: Seq[String])
   extends Task[MapStatus](stageId)
   with Externalizable
   with Logging {
 
   protected def this() = this(0, null, null, 0, null)
 
+  // data locality is on a per host basis, not hyper specific to container (host:port). Unique on set of hosts.
```
rxin, line 87: Even if the partitions are on the same host, the worker would still need to serialize / deserialize those data when they are in different jvm processes. Often that is very costly ...

mridulm, line 87: Yes, but it minimizes the cost of network IO - which can be non trivial
when the cluster is relatively loaded.
And if the task scheduling ends up running the job on the same container
which is hosting it - then the network IO cost goes done to zero.
- Mridul

mridulm:
I just ran through the diffs again to ensure there is no logger configuration added - could not find any.
Is it still a problem ?

core/src/main/scala/spark/scheduler/ShuffleMapTask.scala
rxin, line 87: Does it make sense to have 3 levels of locality?

In-process locality, node locality, rack locality ...

rxin:
I think it was some problem introduced by Hadoop 2.

Matei's commit changed it to build by default with Hadoop 1 and I can see the logs again in spark-shell:

https://github.com/mesos/spark/commit/f1f92c88eb2960a16d33bf7dd291c8ce58f665de

core/src/main/scala/spark/scheduler/ShuffleMapTask.scala
mridulm, line 87: The performance difference is significant - particularly when the data locality is reasonably good.
The first preference would be for hyper local (or in process locality as you called it), next for node locality - both of which qualify for host locality (along with no prefs tasks).
Next comes rack local tasks - after which any node would do.

Obviously between host local (both kinds) and rack local/any, we would definitely prefer host local; but if choice is between hyper local and any container on that node, we would prefer hyper local to completely bypass the IO need entirely.
If we do not support hyper local scheduling, for each rdd block, we can keep ending up transferring data between containers on the same node - the cost for single block might not be very high, but cummulatively adds up to being pretty expensive : not to mention completely thrashing the memory and disk caches on that node.

Unfortunately, there are a bunch of other factors which interact with this in non trivial ways to always end up with qualitative figures - 
- whether data is still in memory or has been pushed to disk, 
- how many other slaves are pulling data from this worker, 
- how many other slaves are scheduled on the same node, 
  etc

To give an example, time to complete for some of our machine algorithms went down from 18+ minutes to 3.x - 4.x minutes after all the scheduling changes were done on a small 40 node cluster.
Ofcourse it was not just this change though - there are a few other flags which control the full set : 
TASK_REVIVAL_INTERVAL and TASK_SCHEDULING_AGGRESSION in ClusterScheduler have significant impact on the runtime when the cluster size goes up. The default value has been set to preserve existing behavior, but in our clusters we have TASK_SCHEDULING_AGGRESSION set to RACK_LOCAL and TASK_REVIVAL_INTERVAL set to 200 or so.

mridulm, line 87: Hmm, looking at the submitted diffs - I think we left hyper local scheduling in, but removed it from locality preference from RDD and Task. Dont recall why this was done - but I suspect it was done to minimize disruption since we had other significant issues to resolve (too many host -> hostport changes, and there was no way to validate this easily since it can be both host or hostPort).

So the final code which was committed into spark yarn branch, and now subsequently to spark master does not have this change. Apologies for my oversight in the response above !

The change to enable this is actually pretty simple - once the patch stabilizes, I can submit a pull request which adds it back in. We have does this before anyway :-)

Thanks for looking into this and bringing it up !

PR #578, opened 2013-04-22
Add option to resize root volume of master node

jey:
Adds option to EC2 scripts to resize the master node's root volume. (I found this useful for running the performance tests.)


jey:
@JoshRosen - ah wow, that looks like the likely culprit, >5GB of logs in there.

mateiz:
Hmm, maybe instead of increasing the root volume, we should move the Ganglia logs. Mind sending a pull request for that? The reason is that the root volume is slower to access than others and is not really meant to store large amounts of data, so we should just discourage users from using it.

rxin:
Closing this one. Jey, if you have a chance, please submit a pull request for moving the Ganglia logs.

PR #581, opened 2013-04-26
fix [SPARK-740] block manage UI throws exception when enabling Spark Streaming

jerryshao:
Because of Spark Streaming's input block naming format (input-xx-xx) is different from Spark (rdd_xx_xx), so block manager UI will throw exception when group this block using block name, detailed thing is illustrated in [[SPARK-740]](https://spark-project.atlassian.net/browse/SPARK-740).

Currently I add a key filter to filter out all the Spark Streaming's input block RDD, so the UI will only display processed RDD. Also I sort RDDInfo by RDD's id, so it is clear to see in the UI.

### DIFF
--- a/core/src/main/scala/spark/storage/StorageUtils.scala
+++ b/core/src/main/scala/spark/storage/StorageUtils.scala
@@ -4,9 +4,9 @@ import spark.{Utils, SparkContext}
 import BlockManagerMasterActor.BlockStatus
 
 private[spark]
-case class StorageStatus(blockManagerId: BlockManagerId, maxMem: Long, 
+case class StorageStatus(blockManagerId: BlockManagerId, maxMem: Long,
   blocks: Map[String, BlockStatus]) {
-  
+
   def memUsed(blockPrefix: String = "") = {
     blocks.filterKeys(_.startsWith(blockPrefix)).values.map(_.memSize).
       reduceOption(_+_).getOrElse(0l)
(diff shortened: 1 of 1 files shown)


core/src/main/scala/spark/storage/StorageUtils.scala
```diff
@@ -22,35 +22,41 @@ case class StorageStatus(blockManagerId: BlockManagerId, maxMem: Long,
...
     sc: SparkContext) : Array[RDDInfo] = {
 
     // Group by rddId, ignore the partition name
-    val groupedRddBlocks = infos.groupBy { case(k, v) =>
-      k.substring(0,k.lastIndexOf('_'))
-    }.mapValues(_.values.toArray)
+    val groupedRddBlocks = infos.filterKeys(_.startsWith("rdd_")).
+      groupBy { case(k, v) =>
+        k.substring(0,k.lastIndexOf('_'))
+      }.mapValues(_.values.toArray)
```
mateiz: To match the style of the rest of the codebase, format this like this:

```
val groupedRddBlocks = infos.filterKeys(_.startsWith("rdd_")).groupBy {
  ...
}.mapValues(...)
```

mateiz:
Looks great, thanks! I just made one small comment on formatting.

tdas:
I have taken a look as well. Looks good.

jerryshao:
I've changed the code formatting, please review, Thanks.

mateiz:
Great, thanks.

PR #582, opened 2013-04-27
Add zip partitions interface

shivaram:
This pull request adds support for zipping RDDs by partitions. Right now, this supports zipping 2, 3 or 4 RDDs together, and we can extend this if required.

Motivation for this was drawn primarily from MLBase algorithms and through discussions with @etrain and @rxin


core/src/main/scala/spark/RDD.scala
```diff
@@ -436,6 +439,25 @@ abstract class RDD[T: ClassManifest](
    */
   def zip[U: ClassManifest](other: RDD[U]): RDD[(T, U)] = new ZippedRDD(sc, this, other)
 
+  def zipAndMapPartitions[B: ClassManifest, V: ClassManifest](
+    f: (Iterator[T], Iterator[B]) => Iterator[V],
+    rdd2: RDD[B]) =
```
mateiz: This style doesn't match the rest of the codebase (should have 4 indents for arguments)

core/src/main/scala/spark/rdd/MapZippedPartitionsRDD.scala
```diff
@@ -0,0 +1,118 @@
+package spark.rdd
+
+import spark.{OneToOneDependency, RDD, SparkContext, Partition, TaskContext}
+import java.io.{ObjectOutputStream, IOException}
+
+private[spark] class MapZippedPartition(
+    idx: Int, 
+    @transient rdds: Seq[RDD[_]]
+  ) extends Partition {
```
mateiz: Ditto on code style here, should look like

```
class Whatever(
    foo,
    bar)
  extends Blah {
```

mateiz:
This looks good overall, Shivaram, but I made a few comments on the style, and I'd recommend changing the name to just zipPartitions.

shivaram:
Thanks for taking a look. I made the style changes and renamed the function to to zipPartitions. Also a couple of lines like "class MapZippedPartitionsRDD4[A: ClassManifest, B: ClassManifest, C: ClassManifest, D:ClassManifest, V: ClassManifest]" were over 100 characters long -- I ended up breaking them into two lines.

core/src/main/scala/spark/RDD.scala
```diff
@@ -436,6 +439,25 @@ abstract class RDD[T: ClassManifest](
    */
   def zip[U: ClassManifest](other: RDD[U]): RDD[(T, U)] = new ZippedRDD(sc, this, other)
 
+  def zipPartitions[B: ClassManifest, V: ClassManifest](
+      f: (Iterator[T], Iterator[B]) => Iterator[V],
+      rdd2: RDD[B]) =
```
mateiz: Specify the return type for these as `: RDD[V]`. If you let it be inferred, it will be MapZippedPartitionsRDD2[T, B, V], and we don't want to make that type exposed in case we later change the implementation.

core/src/main/scala/spark/rdd/MapZippedPartitionsRDD.scala
```diff
@@ -0,0 +1,120 @@
...
+
+  @throws(classOf[IOException])
+  private def writeObject(oos: ObjectOutputStream) {
+    // Update the reference to parent split at the time of task serialization
+    partitionValues = rdds.map(rdd => rdd.partitions(idx))
+    oos.defaultWriteObject()
+  }
+}
+
+abstract class MapZippedPartitionsBaseRDD[V: ClassManifest](
```
mateiz: Rename these classes to just ZippedPartitionsBaseRDD, ZippedPartitionsRDD1, etc.

mateiz:
Noticed a couple of other minor issues.

shivaram:
Renamed the classes and the file as well. The only tricky bit is that the partition class in ZippedRDD is called ZippedPartition -- I called the one here ZippedPartitions. Let me know if you have a better suggestion for that.

mateiz:
Hmm, as bad as this sounds, it might be good to call it ZippedPartitionsPartition. Otherwise it's kind of confusing.

core/src/main/scala/spark/RDD.scala
```diff
@@ -436,6 +439,25 @@ abstract class RDD[T: ClassManifest](
    */
   def zip[U: ClassManifest](other: RDD[U]): RDD[(T, U)] = new ZippedRDD(sc, this, other)
 
+  def zipPartitions[B: ClassManifest, V: ClassManifest](
```
JoshRosen, line 448: Add Scaladoc describing `zipPartitions` here?

shivaram, line 448: Good suggestion -- Added a description now

shivaram:
Hmm ZippedPartitionsPartition is a mouthful, but at least consistent with the rest of the classes. Renamed now.

mateiz:
Great, thanks for updating it.

PR #583, opened 2013-04-28
Fix issues with streaming test cases after yarn branch merge

mridulm:
This is an attempt to isolate and resolve issues which surfaced with streaming testcases after yarn branch merge.


mridulm:
There are a few log messages added which we will remove once it is confirmed that the bug is resolved.
I am not able to reproduce the issue after this patch (the mail I had sent earlier had incomplete diffs, apologies !).

But I am trying on other machines in case it is just a local issue with not reproducing the bug.

Regards,
Mridul

mridulm:
This is work in progress, please do not merge it yet.

mateiz:
Cool, thanks for looking at it. This does seem like a good fix. It also seems to solve the issues on my machine, though I'll keep testing it to make sure.

mridulm:
There seems to be still some issue, I am trying to nail it down.
Will keep this pull request updated.

Apologies for introducing this issue !
Regards,
Mridul

mridulm:
Hi Matei,

  I cant seem to find any other reproducible issue with the codebase.
I did see a failure once out of about 50 tries, but unfortunately, it was is not reproducible and the log messages were not sufficient to debug it.

As we discussed, that problem might be elsewhere actually ... I suspect related to serialization of a DStream whose state might be changing while serialization was happening (which is not something we touched in yarn merge).

For now, I think this PR is good to go pending your review. In case I find a reproducible testcase, I will definitely try to debug this issue more.

Regards,
Mridul

mateiz:
Thanks Mridul. I'll keep looking into it as well, but these are good fixes to have.

PR #584, opened 2013-04-29
Spark-742: TaskMetrics should not employ per-record timing.

pwendell:
This patch does three things:
1. Makes TimedIterator a trait with two implementations (one a no-op)
2. Makes the default behavior to use the no-op implementation
3. Removes DelegateBlockFetchTracker. This is just cleanup, but it seems like
   the triat doesn't really reduce complexity in any way.

In the future we can add other implementations, e.g. ones which perform sampling.


pwendell:
Going to re-open this with a different approach shortly.

PR #585, opened 2013-04-29
[Fix SPARK-742] Task Metrics should not employ per-record timing by default

pwendell:
This just removes the metric that was being tracked per-record. There is also a minor re-factoring which removes DelegateBlockFetchTracker.scala.


core/src/main/scala/spark/BlockStoreShuffleFetcher.scala
```diff
@@ -4,8 +4,8 @@ import executor.{ShuffleReadMetrics, TaskMetrics}
 import scala.collection.mutable.ArrayBuffer
 import scala.collection.mutable.HashMap
 
-import spark.storage.{DelegateBlockFetchTracker, BlockManagerId}
-import util.{CompletionIterator, TimedIterator}
+import spark.storage.BlockManagerId
+import util.CompletionIterator
```
rxin: Mind changing util to fully qualified package name?

Everything else looks good.

pwendell: fixed in latest commit

mateiz:
Thanks Patrick. Can you also cherry pick this into branch-0.7?

Matei

PR #586, opened 2013-04-30
Pull request to address issues Reynold Xin reported

mridulm:
1) The ConnectionManager change addresses an exception I saw in the logs as part of debugging issue reported by Reynold Xin.
The actual bug exists even with previous codebase - but due to the increased MT nature of spark post yarn fix, this might be more clearly manifesting now.

Either case, even if this is not the root cause, this needs to be addressed.

2) The run script change addresses an issue with setting up classpath.
Incorrect directory was being probed for shaded jar.

Please do let me know if this fixes the issues you saw Reynold.
In meantime, I am going to try to keep running tests under various loads to see if spark fails in any of them.

Thanks a lot for the concise testcase to help debug this !

Regards,
Mridul

### DIFF
--- a/core/src/main/scala/spark/network/ConnectionManager.scala
+++ b/core/src/main/scala/spark/network/ConnectionManager.scala
@@ -188,6 +188,38 @@ private[spark] class ConnectionManager(port: Int) extends Logging {
     } )
   }
 
+  // MUST be called within selector loop - else deadlock.
+  private def triggerForceCloseByException(key: SelectionKey, e: Exception) {
+    try {
+      key.interestOps(0)
+    } catch {
+      // ignore exceptions
+      case e: Exception => logDebug("Ignoring exception", e)
+    }
+
+    val conn = connectionsByKey.getOrElse(key, null)
+    if (conn == null) return
+
+    // Pushing to connect threadpool
+    handleConnectExecutor.execute(new Runnable {
+      override def run() {
+        try {
+          conn.callOnExceptionCallback(e)
+        } catch {
+          // ignore exceptions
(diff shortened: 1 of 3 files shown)


core/src/main/scala/spark/network/ConnectionManager.scala
```diff
@@ -235,18 +267,26 @@ private[spark] class ConnectionManager(port: Int) extends Logging {
...
+              if (key.isReadable) {
+                triggerRead(key)
+              } else
+              if (key.isWritable) {
+                triggerWrite(key)
+              }
+            }
+          } catch {
+            // weird, but we saw this happening - even though key.isValid was true, key.isAcceptable would throw CancelledKeyException.
+            case e: CancelledKeyException => {
```
mridulm: I saw this specifically happening only for CancelledKeyException, but do we want to generalize it to Exception ?
An exception here will mean the selector loop exits ! Which could be one of the reasons rxin observed a hang.

mridulm:
So I am explicitly catching only CancelledKeyException, should we change it to Exception ?
If there are any uncaught exceptions, the selector loop will exit - which essentially means unresponsive slave (but not dead - just unresponsive).

rxin:
Maybe it makes sense to do a error level log for general exceptions?

mridulm:
Sounds good, will trigger the same codepath for exceptions too - except that if CancelledKeyException, will do debug logging, else error logging.

mridulm:
Done, can you please review it Reynold/Matei and commit it ?

Thanks,
Mridul

mateiz:
Jenkins, test this please.

mateiz:
Thanks for the fix, Mirdul.

PR #587, opened 2013-04-30
A set of shuffle map output related changes

rxin:
1. Added a BlockObjectWriter interface for writing out a series of jvm objects to disk.
2. Added a getBlockWriter interface to DiskStore.
3. Added a new method to BlockManager for writing out shuffle files. This method provides a short-circuited way to write shuffle data out directly to disk.
4. Added ShuffleBlockManager and updated ShuffleMapTask to use the manager to write map outputs out to disk (ultimately via the BlockObjectWriter interface).
5. Allow specifying a shuffle serilaizer on a per-shuffle basis.


core/src/main/scala/spark/storage/ShuffleBlockManager.scala
```diff
@@ -0,0 +1,53 @@
...
+
+
+private[spark]
+class ShuffleWriterGroup(val id: Int, val writers: Array[BlockObjectWriter])
+
+
+private[spark]
+class ShuffleBlockManager(blockManager: BlockManager) {
+
+  val shuffles = new ConcurrentHashMap[Int, Shuffle]
```
ryanlecompte: This map doesn't appear to be used in this class?

rxin: Ah let me remove it. It was for a separate design that I decided to pull out last minute.

core/src/main/scala/spark/storage/BlockManager.scala
```diff
@@ -293,22 +291,23 @@ class BlockManager(
   }
 
   /**
+   * A short-circuited method to get blocks directly from disk. This is used for getting
+   * shuffle blocks. It is safe to do so without a lock on block info since disk store
+   * never deletes (recent) items.
+   */
+  def getLocalFromDisk(blockId: String, serializer: Serializer): Option[Iterator[Any]] = {
+    diskStore.getValues(blockId, serializer) match {
```
ryanlecompte: Minor, but you can avoid the extra Some object allocation with just:

``` scala
  diskStore.getValues(blockId, serializer) orElse sys.error("Block " + blockId + " not found on disk, though it should be")
```

rxin:
I think the above jenkin test failure was due to a hang (fixed in #586, already merged)not caused by this pull request.

core/src/main/scala/spark/serializer/Serializer.scala
```diff
@@ -14,6 +18,48 @@ trait Serializer {
   def newInstance(): SerializerInstance
 }
 
+
+/**
+ * A singleton object that can be used to fetch serializer objects based on the serializer
+ * class name. If a previous instance of the serializer object has been created, the get
+ * method returns that instead of creating a new one.
+ */
+object Serializer {
```
mateiz: I'd prefer to avoid singleton objects so that we can eventually run multiple instances of Spark in the same JVM (e.g. for multithreaded test execution). Can you either make these methods be part of SparkEnv, or possibly add a "serializer manager" in SparkEnv?

core/src/main/scala/spark/scheduler/ShuffleMapTask.scala
```diff
@@ -129,34 +137,52 @@ private[spark] class ShuffleMapTask(
 
     val taskContext = new TaskContext(stageId, partition, attemptId)
     metrics = Some(taskContext.taskMetrics)
+
+    val blockManager = SparkEnv.get.blockManager
+    var shuffle: ShuffleBlockManager#Shuffle = null
```
mateiz: Can you access this as ShuffleBlockManager.Shuffle instead? It might work if you put it inside an object ShuffleBlockManager instead.

core/src/main/scala/spark/storage/DiskStore.scala
```diff
@@ -20,8 +21,59 @@ import spark.Utils
...
+
+    // Flush the partial writes, and set valid length to be the length of the entire file.
+    // Return the number of bytes written for this commit.
+    override def commit(): Long = {
+      bs.flush()
+      validLength = repositionableStream.position()
+      validLength
+    }
+
+    override def revertPartialWrites() {
```
mateiz, line 65: So right now, this deletes the whole file, but when you add support for merging shuffle files you'll change that?

mateiz:
Hey Reynold,

This looks good except for the singleton object and the semantics of revertPartialWrites on disk. For the latter, do you want to keep this interface to support merging shuffle files in the future? If so, we should at least document that it can cause previously committed writes to be deleted, but it might also be good to just make it resize the file to right before it started writing stuff.

mateiz:
By the way, this is how you can truncate the file: http://docs.oracle.com/javase/1.4.2/docs/api/java/nio/channels/FileChannel.html#truncate%28long%29. You can probably use FileOutputStream.getChannel.truncate() for example.

core/src/main/scala/spark/scheduler/ShuffleMapTask.scala
rxin: Ok I created a trait ShuffleBlocks for that, which makes it easier since we can swap out different implementations later.

core/src/main/scala/spark/serializer/Serializer.scala
rxin: Done.

core/src/main/scala/spark/storage/DiskStore.scala
rxin, line 65: Yes

rxin:
Done. PTAL.

I actually added the append feature fully back to DiskStore (although the higher level shuffle block manager doesn't yet do consolidation).

rxin:
Again - don't think the above test failures were caused by this pr. It is a little bit annoying that Jenkins just keeps saying PRs are failing tests ...

mateiz:
Yeah, we should probably disable that for now.

mateiz:
Thanks Reynold, I've merged this.

PR #589, opened 2013-05-01
Add support for instance local scheduling

mridulm:
There are a couple of things which require review (among others).

a) Resolving executorId to executor host + port.
There is a change (hack ?) introduce to SparkEnv to achieve this due to lack of better way to do this.
This implicitly pulls ClusterScheduler as reference (cant be explicitly referenced though) - I am not very familiar with the design choice of splitting SparkEnv from SparkContext : so this change might be violating that ?

b) I am not familiar with ZippedRDD (or use of zip in general in scala) and what the expectation for preferredLocation for ZippedRDD is.
I suspect current code for preferredLocation is substandard for ZippedRDD : I have a TODO note there since I am not sure how to handle it properly.

This has been tested on a yarn cluster and on via maven test (and some tests modified to ensure they pass).

This had been initially removed from the yarn merge due to the disruptive nature of the change (more than yarn support change :) ). Since that patch is reasonably tested, we can take a shot at this now.

Any and all feedback more than welcome !

Thanks,
Mridul


mridulm:
Just happened to come across these stats while running a few jobs :

Tasks scheduling distribution :

HYPER_LOCAL == 18631
HOST_LOCAL == 2105
RACK_LOCAL == 1914
ANY == 1830

So predominently HYPER_LOCAL tasks (this is inspite of running with -Dspark.tasks.schedule.aggression=ANY !)

Regards,
Mridul

ash211:
I might consider calling it something like instance local over hyper local to imply that it's local to the same instance vs the same host or same rack.  Hyper local doesn't seem like the most descriptive name to me.

pwendell:
Hi, is the overall intent here discussed anywhere? I'm a little confused why in the YARN context a user would want to run multiple spark slaves on the same host within a single SparkContext - (why not just run one larger slave)? Is this discussed already at a higher level in some other PR or JIRA? Thanks.

ash211:
I'm not sure what @mridulm's use case is, but my biggest use case is huge-RAM machines.  I have a machine with 192GB of RAM, and want to dedicate the entire thing to running Spark. But putting all that into one heap is asking for GC disaster.  So I'd like to run several instances of Spark on the same machine, each with say a 32GB heap.

Spark has the ability to take into account data locality when running jobs.  You'd rather process data where it is than ship it somewhere else to be processed on a different machine.  You'd rather ship data within the same rack than across rocks (rack local), within the same machine rather than across machines (host local).  But if you have multiple instances of Spark on one machine, you'd rather ship data within an instance than across instances (instance local).

I see it as somewhat similar to [Cassandra's virtual nodes](http://www.datastax.com/dev/blog/virtual-nodes-in-cassandra-1-2).  You also get the advantage of better supporting heterogeneous machine specs (2x32GB heaps on one machine but 4x32GB heaps on another).

core/src/main/scala/spark/SparkEnv.scala
```diff
@@ -162,7 +177,7 @@ object SparkEnv extends Logging {
       blockManager,
       connectionManager,
       httpFileServer,
-      sparkFilesDir)
+      sparkFilesDir,
+      null)
```
markhamstra: I don't believe that SparkEnv is ever directly exposed to non-Scala code, so is there a good reason to use a nullable var instead of var executorIdToHostPort: Option[(String, String) => String] initialized to None here, with the null check in resolveExecutorIdToHostPort becoming Option.getOrElse?  (Ideally, I'd also like to see stronger typing than 'String', but it is pretty late in the game to start requiring that level of type safety in Spark.)

mridulm:
To give consolidated response to the queries above :

a) @ash211 instance local vs hyper local.
I am fine with renaming it to something more appropriate - hyper local just happened to be how we were naming it internally :-) As you mentioned, outside of our context, probably not very descriptive !

b) @pwendell About multiple slaves per node.
Paraphrasing how resource allocation in YARN works (apologies for lack of sufficient detail, more detailed descriptions exist in yarn docs) :
You express resource constraints - N slaves of X cpu's (not yet supported by us) and Y ram and correspondingdata locality constraints for the job.
Depending on the cluster config, availability of resources, etc - YARN allocates and assigns containers to the application - and you can end up with multiple containers on same node based on resource node configuration.
So the application as such does not have control over which containers will be assigned to it - YARN will ensure that the resource constraints are satisfied for the containers returned, and will try to make a best case effort at maximising satisfaction of locality constraints.

So the assumption that application can always 'occupy' the entire node is not practical (particularly when there are quota restrictions ! It might not even be possible to do so).
Hope this clarifies.

c) @ash211 About huge ram machines, etc.
Our initial motivation was actually (b) :-) To support running spark on YARN required for us to support multiple slaves per node.
But once the initial prototype was done, as you very rightly mentioned, we could extend it to other requirements.
The initial patch did not have support for hyper local scheduling due to the initial fragility and necessity of a bunch of hacks to get it working. But now that it has gone through multiple reviews, and is much more stable, based on @rxin feedback, I re-did our earlier work and ported it to trunk. 

Please note that there are side effects to having multiple containers on the same node - particularly default gc configuration tries to use all core on a node - which, in mulch-tennet deployments would have unwanted side-effects (unless you move to a yarn deployment which uses cpusets).
There are config params which can be used to alleviate this though (to control gc).

core/src/main/scala/spark/SparkEnv.scala
mridulm: Looks like a good idea, I will make the change, thanks.
But in general, is this a good idea to add this to SparkEnv ? Or is there a better way to accomplish this ?

I am not very happy with the need to do executorIdToHostPort in SparkEnv ... and I am hoping I am missing something simpler which already exists and works on the master.

rxin:
Yes - I think we should rename this to instance local or process local, rather than "hyper" local.

There are a lot of benefits from running multiple JVMs on the same node. For example, there is a limit on the amount of memory a single JVM process can handle. Another example, as reported in the Tachyon paper, is that the out-of-process memory read throughput (ramdisk memory map) is limited in a single JVM (9.5GB/s using a single JVM versus 26GB/s using multiple JVMs).

One paper worth reading on this topic is

http://software.intel.com/sites/default/files/m/0/9/9/8/2/22719-Java_Application_Server_Optimization_for_Multi-Core.pdf

mateiz:
I'd recommend renaming this to PROCESS_LOCAL, and renaming HOST_LOCAL to NODE_LOCAL (node is a more commonly used name for a machine in the rest of the code). I agree that there's a huge benefit to multiple JVMs per node, though interestingly I think the sweet spot will be to have 3-4 corers per JVM, because that gives a lot more scheduling flexibility (more threads that can access the same pool of data). This might change for serialized data, which we can memory-map into every JVM on the node.

mridulm:
Have made changes mentioned above -
- Renaming from hyper local to instance local.
- Change to Option from use of null

Please let me know if I am missing anything, or there are additional comments and suggestions !
In particular, please do advise on how to proceed with ZippedRDD and whether changes to SparkEnv are fine.

Thanks,
Mridul

mridulm:
Did not see Matei's comment - will make the changes mentioned.

mridulm:
Have made changes mentioned above -
- Renaming from hyper local to process local.
- Renaming from host local to node local.
- Change to Option from use of null

Restating from above :
Please let me know if I am missing anything, or there are additional comments and suggestions !
In particular, please do advise on how to proceed with ZippedRDD and whether changes to SparkEnv are fine.

Regards,
Mridul

mateiz:
Hey Mridul,

For ZippedRDD, I think we should return a location where both parent partitions are available, if possible, or otherwise a location where any of them is available. The current code did an intersect() which is probably slightly wrong (it's only handling the first case). I would do the following:
- If both are on the same host:port, return that
- If both are on the same host (but not host:port), return the port that one of them is on, or just the host if no such port is set.
- Otherwise, return any of the parent partitions' locations.

For SparkEnv, the change seems fine, though I'll think about whether there's a better way to do this.

Thanks for taking a look at this!

mridulm:
I have fixed it as per your idea Matei.
If there is anything we can do to improve, please do let us know.

Thanks !

shivaram:
ZippedPartitionsRDD uses a similar intersect call right now. It would be great if you could also fix that using the same approach. Thanks !

mridulm:
I am not able to find this class -

$ find core -type f -name '*.scala' | xargs grep 'getPreferredLocations' | grep -i zip | grep -i parti
core/src/main/scala/spark/rdd/ZippedRDD.scala:  override def getPreferredLocations(s: Partition): Seq[String] = {

Is this present in master ?

shivaram:
It is in the master branch at https://github.com/mesos/spark/blob/master/core/src/main/scala/spark/rdd/ZippedPartitionsRDD.scala#L40

mridulm:
I think this was updated in master after I had pulled off it - thanks for pointing it out !
Did a merge from master to get to it.

Unfortunately, I am not able to make much sense of that method : my scala knowledge is fairly limited.
Would it be possible for someone else who understands that code better to take a crack at it ?

Thanks,
Mridul

shivaram:
Okay. I'll fix it sometime later - Actually as the rdd has multiple parents that method can be a little more clever and prefer locations where most parent rdds are located etc.

mridulm:
Probably my response was influenced too much by my headache :-) My
apologies.
But the impression I got was, that this rdd tries to do an intersection of
all preferred locations of all parents it has : and as number of parents
increase, I was not sure if there will be any preferred location (at host
local level).

And since we have not yet added rack locality to rdd's preferred location
yet (and probably should not); I was not sure if either my understanding
was incorrect, or it was supposed to handle a very niche case (we had not
encoutered) anyway.
For example, even for ZippedRDD, the colocation of two partitions on same
host goes down pretty drastically with increase in cluster size : with it
becoming a no-pref task very quickly : which obviously has a performance
penalty.

My hope was that sometime (as part of some future work), if there was no
preferred location for ZippedRDD, then based on relative sizes of
partitions (if this was inexpensive to compute), simply pick the largest
host.
Ideally we might also want to weight it against rack locality.

Regards,
Mridul

mridulm:
@shivaram I have added the basic implementation - which does full intersection (instance local first and then node local).
I did add comments to both ZippedRDD and ZippedPartitionsBaseRDD to describe the exact problem you mention above (though did not see it until now :-) ).

For now, that is something which is left as a future enhancement unfortunately (ideally, imo, would need rdd block size estimate to do a good job of it - which might not be available all the time (if I understand correctly); so brings up a bunch of corner cases to deal with that, etc).

Please do take a look at it though - in case I am misinterpreting the code or messing something up !

mridulm:
There is a test failure, though I dont think it is related to this PR.
- unpersist RDDs **\* FAILED ***
  java.util.NoSuchElementException:
  at spark.util.TimeStampedHashMap.apply(TimeStampedHashMap.scala:64)
  at spark.storage.StorageUtils$$anonfun$6.apply(StorageUtils.scala:67)
  at spark.storage.StorageUtils$$anonfun$6.apply(StorageUtils.scala:58)
  at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:233)
  at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:233)
  at scala.collection.immutable.MapLike$$anon$2$$anonfun$foreach$3.apply(MapLike.scala:109)
  at scala.collection.immutable.MapLike$$anon$2$$anonfun$foreach$3.apply(MapLike.scala:109)
  at scala.collection.immutable.Map$Map1.foreach(Map.scala:119)
  at scala.collection.immutable.MapLike$$anon$2.foreach(MapLike.scala:109)
  at scala.collection.TraversableLike$class.map(TraversableLike.scala:233)

I have added a note to the unpersist RDD commit PR too ..

mateiz:
With the introduction of multiple levels of locality, it might actually be good in the long term to add a "Location" class that can be either a rack, node, or process, and make getPreferredLocations return a list of those. But we can handle that later.

shivaram:
Thanks for updating zipPartitions. The change looks good to me with respect to your description. The only comment I have is that I interpreted Matei's original suggestion to mean - use host local intersection if the instance local intersection is empty. Right now it looks like we are appending the instance local intersections and any other host local ones (on different hosts) ?

mridulm:
We will try exact intersection first (which would consider instance local
or node local) : this is the best we can get from the info provided by
rdd's.
Next, to account for instance local blocks where we would also need to
consider node local; we go over the same list, remove everything we
considered above, project only the hostname, and then do an intersection of
the lists (which contains only hostname).
So if two different instance on a single node contain all relevant blocks,
this will make it node local - while if all blocks are in same instance, it
will qualify only for instance local -> for all the block locations.
At schedule time, we give higher priority to instance local scheduling
compared to host local scheduling.

This is to ensure that in case an instance or host already is computing
other tasks and has no available cpu's, we can ensure that it does not need
to wait needlessly when the cost of scheduling across instances is fairly
low (and relatively quick).

Actually in our experiments, the main slowdown in spark was the inability
to make use of the workers efficiently (which is why we added an scheduling
'aggression' parameter) : due to the wait before considering non host local
options.

Regards,
Mridul

mridulm:
Hi Matei, 

  Is there anything pending on this PR ? Not sure if I missed anything.
If there is something to be fixed/changed, please do let us know !

Thanks,
Mridul

mridulm:
Fixed a couple of bugs which slipped through the initial merge. The commit message(s) should hopefully clarify.

mateiz:
Hey Mridul,

Sorry for the late reply, I've been traveling again. I looked at this more carefully and I'm wondering about one thing: should we add a "scheduling aggression" parameter of PROCESS_LOCAL too, and add a locality wait for it in TaskSetManager? I think this is necessary because otherwise we will greedily schedule tasks to non-process-local executors on the same host instead of waiting for a process-local one to finish. Am I missing something on this front?

mridulm:
Hi Matei,

  The scheduling aggression is to do the reverse actually - minimize waiting for optimal scheduling location and ensure that tasks get pushed to workers as soon as possible : tradeoff IO with faster scheduling.
The way TaskSetManager and ClusterScheduler work right now - even if one task gets scheduled optimally, a further wait-time is introduced before others can be scheduled non-optimally : which means with higher number of tasks and higher number of nodes, the number of workers & unscheduled tasks which sit around idle goes up drastically; with very severe performance degradation (even if wait time is 1 ms : cumulatively, the effect is very bad). This parameter allows us to control this tradeoff.

Note that within a collection of one or more available nodes, the process local tasks will get picked first : followed by node local. But since the cost of transferring data between both is minimal in comparison to dong so over the wire; we sort of treat them as the same - as opposed to anything else which goes over the wire for rack local, etc.

Additionally, as soon as the Codec dfs bug is available on all hadoop distributions, we can enable multiple cores per slave : so it would alleviate this problem a bit (each worker will have multiple cores, and so more slots for better process local scheduling available).

Ofcourse, this reflects our usecases and experiments : and we have, unfortunately, not tested the variety of other scenarios spark is used under !

Thanks,
Mridul

mateiz:
The problem is that the cost of transferring data is not minimal. In some applications we tested, serializing the data took a lot longer than the computation. So I'd prefer to have a wait option for instance-local scheduling. You can just set it to 0 in your cluster if it doesn't make a difference.

On YARN, the performance of waiting for instance-local scheduling is probably bad only because you have one core per worker. In all our other workloads, our current node-local scheduling does just fine, because you only need to wait for one of the 4-8 cores to free up, as you said.

By the way, maybe another way to control this would be to have three wait times: process-local, node-local and rack-local. We would then go "up" and "down" in allowed locality level according to these, and you could set some of the to 0 if you want to ignore that level. But for now, just adding that aggression option would probably be fine; I can implement this later. This multiple wait time design is how the Hadoop fair scheduler worked too.

mridulm:
Ah ! This is indeed the disconnect : in almost all our usecases, the data to be processed is larger than the amount of memory available in the cluster (and sometimes, much larger) - or will become larger within a few iterations.
So serialization cost is something we have to pay just to run the job : given this, the overhead we see from process local to node local is negligible when compared to going across nodes (ofcourse, there are other factors like polluting cache, etc - ignoring those).

What complicates things is that not all tasks have process local granularity of preferred location - but just have node local (hdfs blocks, inferred locality in ZippedRDD, etc for example). So adding this would be tricky at best.

Can we defer addition of this requirement to a later PR ? What we currently have here improves scheduling without regressions (afaik) - while exposing point of extensions which we can use to enhance scheduling further.

Regards,
Mridul

rxin:
Mridul,

Do you know what the breakdown is for jobs that are too big to fit in
memory vs jobs that can't fit in memory after a few iterations. If you use
the newly checked-in unpersist feature, would the second class of jobs
still have problems fitting data in memory? And what if I give you the
ability to do update in place?

mridulm:
Hi Reynold,

  The implementations are designed to work on datasets which need not fit into memory.
It just so happens that for some of our experiments, the train dataset is small enough to fit into memory - but depending on the algo : this remains so until end of the job or just for the first few iterations before they dont anymore (decision tree on spark for example).
Additionally, given length of time the jobs run for, we are more interested in ensuring fault tolerance since quite a few containers and/or nodes do fail in our larger clusters.

While it might be possible to optimize it for the higher memory usecase, currently the implementations are not targeting that ... if things do fit into memory, even given the serialization cost, the lack of needing to hit the disk to read is a good performance boost : while ensuring graceful degradation when it has to hit the disk. 

Btw, I do not actually do the ML implementations, so can speak only so much with authority on that subject :-)
I will pass this on to others !

Regards,
Mridul

mateiz:
Thanks Mridul, I've now merged this. We can implement the instance-local delay later since that doesn't have an impact in non-YARN clusters.

PR #590, opened 2013-05-01
Refactor Kryo serializer to use the Twitter chill library

ryanlecompte:
This pull request refactors the existing KryoSerializer to use Twitter's [chill](https://github.com/twitter/chill) Kryo library. This significantly simplifies Spark's Kryo serializer implementation and removes a lot of the hacks around handling Scala's maps and other types. We also found an issue with the existing Kryo serializer implementation where ragged arrays were not serialized / deserialized properly. A test case for this has been added. I've tested this new implementation extensively on top of the 0.7 branch and it performs well in our testing. If accepted, it would be nice to apply these changes not only to master but also for the upcoming 0.7.1 release if possible.


core/src/test/scala/spark/KryoSerializerSuite.scala
```diff
@@ -106,35 +103,35 @@ class KryoSerializerSuite extends FunSuite {
...
 package test {
-  case class CaseClass(i: Int, s: String) {}
-  
-  class ClassWithNoArgConstructor {
-    var x: Int = 0
-    override def equals(other: Any) = other match {
-      case c: ClassWithNoArgConstructor => x == c.x
-      case _ => false
-    }
+case class CaseClass(i: Int, s: String) {}
```
rxin: Indent this - and while you at it, perhaps change package test to some more meaningful name object KryoTest?

core/src/main/scala/spark/KryoSerializer.scala
```diff
@@ -2,24 +2,15 @@ package spark
...
 import java.io._
 import java.nio.ByteBuffer
-import java.nio.channels.Channels
-
-import scala.collection.immutable
-import scala.collection.mutable
-
-import com.esotericsoftware.kryo._
-import com.esotericsoftware.kryo.{Serializer => KSerializer}
+import com.twitter.chill.KryoBijection
```
rxin: Mind moving this after esotericsoftware since t comes before e?

core/src/main/scala/spark/KryoSerializer.scala
```diff
@@ -91,130 +79,55 @@ private[spark] class KryoSerializerInstance(ks: KryoSerializer) extends Serializ
...
-      logInfo("Running user registrator: " + regCls)
-      val classLoader = Thread.currentThread.getContextClassLoader
-      val reg = Class.forName(regCls, true, classLoader).newInstance().asInstanceOf[KryoRegistrator]
-      reg.registerClasses(kryo)
+    try {
+      Option(System.getProperty("spark.kryo.registrator")).foreach { regCls =>
+        logInfo("Running user registrator: " + regCls)
+        val reg = Class.forName(regCls, true, classLoader).newInstance().asInstanceOf[KryoRegistrator]
+        reg.registerClasses(kryo)
+}
```
rxin: The closing curly bracket is not indented.

rxin:
Thanks, Ryan. Looks good in general. I might have to run this on Shark to make sure there is no problem on that end.

ryanlecompte:
Thanks for taking a look! I just made a commit to address your review
comments. Thanks.

ryanlecompte:
FYI - Just updated this to use ThreadLocal again. This was lost in the new implementation and was causing unnecessary instantiations of the Kryo-related objects to occur. This is fixed now.

rxin:
Ryan, I tried this on Shark and saw the following error:

java.lang.NoSuchMethodError: com.esotericsoftware.kryo.Kryo.setInstantiatorStrategy(Lorg/objenesis/strategy/InstantiatorStrategy;)V
    at com.twitter.chill.KryoBase.setInstantiatorStrategy(KryoBase.scala:74)
    at com.twitter.chill.KryoBijection$class.getKryo(KryoBijection.scala:39)
    at com.twitter.chill.KryoBijection$.getKryo(KryoBijection.scala:33)
    at spark.KryoSerializer.newKryo(KryoSerializer.scala:96)
    at spark.KryoSerializerInstance.<init>(KryoSerializer.scala:44)
    at spark.KryoSerializer.newInstance(KryoSerializer.scala:130)
    at spark.broadcast.HttpBroadcast$.write(HttpBroadcast.scala:113)
    at spark.broadcast.HttpBroadcast.<init>(HttpBroadcast.scala:28)
    at spark.broadcast.HttpBroadcastFactory.newBroadcast(HttpBroadcast.scala:54)
    at spark.broadcast.HttpBroadcastFactory.newBroadcast(HttpBroadcast.scala:50)
    at spark.broadcast.BroadcastManager.newBroadcast(Broadcast.scala:50)
    at spark.SparkContext.broadcast(SparkContext.scala:469)
    at spark.rdd.HadoopRDD.<init>(HadoopRDD.scala:50)
    at spark.SparkContext.hadoopRDD(SparkContext.scala:280)

rxin:
Although, I thought it'd be good to initialize the buffer (input vs output) on each new kryo instance, but we can probably use a thread local kryo object for reuse. What is the performance problem you saw?

ryanlecompte:
Hey Reynold,

I've seen this before when there is a conflicting version of the underlying
Kryo library in the classpath. Can you make sure that kaffee isn't
specified as an explicit dependency and that only "chill" is listed?

ryanlecompte:
I found in testing that using a ThreadLocal version vs. creating a new
instance every time yielded much worse performance. In my testing creating
a new instance each time yielded the best performance. I don't have exact
numbers on hand right now, but I recall the job being much slower when I
switched it to using ThreadLocal.

rxin:
This is pretty strange. When I did publish-local with your changes from Spark, and then use Shark to pull in the 0.8.0-SNAPSHOT, I always see the de.javakaffee jar in lib_managed. I already rm -rf ~/.m2 ~/.ivy2 many times, and still can't get rid of the jars. Is there a hidden dependency somewhere ...

```
> find . -name "*kryo*"
./lib_managed/bundles/de.javakaffee/kryo-serializers
./lib_managed/bundles/de.javakaffee/kryo-serializers/kryo-serializers-0.22.jar
./lib_managed/jars/com.esotericsoftware.kryo
./lib_managed/jars/com.esotericsoftware.kryo/kryo
./lib_managed/jars/com.esotericsoftware.kryo/kryo/kryo-2.20-shaded.jar
```

ryanlecompte:
Hmm, it may be some hidden dependency with another library in 0.8 (master)
branch. I don't see this issue at all on 0.7 (we have this patch applied to
the 0.7 branch in production here). In 0.7 I see this:

➜  lib_managed git:(qf-0.7.0-kryo)   find . -name "_kryo_"
./jars/kryo-2.17.jar

I'll try to see where the dependencies changed from 0.7 to 0.8.

jdanbrown:
Maybe https://github.com/jrudolph/sbt-dependency-graph would be helpful to debug the transitive dependencies? It produces output similar to [`mvn dependency:tree`](https://maven.apache.org/plugins/maven-dependency-plugin/examples/resolving-conflicts-using-the-dependency-tree.html).

ryanlecompte:
Great suggestion, Dan. I generated the dependency graph and actually didn't
see de.javakaffee anywhere in the list, just chill and the dependency on
the core underlying kryo library. Here's a link to the generated tree:
https://gist.github.com/ryanlecompte/5608353

rxin:
bummer. It is Tachyon.

[info]   +-org.tachyonproject:tachyon:0.2.1
[info]     +-commons-io:commons-io:2.4
[info]     +-de.javakaffee:kryo-serializers:0.22
[info]     | +-com.esotericsoftware.kryo:kryo:2.20
[info]     |   +-com.esotericsoftware.minlog:minlog:1.2
[info]     |   +-com.esotericsoftware.reflectasm:reflectasm:1.07
[info]     |   | +-org.ow2.asm:asm:4.0
[info]     |   | 
[info]     |   +-org.objenesis:objenesis:1.2
[info]     |   

@haoyuan - how come Kryo is used in Tachyon?

haoyuan:
Tachyon master uses Kryo. In the future, we will separate master and client jar, and the frameworks using client dont need to worry about many dependency issue anymore. For now, you can let sbt to not link the Tachyon's kryo dependency from shark's build file.

ryanlecompte:
Ah, makes sense! Let me know if I should modify this pull request in any way, although it seems like we can get around the issue by modifying Shark's build file like Haoyuan suggested.

rxin:
Ryan,

Haoyuan was suggesting modifying Tachyon's build file so Shark only needs to depend on the client jars, which won't contain kryo. 

It worked once I removed Shark's dependency on Tachyon - but this is only to verify the Spark patch would work, and is not a realistic solution. 

I took a look at the Chill library. It looks like it does a great job with the scala classes, but not so much for Java collections. For example, this particular one is missing:

http://grepcode.com/file/repo1.maven.org/maven2/de.javakaffee/kryo-serializers/0.9/de/javakaffee/kryoserializers/ArraysAsListSerializer.java

ryanlecompte:
That's a great point. To be honest our primary use case has been purely Scala projects, so support for serializing Java collections never really affected us in testing this pull request. @sritchie and @johnynek might have some more ideas here. Is the goal of chill to only provide support for Scala collections, or should the library also provide support for Java collections (e.g., http://grepcode.com/file/repo1.maven.org/maven2/de.javakaffee/kryo-serializers/0.9/de/javakaffee/kryoserializers/ArraysAsListSerializer.java) like the de.javakaffee library does? In theory you could have chill and javakaffee running side-by-side, however they both rely on the underlying com.esotericsoftware.kryo library and may not always require the same version (as is the case that you initially ran into, Reynold).

johnynek:
So, chill can easily add some java specific serializers (we have talked about chill-java).  That said, I think that Kryo has collection serializers registered by default now and I think this javakaffee stuff may not be needed any more.

We're happy to make chill a chokepoint for a lot of serialization work (we are planning on doing chill-guava for instance) so it would be great to work together on this.

ryanlecompte:
Thanks for chiming in on the issue, @johnynek. @rxin, do you have any specific Java collections in mind that you'd like to test (or even better, have any existing jobs / test cases that would possibly fail with chill and not with javakaffee)? Let me know if there's something that I can do to make this pull request work for Spark. I definitely would prefer having chill be the core "engine" behind Spark's kryo serializer if at all possible.

rxin:
I tested the Arrays.asList serialization. Without de.javakaffee.kryoserializers.ArraysAsListSerializer, the current Kryo would still fail:

```
object Test {
    System.setProperty("spark.kryo.registrator", classOf[KryoRegistrator].getName)
    val serializer = new spark.KryoSerializer
    val a = java.util.Arrays.asList[Int](1, 2)
    val d = serializer.newInstance().serialize(a)
    println(serializer.newInstance().deserialize(d))
  }
}
```

This prints to stdout

```
13/05/22 22:29:50 INFO spark.KryoSerializer: Running user registrator: shark.KryoRegistrator
[1, 2]
```

But if I comment out "System.setProperty("spark.kryo.registrator", classOf[KryoRegistrator].getName)",

```
> ./run shark.Test
java.lang.NullPointerException
at java.util.Arrays$ArrayList.size(Arrays.java:3362)
at java.util.AbstractList.add(AbstractList.java:91)
at com.esotericsoftware.kryo.serializers.CollectionSerializer.read(CollectionSerializer.java:109)
at com.esotericsoftware.kryo.serializers.CollectionSerializer.read(CollectionSerializer.java:18)
at com.esotericsoftware.kryo.Kryo.readClassAndObject(Kryo.java:732)
at spark.KryoSerializerInstance.deserialize(KryoSerializer.scala:68)
at shark.Test$.main(Test.scala:10)
at shark.Test.main(Test.scala)
at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:39)
at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:25)
at java.lang.reflect.Method.invoke(Method.java:597)
at scala.tools.nsc.util.ScalaClassLoader$$anonfun$run$1.apply(ScalaClassLoader.scala:78)
at scala.tools.nsc.util.ScalaClassLoader$class.asContext(ScalaClassLoader.scala:24)
at scala.tools.nsc.util.ScalaClassLoader$URLClassLoader.asContext(ScalaClassLoader.scala:88)
at scala.tools.nsc.util.ScalaClassLoader$class.run(ScalaClassLoader.scala:78)
at scala.tools.nsc.util.ScalaClassLoader$URLClassLoader.run(ScalaClassLoader.scala:101)
at scala.tools.nsc.ObjectRunner$.run(ObjectRunner.scala:33)
at scala.tools.nsc.ObjectRunner$.runAndCatch(ObjectRunner.scala:40)
at scala.tools.nsc.MainGenericRunner.runTarget$1(MainGenericRunner.scala:56)
at scala.tools.nsc.MainGenericRunner.process(MainGenericRunner.scala:80)
at scala.tools.nsc.MainGenericRunner$.main(MainGenericRunner.scala:89)
at scala.tools.nsc.MainGenericRunner.main(MainGenericRunner.scala)
```

ryanlecompte:
Your'e right - it looks like chill is using the latest version of the underlying kryo library and is unable to deserialize that Java collection:

``` scala
scala> import com.twitter.chill._
import com.twitter.chill._

scala> val a = java.util.Arrays.asList[Int](1, 2)
a: java.util.List[Int] = [1, 2]

scala> KryoInjection(a)
res0: Array[Byte] = Array(1, 0, 106, 97, 118, 97, 46, 117, 116, 105, 108, 46, 65, 114, 114, 97, 121, 115, 36, 65, 114, 114, 97, 121, 76, 105, 115, -12, 1, 2, 2, 2, 2, 4)

scala> KryoInjection.invert(res0)
res1: Option[AnyRef] = None
```

@johnynek @sritchie any thoughts? It looks like javakaffee has a bunch of custom serializers that we'd need to either support directly in chill (or a new chill-java), or perhaps just make javakaffee a dependency of chill and register all of these custom serializers directly? The serializers are here: https://github.com/magro/kryo-serializers

rxin:
For Shark, the only thing I need is the Arrays.asList. But since Spark supports Java API, it would be very useful to be able to serialize the rarer Java collections and classes out of the box.

It would make sense to either make Chill work with javakaffee, or have a chill-java, or just build the special serializations into Chill itself.

koertkuipers:
ryan,
i was excited to try this patch since we do a lot of work in scala and i am familiar with chill serialization thanks to scalding. i tried the patch on both spark 0.7.3-SNAPSHOT and spark 0.8.0-SNAPSHOT. in both cases all seems well until i try to connect to the cluster with a spark-shell (i havent tried to run a driver program yet). the executors get created but immediately disconnect from the shell with a IncompatibleClassChangeError. see below. when i revert the patch and rebuild and redistribute across the cluster the error is gone.
best,
koert

In the shell i see the executors connecting and immediately disconnecting:
13/06/15 23:49:27 INFO cluster.SparkDeploySchedulerBackend: Executor 4 disconnected, so removing it

When i then go check the logs for the executor i see:
13/06/16 00:53:35 INFO executor.Executor: Using REPL class URI: http://192.168.3.166:60567
Uncaught error from thread [sparkExecutor-akka.actor.default-dispatcher-1] shutting down JVM since 'akka.jvm-exit-on-fatal-error' is enabled for ActorSystem[sparkExecutor]
java.lang.IncompatibleClassChangeError: Implementing class
    at java.lang.ClassLoader.defineClass1(Native Method)
    at java.lang.ClassLoader.defineClassCond(ClassLoader.java:631)
    at java.lang.ClassLoader.defineClass(ClassLoader.java:615)
    at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:141)
    at java.net.URLClassLoader.defineClass(URLClassLoader.java:283)
    at java.net.URLClassLoader.access$000(URLClassLoader.java:58)
    at java.net.URLClassLoader$1.run(URLClassLoader.java:197)
    at java.security.AccessController.doPrivileged(Native Method)
    at java.net.URLClassLoader.findClass(URLClassLoader.java:190)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:306)
    at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:301)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:247)
    at java.lang.ClassLoader.defineClass1(Native Method)
    at java.lang.ClassLoader.defineClassCond(ClassLoader.java:631)
    at java.lang.ClassLoader.defineClass(ClassLoader.java:615)
    at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:141)
    at java.net.URLClassLoader.defineClass(URLClassLoader.java:283)
    at java.net.URLClassLoader.access$000(URLClassLoader.java:58)
    at java.net.URLClassLoader$1.run(URLClassLoader.java:197)
    at java.security.AccessController.doPrivileged(Native Method)
    at java.net.URLClassLoader.findClass(URLClassLoader.java:190)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:306)
    at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:301)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:247)
    at java.lang.Class.forName0(Native Method)
    at java.lang.Class.forName(Class.java:171)
    at spark.executor.Executor.createClassLoader(Executor.scala:159)
    at spark.executor.Executor.<init>(Executor.scala:39)
    at spark.executor.StandaloneExecutorBackend$$anonfun$receive$1.apply(StandaloneExecutorBackend.scala:39)
    at spark.executor.StandaloneExecutorBackend$$anonfun$receive$1.apply(StandaloneExecutorBackend.scala:36)
    at akka.actor.Actor$class.apply(Actor.scala:318)
    at spark.executor.StandaloneExecutorBackend.apply(StandaloneExecutorBackend.scala:16)
    at akka.actor.ActorCell.invoke(ActorCell.scala:626)
    at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:197)
    at akka.dispatch.Mailbox.run(Mailbox.scala:179)

koertkuipers:
i noticed lib_managed now has multiple versions of asm. this could have something to do with it...

not sure why lib_managed would have multiple versions of some jars. this has bitten me before (with jackson). isn't the whole idea of version management that it picks the best version for you given the requirements?

koertkuipers:
chill depends on asm 4.0 while spark uses asm 3.3.1. because of the usage of asm-all jar the dependency resolution did not see a relation and allowed both dependencies. i suspect this causes my issues. however spark code breaks against asm 4.0 (in ClosureCleaner because asm 4.0 has no more EmptyVisitor, at least thats the first issue i ran into). i think i am missing something here since ryan obviously did not run into this.

ryanlecompte:
@koertkuipers - really strange that you're running into these issues. Did you try manually applying the patch to latest master or 0.7.x? We're running this patch successfully in production. One thing to make sure of is that you remove the old reference to the other kryo library from the spark build file. Also, make sure that the classpath for your job is including the same version of Spark with this patch applied. It sounds like you may have different versions of spark or something like that.

koertkuipers:
yes i applied the patch to both latest 0.7.x and latest 0.8.x, and i
removed the de.javakaffee.kryo-serializers dependency.
the patch works fine except for a minor incompatibility in
project/SparkBuild.scala which i easily fixed by hand.

i was running spark-shell as a test on same machine as master from same
build, so that should mean i used exactly same version of spark, however i
will double check now to make sure i didnt mess that up (we do have other
versions laying around).

what do you have in lib_managed?
i see this for asm:

$ ls lib_managed/jars | grep asm
asm-4.0.jar
asm-all-3.3.1.jar
asm-commons-4.0.jar
asm-tree-4.0.jar
reflectasm-1.07-shaded.jar

chill is pulling is asm 4.0 while spark-core depends on asm 3.3.1. i
suspect that is the issue but that is just a guess.

ryanlecompte:
Here's our lib_managed (sorry for the bad formatting!):

``` scala
akka-actor-2.0.3.jar              commons-cli-1.2.jar                  core-3.1.1.jar                 jets3t-0.7.1.jar                        kfs-0.3.jar                   servlet-api-2.5-6.1.14.jar
akka-remote-2.0.3.jar             commons-codec-1.7.jar                dispatch-json_2.9.1-0.8.5.jar  jetty-6.1.26.jar                        kryo-2.17.jar                 servlet-api-2.5.jar
akka-slf4j-2.0.3.jar              commons-collections-3.2.1.jar        easymock-3.1.jar               jetty-continuation-7.5.3.v20111011.jar  mesos-0.9.0-incubating.jar    sjson_2.9.1-0.15.jar
ant-1.6.5.jar                     commons-configuration-1.6.jar        fastutil-6.4.4.jar             jetty-http-7.5.3.v20111011.jar          mimepull-1.6.jar              slf4j-api-1.6.4.jar
asm-4.0.jar                       commons-digester-1.8.jar             guava-11.0.1.jar               jetty-io-7.5.3.v20111011.jar            minlog-1.2.jar                slf4j-log4j12-1.6.1.jar
asm-all-3.3.1.jar                 commons-el-1.0.jar                   h2-lzf-1.0.jar                 jetty-server-7.5.3.v20111011.jar        objenesis-1.2.jar             spray-base-1.0-M2.1.jar
asm-commons-4.0.jar               commons-httpclient-3.0.1.jar         hadoop-core-1.0.4.jar          jetty-util-6.1.26.jar                   oro-2.0.8.jar                 spray-can-1.0-M2.1.jar
asm-tree-4.0.jar                  commons-io-1.4.jar                   hsqldb-1.8.0.10.jar            jetty-util-7.5.3.v20111011.jar          parboiled-core-1.0.2.jar      spray-io-1.0-M2.1.jar
bijection-core_2.9.2-0.3.0.jar    commons-lang-2.6.jar                 httpclient-4.1.jar             jsp-2.1-6.1.14.jar                      parboiled-scala-1.0.2.jar     spray-json_2.9.2-1.1.1.jar
cglib-nodep-2.2.2.jar             commons-logging-1.1.1.jar            httpcore-4.1.jar               jsp-api-2.1-6.1.14.jar                  protobuf-java-2.4.1.jar       spray-server-1.0-M2.1.jar
chill_2.9.2-0.2.2.jar             commons-math-2.1.jar                 jackson-core-asl-1.0.1.jar     jsr305-1.3.9.jar                        reflectasm-1.07-shaded.jar    spray-util-1.0-M2.1.jar
colt-1.2.0.jar                    commons-net-1.4.1.jar                jackson-mapper-asl-1.0.1.jar   junit-3.8.1.jar                         scalacheck_2.9.2-1.9.jar      test-interface-0.5.jar
commons-beanutils-1.7.0.jar       concurrent-1.3.4.jar                 jasper-compiler-5.5.12.jar     junit-4.8.2.jar                         scalatest_2.9.2-1.8.jar       twirl-api_2.9.2-0.5.2.jar
commons-beanutils-core-1.8.0.jar  concurrentlinkedhashmap-lru-1.2.jar  jasper-runtime-5.5.12.jar      junit-interface-0.8.jar                 servlet-api-2.5-20081211.jar  xmlenc-0.52.jar
```

koertkuipers:
thanks. did a diff from your lib_managed to mine. i believe all the changes are simply updates in spark, so nothing suspicious. ok i will go double check i wasn't mixing spark versions...

```
> akka-zeromq-2.0.3.jar
> algebird-core_2.9.2-0.1.11.jar
8a11,12
> avro-1.6.3.jar
> avro-ipc-1.6.3.jar
22a27
> commons-io-2.1.jar
32a38
> flume-ng-sdk-1.2.0.jar
35a42
> hamcrest-core-1.1.jar
39,40c46,47
< jackson-core-asl-1.0.1.jar
< jackson-mapper-asl-1.0.1.jar
---
> jackson-core-asl-1.9.3.jar
> jackson-mapper-asl-1.9.3.jar
42a50
> JavaEWAH-0.6.6.jar
45,48c53,56
< jetty-continuation-7.5.3.v20111011.jar
< jetty-http-7.5.3.v20111011.jar
< jetty-io-7.5.3.v20111011.jar
< jetty-server-7.5.3.v20111011.jar
---
> jetty-continuation-7.6.8.v20121106.jar
> jetty-http-7.6.8.v20121106.jar
> jetty-io-7.6.8.v20121106.jar
> jetty-server-7.6.8.v20121106.jar
50c58,61
< jetty-util-7.5.3.v20111011.jar
---
> jetty-util-7.6.8.v20121106.jar
> jline-0.9.94.jar
> jna-3.0.9.jar
> jnr-constants-0.8.2.jar
55,56c66,67
< junit-4.8.2.jar
< junit-interface-0.8.jar
---
> junit-dep-4.8.2.jar
> junit-interface-0.9.jar
63a75
> paranamer-2.3.jar
68,69c80,81
< scalacheck_2.9.2-1.9.jar
< scalatest_2.9.2-1.8.jar
---
> scalacheck_2.9.2-1.10.0.jar
> scalatest_2.9.2-1.9.1.jar
70a83
> servlet-api-2.5-20110124.jar
72d84
< servlet-api-2.5.jar
83c95,98
< twirl-api_2.9.2-0.5.2.jar
---
> twirl-api_2.9.2-0.6.1.jar
> twitter4j-core-3.0.3.jar
> twitter4j-stream-3.0.3.jar
> velocity-1.7.jar
84a100,102
> zeromq-scala-binding_2.9.1-0.0.6.jar
> zkclient-0.1.jar
> zookeeper-3.3.3.jar
```

koertkuipers:
i did a pseudo-distributed installation in a VM and logged in with spark-shell inside the VM and i get the same error. so its not a mixing of spark versions i think.

koertkuipers:
in a clean pseudo distributed VM spark-0.7-SNAPSHOT works fine with a master, slave and a spark-shell

if i then apply the patch the master and slave still run, but the spark-shell causes the executors to bomb with the error described.

if i then simply go in and remove asm-4.0.jar, asm-commons-4.0.jar and asm-tree-4.0.jar from lib_managed/jars and restart the spark daemons, the executors no longer bomb (but kryo will not work i assume, haven't tested this).

so from this it seems to me it's a asm versions conflict causing this. the most logical fix would be to upgrade spark to asm 4 which would involve changes in ClosureCleaner

sritchie:
Ah! Yes, this makes sense. ASM changed its jar org name up the upgrade to 4.0, which is hell for dependency resolution.

We've actually cleaned up the `ClosureCleaner` and upgraded to ASM 4.0 in `chill`:

https://github.com/twitter/chill/blob/develop/chill-scala/src/main/scala/com/twitter/chill/ClosureCleaner.scala#L46

Would it make sense to upgrade ClosureCleaner usages to this version in this pull req?

ryanlecompte:
Ah, interesting! I think if we are going to rely on chill, then it might make sense to use the chill ClosureCleaner too. Did you guys make any further improvements over Spark's version?

Regardless, I'm still a bit puzzled as to how this is currently working without problems for us.

koertkuipers:
i think move to the closusre cleaner in chill , and add a chill-java project to deal with @rxin 's need for handling serialization of Arrays.asList

koertkuipers:
this should probably target spark 0.8.0?

mateiz:
I'd like to understand the changes made in the closure cleaner before including it. Since it's such an important thing to get right, I don't want to depend on an external library for it unless it will clearly evolve in a reasonable way and continue working with all the tests and use cases we have (e.g. closures from Spark's modified REPL). So let's do that as a separate pull request.

However, if you guys want Spark to depend on ASM 4.0 instead of 3.x so we can depend on Chill, that would be fine. I can make that change separately. When will Chill have a release we can link to that includes Java support?

johnynek:
@mateiz The closure cleaner does not appear to do what we think it should. You don't have to use ours. I think changes to the way scala compiles the code broke an assumption of the earlier one I copied from you guys. We need more tests to verify it before I trust it.

If you guys spell out what is needed, I can be sure that chill supports that in the next release (2 weeks).

I really think collaborating on the kryo and serialization issues will be a win for all of us.

mateiz:
Hey Oscar,

I think a good start would just be to have chill-java in there, depending on ASM 4 as it does now. We can figure out the closure stuff later, since it's always possible to switch to a newer one. I'd love to use Chill for serialization though.

For the cleaner stuff, I'm happy to talk about it sometime. It definitely has limitations, especially when you use it inside a class, but it was a very difficult thing to get right in general once you escape to calling methods from the external class. Our main approach has been to have a clear recommendation on how to avoid the issues (namely defining local vals). I was also hoping that in Scala 2.10 or 2.11 we would use macros or another new Scala feature (spores) to capture the environment.

mateiz:
(And just to be clear on what I mean by chill-java, it's basically serializers for Java collection types, and specifically Arrays.asList for Shark; it's what was covered by the de.javakaffee/kryo-serializers project mentioned above.)

koertkuipers:
Kryo will handle most java collections very well out of the box. For the
notable exceptions such as Arrays.asList we could add a chill-java project
(the logical choice in long term) or as a stop-gap just add a dedicated
kryo serializer to Spark for now.

The biggest win of chill will be how it serializes scala collections (and
Option, Either, singletons, etc.) as Kryo by itself does not handle those
well.

To be able to use chill the patch is ready to go, but spark's
ClosureCleaner will need to be worked on to compile against ASM 4 (there
are breaking changes). Basically the patch needs one more change (remove
the ASM 3 dependency) and then ClosureCleaner no longer compiles.

koertkuipers:
To be clear: the patch currently compiles, but that's only because both ASM
3 and ASM 4 are in the classpath (due to a change in the org of the
artifact). but spark does not actually run like that (at least not when i
tried it). So once you remove the ASM 3 dependency ClosureCleaner no longer
compiles.

So to get the patch to pass (and not break Shark) we have to 1) add
serialization support for Arrays.asList and 2) make ClosureCleaner compile
against ASM 4

mateiz:
OK, got it. So if you guys release a chill-java in the next Chill version, I'm going to work on the ASM 4 stuff here and make sure this pull request becomes mergeable.

squito:
just fyi, I was talking about this with Ryan, and we realized why we haven't had any issues w/ serializing closures.  We should have realized this earlier, the reason we didn't have problems is b/c we are still using the Java serializer for closures.  We forgot that the closure serializer was controlled by a different param, `spark.closure.serializer`.  @koertkuipers you could try using this patch and only setting `spark.serializer`.  We've been using this patch for a while now like that and haven't had any issues.

koertkuipers:
do you use the spark-shell at all? my main issue was that the spark-shell
simply blew up all executors immediately. i didn't even get to any jobs yet

squito:
We actually haven't run the repl at all with that patch -- turns out we haven't even deployed the repl project :)  Our tests have entirely been with jobs we already had written.  Can you try running a non-repl program?  I'll also try out the repl when I get a bit of time to deploy it.

We definitely want to get this to work w/ the repl & for closures, but for the moment I just want to figure out why this works for us at all ...

rxin:
One guess is that on your deployment asm 3 was loaded  instead of 4. And
the chill code path might work with asm 3 or you have simply not triggered
the code path that would require asm 4.

koertkuipers:
if asm 3 was loaded on it's own or before asm4 i dont think chill would run...

looks to me like like ClosureCleaner and ExecutorClassLoader (in the REPL) are the only pieces that use ASM 3 in spark. so my guess is that they somehow compiled against a mixture of ASM3 and ASM4 classes but they wont run like that, and you are not using those anywhere in your deployment. and your actual deployment has ASM4 on the classpath before ASM3 which means chill works fine for you.

@mateiz ExecutorClassLoader will also need to be upgraded to ASM 4

mateiz:
Hey, so I've now updated Spark to use ASM 4.0. We should be good to go as soon as the Java stuff is also in a release.

ryanlecompte:
Awesome! Thanks for everyone's collaboration to get this pull request ready to be merged in.

johnynek:
See: https://github.com/twitter/chill/pull/70 How does this look?

I just took our old java serializers (which we also sent to magro) and the ArrayAsList serializer.  We have the same license so any others needed we will take.

I want to keep chill-java, java only (scala in tests is okay) and not depend on anything but the java stdlib.

rxin:
This looks great!

ryanlecompte:
Awesome! Once chill-java is pushed out to Sonatype I can update this pull request with that new dependency, and also register the custom Java serializers in `spark.serializer.KryoSerializer`.

koertkuipers:
i tried this patch with spark updated to ASM 4.0 and it works well. both for spark-shell and for jobs.

the real good news is that kryo serialization is now reliable for me. before i had this when i used certain scala classes in RDDs and i ran jobs on a cluster with kryo enabled:
scala> x.count
res25: Long = 12
scala> x.partitionBy(new spark.HashPartitioner(8)).count
res26: Long = 1

i suspect this had to do with silent kryo errors. now this behavior is gone :)

in SparkBuild.scala i had to add the exludeAsm to all hadoop artifacts to stop them from bringing in asm 3. although i believe this to be specific to my hadoop version i think it is generally a good idea.

ryanlecompte:
Awesome news! I'm glad this is working for you. Hopefully we can get this
merged in to master soon.

sritchie:
@ryanlecompte, do we need any further serializers in `chill-java` before we publish a version and get this in?

ryanlecompte:
@rxin, what do you think? Is chill-java currently sufficient for Shark's needs?

rxin:
Yup - the main thing for Shark was arraysAsList

ryanlecompte:
Awesome, thanks. @sritchie - we're all good for chill-java to be released. Once that's up, we can merge this pull request and then I'll submit another one that adds the chill-java dependency + java serializer registrations.

johnynek:
@mateiz or anyone else: I'd love to the closure cleaner in chill so we can use it in scalding/summingbird.  Anyone willing to make a pull req to replace ours with this one and include tests as well?

mateiz:
Sorry, I think you accidentally a word in there, but you mean can we submit ours as a PR? That would be good; I can try to do later next week. Another option is to have a separate closure cleaner project, but maybe that's overkill.

johnynek:
lol, glad you got the meaning of the message. In my experience, having really small libraries amplifies versioning and publishing pain beyond the point that it is worth it. We have been consolidating projects into medium repos of highly related code, and then sbt makes it easy to publish in bulk.

So, the win with having it in chill is that:
1) We already have an ASM dep, so keeping them the same helps diamond dependency problems.
2) closure-cleaning is related to serialization (and GC efficiency).

I guess a closure-cleaner module in chill could be useful for those that don't want the kryo dep. I'm not strongly opinionated aside from having them in one place to easily publish/depend on the same ASM.

ryanlecompte:
bump @johnynek @sritchie - any idea when we can get an initial chill-java released so we can get this pull req merged? thanks!

sritchie:
I think we're planning on getting this out today. Thanks for the bump, we're on it!

mateiz:
@sritchie Did you guys end up releasing this? I tried using Chill 0.2.3 and I still get the problem from before with serializing the result of Arrays.asList.

For reference, the problem is:

```
java.lang.NullPointerException
    at java.util.Arrays$ArrayList.size(Arrays.java:3362)
    at java.util.AbstractList.add(AbstractList.java:91)
    at com.esotericsoftware.kryo.serializers.CollectionSerializer.read(CollectionSerializer.java:109)
    at com.esotericsoftware.kryo.serializers.CollectionSerializer.read(CollectionSerializer.java:18)
    at com.esotericsoftware.kryo.Kryo.readClassAndObject(Kryo.java:729)
```

sritchie:
Hey @mateiz, sorry for the delay -- we're almost finished with one last issue we wanted to get resolved before we released the code. @johnynek and I are planning on getting the new chill out in the next week, we hope.

mateiz:
Alright, great. Just post a note here when it's done and I'll try it out.

sritchie:
@ryanlecompte, @mateiz -- we (mostly @johnynek) have just released chill 0.3.0 to sonatype. It was a bit of a refactor:

https://github.com/twitter/chill/releases/tag/0.3.0

@ryanlecompte, can I beg you for one more round on this new release? I'm really excited to see this one merged.

johnynek:
https://github.com/twitter/chill/releases/tag/0.3.0

Done.

You guys probably want to look at ScalaKryoInstantiator in the chill jar:
https://github.com/twitter/chill/blob/develop/chill-scala/src/main/scala/com/twitter/chill/ScalaKryoInstantiator.scala

Look forward to sharing serialization code. We are going to try to get cascalog and storm on this stack so we can all solve these issues once and for all.

ryanlecompte:
Thanks for doing this, guys! I'm actually on vacation for the rest of this week, but will get to this ASAP.

ryanlecompte:
I'm closing out this pull request and opening up another with the new changes for chill 0.3.0. My branch that I had based this pull req on somehow appears to be in a corrupted state.

ryanlecompte:
New pull req here: https://github.com/mesos/spark/pull/732

PR #591, opened 2013-05-01
RDD.unpersist: probably the most desired feature of Spark

rxin:
This builds upon the block removal ability I added a while ago, and also on the RDD removal code Ram added. 

I pulled his code and cleaned it up and made sure the tests are properly coded against concurrency. 

Side note: I thought this would only take 5 mins when I decided to do it. Turned out it took over 2 hrs :(


JoshRosen:
It looks like this will fix [SPARK-621](https://spark-project.atlassian.net/browse/SPARK-621).

After I call `upersist()`, should it be legal for me to call `persist()` again?  If so, should we update the comment on `persist()`, since it will no longer be strictly true that `persist()` "can only be called once on each RDD"?

mateiz:
Looks good, but two small suggestions:
- Move the body of unpersist into a BlockManagerMaster.removeRDD method; it contains a lot of implementation specific stuff that we shouldn't expose outside the BlockManager interfaces
- Don't add the parens on the get\* methods in SparkContext -- I think they'll cause old code to complain when compiling.

JoshRosen:
Maybe add an alias in JavaRDD, so this is available through the Java API?  And in PySpark too?

rxin:
k done. PTAL.

rxin:
Pretty sure the above Jenkins test error is not from this pull request.

mateiz:
Thanks for the patch.

mridulm:
The tests fail with this :
- unpersist RDDs **\* FAILED ***
  java.util.NoSuchElementException:
  at spark.util.TimeStampedHashMap.apply(TimeStampedHashMap.scala:64)
  at spark.storage.StorageUtils$$anonfun$6.apply(StorageUtils.scala:67)
  at spark.storage.StorageUtils$$anonfun$6.apply(StorageUtils.scala:58)
  at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:233)
  at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:233)
  at scala.collection.immutable.MapLike$$anon$2$$anonfun$foreach$3.apply(MapLike.scala:109)
  at scala.collection.immutable.MapLike$$anon$2$$anonfun$foreach$3.apply(MapLike.scala:109)
  at scala.collection.immutable.Map$Map1.foreach(Map.scala:119)
  at scala.collection.immutable.MapLike$$anon$2.foreach(MapLike.scala:109)
  at scala.collection.TraversableLike$class.map(TraversableLike.scala:233)

Essentially, rddInfoFromBlockStatusList assumes the rdd will always be present in sc.persistentRdds - which might not be correct ?
Since I am not sure about this codepath, I have not attempted to fix it : but saw this as part of running the tests via 
$ mvn  package

rxin:
Yea - I fixed it by removing the last line. That function was not properly implemented. It is probably useful to fix that function in the future. Meantime, let's just fix the test by not running that particular function outside of the timeout block.

https://github.com/mesos/spark/pull/598

PR #592, opened 2013-05-02
Don't accept generated local directory names that can't be created

charlesreiss:
Keep trying to create local dirs until one is created, not just that it did not exist before we attempted to create it. This should cause us to fail if spark.local.dir is unwritable and avoid a rare race if two spark executors are starting at the same time on the same machine.


mateiz:
Thanks Charles!

PR #593, opened 2013-05-03
Master UI has link to Application UI

squito:
Its a pain to go through the logs to find the link to the Storage Dashboard, so now the master UI has a link to it.

Even though its just the BlockMangerUI right now, I called it Application UI in the code & the link, b/c I assume eventually we'll want it to serve more than just storage info.

### DIFF
--- a/core/src/main/scala/spark/deploy/ApplicationDescription.scala
+++ b/core/src/main/scala/spark/deploy/ApplicationDescription.scala
@@ -5,7 +5,9 @@ private[spark] class ApplicationDescription(
     val cores: Int,
     val memoryPerSlave: Int,
     val command: Command,
-    val sparkHome: String)
+    val sparkHome: String,
+    val appUIHost: String,
+    val appUIPort: Int)
   extends Serializable {
 
   val user = System.getProperty("user.name", "<unknown>")
--- a/core/src/main/scala/spark/deploy/ApplicationDescription.scala
+++ b/core/src/main/scala/spark/deploy/ApplicationDescription.scala
@@ -6,8 +6,7 @@ private[spark] class ApplicationDescription(
     val memoryPerSlave: Int,
     val command: Command,
     val sparkHome: String,
-    val appUIHost: String,
-    val appUIPort: Int)
+    val appUiUrl: String)
   extends Serializable {
 
   val user = System.getProperty("user.name", "<unknown>")
--- a/core/src/main/scala/spark/scheduler/cluster/SparkDeploySchedulerBackend.scala
+++ b/core/src/main/scala/spark/scheduler/cluster/SparkDeploySchedulerBackend.scala
@@ -32,7 +32,7 @@ private[spark] class SparkDeploySchedulerBackend(
     val sparkHome = sc.getSparkHome().getOrElse(
       throw new IllegalArgumentException("must supply spark home for spark standalone"))
     val appDesc =
-      new ApplicationDescription(appName, maxCores, executorMemory, command, sparkHome, sc.ui.host, sc.ui.port)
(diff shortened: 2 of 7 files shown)


core/src/main/scala/spark/deploy/ApplicationDescription.scala
```diff
@@ -5,7 +5,9 @@ private[spark] class ApplicationDescription(
     val cores: Int,
     val memoryPerSlave: Int,
     val command: Command,
-    val sparkHome: String)
+    val sparkHome: String,
+    val appUIHost: String,
+    val appUIPort: Int)
```
markhamstra: This applies more broadly than to just your additions, but is there any reason to keep appUIHost and appUIPort as separate elements?  If they always need to go together as a consistent pair, I think I'd rather see them packaged as at least a tuple.

squito: yeah, this is a great point.  I wondered about this myself, but made it this way just for consistency.

mateiz: I agree on this; in fact it should just be an appUiUrl or something like that.

squito:
updated to use combined `appUiUrl`

core/src/main/scala/spark/storage/BlockManagerUI.scala
```diff
@@ -20,19 +20,20 @@ class BlockManagerUI(val actorSystem: ActorSystem, blockManagerMaster: ActorRef,
   val STATIC_RESOURCE_DIR = "spark/deploy/static"
 
   implicit val timeout = Duration.create(System.getProperty("spark.akka.askTimeout", "10").toLong, "seconds")
+  val host = Utils.localHostName()
+  val port = if (System.getProperty("spark.ui.port") != null) {
+    System.getProperty("spark.ui.port").toInt
+  } else {
+    // TODO: Unfortunately, it's not possible to pass port 0 to spray and figure out which
```
rxin, line 27: What is the TODO item here?

rxin:
Jenkins, test this please.

core/src/main/scala/spark/scheduler/cluster/SparkDeploySchedulerBackend.scala
```diff
@@ -31,7 +31,8 @@ private[spark] class SparkDeploySchedulerBackend(
     val command = Command("spark.executor.StandaloneExecutorBackend", args, sc.executorEnvs)
     val sparkHome = sc.getSparkHome().getOrElse(
       throw new IllegalArgumentException("must supply spark home for spark standalone"))
-    val appDesc = new ApplicationDescription(appName, maxCores, executorMemory, command, sparkHome)
+    val appDesc =
+      new ApplicationDescription(appName, maxCores, executorMemory, command, sparkHome, sc.ui.appUIAddress)
```
rxin: Mind wrapping the parameter line to make them fit in 100 characters wide?

i.e.

val appDesc = new ApplicationDescription(appName, maxCores, executorMemory, command, sparkHome,
  sc.ui.appUIAddress)

rxin:
Thanks, this looks pretty good. I had two minor comments ...

core/src/main/scala/spark/storage/BlockManagerUI.scala
JoshRosen, line 27: I think this TODO was added by Matei in the the code that was moved here (see the lines below this in the diff).

squito:
I fixed the line break.  As Josh said, I just copied the TODO from earlier.  I dunno why its a TODO (is there some alternate solution??) -- I could leave the comment and remove the "TODO".

rxin:
Thanks, Imran.

PR #594, opened 2013-05-03
Add zip partitions to Java API

shivaram:
Note that this change only adds support for zipping 2 partitions to the Java API. Supporting the 3 and 4 argument case requires more glue code (Function3, WrappedFunction3 etc.) and IMHO we can postpone that till we have use cases.


mateiz:
Thanks for adding this.

PR #595, opened 2013-05-04
Fix flaky test by changing catch and adding sleep

shivaram:


shivaram:
Actually the test is still flaky on my machine. Please hold off on this -- I'm digging further into this.

rxin:
I submitted one that ignored the last failing line. I think the problem is timestamped hashmap could actually cleaned up the hashmap while it is being iterated through ...

https://github.com/mesos/spark/pull/598

PR #596, opened 2013-05-04
hBase example

esjewett:
This pull request provides an example of creating a NewHadoopRDD based on an hBase table. The HBaseTest class has a "main" method that takes two arguments: (1) the first SparkContext configuration parameter and (2) the hBase table name in String format. If the table does not already exist, the example will create it.

hBase resources are also added to SparkBuild.scala under the section for the examples project.

I wrote this code myself and I'm the copyright holder. I license it to the Spark/Mesos project under the project's license (BSD).

### DIFF
--- a/examples/src/main/scala/spark/examples/HBaseTest.scala
+++ b/examples/src/main/scala/spark/examples/HBaseTest.scala
@@ -1,0 +1,34 @@
+package spark.examples
+
+import spark._
+import spark.rdd.NewHadoopRDD
+import org.apache.hadoop.hbase.{HBaseConfiguration, HTableDescriptor, HColumnDescriptor}
+import org.apache.hadoop.hbase.client.HBaseAdmin
+import org.apache.hadoop.hbase.mapreduce.TableInputFormat
+
+object HBaseTest {
+  def main(args: Array[String]) {
+    val sc = new SparkContext(args(0), "HBaseTest",
+      System.getenv("SPARK_HOME"), Seq(System.getenv("SPARK_EXAMPLES_JAR")))
+
+    val conf = HBaseConfiguration.create()
+    conf.set(TableInputFormat.INPUT_TABLE, args(1))
+
+    // Initialize hBase tables if necessary
+    val admin = new HBaseAdmin(conf)
+    if(!admin.isTableAvailable(args(1))) {
+      val colDesc = new HColumnDescriptor(args(2))
+      val tableDesc = new HTableDescriptor(args(1))
+      tableDesc.addFamily(colDesc)
+      admin.createTable(tableDesc)
+    }
+
+    val hBaseRDD = new NewHadoopRDD(sc, classOf[TableInputFormat], 
+                        classOf[org.apache.hadoop.hbase.io.ImmutableBytesWritable],
+                        classOf[org.apache.hadoop.hbase.client.Result], conf)
+
+    hBaseRDD.count()
+
+    System.exit(0)
+  }
+}
--- a/examples/src/main/scala/spark/examples/HBaseTest.scala
+++ b/examples/src/main/scala/spark/examples/HBaseTest.scala
@@ -12,6 +12,9 @@ object HBaseTest {
       System.getenv("SPARK_HOME"), Seq(System.getenv("SPARK_EXAMPLES_JAR")))
 
     val conf = HBaseConfiguration.create()
+
+    // Other options for configuring scan behavior are available. More information available at 
+    // http://hbase.apache.org/apidocs/org/apache/hadoop/hbase/mapreduce/TableInputFormat.html
     conf.set(TableInputFormat.INPUT_TABLE, args(1))
 
     // Initialize hBase table if necessary
@@ -22,8 +25,8 @@ object HBaseTest {
     }
 
     val hBaseRDD = new NewHadoopRDD(sc, classOf[TableInputFormat], 
-                        classOf[org.apache.hadoop.hbase.io.ImmutableBytesWritable],
-                        classOf[org.apache.hadoop.hbase.client.Result], conf)
+      classOf[org.apache.hadoop.hbase.io.ImmutableBytesWritable],
+      classOf[org.apache.hadoop.hbase.client.Result], conf)
 
     hBaseRDD.count()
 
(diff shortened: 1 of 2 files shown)


rxin:
Nice. I think this will be very useful to a lot of people.

That aside, while examples like this are tremendously valuable, I feel the spark project probably should not build examples by default (or move it to the contrib folder or even a separate project), given the growing dependencies in the examples subproject (imagine if someone adds Cassandra next, and other databases too).

JoshRosen:
This will fix [SPARK-699](https://spark-project.atlassian.net/browse/SPARK-699).

esjewett:
@JoshRosen Indeed. Sorry, I forgot to mention that.

@rxin I appreciate the problem. Should I change anything about this pull request, or is there somewhere else I should put this? I also wonder if it would be possible in SBT to make building the examples optional. I'm not really an SBT wizard, so I don't know, but it seems this should be doable.

rxin:
No worry. I was initiating a new discussion. That should probably go into a separate PR.

examples/src/main/scala/spark/examples/HBaseTest.scala
```diff
@@ -0,0 +1,32 @@
...
+
+    // Initialize hBase table if necessary
+    val admin = new HBaseAdmin(conf)
+    if(!admin.isTableAvailable(args(1))) {
+      val tableDesc = new HTableDescriptor(args(1))
+      admin.createTable(tableDesc)
+    }
+
+    val hBaseRDD = new NewHadoopRDD(sc, classOf[TableInputFormat], 
+                        classOf[org.apache.hadoop.hbase.io.ImmutableBytesWritable],
```
pwendell: Would you mind indenting these wrapped lines only 2 spaces? That's consistent with the Scala style we use.

examples/src/main/scala/spark/examples/HBaseTest.scala
```diff
@@ -0,0 +1,32 @@
...
+import org.apache.hadoop.hbase.client.HBaseAdmin
+import org.apache.hadoop.hbase.mapreduce.TableInputFormat
+
+object HBaseTest {
+  def main(args: Array[String]) {
+    val sc = new SparkContext(args(0), "HBaseTest",
+      System.getenv("SPARK_HOME"), Seq(System.getenv("SPARK_EXAMPLES_JAR")))
+
+    val conf = HBaseConfiguration.create()
+    conf.set(TableInputFormat.INPUT_TABLE, args(1))
```
pwendell, line 18: I noticed that the TableInputFormat lets people set many other options to narrow the range of rowKeys or columns:
http://hbase.apache.org/apidocs/org/apache/hadoop/hbase/mapreduce/TableInputFormat.html

It might be cool to add notes for a few popular options, even if these aren't used in the example. It might not be obvious to people that these options are also configured by setting a conf value with a TableInputFormat key. E.g.:

// Other options:
// conf.set(TableInputFormat.ROW_START, "yourStartKey") 
// conf.set(TableInputFormat.ROW_END, "yourEndKey")

pwendell:
Thanks a ton ,this is really valuable - we get asked all the time how to do this! A few smaller comments inline in the code.

pwendell:
Also, once this gets ready to merge, we should try to cherry-pick it into 0.7. Ethan, that just means submitting another pull request against the branch-0.7 branch once this request gets finalized (assuming the code doesn't need to change).

examples/src/main/scala/spark/examples/HBaseTest.scala
esjewett, line 18: I'm going to play around with this eventually for the project where I'm using this, but it's not clear to me how all of these options interact at the moment. For example, can I just define the ROW_START and ROW_END or do I need to pass in a Scan instance converted to a String? I really don't know. How about a comment as follows?

// Other options for configuring scan behavior are available. More information
// available at http://hbase.apache.org/apidocs/org/apache/hadoop/hbase/mapreduce/TableInputFormat.html

esjewett:
@pwendell You're welcome. Thanks for the prompt to just do it the right way, as an example. I've added another commit that hopefully addresses your comments.

pwendell:
Jenkins, test this please.

mateiz:
Unfortunately I'm not sure that command makes Jenkins test it -- ask Matt about how to do it.

examples/src/main/scala/spark/examples/HBaseTest.scala
```diff
@@ -0,0 +1,35 @@
...
+
+    // Initialize hBase table if necessary
+    val admin = new HBaseAdmin(conf)
+    if(!admin.isTableAvailable(args(1))) {
+      val tableDesc = new HTableDescriptor(args(1))
+      admin.createTable(tableDesc)
+    }
+
+    val hBaseRDD = new NewHadoopRDD(sc, classOf[TableInputFormat],
```
holdenk: Would it make sense to use  sc.newAPIHadoopRDD(conf,
             classOf[TableInputFormat], 
             classOf[org.apache.hadoop.hbase.io.ImmutableBytesWritable],
             classOf[org.apache.hadoop.hbase.client.Result]) instead of new NewHadoopRDD(...)?

pwendell:
Trying again (we're still working through new test setup): test this please

esjewett:
@holdenk That should work as well. Is that the preferred way to create RDDs? If so, I'm happy to change it, though I will often use the actual RDD classes rather than the methods on the SparkContext object.

pwendell:
@esjewett @holdenk It would be preferable to use the SparkContext version rather than instantiating the RDD directly. In general, we try to make it be possible for people to instantiate all of the included (i.e. non-user-defined) RDD types through SparkContext.

Other than that, this looks good to me! test this please

esjewett:
@pwendell @holdenk Cool. I've made the change.

pwendell:
Okay I just built this locally and confirmed it compiles (Jenkins stuff doesn't seem to catch this commit).

@mateiz - I think this is ready to go if you want to look at it and/or commit.

mateiz:
Great, thanks!

rxin:
FYI I reverted this merge because the additional dependency on hbase introduced an older version of Netty, which caused lots of test failures. I will reopen the pull request.

rxin:
I guess I can't re-open a merged PR.... Patrick, can you handle this one?

pwendell:
Hey @esjewett  can you re-submit this PR? We'll have to workout the hbase dependency issue...

esjewett:
Sorry about that. I think I should have just pulled in the hbase-client dependencies instead of all of hbase. I'll try it that way and see if everything works before re-submitting the PR.

PR #597, opened 2013-05-05
Two minor bug fixes for Spark Web UI

JoshRosen:
This pull request fixes two minor issues in the Spark Web UI:
- [SPARK-630](https://spark-project.atlassian.net/browse/SPARK-630): App details page lists finished/killed executors as RUNNING.
- [SPARK-629](https://spark-project.atlassian.net/browse/SPARK-629): App details page has strange value for number of cores.

I also fixed a minor HTML issue where an `<a>` tag was closed with `</href>`.

### DIFF
--- a/core/src/main/scala/spark/deploy/ApplicationDescription.scala
+++ b/core/src/main/scala/spark/deploy/ApplicationDescription.scala
@@ -2,7 +2,7 @@ package spark.deploy
 
 private[spark] class ApplicationDescription(
     val name: String,
-    val cores: Int,
+    val maxCores: Int, /* Integer.MAX_VALUE denotes an unlimited number of cores */
     val memoryPerSlave: Int,
     val command: Command,
     val sparkHome: String)
--- a/core/src/main/scala/spark/deploy/JsonProtocol.scala
+++ b/core/src/main/scala/spark/deploy/JsonProtocol.scala
@@ -26,7 +26,7 @@ private[spark] object JsonProtocol extends DefaultJsonProtocol {
       "starttime" -> JsNumber(obj.startTime),
       "id" -> JsString(obj.id),
       "name" -> JsString(obj.desc.name),
-      "cores" -> JsNumber(obj.desc.cores),
+      "cores" -> JsNumber(obj.desc.maxCores),
       "user" -> JsString(obj.desc.user),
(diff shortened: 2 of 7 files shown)


core/src/main/scala/spark/deploy/ApplicationDescription.scala
```diff
@@ -2,7 +2,7 @@ package spark.deploy
 
 private[spark] class ApplicationDescription(
     val name: String,
-    val cores: Int,
+    val maxCores: Int, /* Integer.MAX_VALUE denotes an unlimited number of cores */
```
JoshRosen, line 5: The issue with the weird number of cores in the web UI seems to stem from the fact that this seems to be the _maximum number of_ cores that the application may use, but `Integer.MAX_VALUE` wasn't being consistently treated as "unlimited" in the web UI code.  I think that renaming this to `maxCores` makes the code clearer.

core/src/main/scala/spark/deploy/JsonProtocol.scala
```diff
@@ -26,7 +26,7 @@ private[spark] object JsonProtocol extends DefaultJsonProtocol {
       "starttime" -> JsNumber(obj.startTime),
       "id" -> JsString(obj.id),
       "name" -> JsString(obj.desc.name),
-      "cores" -> JsNumber(obj.desc.cores),
+      "cores" -> JsNumber(obj.desc.maxCores),
```
JoshRosen, line 29: Although I renamed this to `maxCores` in the internal API, I left the JSON API unmodified.

core/src/main/scala/spark/deploy/master/Master.scala
```diff
@@ -275,6 +275,7 @@ private[spark] class Master(host: String, port: Int, webUiPort: Int) extends Act
       for (exec <- app.executors.values) {
         exec.worker.removeExecutor(exec)
         exec.worker.actor ! KillExecutor(exec.application.id, exec.id)
+        exec.state = ExecutorState.KILLED
```
JoshRosen, line 278: `exec.state` is ordinarily updated in the `ExecutorStateChanged` message handler, but that isn't fired when we kill the executors.  As a result, killed executors were still marked as `RUNNING` in the master Web UI.

mateiz:
Thanks for the fixes, Josh.

PR #599, opened 2013-05-05
Fix SPARK-670: EC2 'start' command should require -i option.

JoshRosen:
This fixes [SPARK-670](https://spark-project.atlassian.net/browse/SPARK-670): "spark-ec2 should warn if you use the "start" command without passing a SSH key file".


rxin:
Thanks, Josh!

PR #600, opened 2013-05-06
Added EmptyRDD (branch-0.7).

rxin:
Ideally I would like RDD.empty to create an empty RDD too, but it is hard to do that because each RDD is SparkContext specific.

### DIFF
--- a/core/src/main/scala/spark/rdd/EmptyRDD.scala
+++ b/core/src/main/scala/spark/rdd/EmptyRDD.scala
@@ -1,0 +1,16 @@
+package spark.rdd
+
+import spark.{RDD, SparkContext, SparkEnv, Partition, TaskContext}
+
+
+/**
+ * An RDD that is empty, i.e. has no element in it.
+ */
+class EmptyRDD[T: ClassManifest](sc: SparkContext) extends RDD[T](sc, Nil) {
+
+  override def getPartitions: Array[Partition] = Array.empty
+
+  override def compute(split: Partition, context: TaskContext): Iterator[T] = {
+    throw new UnsupportedOperationException("empty RDD")
+  }
+}
--- a/core/src/test/scala/spark/RDDSuite.scala
+++ b/core/src/test/scala/spark/RDDSuite.scala
@@ -3,7 +3,7 @@ package spark
 import scala.collection.mutable.HashMap
 import org.scalatest.FunSuite
 import spark.SparkContext._
-import spark.rdd.{CoalescedRDD, CoGroupedRDD, PartitionPruningRDD, ShuffledRDD}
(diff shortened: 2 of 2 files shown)


mridulm:
Hi, 
  Curious what is the intention behind empty RDD - is it to ensure it is never computed ? (like an assertion check).
Or as a marker to check for equality, etc ?

Thanks,
Mridul

rxin:
Just a convenience class when empty should be used, similar to Scala's Iterator.empty or Array.empty.

My particular use case in Shark is for LimitOperator to return the EmptyRDD directly if limit is 0. That would save a round of task launching.

core/src/main/scala/spark/rdd/EmptyRDD.scala
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * An RDD that is empty, i.e. has no element in it.
+ */
+class EmptyRDD[T: ClassManifest](sc: SparkContext) extends RDD[T](sc, Nil) {
+
+  override def getPartitions: Array[Partition] = Array.empty
+
+  override def compute(split: Partition, context: TaskContext): Iterator[T] = {
+    throw new UnsupportedOperationException("empty RDD")
```
JoshRosen, line 14: Why should it be illegal to call compute() on an empty RDD?  In Scala, I can still perform ordinary operations on an empty collection.  Why not just return Iterator[T].empty?

rxin, line 14: There is no partition for you to run compute on though.

rxin, line 14: You can still do count or map, but the result will simply be an RDD with no partitions. compute should never been called.

rxin:
I am pretty sure Jenkins hates me. Every pull request I submitted so far trigged some failures in Jenkins, and none of them were caused by my commits ...

core/src/main/scala/spark/rdd/EmptyRDD.scala
JoshRosen, line 14: Ah, right.  Good idea to keep this exception in, since calling `compute()` with a bogus partition would be a serious bug.

mridulm:
Thanks for clarifying ! This makes a lot of sense.
I missed that there were no partitions (like @JoshRosen did above :-) ) - and seeing the exception being thrown from compute; was wondering how this would be used.

Btw, just wondering what happens to join/cogroup/cross/etc with one operand is 'EmptyRDD' : will it cause zero tasks to be launched ? (I have not looked at how tasks are actually created yet - apologies if this is trivial).

Thanks,
Mridul

rxin:
They should all just work. I added tests for those in the other pull request: https://github.com/mesos/spark/pull/601

I am going to close this one.

Matei - if you want this to go into branch 0.7, just do a cherry-pick. The changes are very isolated so it is very easy to cherry-pick.

PR #557, opened 2013-04-07
Shuffle Performance Optimization

shane-huang:
The patch includes below optimizations for shuffle performance. 
1. Use netty embeded OIO file server instead of Connection Manager for block shuffling.
2. Skip 0-byte block requests to reduce the network transport in shuffle

A system property "spark.shuffle.use.netty" is provided for user to choose whether enable or disable this optimization.  If spark.shuffle.use.netty=true, the optimization is enabled, otherwise it is disabled. By default it is disabled. 

This patch is moved from against branch-0.6 to master. The original pull request is closed( https://github.com/mesos/spark/pull/522).

PR #602, opened 2013-05-08
Manual merge & cleanup of Shane's Shuffle Performance Optimization

rxin:
Merged #557 (originally #522) to master branch.

Did some clean ups too. Note that most of the classes in spark.network are moved into their own files without any changes.

Quoting #557:

The patch includes below optimizations for shuffle performance. 
1. Use netty embeded OIO file server instead of Connection Manager for block shuffling.
2. Skip 0-byte block requests to reduce the network transport in shuffle

A system property "spark.shuffle.use.netty" is provided for user to choose whether enable or disable this optimization. If spark.shuffle.use.netty=true, the optimization is enabled, otherwise it is disabled. By default it is disabled.


rxin:
I updated the pull request so it can now be automatically merged again (previous had conflict with the instance local scheduling pull request).

core/src/main/scala/spark/storage/DiskStore.scala
```diff
@@ -262,10 +274,48 @@ private class DiskStore(blockManager: BlockManager, rootDirs: String)
...
         } catch {
           case t: Throwable => logError("Exception while deleting local spark dirs", t)
         }
       }
     })
   }
+
+  private def startShuffleBlockSender() {
+    try {
+      val port = System.getProperty("spark.shuffle.sender.port", "6653").toInt
```
mateiz, line 288: It would be good if this bound to an empty port instead, but we can fix it later.

mateiz:
Looks good; I've merged this now, though to really release this we should make it use a free port instead of a hardcoded one, for multitenant clusters.

PR #603, opened 2013-05-09
Several Improvements to EC2 Scripts

pwendell:
This pull request is related to an equivalent request in the shivaram/spark-ec2 repository (https://github.com/shivaram/spark-ec2/pull/1). It should be committed atomically with that. Also - some of the changes are on the AMI image itself.

Together with that change and the new AMI, several improvements have been made:
- It is now possible for users to launch both PVM (standard) and HVM (high capacity) instance architectures. The type will be inferred based on the specific instance type given
- It is now possible for users to launch in any EC2 region
  - There is a developer script to auto-deploy new AMI's to all regions
- There is better support for versioning EC2 images as we go forward
- AMI id's are now tracked inside of github
- The image itself has been improved
  - Hadoop HDFS and MapReduce upgraded to CDH4
  - Hadoop native libraries setup correctly (both for HVM and PVM) along with Snappy compresison
  - Matei's public key is no longer present on the image
  - Java version 1.7 is added to the machines and made the default

Note that I did not update the Spark/Shark version on the image. They are still at 0.7.0/0.2.0 so this should be a "drop in" replacement for the current AMI.


ec2/spark_ec2.py
```diff
@@ -34,8 +34,8 @@
 from boto.ec2.blockdevicemapping import BlockDeviceMapping, EBSBlockDeviceType
 from boto import ec2
 
-# A static URL from which to figure out the latest Mesos EC2 AMI
-LATEST_AMI_URL = "https://s3.amazonaws.com/mesos-images/ids/latest-spark-0.7"
+# A URL prefix from which to fetch AMI information
+AMI_PREFIX = "https://raw.github.com/pwendell/spark-ec2/ec2-updates/ami-list"
```
pwendell: NOTE: This URL is set to my own repository to allow people to test this change on their own. Before we commit, I'll need to change this to 

https://raw.github.com/mesos/spark-ec2/ec2-updates/ami-list

However, that requires updating the code there as well.

ec2/spark_ec2.py
```diff
@@ -64,9 +64,9 @@ def parse_args():
       help="Availability zone to launch instances in, or 'all' to spread " +
            "slaves across multiple (an additional $0.01/Gb for bandwidth" +
            "between zones applies)")
-  parser.add_option("-a", "--ami", default="latest",
-      help="Amazon Machine Image ID to use, or 'latest' to use latest " +
-           "available AMI (default: latest)")
+  parser.add_option("-a", "--ami", default="v0.7.0",
```
shivaram: I'd still suggest supporting `latest` to handle new spark versions without any changes to this file.

ec2/spark_ec2.py
```diff
@@ -156,6 +156,51 @@ def wait_for_instances(conn, instances):
...
+    instance_type = "pvm"
+    print >> stderr,\
+        "Don't recognize %s, assuming type is pvm" % opts.instance_type
+  if version != "v0.7.0":
+    print >> stderr, \
+      "Don't know how to resolve AMI for version: %s" % version
+  # TODO(pwendell) Once we have multiple Spark AMI versions, we should let 
+  # people give a version flag here in place of just saying 'latest'.
+  version = version[1:]
+  ami_path = "%s/%s/%s/%s" % (AMI_PREFIX, version, opts.region, instance_type)
```
shivaram: ~~So it looks like we loose the ability to specify an arbitrary AMI ID with this change. I think that is still valuable to support (say I have my own AMI which has additional packages etc.). I guess you could check for the first character being 'v' and if not just use the given id ?~~

EDIT: my bad - just noticed this was handled

Also it'll be good to have `latest` be a symlink in the github repo to 0.7.0 so the url resolution will work with latest too.

pwendell:
Okay I amended this to continue supporting the `latest` tag.

ec2/spark_ec2.py
```diff
@@ -207,15 +252,9 @@ def launch_cluster(conn, opts, cluster_name):
...
-  if opts.ami == "latest":
-    try:
-      opts.ami = urllib2.urlopen(LATEST_AMI_URL).read().strip()
-      print "Latest Spark AMI: " + opts.ami
-    except:
-      print >> stderr, "Could not read " + LATEST_AMI_URL
-      sys.exit(1)
-
+  # Figure out Spark AMI
+  if opts.ami[0] == "v":
```
shivaram: This should also check for latest ?

ec2/spark_ec2.py
```diff
@@ -156,6 +156,51 @@ def wait_for_instances(conn, instances):
...
+  else:
+    instance_type = "pvm"
+    print >> stderr,\
+        "Don't recognize %s, assuming type is pvm" % opts.instance_type
+  if version not in ["latest", "v0.7.0"]:
+    print >> stderr, \
+      "Don't know how to resolve AMI for version: %s" % version
+  # TODO(pwendell) Once we have multiple Spark AMI versions, we should let 
+  # people give a version flag here in place of just saying 'latest'.
+  version = version[1:]
```
shivaram: I think we need to special case this for `latest` i.e not remove the first character

JoshRosen: I think it makes more sense to first normalize the version, then check whether it's in a normalized list of known versions, because I suspect that a lot of users will supply `0.7.0` instead of `v0.7.0`

ec2/spark_ec2.py
JoshRosen: Yeah, I think so too.  When I tried to run

```
./spark-ec2 -i ~/.ssh/berkeley-laptop.pem -k berkeley-laptop -s 10 -t m1.large --spot-price 0.25 --region us-west-1 launch test_cluster
```

I received the error

```
Setting up security groups...
Searching for existing cluster test_cluster...
Launching instances...
ERROR:boto:400 Bad Request
ERROR:boto:<?xml version="1.0" encoding="UTF-8"?>
<Response><Errors><Error><Code>InvalidAMIID.Malformed</Code><Message>Invalid id: "latest" (expecting "ami-...")</Message></Error></Errors><RequestID>6c4888ae-e4b6-4539-b0be-047268fbcdd5</RequestID></Response>
Could not find AMI latest
```

JoshRosen: Aren't AMI ids supposed to be of the form `ami-...`?  If that's true, you can change this to

``` python
if not opts.ami.startswith("ami-"):
```

JoshRosen:
For anyone testing this, here are some nice instructions on creating a single SSH keypair for use in multiple EC2 regions: http://stackoverflow.com/questions/10207416/single-keypair-for-all-regions

ec2/spark_ec2.py
pwendell: Josh - good point. I added a commit that first sanitizes the version then deals with this check.

pwendell:
Just FYI to anyone watching this, we are waiting for the next spark release so we can do some refactoring. The high level plan is to actually pull spark binaries for hadoop1/hadoop2 instead of pre-loading those on different images.

pwendell:
Okay, this update makes the following changes:
- Rather than rely on AMI's for Spark/Shark versioning, this accepts an argument for the Spark version. Spark and Shark have become modules and those modules initialize themselves by downloading binary releases. This is still not "bare bones". That will require quite a bit more effort, but this is a first step.
- Because of this, we do not have to create new AMI's for each version. We'll just need to create an AMI for each region and HVM/PVM. There is still a tracked registry for those AMI's in the spark-ec2 repo. The registry only has one entry in it right now - I'll create copies once this review goes through...
- Right now, only one version cut is supported 0.7.2/0.7.0, but this will be easy to extend going forward once we produce other binary releases.
- Right now, only Hadoop2 is supported. Allowing both multiple Spark versions and multiple Hadoop versions gets a bit unwieldy. We could add that though if there is substantial support for it. 
- To make this backwards compatible, I propose creating a new branch in spark-ec2. Anyone with an old script will just checkout that version.

mateiz:
Hey Patrick, is Shivaram planning to review this, or do you want me to do it?

shivaram:
I reviewed this part of the change and it looked fine. I had a few comments for the corresponding changes to spark-ec2 which is at https://github.com/shivaram/spark-ec2/pull/1 -- I will take another look at that today.

ec2/spark_ec2.py
```diff
@@ -156,66 +154,94 @@ def wait_for_instances(conn, instances):
 def is_active(instance):
   return (instance.state in ['pending', 'running', 'stopping', 'stopped'])
 
+# Return correct versions of Spark and Shark, given the supplied Spark version
+def get_spark_shark_version(opts):
+  spark_shark_map = {"0.7.2": "0.7.0"}
+  version = opts.spark_version.replace("v", "")
+  if version not in ["0.7.2"]:
```
shivaram: Should this be `if version not in spark_shark_map` ?

shivaram:
Had a minor comment -- but otherwise this part of the change looks good to me. However I guess we will merge this after the spark-ec2 change is merged in a new branch.

PR #604, opened 2013-05-09
hBase example (retry of pull request #596)

esjewett:
I've added an exclusion for Netty as is done in the streamingSettings. However I haven't been able to get the test suite to run properly with or without my changes. If this addresses the issue with the test suite for others, great. Otherwise I'll keep trying to get the tests running properly so that I can fix this.


esjewett:
Sorry, that did not work properly....

PR #605, opened 2013-05-10
Add hBase example (retry of pull request #596)

esjewett:
I've added an exclusion for Netty as is done in the streamingSettings. However I haven't been able to get the test suite to run properly with or without my changes. If this addresses the issue with the test suite for others, great. Otherwise I'll keep trying to get the tests running properly so that I can fix this.

### DIFF
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -67,6 +67,11 @@
           <artifactId>hadoop-core</artifactId>
           <scope>provided</scope>
         </dependency>
+        <dependency>
+          <groupId>org.apache.hbase</groupId>
+          <artifactId>hbase</artifactId>
+          <version>0.94.6</version>
+        </dependency>
       </dependencies>
       <build>
         <plugins>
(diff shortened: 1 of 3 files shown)


rxin:
Jenkins, test this please.

esjewett:
Realized today while converting my application that if we want to cherry-pick this to 0.7 per @pwendell's suggestion, we will also need to cherry-pick the calls to the setConf method of InputFormat that have been added to NewHadoopRDD.scala since 0.7.

rxin:
It is probably best to just put this in 0.8 and not cherry-pick into 0.7 ...

rxin:
@esjewett Is there a way to only get the client hbase library?

esjewett:
@rxin No, unfortunately not, as we need classes available from org.apache.hadoop.hbase and org.apache.hadoop.hbase.mapreduce.

mateiz:
@esjewett can you patch the Maven build file for "examples" to include this as well?

esjewett:
@mateiz Done ... I think. "mvn -Phadoop1 compile" and "mvn -Phadoop2 compile" both return successes. However, I can't verify that "mvn -Phadoop\* package" works properly because I still get failing tests due to my local setup being messed up. Hopefully Jenkins can verify.

pwendell:
Jenkins, test this please

mateiz:
Hey, sorry for the long delay. Going to merge this in. Thanks!

PR #607, opened 2013-05-10
Jdbc

koeninger:
RDD for jdbc interop

https://groups.google.com/forum/?fromgroups=#!topic/spark-developers/fhj6JUzJny4

### DIFF
--- a/core/src/main/scala/spark/rdd/JdbcRDD.scala
+++ b/core/src/main/scala/spark/rdd/JdbcRDD.scala
@@ -5,23 +5,27 @@ import java.sql.{Connection, ResultSet}
 import spark.{Logging, Partition, RDD, SparkContext, TaskContext}
 import spark.util.NextIterator
 
+private[spark] class JdbcPartition(idx: Int, val lower: Long, val upper: Long) extends Partition {
+  override def index = idx
+}
+
 /**
-  An RDD that executes an SQL query on a JDBC connection and reads results.
-  @param getConnection a function that returns an open Connection.
-    The RDD takes care of closing the connection.
-  @param sql the text of the query.
-    The query must contain two ? placeholders for parameters used to partition the results.
-    E.g. "select title, author from books where ? <= id and id <= ?" 
-  @param lowerBound the minimum value of the first placeholder
-  @param upperBound the maximum value of the second placeholder
-    The lower and upper bounds are inclusive.
-  @param numPartitions the number of partitions.
-    Given a lowerBound of 1, an upperBound of 20, and a numPartitions of 2,
-    the query would be executed twice, once with (1, 10) and once with (11, 20)
-  @param mapRow a function from a ResultSet to a single row of the desired result type(s).
-    This should only call getInt, getString, etc; the RDD takes care of calling next.
-    The default maps a ResultSet to an array of Object.
-*/
+ * An RDD that executes an SQL query on a JDBC connection and reads results.
+ * @param getConnection a function that returns an open Connection.
+ *   The RDD takes care of closing the connection.
+ * @param sql the text of the query.
+ *   The query must contain two ? placeholders for parameters used to partition the results.
+ *   E.g. "select title, author from books where ? <= id and id <= ?" 
+ * @param lowerBound the minimum value of the first placeholder
+ * @param upperBound the maximum value of the second placeholder
+ *   The lower and upper bounds are inclusive.
+ * @param numPartitions the number of partitions.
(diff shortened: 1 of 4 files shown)


core/src/main/scala/spark/rdd/JdbcRDD.scala
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+}
+
+private[spark] class JdbcPartition(idx: Int, val lower: Long, val upper: Long) extends Partition {
+  override def index = idx
+}
+
+object JdbcRDD {
+  val resultSetToObjectArray = (rs: ResultSet) =>
```
rxin: Change this to a function directly instead of a val for a closure. i.e.

```
def resultSetToObjectArray(rs: ResultSet) = {
  Array.tabulate[Object](rs.getMetaData.getColumnCount)(i => rs.getObject(i + 1))
}
```

core/src/main/scala/spark/rdd/JdbcRDD.scala
```diff
@@ -0,0 +1,94 @@
...
+      zipWithIndex.
+      map(x => new JdbcPartition(x._2, x._1.head, x._1.last)).
+      toArray
+
+  override def compute(thePart: Partition, context: TaskContext) = new NextIterator[T] {
+    context.addOnCompleteCallback{ () => closeIfNeeded() }
+    val part = thePart.asInstanceOf[JdbcPartition]
+    val conn = getConnection()
+    val stmt = conn.prepareStatement(sql, ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY)
+    // force mysql driver to stream rather than pull entire resultset into memory
```
rxin: Do you mind adding a comment saying why this is not done for other dbs, as you did in the pull request comment?

core/src/main/scala/spark/rdd/JdbcRDD.scala
```diff
@@ -0,0 +1,94 @@
...
+        logInfo("closed connection")
+      } catch {
+        case e: Exception => logWarning("Exception closing connection", e)
+      }
+    }
+  }
+
+}
+
+private[spark] class JdbcPartition(idx: Int, val lower: Long, val upper: Long) extends Partition {
```
rxin: Move the jdbc partition definition to the top to match other RDD classes ...

core/src/main/scala/spark/rdd/JdbcRDD.scala
```diff
@@ -0,0 +1,94 @@
...
+    getConnection: () => Connection,
+    sql: String,
+    lowerBound: Long,
+    upperBound: Long,
+    numPartitions: Int,
+    mapRow: (ResultSet) => T = JdbcRDD.resultSetToObjectArray)
+  extends RDD[T](sc, Nil) with Logging {
+
+  override def getPartitions: Array[Partition] =
+    ParallelCollectionRDD.slice(lowerBound to upperBound, numPartitions).
```
rxin: mhmm Don't use a ParallelCollectionRDD here. I think you are just trying to generate the splits, which can be done with Scala local collections entirely.

core/src/main/scala/spark/rdd/JdbcRDD.scala
```diff
@@ -0,0 +1,94 @@
+package spark.rdd
+
+import java.sql.{Connection, ResultSet}
+
+import spark.{Logging, Partition, RDD, SparkContext, TaskContext}
+import spark.util.NextIterator
+
+/**
```
rxin, line 12: Match the commenting style in other class definitions, i.e.

```
/**
 * line 1
 * line 2
 */
```

rxin:
Thanks for doing this. I left a few comments, mostly style related other than the ParallelCollectionRDD part.

koeninger:
Thanks for the feedback, I think this commit should address your comments.  Let me know if you see anything else.

rxin:
Thanks - this looks good.

One more thing: Do you mind adding a test case for this, as it will help to make sure this works properly in the future?

You can add Derby as a test dependency in project/SparkBuild.scala file:

"org.apache.derby" % "derby" % "10.4.2.0" % "test"

And use Derby to do a db testing ...

koeninger:
No problem, added a derby test case.

My approach to dealing with derby files was to just create the db under target/ and add derby.log to gitignore; if you think it'd be better to disable derby.log altogether let me know.

rxin:
Thanks - I did a manual merge of this because github couldn't auto merge it. 

I also added a dependency on derby to the maven pom files.

PR #608, opened 2013-05-11
SPARK-738: Spark should detect and wrap nonserializable exceptions

pwendell:


pwendell:
Jenkins, test this please.

mridulm:
Since Throwable is Serializable (and so all Error's and Exception's are
too), the root cause here are buggy extensions : we should be fixing that
if it is from within spark.
This hack should be required only when what is thrown is outside of our
codebase, and cant influence a fix.

Regards,
Mridul

rxin:
Yes this should catch user generated exceptions that are not serializable.

core/src/main/scala/spark/executor/Executor.scala
```diff
@@ -123,7 +123,19 @@ private[spark] class Executor(executorId: String, slaveHostname: String, propert
 
         case t: Throwable => {
           val reason = ExceptionFailure(t)
-          context.statusUpdate(taskId, TaskState.FAILED, ser.serialize(reason))
+          val serReason =
+            try {
```
rxin: put catch on the same line as the closing bracket

} catch {

also don't use abbreviated "exn" in the exception message.

core/src/main/scala/spark/executor/Executor.scala
```diff
@@ -1,6 +1,6 @@
 package spark.executor
 
-import java.io.{File, FileOutputStream}
+import java.io.{NotSerializableException, File, FileOutputStream}
```
rxin: Put NotSerializableException after File / FileOutputStream

mateiz:
Patrick, as a high-level approach, why don't we just send the exception's message and stack trace (array of StackTraceElement) instead of the exception itself? I'd prefer not to have a special case for this.

mateiz:
(At the end of the day, those are the only pieces of information we need to display.)

rxin:
That's an excellent idea - we should return the exception class name, message, and stack trace.

pwendell:
I agree - let me refactor this and post tomorrow.

pwendell:
Okay I changed it to use the new approach. I didn't include classname (it is not used downstream and there are other fields in exception we could include also - so I figured keep it minimal for now).

core/src/main/scala/spark/TaskEndReason.scala
```diff
@@ -14,9 +14,16 @@ private[spark] case object Success extends TaskEndReason
 private[spark] 
 case object Resubmitted extends TaskEndReason // Task was finished earlier but we've now lost it
 
-private[spark] 
-case class FetchFailed(bmAddress: BlockManagerId, shuffleId: Int, mapId: Int, reduceId: Int) extends TaskEndReason
+private[spark] case class FetchFailed(
```
pwendell, line 17: This is code cleanup - it exceeded 100 characters before.

mridulm:
Can you please also include the exception/error classname too ?
Particularly with generated code, the exception name is more useful than the stack trace (and reverse in bunch of other cases; unfortunately).

Thanks !
Mridul

pwendell:
Mridul - do you mean exn.getClass()? I thought most exceptions report their own type in toString(), could you give an example where it's not sufficient?

scala> val x = new IOException("test")
x: java.io.IOException = java.io.IOException: test

scala> x.toString()
res4: java.lang.String = java.io.IOException: test

scala> x.getClass().toString()
res5: java.lang.String = class java.io.IOException

scala> class MyException extends Exception
defined class MyException

scala> val f = new MyException()
f: MyException = MyException

scala> f.toString()
res6: java.lang.String = MyException

mridulm:
This is just a convention; and not mandated by the api contract. (Unlike
Throwable implements Serializable - which, as we see, is also being
violated !).
All we need is just exn.getClass().getName() - not necessary to serialize
the entire class across ofcourse.

Thanks,
Mridul

pwendell:
Okay - since Mridul and Reynold both seem to support - I added the exception class name.

pwendell:
Jenkins, test this please (just trying to debug something with Matt, this still isn't working for me)

mateiz:
Thanks Patrick!

PR #610, opened 2013-05-14
Throw exception if TaskResult exceeds Akka frame size

JoshRosen:
This addresses [SPARK-747](https://spark-project.atlassian.net/browse/SPARK-747): "Throw an exception on slaves if a message is sent that is larger than Akka's max frame size". 

If serialized task results are within 1024 bytes of the maximum frame size, the Executors die with an appropriate exception.  I didn't know how much overhead Akka adds to the message payload, but I assume that it won't be more than 1024 bytes.

I decided to pull the frame size directly from the ActorSystem's actual configuration since that's the actual limit that matters.

A better long-term approach might be to collect large task results through the BlockManager (as proposed in [SPARK-669](https://spark-project.atlassian.net/browse/SPARK-669)), but that involves a number of subtleties that I'll need more time to think through.

### DIFF
--- a/core/src/main/scala/spark/TaskEndReason.scala
+++ b/core/src/main/scala/spark/TaskEndReason.scala
@@ -20,3 +20,5 @@ case class FetchFailed(bmAddress: BlockManagerId, shuffleId: Int, mapId: Int, re
 private[spark] case class ExceptionFailure(exception: Throwable) extends TaskEndReason
 
 private[spark] case class OtherFailure(message: String) extends TaskEndReason
+
+private[spark] case class TaskResultTooBigFailure() extends TaskEndReason
--- a/core/src/main/scala/spark/executor/Executor.scala
+++ b/core/src/main/scala/spark/executor/Executor.scala
@@ -72,6 +72,7 @@ private[spark] class Executor(executorId: String, slaveHostname: String, propert
   // Initialize Spark environment (using system properties read above)
   val env = SparkEnv.createFromSystemProperties(executorId, slaveHostname, 0, false, false)
   SparkEnv.set(env)
+  private val akkaFrameSize = env.actorSystem.settings.config.getBytes("akka.remote.netty.message-frame-size")
 
   // Start worker thread pool
   val threadPool = new ThreadPoolExecutor(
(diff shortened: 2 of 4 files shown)


rxin:
Maybe we shouldn't kill the executor, but just return the exception back to the master? Create a new runtime exception, e.g. TaskResultTooBigException?

Also it would be great to add a test case.

JoshRosen:
I modified the code to abort the job if any task crashes due to a message exceeding the frame size, since there's no point in re-running a task that we know is doomed to fail repeatedly.  I also added a test case for this in DistributedSuite.

JoshRosen:
Jenkins: this is ok to test

mridulm:
I have not looked into solving this : but this is something we had hit in the past too - and wasted a lot of time before we identified what was the problem.

Trying to understand from folks who might have looked into this before : what are the issues with solving this properly (as opposed to warn + fail, repeat with bigger value) ?
For example - not send response in a single message, but split it into multiple messages (shuffle already does this).

Regards,
Mridul

JoshRosen:
@mridulm There are several ways that we could fix the root problem.

I started to write [a patch](https://github.com/JoshRosen/spark/tree/results-through-blockmanager) to have Executors push results into the driver's BlockManager, but I'm no longer convinced that its push-approach is a good idea (note that there's currently a small thread-safety issue in that branch's code, so I don't recommend trying it yet).  With a push mechanism, the master may be forced to buffer large amounts of partially-received task results, whereas a pull-based approach would let the master control the scheduling of those fetches.  I don't think that it's a good idea to perform expensive blocking fetches inside the TaskSetManager's task completion event handling code, so this could be tricky to implement efficiently.

More generally, I'm not convinced that the frame size limitation is the biggest problem facing Spark's `reduce` code.  Right now, all reduce tasks send their results to the driver (the same applies to accumulator updates, since these are essentially reduces).  This typically means that the number of intermediate reduce outputs received by the master is proportional to the _number of partitions_ instead of the number of machines.  In many machine learning applications, where you're frequently reducing large dense vectors (in the range of hundreds of megabytes to gigabytes), this can be a huge bottleneck.

This bottleneck can be avoided using techniques like machine-local aggregation of task results ([SPARK-664](https://spark-project.atlassian.net/browse/SPARK-664)) and aggregation trees.  You can also use AllReduce-style aggregation tricks to eliminate the "driver-as-a-middleman" (I have a [_very_ rough example](https://github.com/JoshRosen/spark/blob/8e558d852ec4d6026e38888c338f7304a01ba6b3/examples/src/main/scala/spark/examples/SparkLRCollectiveCommunication.scala) of this written as a user-level Spark program).

Automatically performing those aggregation optimizations will require bigger architectural changes to how we collect task results (I have some detailed notes on the challenges involved in this, which I may post to the `spark-developers` mailing list at some point).  In the long run, I think it would be better to address this larger issue in whatever TaskResult modification we make.

Getting back to this pull request, we might want to apply this patch in `branch-0.7` in case we don't want to backport improvements to result-fetching code.  I think we're going to release Spark 0.7.1 pretty soon, so I'd like at least an improved warning message to make in into that branch.

rxin:
Thanks.

PR #613, opened 2013-05-15
Regularly poll executors to track their utilization

squito:
This isn't ready to be merged yet (needs tests & docs), but wanted to get some feedback.

The point of this PR is to provide a really high-level metric on how well the cluster is being utilized, by simply polling what percentages of cores are active.  I'm hopefully this will diagnose a lot of cases where jobs are slow b/c spark is being used incorrectly -- imbalanced jobs, driver spending all its time merging accumulator values, long breaks between stages, etc.

The ugly part of this is plumbing the events from StandaloneSchedulerBackend --> ClusterScheduler --> SparkContext --> DAGScheduler --> SparkListener.  I'm not sure what the right architecture is in general to let any arbitrary component get an event to the SparkListeners.


core/src/main/scala/spark/scheduler/SparkListener.scala
```diff
@@ -44,6 +73,13 @@ class StatsReportListener extends SparkListener with Logging {
     showDistribution("other time pct: ", Distribution(runtimePcts.map{_.other * 100}), "%2.0f %%")
   }
 
+  def onExecutorStatusUpdate(executorStatus: ExecutorStatus) {
+    //update ALL active stages
+    activeStageToExecutorStatus.foreach{case(k,v) =>
+      activeStageToExecutorStatus += k -> (v + executorStatus)
+    }
```
squito, line 80: Note that `ExecutorStatus` messages even happen when there aren't any active stages.  This means that you can also measure the cluster utilization across the lifetime of a spark context, not just within a stage.  Eg., this would help diagnose if every stage is very well distributed, but between stages there is a lot of work happening on the master.  I just decided to not include that in `StatsReportListener`.

(You could achieve the same effect w/out actually sending all the messages when there are no active stages, but the cluster is idle anyway, so why not.)

rxin:
Why not ganglia or other external monitoring tools?

core/src/main/scala/spark/scheduler/ExecutorStatusPoller.scala
```diff
@@ -0,0 +1,59 @@
...
+
+import scala.collection.mutable
+import spark.scheduler.cluster.StandaloneSchedulerBackend
+import spark.scheduler.local.LocalScheduler
+import spark.Logging
+
+/**
+ *
+ */
+abstract class ExecutorStatusPoller extends Logging {
```
ryanlecompte, line 12: I'd suggest using a java.util.concurrent.ScheduledThreadPoolExecutor instead of an infinite while loop. This would also let you schedule the poller at a fixed interval without having to manage the sleep "catchup" time yourself, e.g.:

``` scala
  val pool = Executors.newSingleThreadScheduledExecutor()
  val poller = new Runnable() {
    override def run() { // poll each executorId here }
  }

  // schedule repeated task
  pool.scheduleAtFixedRate(poller, 0, waitBetweenPolls, TimeUnit.MILLISECONDS)
```

This also lets you gracefully stop the poller via:

``` scala
  // gracefully shutdown the poller
  pool.shutdown()
  pool.awaitTermination(30, TimeUnit.SECONDS)
```

core/src/main/scala/spark/scheduler/cluster/StandaloneSchedulerBackend.scala
```diff
@@ -23,12 +23,14 @@ class StandaloneSchedulerBackend(scheduler: ClusterScheduler, actorSystem: Actor
 
   // Use an atomic variable to track total number of cores in the cluster for simplicity and speed
   var totalCoreCount = new AtomicInteger(0)
+  var allExecutors = new HashSet[String]
```
ryanlecompte: Instead of having two state tracking variables (allExecutors and totalCores), could we not just have a single Map[String, Int] where the key is the executor id and the value would be # of cores? Something like:

``` scala
  val allExecutors = Map.empty[String, Int].withDefaultValue(0)
```

squito:
@rxin yeah ganglia could provide something pretty close to this.  But I thought this was useful b/c

(1) this is so simple, its useful to have even if you don't setup ganglia. And as this is integrated right into spark, its easier to connect these measures w/ whats going on in your code (you don't just have a ganglia graph w/ times, which then you've got connect back to what was going on in your code).  Its not measuring the exact same thing as ganglia would w/ core utilization, but I can't make a really strong case why this is particularly better.  If there is eventually really tight integration w/ ganglia, then maybe this could get ripped out.

(2) I was hoping that this might get used for more thorough polling of the executors, eg. stack trace sampling, task progress, etc.  So might be a useful stepping stone even if the "core utilization" part is dropped eventually.

core/src/main/scala/spark/scheduler/ExecutorStatusPoller.scala
squito, line 12: good point, I will make that change.  This also got me thinking -- do I even want to create a new thread at all? Is there an appropriate thread pool for these repeated tasks already?

mateiz:
Imran, this approach looks good to me, but I'm going to send it to Patrick, who's been looking at monitoring stuff too. I think these are reasonable API calls to add to the listener though.

pwendell:
Hey Imran,

Just a high level question (haven't done a close look yet). If this is all of the information we are collecting - why do you need to poll the executors in the first place?

The information of which tasks are running on which executor when is available directly at the driver. You could actually get much finer grained utilization statistics using that information without the need to add RPC's.
- Patrick

squito:
Ah well, part of the motivation for this is that we noticed huge delays between when the executor thinks its finished a task, and when the driver fully registered it -- over 50% of the time actually for some of our workloads.  We were able to fix this when we discovered it (in our case it seemed to be mostly the cost of merging the accumulator results on the driver), but I'd really like to make this inefficiency more obvious.

If we really wanted to, we could probably do this entirely within the driver, but I guess I did it this way b/c I was hoping to piggyback more info on top of those RPCs in the future -- eg., maybe jvm metrics, shuffle status, counters could also come via the same mechanism.

stephenh:
FWIW I've wanted a quick & dirty "what's the cluster utilization like?" UI for a long time--agreed that other tools should be used for more extensive monitoring, but it would be nice to have some really basic info available for free/out of the box.

(I haven't looked at the code, so can't comment on the approach/implementation, just chiming in to say I'd enjoy seeing this happen.)

PR #614, opened 2013-05-16
Running spark on yarn

mridulm:
Documentation change - I see folks tripping on this in the spark-user group.
Hopefully, this might help things !

### DIFF
--- a/core/src/main/scala/spark/SparkContext.scala
+++ b/core/src/main/scala/spark/SparkContext.scala
@@ -537,6 +537,8 @@ class SparkContext(
    * filesystems), or an HTTP, HTTPS or FTP URI.
    */
   def addJar(path: String) {
+    // weird - debug why this is happening.
+    if (null == path) return
     val uri = new URI(path)
     val key = uri.getScheme match {
       case null | "file" => env.httpFileServer.addJar(new File(uri.getPath))
--- a/core/src/main/scala/spark/SparkContext.scala
+++ b/core/src/main/scala/spark/SparkContext.scala
@@ -537,8 +537,10 @@ class SparkContext(
    * filesystems), or an HTTP, HTTPS or FTP URI.
    */
   def addJar(path: String) {
-    // weird - debug why this is happening.
-    if (null == path) return
+    if (null == path) {
+      logInfo("null specified as parameter to addJar")
+      return
+    }
     val uri = new URI(path)
     val key = uri.getScheme match {
       case null | "file" => env.httpFileServer.addJar(new File(uri.getPath))
--- a/docs/running-on-yarn.md
+++ b/docs/running-on-yarn.md
@@ -11,14 +11,32 @@ Ex:  mvn -Phadoop2-yarn clean install
 
 # Building spark core consolidated jar.
 
-Currently, only sbt can buid a consolidated jar which contains the entire spark code - which is required for launching jars on yarn.
-To do this via sbt - though (right now) is a manual process of enabling it in project/SparkBuild.scala.
+We need a consolidated spark core jar (which bundles all the required dependencies) to run Spark jobs on a yarn cluster.
+This can be built either through sbt or via maven.
+
+- Building spark assembled jar via sbt.
+It is a manual process of enabling it in project/SparkBuild.scala.
 Please comment out the
   HADOOP_VERSION, HADOOP_MAJOR_VERSION and HADOOP_YARN
 variables before the line 'For Hadoop 2 YARN support'
 Next, uncomment the subsequent 3 variable declaration lines (for these three variables) which enable hadoop yarn support.
 
-Currnetly, it is a TODO to add support for maven assembly.
+Assembly of the jar Ex:  
+./sbt/sbt clean assembly
+
+The assembled jar would typically be something like :
+./streaming/target/spark-streaming-<VERSION>.jar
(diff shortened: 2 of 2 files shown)


mridulm:
The fixes are done, waiting for review/merge.

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -537,6 +537,8 @@ class SparkContext(
    * filesystems), or an HTTP, HTTPS or FTP URI.
    */
   def addJar(path: String) {
+    // weird - debug why this is happening.
```
rxin: Do you have a stack trace on when this happens?

mridulm: Not right now (at home)
But it is an NPE coming from addJar usage from SparkContext init ... I will
need to dig into why that is happening.
But that is decoupled from this fix ... Since users can add jar incorrectly
with null as param, it solves that here while allowing examples to run on
yarn.

Regards
Mridul

mridulm: Ok, figured out the issue out - SparkPi tries to add a jar from env variable 'SPARK_EXAMPLES_JAR'.
Which need not (or wont) be set when running in yarn mode (we have to use the --jar option).

## So the code snippet from SparkPi

```
val spark = new SparkContext(args(0), "SparkPi",
  System.getenv("SPARK_HOME"), Seq(System.getenv("SPARK_EXAMPLES_JAR")))
```

## 

will cause jars seq to have a null within it : hence the addJar null check is required and solves the issue.

JoshRosen: It looks like `SPARK_EXAMPLES_JAR` becoming null has caused problems for other users: https://groups.google.com/d/msg/spark-users/x5UczgI-Xm8/wzMm3Mb77-oJ

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -537,6 +537,8 @@ class SparkContext(
...
+    if (null == path) return
```
JoshRosen: I would log a warning warning here, since it will help users to debug their programs if they crash because JARs that they expected to add were not copied to the workers.

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -537,6 +537,10 @@ class SparkContext(
    * filesystems), or an HTTP, HTTPS or FTP URI.
    */
   def addJar(path: String) {
+    if (null == path) {
+      logInfo("null specified as parameter to addJar")
```
rxin, line 541: Can we change this to logWarning, and also print out the stack trace?

rxin, line 541: Actually - do you mind doing this in a separate PR? I am not entirely sure if it is the right behavior to ignore null paths, because it suggests a programming error.

The documentation changes should definitely go in.

mridulm, line 541: The reason this PR was submitted was to fix issues observed by users trying to run spark examples on yarn.

It was initially meant to be only a documentation change : until we observed that SparkPi was making assumptions about environment variables - which are not relevant to yarn mode, but is in other modes.

Actually this api, addJar, is not useful in yarn mode imo.

While it could be argued that all usage of the api (explicitly or implicitly) should defend against this sort of bug, practically (particularly when there are different modes of usage) it is not always possible. Ignore if error is deemed beneign or fail with exception if it is severe - in this case, ignore since it is expected failure.

Ofcourse, this is just my opinion to ensure that exposed api is robust !

Please feel free to cherry pick the changes as required ...

mridulm, line 541: The formatting is getting messed up and resulting in a single blob of text ... sigh.

docs/running-on-yarn.md
```diff
@@ -11,14 +11,32 @@ Ex:  mvn -Phadoop2-yarn clean install
...
   HADOOP_VERSION, HADOOP_MAJOR_VERSION and HADOOP_YARN
 variables before the line 'For Hadoop 2 YARN support'
 Next, uncomment the subsequent 3 variable declaration lines (for these three variables) which enable hadoop yarn support.
 
-Currnetly, it is a TODO to add support for maven assembly.
+Assembly of the jar Ex:  
+./sbt/sbt clean assembly
+
+The assembled jar would typically be something like :
+./streaming/target/spark-streaming-<VERSION>.jar
```
rxin: Why are we using streaming here?

core/src/main/scala/spark/SparkContext.scala
rxin, line 541: Thanks. I will do a manual merge. One more issue with the doc update. Can you take a look?

docs/running-on-yarn.md
mridulm: That was a typo - I responded with the right path to the users and
copy-pasted wrong path to doc, sigh ...
Good catch - thanks ! Committed it.

On a related note though : the assembly happening in sbt is actually wrong
- ideally we would need to include all aspects of spark in the assembled
  jar - like what is in maven build.
  streaming, repl (others ?) are not included in sbt build ...

Regards,
Mridul

rxin:
I merged this. Thanks!

PR #615, opened 2013-05-16
Maven build fix & two other small changes

rxin:
1. Added netty-all dependency to Maven build file.
2. Removed the duplicated netty dependency in sbt build file.
3. Changed run script to check for repl classes only if repl is used. In the case repl is not used, check for core classes only.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -88,6 +88,10 @@
       <artifactId>mesos</artifactId>
     </dependency>
     <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-all</artifactId>
+    </dependency>
+    <dependency>
       <groupId>log4j</groupId>
       <artifactId>log4j</artifactId>
     </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -257,6 +257,11 @@
         <version>${mesos.version}</version>
       </dependency>
       <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-all</artifactId>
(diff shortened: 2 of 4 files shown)


mridulm:
Netty has been explicitly added to sbt to ensure that the highest version which is backwardly compatible is picked up - and the others (added via dependencies) are ignored. Picking the wrong version as part of assembly causes runtime failures since various components which spark depend on pull various versions.

Since I do not follow netty much, I am not sure if there are any backwardly incompatible changes in 4.0.0.Beta2 ... if yes and relevant, then the various dependencies we have will break in various corner cases.

rxin:
netty-all is different from netty.

Without that dependency, master branch as is won't build in maven.

mridulm:
From the http://netty.io/news/2012/07/13/4-0-0-alpha1.html and from http://netty.io/downloads.html, it looks like netty-all is 4.0.\* version.
And 4.0, based on cursory glace at http://netty.io/news/index.html , has backwardly incompatible changes.

Whether these are relevant to spark or not (directly or indirectly for the dependencies), I am not sure of.

rxin:
Compatibility wise it is most likely ok because netty changed all the package names (from org.jboss to io.netty). So the part of Spark that depends on Netty 4 won't conflict with the other Spark dependencies that use Netty 3.x.

mridulm:
Good point. Hopefully, there are no other interactions ! (env, etc)

Thanks,
Mridul

shivaram:
Looks good to me

PR #616, opened 2013-05-19
Exclude old versions of Netty from Maven-based build

jey:
Ports over the Netty exclusion changes in 1b169f190c5c5210d088faced86dee1007295ac8 to Maven


rxin:
I think this gets rid of the netty dependency in streaming. But to be safe, maybe we should follow the way sbt is doing it?
1. Add an explicit dependency on 3.5.x version of Netty in pom.xml
2. Add the netty dependency and exclusions to pom.xml for both core and streaming.

mateiz:
Thanks for submitting this.

PR #621, opened 2013-05-24
Use ec2-metadata in start-slave.sh to detect if running on EC2

JoshRosen:
This should finally fix [SPARK-613](https://spark-project.atlassian.net/browse/SPARK-613), the issue where the standalone web UI would link to internal addresses.

It turns out that #419 applied the same change, but only to `start-master.sh`, so some workers were still starting their web UI's using internal addresses.

After merging, please cherry-pick this into branch-0.7.

### DIFF
--- a/bin/start-slave.sh
+++ b/bin/start-slave.sh
@@ -6,7 +6,8 @@ bin=`cd "$bin"; pwd`
 # Set SPARK_PUBLIC_DNS so slaves can be linked in master web UI
 if [ "$SPARK_PUBLIC_DNS" = "" ]; then
     # If we appear to be running on EC2, use the public address by default:
-    if [[ `hostname` == *ec2.internal ]]; then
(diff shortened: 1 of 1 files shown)


massie:
This failed because it was aborted, Josh. We'll rerun it. Jenkins, test this please.

JoshRosen:
I don't think that the tests even exercise the `start-master.sh` and `start-slave.sh` scripts (let alone testing them on EC2), so we probably don't need to run them for this PR.

Honestly, this PR is just for formality: the change is identical to #419 and should be uncontroversial.  I probably should have just immediately merged my own PR (that's nicer than just committing it to master myself, since it creates a thread for discussion and notifies repo notification subscribers).

JoshRosen:
Alright, merged into master and cherry-picked into branch-0.7 as 2d84917d7be86051fc6ff56563cf72c0b1541bdf

PR #622, opened 2013-05-24
bug fix: Shuffle block iterator is ignoring the shuffle serializer setting.

rxin:
This was introduced when we merged Shane's shuffle change. 

I also added a unit test to make sure we can catch this bug in the future.


mateiz:
Thanks, looks good

rxin:
BTW this exposed a problem in the netty shuffle code, which always bind to the same port. Shiv and I are fixing that now ...

PR #623, opened 2013-05-24
Automatically configure Netty port.

rxin:
This makes unit tests using local-cluster pass. Previously they were failing because Netty was trying to bind to the same port for all processes.

Pair programmed with @shivaram.


mateiz:
Cool, thanks for fixing this.

PR #625, opened 2013-05-28
Fix start-slave not passing instance number to spark-daemon.

stephenh:
We updated to a snapshot version of master and our calls to start-slave failed because spark-daemon now expects number-of-instances as a parameter before the command args.


mateiz:
Good catch, thanks.

PR #626, opened 2013-05-28
Remove unused addIfNoPort.

stephenh:
Was reading around the new host/port code and noticed this was unused.

### DIFF
--- a/core/src/main/scala/spark/Utils.scala
+++ b/core/src/main/scala/spark/Utils.scala
@@ -403,17 +403,6 @@ private object Utils extends Logging {
     hostPortParseResults.get(hostPort)
   }
 
-  def addIfNoPort(hostPort: String,  port: Int): String = {
-    if (port <= 0) throw new IllegalArgumentException("Invalid port specified " + port)
-
-    // This is potentially broken - when dealing with ipv6 addresses for example, sigh ... but then hadoop does not support ipv6 right now.
-    // For now, we assume that if port exists, then it is valid - not check if it is an int > 0
-    val indx: Int = hostPort.lastIndexOf(':')
-    if (-1 != indx) return hostPort
-
-    hostPort + ":" + port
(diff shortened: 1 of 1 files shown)


rxin:
Jenkins, test this please.

stephenh:
Huh. Seems odd that the tests would have failed for this? The string "addIfNoPort" does not exist anywhere in the spark repo, AFAICT, and if it did, it should result in a compile failure, not the tests timing out...

rxin:
Our Jenkins has been flaky lately... it complains almost every time a new pull request comes.

Jenkins, test this please.

Jenkins, test this again please.

jey:
The root cause is a non-deterministic hang in the `DistributedSuite` tests.

rxin:
But the error message seems to suggest it failed in compilation?

jey:
When I click the link that Jenkins posted and select "console output" from the left-hand menu, I'm taken [here](https://amplab.cs.berkeley.edu/jenkins/job/SparkPullRequestBuilder/67/console), and it looks like it is hanging sometime after completing `ShuffleSuite` (probably during `DistributedSuite`?).

rxin:
Ah ok - because the message says

```
Build timed out (after 90 minutes). Marking the build as aborted.
Build was aborted
Recording test results
Finished: ABORTED
```

I actually thought it was running the build itself rather than running the tests. 

How did you tell it was running Shuffle or Distributed? Neither appeared in the log. Did you figure it out just by looking at the log output and pattern-matching the output?

jey:
There's a tiny "Full Log" link at the top of the "Console Output" page

mridulm:
addIfNoPort used to be relevant before the move to executorId - you are
right, it is not used anymore.

Regards,
Mridul

mateiz:
Good catch, going to merge it then.

PR #627, opened 2013-05-29
Netty and shuffle  bug fixes

shivaram:
This patch fixes some bugs in the recent netty shuffle refactoring. This also fixes a DiskBlockObjectWriter bug where some blocks were reported to have zero size when using kryo. Thanks to @jey for catching this and @rxin for various suggestions.

### DIFF
--- a/core/src/main/java/spark/network/netty/FileServer.java
+++ b/core/src/main/java/spark/network/netty/FileServer.java
@@ -37,29 +37,33 @@ class FileServer {
         .childHandler(new FileServerChannelInitializer(pResolver));
     // Start the server.
     channelFuture = bootstrap.bind(addr);
-    this.port = addr.getPort();
+    try {
+      // Get the address we bound to.
+      InetSocketAddress boundAddress =
+        ((InetSocketAddress) channelFuture.sync().channel().localAddress());
+      this.port = boundAddress.getPort();
+    } catch (InterruptedException ie) {
+      this.port = 0;
+    }
   }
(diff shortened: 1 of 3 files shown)


mateiz:
Jenkins: this is ok to test.

rxin:
To provide a little bit more context, there are three problems this pull request solved:
1. Our previous unit tests didn't actually test the Netty shuffle code path because it was starting a local cluster with 1 process (2 cores) instead of 2 processes (1 core/each). There was no cross-process shuffle block fetches at all ...
2. The Netty FileServer gets started, and then immediately shut down. This pull request fixes that too.
3.  BlockObjectWriter's commit method doesn't force the serializer to flush before checking the size. This works fine for Java, but would fail for Kryo because Kryo manages its own output buffer.

shivaram:
I also updated `revertPartialWrites` to similarly flush the serializer when we have a task failure.

rxin:
Jenkins: this is ok to test.

Jenkins: test this please.

Jenkins: test this again please.

jey:
Jenkins, retest this please.

rxin:
Is that the magic word? :)

rxin:
Actually since this no longer merges, it doesn't make sense for Jenkins to run the test. Shivaram will update the pr to do the merge.

shivaram:
Fixed the merge. I am also running tests locally right now

rxin:
Thanks!

PR #628, opened 2013-05-29
Skip fetching zero-sized blocks in NIO.

shivaram:
This avoids making RPCs if blocks we need to fetch have zero bytes. Also this change unifies the `splitLocalRemoteBlocks` method for Netty/NIO which had the same code


rxin:
While you are at it already, do you think if we can combine the fetch request generation logic in Netty and the non-Netty code path? As far as I can tell, the fetch request generation is the same. The only difference is in the way fetches are done.

shivaram:
The fetch requests are created in the `splitLocalRemoteBlocks` method that is combined in this change. They are then enqueued in `initialize()`  which is not shared in netty/NIO. However `initialize` also starts the shuffle copiers and uses a linked blocking queue in netty, so I don't think its easy to combine that.

rxin:
Ah that looks good then. I missed the part where you deleted the huge trunk of the code.

rxin:
Jenkins: this is ok to test.

rxin:
Thanks.

PR #629, opened 2013-05-30
Sometime Maven build runs out of PermGen space.

ghost:
I have see PermGen issues on a number of occasions in the Maven build. This patch seems to fix it for good.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,6 +59,9 @@
     <slf4j.version>1.6.1</slf4j.version>
     <cdh.version>4.1.2</cdh.version>
     <log4j.version>1.2.17</log4j.version>
+
+    <PermGen>0m</PermGen>
+    <MaxPermGen>512m</MaxPermGen>
   </properties>
 
   <repositories>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -392,6 +392,10 @@
             <jvmArgs>
               <jvmArg>-Xms64m</jvmArg>
               <jvmArg>-Xmx1024m</jvmArg>
+              <jvmArg>-XX:PermSize</jvmArg>
+              <jvmArg>1024m</jvmArg>
+              <jvmArg>-XX:MaxPermSize</jvmArg>
+              <jvmArg>1024m</jvmArg>
```
mateiz: This is a bit extreme.. I would set the MaxPermSize to 256m and have no initial PermSize setting. That's the same thing we have in SBT.

ghost: 1GB dev machine doesn't sound that extreme ;) But I will post a different patch to allow a smaller default value with a runtime parametrization via sys.prop.

pwendell: A lot of use develop on our laptops, and will have memory hogs running
already (browser, IDE, etc) so I'd also prefer to prevent excessive memory
usage during the build.

ghost: That's what newer version of the pull request addresses hopefully.

mateiz:
Alright, thanks!

PR #630, opened 2013-05-30
Add ASF release repo so Hadoop artifacts can be downloaded from source.

ghost:
It seems to be more appropriate to use real source of the artifacts rather than a mirror


rxin:
Cos, is this actually only a "staging" repo, as the name suggests ("ASF staging Repository")?

ghost:
My bad - I was thinking of something else. This is the release repository - I have fixed the name. Sorry.

mateiz:
I'm still confused as to why we'd add this -- won't the same releases make it into Maven Central? Is this just for alpha or nightly releases? Basically I don't want to force all users to depend on this repo -- if they want it, they could link it in their own project (especially when we change Spark to work with any Hadoop client version).

ghost:
Well, you are right in this regard: anything can be configured within ~/.m2/settings.xml if needed. My original though was that it might be convenient to pick up recent releases before they got propagated to http://mvnrepository.com or else.

On the second thought, the idea doesn't look that good. Let's drop it. Appreciate your looking into this!

PR #631, opened 2013-05-30
Fix block manager UI display issue when enable spark.cleaner.ttl

jerryshao:
When spark.cleaner.ttl is enabled, RDD id got from BlockManger's block id may not have correspondence in SparkContext's `persistentRdds`, `persistentRdds` has already clean up this RDD, which will make `val rdd = sc.persistentRdds(rddId)` throw `NoSuchElementException` . So here cleaned RDD should be filted out.

### DIFF
--- a/core/src/main/scala/spark/storage/StorageUtils.scala
+++ b/core/src/main/scala/spark/storage/StorageUtils.scala
@@ -64,12 +64,12 @@ object StorageUtils {
       // Find the id of the RDD, e.g. rdd_1 => 1
       val rddId = rddKey.split("_").last.toInt
       // Get the friendly name for the rdd, if available.
-      val rdd = sc.persistentRdds(rddId)
-      val rddName = Option(rdd.name).getOrElse(rddKey)
-      val rddStorageLevel = rdd.getStorageLevel
-
-      RDDInfo(rddId, rddName, rddStorageLevel, rddBlocks.length, rdd.partitions.size, memSize, diskSize)
-    }.toArray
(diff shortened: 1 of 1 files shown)


core/src/main/scala/spark/storage/StorageUtils.scala
```diff
@@ -64,12 +64,15 @@ object StorageUtils {
...
-    }.toArray
+      val rddInfo = sc.persistentRdds.get(rddId) match {
+        case Some(r) =>
+          val rddName = Option(r.name).getOrElse(rddKey)
+          val rddStorageLevel = r.getStorageLevel
+          Some(RDDInfo(rddId, rddName, rddStorageLevel, rddBlocks.length, r.partitions.size, memSize, diskSize))
+        case None => None
+      }
+      rddInfo
+    }.flatMap(x => x).toArray
```
markhamstra, line 73: Can we tighten this up a bit?  Instead of

``` scala
val rddInfo = sc.persistentRdds.get(rddId) match {...
  case None => None
}
rddInfo
```

just

``` scala
sc.persistentRdds.get(rddId).map {r =>
  val rddName = Option(r.name).getOrElse(rddKey)
  val rddStorageLevel = r.getStorageLevel
  RDDInfo(rddId, rddName, rddStorageLevel, rddBlocks.length, r.partitions.size, memSize, diskSize)
  }
```

jerryshao:
Hi, thanks for your comments, I've already updated this, please review. Thanks

mateiz:
Thanks for the fix.

PR #632, opened 2013-05-30
Implement fair scheduler for spark local mode

xiajunluan:
1. Abstract the TaskSetManager interface for ClusterTaskSetManager and LocalTaskSetManager
2. Use original thread pool for scheduler resource
3. Finish the local end-to-end test for scheduler algorithm


mateiz:
Jenkins, this is ok to test.

core/src/main/scala/spark/scheduler/cluster/ClusterTaskSetManager.scala
```diff
@@ -0,0 +1,734 @@
...
+
+      retval
+    } catch {
+      case nEx: NoSuchElementException => {
+        logWarning("Invalid task locality specified '" + str + "', defaulting to NODE_LOCAL");
+        // default to preserve earlier behavior
+        NODE_LOCAL
+      }
+    }
+  }
```
mateiz, line 51: Why do we need to allow invalid locality strings? I'd rather not allow them and set various config options, etc to have good defaults. Printing a warning in this case is too risky, users might miss it.

mridulm, line 51: This is primarily used to parse system property specified user - so in case
of errors : default to node_local after flagging a warning.

mateiz, line 51: Ah, I see, this is older code that was moved. I would still prefer to throw an exception here instead of warning, but we can do it in a different patch since this patch doesn't contain the code that actually parses this. Basically my thinking is that if the user has a mistake in their config file, why continue?

core/src/main/scala/spark/scheduler/local/LocalScheduler.scala
```diff
@@ -2,19 +2,51 @@ package spark.scheduler.local
...
+      launchTask(localScheduler.resourceOffer(freeCores))
+    case LocalStatusUpdate(taskId, state, serializeData) =>
+      freeCores += 1
+      localScheduler.statusUpdate(taskId, state, serializeData)
+      launchTask(localScheduler.resourceOffer(freeCores))
+  }
+
+  def launchTask(tasks : Seq[TaskDescription]) {
+    for (task <- tasks)
+    {
```
mateiz: Put this curly brace on the previous line

core/src/main/scala/spark/scheduler/local/LocalScheduler.scala
```diff
@@ -30,89 +62,128 @@ private[spark] class LocalScheduler(threads: Int, maxFailures: Int, sc: SparkCon
...
+          new FIFOSchedulableBuilder(rootPool)
+        case "FAIR" =>
+          new FairSchedulableBuilder(rootPool)
+      }
+    }
+    schedulableBuilder.buildPools()
 
-  override def start() { }
+    localActor = env.actorSystem.actorOf(
+      Props(new LocalActor(this, threads)), "Test")
```
mateiz: This can fit on one line

core/src/main/scala/spark/scheduler/local/LocalScheduler.scala
```diff
@@ -30,89 +62,128 @@ private[spark] class LocalScheduler(threads: Int, maxFailures: Int, sc: SparkCon
...
+  }
+
+  def resourceOffer(freeCores: Int): Seq[TaskDescription] = {
+    synchronized {
+      var freeCpuCores = freeCores
+      val tasks = new ArrayBuffer[TaskDescription](freeCores)
+      val sortedTaskSetQueue = rootPool.getSortedTaskSetQueue()
+      for (manager <- sortedTaskSetQueue) {
+        logInfo("parentName:%s,name:%s,runningTasks:%s".format(manager.parent.name, manager.name, manager.runningTasks))
+      }
```
mateiz, line 111: This should probably be logDebug instead of logInfo

core/src/main/scala/spark/scheduler/local/LocalTaskSetManager.scala
```diff
@@ -0,0 +1,173 @@
...
+    return None
+  }
+
+  def slaveOffer(execId: String, hostPort: String, availableCpus: Double, overrideLocality: TaskLocality.TaskLocality = null): Option[TaskDescription] = {
+    SparkEnv.set(sched.env)
+    logDebug("availableCpus:%d,numFinished:%d,numTasks:%d".format(availableCpus.toInt, numFinished, numTasks))
+    if (availableCpus > 0 && numFinished < numTasks) {
+      findTask() match {
+        case Some(index) =>
+          logInfo(taskSet.tasks(index).toString)
```
mateiz: This looks like a debug message -- you can remove it

xiajunluan:
Hi Matei

I have changed code based on your comments.

core/src/test/scala/spark/scheduler/LocalSchedulerSuite.scala
```diff
@@ -0,0 +1,171 @@
...
+        override def run() {
+          val ans = nums.map(number => {
+            TaskThreadInfo.threadToRunning(number) = true
+            TaskThreadInfo.threadToLock(number).jobWait()
+            number
+          }).collect()
+          assert(ans.toList === List(threadIndex))
+          sem.release()
+          TaskThreadInfo.threadToRunning(threadIndex) = false
+        }
```
mateiz: Indentation here seems wrong -- should be just two spaces

core/src/test/scala/spark/scheduler/LocalSchedulerSuite.scala
```diff
@@ -0,0 +1,171 @@
...
+    }.start()
+    Thread.sleep(2000)
```
mateiz: Is this to wait for the thread to get started? It would be better to use a condition variable or semaphore instead of sleeping, because this is a long time to sleep in a unit test.

core/src/test/scala/spark/scheduler/LocalSchedulerSuite.scala
```diff
@@ -0,0 +1,171 @@
...
+    createThread(6,null,sc,sem)
+    assert(TaskThreadInfo.threadToRunning(1) === true)
+    assert(TaskThreadInfo.threadToRunning(2) === true)
+    assert(TaskThreadInfo.threadToRunning(3) === true)
+    assert(TaskThreadInfo.threadToRunning(4) === true)
+    assert(TaskThreadInfo.threadToRunning(5) === false)
+    assert(TaskThreadInfo.threadToRunning(6) === false)
+    
+    TaskThreadInfo.threadToLock(1).jobFinished()
+    Thread.sleep(1000)
```
mateiz: This sleep would also be nice to replace with some kind of semaphore or shared state; at the very least have a loop that checks the `ThreadInfo.threadFinished` variable every 100ms, and make that a ConcurrentMap.

core/src/test/scala/spark/scheduler/LocalSchedulerSuite.scala
```diff
@@ -0,0 +1,171 @@
...
+  }
+}
+
+object TaskThreadInfo {
+  val threadToLock = HashMap[Int, Lock]()
+  val threadToRunning = HashMap[Int, Boolean]()
+}
+
+
+class LocalSchedulerSuite extends FunSuite with LocalSparkContext {
```
mateiz, line 49: This test could use a little bit of documentation at the top just to explain how it works. It doesn't need to be long, but just say that we'll launch one task per job and have the task's thread wait for a the test to signal it that it can finish (through TaskThreadInfo).

mateiz:
Hey Andrew,

Thanks for fixing those. I made a few other comments on the unit tests -- in particular, I'd really like to avoid having to sleep(2000) a bunch of times in there, because it will make the test take a long time. Other than that, this looks really good. Let me know when you've updated it!

core/src/test/scala/spark/scheduler/LocalSchedulerSuite.scala
ryanlecompte: You could also just use a simple java.util.concurrent.CountDownLatch with the proper size.

rxin:
Hi Andrew - just want to make sure you are aware that there are a couple more comments Matei/Ryan made on the code section, which is not displayed in the main pull request page here (kind of confusing).

xiajunluan:
Hi Reynold - I have got it, and will update the code after more tests. Thanks for reminder.

xiajunluan:
Hi Matei

I have added completion variable for thread to avoid sleep(2000) to wait job start, but for tasks that will pend for scheduling, I could not think out better solution other than sleep to wait stage to be added to taskSetManager queue for later scheduler when cluster has free resource, I have added some comments in codes for this scenario, pls help me to review codes and call me if there is better solution.

mateiz:
Hi Andrew,

The remaining sleeps seem fine, since there are very few of them. Thanks! I've merged this manually because your branch has diverged a bit from master and was missing two changes in TaskSetManager.

PR #633, opened 2013-05-30
Copy file instead of symlink for add file.

rxin:


rxin:
Also did some minor cosmetic cleanup of Utils.scala

rxin:
Jenkins finally liked me

PR #635, opened 2013-05-31
Consolidations of shuffle files from different map tasks

jason-dai:
A Spark job can have many small tasks, which in turn generates many small shuffle files - one of the major shuffle performance issues. This change combines multiple such files (for the same reducer) into one single file as follows (using the shuffle manager introduced by  Reynold's https://github.com/mesos/spark/pull/587):

1) On each slave, the shuffle manager maintains a shuffle block pool for each shuffle

2) Each pool maintains a list of shuffle block group; a ShuffleMapTask acquires a free group when it needs to write its results, and returns the group when it's done

3) Each group maintains a list of writers, each for a different bucket (reduce partition)


jason-dai:
The patch passes tests on my side following the normal path (i.e., no failures). I also introduced some random failures (i.e., throwing random exceptions during writing ShuffleMapTask's results), which helped catch a few bugs in the failure handling path. I think we need to perform more extensive failure testing - not sure how this fits into the existing test framework though.

rxin:
Jenkins, test this please.

rxin:
Ah it's 1:30am. I am going to bed now and will take a look at this tomorrow. Thanks for doing this.

rxin:
I haven't looked at your changes yet, but can you comment on the impact on fetch parallelism?

jason-dai:
For netty fetcher, there are no changes – its parallelism is controlled by the number of copiers. For basic fetcher, its parallelism is controlled by MaxBytesInFlight, and in this case its parallelism will be lower.

core/src/main/scala/spark/PairRDDFunctions.scala
```diff
@@ -69,12 +69,14 @@ class PairRDDFunctions[K: ClassManifest, V: ClassManifest](
     } else if (mapSideCombine) {
       val mapSideCombined = self.mapPartitions(aggregator.combineValuesByKey(_), true)
       val partitioned = new ShuffledRDD[K, C](mapSideCombined, partitioner, serializerClass)
+      logInfo("serializerClass=" + serializerClass)
```
rxin: Maybe change this to logDebug

core/src/main/scala/spark/MapOutputTracker.scala
```diff
@@ -117,7 +117,7 @@ private[spark] class MapOutputTracker extends Logging {
   private val fetching = new HashSet[Int]
 
   // Called on possibly remote nodes to get the server URIs and output sizes for a given shuffle
-  def getServerStatuses(shuffleId: Int, reduceId: Int): Array[(BlockManagerId, Long)] = {
+  def getServerStatuses(shuffleId: Int, reduceId: Int): Array[(BlockManagerId, Int, Long)] = {
```
rxin, line 122: Add a comment on what the Int and Long represent.

core/src/main/scala/spark/storage/DiskStore.scala
```diff
@@ -68,11 +73,28 @@ private class DiskStore(blockManager: BlockManager, rootDirs: String)
     }
 
     override def revertPartialWrites() {
+      // Revert by discarding current writes, except that if no values have been committed, 
+      // we revert by recreate the file (otherwise there are errors when reading objects from the file later on  
+      if (lastValidPosition == initialPosition)
+        recerateFile()
```
rxin: typo here

core/src/main/scala/spark/storage/DiskStore.scala
```diff
@@ -68,11 +73,28 @@ private class DiskStore(blockManager: BlockManager, rootDirs: String)
     }
 
     override def revertPartialWrites() {
+      // Revert by discarding current writes, except that if no values have been committed, 
+      // we revert by recreate the file (otherwise there are errors when reading objects from the file later on  
+      if (lastValidPosition == initialPosition)
```
rxin, line 78: initialPosition is set to the size of the file when the file is opened. Isn't it problematic if we open an existing file, did some writes and then want to revert, we could delete the old file?

core/src/main/scala/spark/storage/ShuffleBlockManager.scala
```diff
@@ -15,30 +22,158 @@ trait ShuffleBlocks {
...
 
-      override def releaseWriters(group: ShuffleWriterGroup) = {
-        // Nothing really to release here.
+  }
+  
+ /**
+  * Get a Shuffle Pool
+  */
+  private def getPool(shuffleId: Int) : ShuffleBlocksPool = {
+    pools.putIfAbsent(shuffleId,new ShuffleBlocksPool(shuffleId))
```
rxin: Ditto the other comment - don't create a new ShuffleBlocksPool everytime.

core/src/main/scala/spark/storage/ShuffleBlockManager.scala
```diff
@@ -15,30 +22,158 @@ trait ShuffleBlocks {
...
+    val nextGroupID = new AtomicInteger(0)
+  
+    // Check if this pool closed and not available for further use again. 
+    val isClosed = false
+
+   /**
+    * Acquire a new group from pool
+    */
+    def acquireGroup(numBuckets: Int, serializer: Serializer) : ShuffleWriterGroup = {
+        //TODO. throws exception now. This needs to be handled. maybe reopen it.
```
rxin: 2 space indent

core/src/main/scala/spark/storage/ShuffleBlockManager.scala
```diff
@@ -15,30 +22,158 @@ trait ShuffleBlocks {
...
+    */
+    def closeBlock(groupId:Int, bucketId:Int) {
+      if (allGroups(groupId.toInt).writers(bucketId.toInt).isOpen)
+        allGroups(groupId.toInt).writers(bucketId.toInt).close()
+    }
+
+   /**
+    * Close all writers in a group. Reserved for possible optimization
+    */
+    def closeGroup(groupId:Int){
```
rxin: add a whitespace  between ) and {

core/src/main/scala/spark/storage/ShuffleBlockManager.scala
```diff
@@ -15,30 +22,158 @@ trait ShuffleBlocks {
...
+      else
+        false
+    }
+ 
+    /**
+    * Open all writers in a group
+    */
+    private def openGroup(group: ShuffleWriterGroup) : ShuffleWriterGroup = {
+      //open all the writers
+      group.writers.map { writer => { if(!writer.isOpen) writer.open()} }
```
rxin:
```
group.writers.map { writer => if (!writer.isOpen) writer.open() }
```

i.e. whitespae after if, and don't wrap the extra { }

core/src/main/scala/spark/storage/ShuffleBlockManager.scala
```diff
@@ -15,30 +22,158 @@ trait ShuffleBlocks {
...
+      group
+    }
 }
 
+  //keep track of pools for all shuffles indexed by Id
+  val pools = new ConcurrentHashMap[Int,ShuffleBlocksPool]
```
rxin: add whitespace after ","

core/src/main/scala/spark/storage/ShuffleBlockManager.scala
```diff
@@ -15,30 +22,158 @@ trait ShuffleBlocks {
...
+    def closeBlock(groupId:Int, bucketId:Int) {
+      if (allGroups(groupId.toInt).writers(bucketId.toInt).isOpen)
+        allGroups(groupId.toInt).writers(bucketId.toInt).close()
+    }
+
+   /**
+    * Close all writers in a group. Reserved for possible optimization
+    */
+    def closeGroup(groupId:Int){
+      allGroups(groupId.toInt).writers.map { _.close()}
```
rxin: add whitespace after a close()

core/src/main/scala/spark/storage/ShuffleBlockManager.scala
```diff
@@ -15,30 +22,158 @@ trait ShuffleBlocks {
...
+   * retrieved by one thread
+   */
+  def closeBlock(blockId: String) {
+    val regex = "shuffle_([0-9]*)_([0-9]*)_([0-9]*)".r
+    blockId match {
+      case regex(shuffleId, groupId, bucketId) => 
+        logDebug("closeBlock shuffleId: " + shuffleId + ", groupId: " + groupId + ", bucketId: " + bucketId)
+        val pool = getPool(shuffleId.toInt)
+        if (pool != null)
+          pool.closeBlock(groupId.toInt,bucketId.toInt)
```
rxin: space after ","

core/src/main/scala/spark/storage/BlockManager.scala
```diff
@@ -287,6 +287,7 @@ private[spark] class BlockManager(
    * never deletes (recent) items.
    */
   def getLocalFromDisk(blockId: String, serializer: Serializer): Option[Iterator[Any]] = {
+    shuffleBlockManager.closeBlock(blockId)
```
rxin, line 290: Can you add comment to the ShuffleBlockManager explaining the blocks are closed here?

rxin:
Jason,

I made some comments. Most of them are minor style things, but there are two things that stood out:
1. Does revertPartialWrites delete pre-existing file?
2. Add a comment explaining how writers are closed in ShuffleBlockManager.

Also, one other thing I didn't point out: For if statements, just wrap all of them with { } unless it is an one-liner of if/else.

core/src/main/scala/spark/storage/BlockManager.scala
```diff
@@ -382,6 +383,8 @@ private[spark] class BlockManager(
     // As an optimization for map output fetches, if the block is for a shuffle, return it
     // without acquiring a lock; the disk store never deletes (recent) items so this should work
     if (ShuffleBlockManager.isShuffle(blockId)) {
+      //close the shuffle Writers for blockId
```
rxin, line 386: Do shuflfe blocks still walk through this code path? If no, I would just throw an exception if it is a shuffle block here.

jason-dai, line 386: The remote shuffle blocks still go here

core/src/main/scala/spark/storage/DiskStore.scala
jason-dai, line 78: A shuffle file is closed when we get a shuffle request, and it should never be re-opened (and if you re-open the file and append it, you may corrupt the file as it may have a tailer). Maybe we should first delete the shuffle file if it exits when we first open it (e.g., in case we need to re-run the map tasks).

rxin, line 78: But if we re-run the map tasks, wouldn't that wipe out the shuffle outputs for all other previous map tasks?

jason-dai, line 78: The shuffle file is closed when all the map tasks are done (after the shuffle request is received). If we need to re-run a map task afterwards, 
(1) There is no easy way to remove the results of the previous run of this task from the shuffle file
(2) You cannot re-open a closed shuffle file and append to it (e.g., kyro will append some tailer when closing the file) 
(3) If a fetch fails, Spark will re-run all the map tasks (for the same shuffle) whose output locations are at the failing node; so it's OK to delete their outputs.

I actually also tried throwing random exceptions during shuffle and it works fine. Again, is there a good way to include such tests?

rxin, line 78: Thanks for the response. It would be useful to actually put whatever you just wrote into the code comment block.

For failure testing, can we do something similar to what FailureSutie does?

core/src/main/scala/spark/storage/ShuffleBlockManager.scala
rxin: Also move this into ShuffleWriterGroup

rxin:
Hi Jason - any update on the failure paths?

jason-dai:
Sorry I was held up by something and this got delayed. I have not thought through all the failure scenarios, but here are some initial thoughts:

(1) Generally speaking, reverting by flushing and truncating can not be guaranteed to be correct when the encoder (compression or serialization) is stateful - e.g., delta encoding. A simple way to address this is to treat each shuffle block as a complete encoded file, and concatenating block files into a larger shuffle files.

(2) The previous failure handling is simple - the location of the map output is registered at the MapOutputTracker if the map task is successful; if anything goes wrong, it will not be registered (and consequently duplicate or invalid shuffle file will never be fetched). Things get complicated when merging multiple shuffle blocks into a larger file. Say, if there are two speculative map tasks complete at the same time, both their results will be written into (different) shuffle files successfully; it will be very complex to revert one after the map task is finished. It is much simpler if we can still rely on MapOutputTracker as the single source of truth for the map output locations.

Therefore, I would propose that the shuffle file be in the form of: {(map1, map1 output block), (map2, map2 output block), ..., (map1 size, map2 size, ...)}
We can still revert by truncating - but this will be best efforts; the MapOutputTracker will maintain the map output locations, which will be used to skip duplicate or invalid blocks when deserialization.

rxin:
Closing this one because it is subsumed by #669

PR #636, opened 2013-05-31
Unpersist More block manager cleanup.

rxin:
Implemented a removeRdd method in BlockManager, and use that to
implement RDD.unpersist. Previously, unpersist needs to send B akka
messages, where B = number of blocks. Now unpersist only needs to send W
akka messages, where W = the number of workers.

This commit was based on https://github.com/mesos/spark/pull/633 so please merge that one first.

### DIFF
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -1,13 +1,10 @@
 package spark
 
-import java.net.URL
-import java.util.{Date, Random}
-import java.util.{HashMap => JHashMap}
+import java.util.Random
 
 import scala.collection.Map
 import scala.collection.JavaConversions.mapAsScalaMap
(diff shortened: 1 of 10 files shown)


rxin:
@woggling can you take a look at this, since you have been poking block managers.

charlesreiss:
This looks okay to me. I'd prefer it if we didn't have potential (but rare for "good" usage of unpersist()) master/worker inconsistencies, but it doesn't seem like there's a simple fix that doesn't have likely performance issues. It would also be preferable if the master tracked block groups so it didn't have to scan the entire list to find them, but that should be another patch.

rxin:
@woggle I incorporated your comments.

charlesreiss:
LGTM.

mateiz:
This looks good to me too. Going to merge it.

PR #639, opened 2013-06-04
Bump akka and blockmanager timeouts to 60 seconds

pwendell:
Occasionally people experience large GC pauses that cause timeout failures. This  bumps the timeout value to a fairly conservative length - 60 seconds - to avoid this. The trade-off is that if nodes fail during a job it will take a bit longer to detect it. That rarely happens, so I think it's a good trade-off.

Also, this sets spark.akka.timeout to equal the blockmanager timeout.

### DIFF
--- a/core/src/main/scala/spark/storage/BlockManager.scala
+++ b/core/src/main/scala/spark/storage/BlockManager.scala
@@ -950,7 +950,7 @@ object BlockManager extends Logging {
   }
 
   def getHeartBeatFrequencyFromSystemProperties: Long =
-    System.getProperty("spark.storage.blockManagerHeartBeatMs", "5000").toLong
+    System.getProperty("spark.storage.blockManagerHeartBeatMs", "60000").toLong
 
   def getDisableHeartBeatsForTesting: Boolean =
     System.getProperty("spark.test.disableBlockManagerHeartBeat", "false").toBoolean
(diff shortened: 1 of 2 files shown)


mateiz:
Looks good, thanks! Can you send a PR for branch 0.7 also?

core/src/main/scala/spark/storage/BlockManager.scala
```diff
@@ -950,7 +950,7 @@ object BlockManager extends Logging {
   }
 
   def getHeartBeatFrequencyFromSystemProperties: Long =
```
shivaram, line 952: Afaik, this is the frequency with which heart beat messages are sent -- I think there is another flag `spark.storage.blockManagerSlaveTimeoutMs` which is by default `3 * heartBeatFrequency`. One thing we could consider is having frequent hearbeats (say 10s) but detecting a failure only when 10 heartbeats are missed.

mateiz, line 952: Oh, good point -- this version is wrong. The TimeoutMS is what matters. The heartbeat shouldn't even be configurable IMO -- we can just divide timeout by 4 or something.

pwendell, line 952: ah I see, that's strictly better. I'll do that and submit here and also to
0.7.

pwendell, line 952: Ya matei here it actually goes the other way here (the default for the
timeout is a multiplier of the heartbeat) that's why the heartbeat was set
in my PR. But I'll just reverse it to be like you said.

PR #640, opened 2013-06-04
Fixing bug in BlockManager timeout

pwendell:
Shivaram - does this look okay? This changes the timeout and sets the heartbeat as timeout / 4 (same strategy as is used in the standalone scheduler).


mateiz:
Looks good, thanks.

shivaram:
Looks good -- The only thing is we might need to inform users that we've removed the spark.storage.blockManagerHeartBeatMs flag in 0.8

utgarda:
Definitely :) I actually found this page while searching clues why the old settings don't work for me.

PR #641, opened 2013-06-04
Bump akka and blockmanager timeouts to 60 seconds

pwendell:


pwendell:
Matei - would you prefer for this patch to not to deprecate the flag? since it's in 0.7? We could just change only the timeout default.

pwendell:
I changed it so that the existing config option is unchanged...

mateiz:
I doubt anyone uses that flag, so I'm fine removing it. The fewer config options, the better. We never really documented it either.

PR #644, opened 2013-06-08
add Joblogger to Spark (on new Spark code)

shimingfei:
update from Pull-request 473, use interfaces in latest SparkContext

### DIFF
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -114,10 +114,10 @@ abstract class RDD[T: ClassManifest](
     this
   }
 
-  /**User-defined generator of this RDD*/
-  var generator = Utils.getCallSiteInfo._4
+  /** User-defined generator of this RDD*/
+  var generator = Utils.getCallSiteInfo.firstUserClass
   
-  /**reset generator*/
+  /** Reset generator*/
   def setGenerator(_generator: String) = {
     generator = _generator
   }
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -114,6 +114,14 @@ abstract class RDD[T: ClassManifest](
     this
   }
 
+  /**User-defined generator of this RDD*/
+  var generator = Utils.getCallSiteInfo._4
+  
+  /**reset generator*/
+  def setGenerator(_generator: String) = {
+    generator = _generator
+  }
+
   /**
    * Set this RDD's storage level to persist its values across operations after the first time
    * it is computed. This can only be used to assign a new storage level if the RDD does not
@@ -788,7 +796,7 @@ abstract class RDD[T: ClassManifest](
   private var storageLevel: StorageLevel = StorageLevel.NONE
 
   /** Record user function generating this RDD. */
-  private[spark] val origin = Utils.getSparkCallSite
+  private[spark] val origin = Utils.formatSparkCallSite
 
   private[spark] def elementClassManifest: ClassManifest[T] = classManifest[T]
 
--- a/core/src/main/scala/spark/Utils.scala
+++ b/core/src/main/scala/spark/Utils.scala
@@ -528,7 +528,7 @@ private object Utils extends Logging {
    * (outside the spark package) that called into Spark, as well as which Spark method they called.
    * This is used, for example, to tell users where in their code each RDD got created.
    */
-  def getSparkCallSite: String = {
+  def getCallSiteInfo = {
     val trace = Thread.currentThread.getStackTrace().filter( el =>
       (!el.getMethodName.contains("getStackTrace")))
 
@@ -540,6 +540,7 @@ private object Utils extends Logging {
     var firstUserFile = "<unknown>"
     var firstUserLine = 0
     var finished = false
+    var firstUserClass = "<unknown>"
 
     for (el <- trace) {
       if (!finished) {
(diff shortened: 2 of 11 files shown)


core/src/main/scala/spark/scheduler/SparkListener.scala
```diff
@@ -1,27 +1,54 @@
...
+   */
+  def onJobEnd(job: ActiveJob, event: SparkListenerEvents) { }
+  
 }
 
 sealed trait SparkListenerEvents
 
 case class StageCompleted(val stageInfo: StageInfo) extends SparkListenerEvents
 
+case object SparkListenerJobSuccess extends SparkListenerEvents
```
pwendell: Hey Mingfei,

I didn't notice this before. But here, you should have a single case class for each function in SparkListener. That case class should wrap any of the relevant information needed to pass onto the listener.

So there should be case classes:

StageCompleted
StageSubmitted
TaskEnd
JobStart
JobEnd

And each class should wrap the necessary information to pass to the Listener.

pwendell: We want to avoid directly passing DAGScheduler data structures to the Listener, in case they change.

core/src/main/scala/spark/scheduler/SparkListener.scala
```diff
@@ -1,27 +1,54 @@
...
   /**
    * called when a stage is completed, with information on the completed stage
    */
-  def onStageCompleted(stageCompleted: StageCompleted)
+  def onStageCompleted(stageCompleted: StageCompleted) { }
+  
+  /**
+   * called when a stage is submitted
+   */
+  def onStageSubmitted(stage: Stage, info: String = "") { }
```
pwendell: Don't use a string here to pass an argument (the number of tasks). Can't you just access stage.numPartitons? If not, pass the number of tasks as a numeric argument and then deal with formatting the string inside of the JobLogger.

Ideally you should have
case class StageSubmitted(stage: Stage)

...something like that.

core/src/main/scala/spark/Utils.scala
```diff
@@ -528,7 +528,7 @@ private object Utils extends Logging {
    * (outside the spark package) that called into Spark, as well as which Spark method they called.
    * This is used, for example, to tell users where in their code each RDD got created.
    */
-  def getSparkCallSite: String = {
+  def getCallSiteInfo = {
```
mateiz: For this, create a class called CallSiteInfo and specify the return type as being that class, like this:

In ClassSiteInfo.scala:

```
private[spark]
class ClassSiteInfo(val lastSparkMethod: String, ...)
```

And here:

```
def getCallSiteInfo: CallSiteInfo = {
```

We try to include return types in all of our method signatures so that we get a warning if we ever change the code to return another type.

core/src/main/scala/spark/RDD.scala
```diff
@@ -114,6 +114,14 @@ abstract class RDD[T: ClassManifest](
     this
   }
 
+  /**User-defined generator of this RDD*/
+  var generator = Utils.getCallSiteInfo._4
+  
+  /**reset generator*/
```
mateiz: For doc comments, always start the comment with uppercase, and make sure to include spaces after the `/**`. So like this:

```
/** Reset generator */
```

core/src/main/scala/spark/executor/TaskMetrics.scala
```diff
@@ -2,6 +2,11 @@ package spark.executor
 
 class TaskMetrics extends Serializable {
   /**
+   * host's name the task runs on
```
mateiz: Same here, make these comments start with an uppercase latter to be consistent with the rest of the codebase

core/src/main/scala/spark/scheduler/SparkListener.scala
```diff
@@ -1,27 +1,62 @@
...
+
+  /**
+   * called when a job starts
+   */
+  def onJobStart(jobStart: SparkListenerJobStart) { }
+  
+  /**
+   * called when a job ends
+   */
+  def onJobEnd(jobEnd: SparkListenerEvents) { }
```
mateiz: I don't like that this returns a SparkListenerEvents object. The user will have to do instanceOf on it, or try to match it with other types. Instead, change the events so that there's only one SparkListenerJobEnded event, and inside, it contains some kind of status, which is either SUCCESS, FAILED or CANCELLED, as well as optional fields for the reasons why it failed.

core/src/main/scala/spark/scheduler/SparkListener.scala
```diff
@@ -1,27 +1,62 @@
...
-  def onStageCompleted(stageCompleted: StageCompleted)
-}
 
 sealed trait SparkListenerEvents
 
+case class SparkListenerStageSubmitted(stage: Stage, taskSize: Int) extends SparkListenerEvents
+
 case class StageCompleted(val stageInfo: StageInfo) extends SparkListenerEvents
 
+case class SparkListenerTaskEnd(event: CompletionEvent) extends SparkListenerEvents
```
mateiz: We shouldn't be exposing the `CompletionEvent` class to people who write a `SparkListener`, because it's an internal class in the scheduler that might change over time. `SparkListener` is a somewhat public API (you can plug in your own) so it would be nice to make it as forward-compatible as possible. Instead of doing this, create a `SparkListenerTaskEnded` class and include the task ID, Stage, machine it ran on, etc (the things you need to log it).

mateiz:
Hi Shiming,

I've now taken a look at this too and made some comments. It looks really great, but there were two small things I was concerned with in the API (`CompletionEvent` and the job-ended callback), as well as some small style changes. Let me know if you have any questions about the comments, and thanks for putting this together!

pwendell:
Hey Siming,

Any idea when you'll be able to get around to Matei's comments? I have another PR that depends on the TaskListener changes, ideally this would go in before that.

shimingfei:
Hi Patrick,
Sorry for the delay, I will push the code today later.

shimingfei:
Hi Mateiz and Patrick,
I have modified the code according to your comments

Thanks
Mingfei

pwendell:
Jenkins, test this please.

mateiz:
Hi Shiming,

Thanks for the changes, but I'm still worried that the API to SparkListener includes some internal Spark classes (ActiveJob and JobResult). However, since there's a lot of utility in this patch and some other people want to depend on it, I'm going to merge it as is and mark the API as experimental. Before 0.8 we can hopefully replace those objects with some public objects (maybe inside a spark.listener package) that we're comfortable with maintaining. The problem is basically that if we need to change ActiveJob for the purpose of the Spark scheduler, we shouldn't be exposing that to users of this interface.

PR #645, opened 2013-06-09
Adding compression to Hadoop save functions

pwendell:
In many HDFS clusters some or all of the data is stored with compression. This patch adds a hook to easily pass a compression codec class when calling saveAsTextFile, saveAsSequencefile, or saveAsHadoopFile, avoiding the need to create a custom JobConf for every save.

There are a few different ways one could factor this, since there is so much overloading and chaining in tall the saveAsXX. I tried to chose a way which didn't  dramatically change what is there now.

Fixes SPARK-763.

### DIFF
--- a/core/src/main/scala/spark/PairRDDFunctions.scala
+++ b/core/src/main/scala/spark/PairRDDFunctions.scala
@@ -10,6 +10,8 @@ import scala.collection.JavaConversions._
 
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.Path
+import org.apache.hadoop.io.compress.CompressionCodec
+import org.apache.hadoop.io.SequenceFile.CompressionType
 import org.apache.hadoop.mapred.FileOutputCommitter
 import org.apache.hadoop.mapred.FileOutputFormat
 import org.apache.hadoop.mapred.HadoopWriter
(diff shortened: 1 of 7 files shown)


core/src/main/scala/spark/PairRDDFunctions.scala
```diff
@@ -583,11 +609,22 @@ class PairRDDFunctions[K: ClassManifest, V: ClassManifest](
...
       valueClass: Class[_],
       outputFormatClass: Class[_ <: OutputFormat[_, _]],
-      conf: JobConf = new JobConf(self.context.hadoopConfiguration)) {
+      conf: JobConf = new JobConf(self.context.hadoopConfiguration),
+      codec: Option[Class[_ <: CompressionCodec]] = None) {
     conf.setOutputKeyClass(keyClass)
     conf.setOutputValueClass(valueClass)
     // conf.setOutputFormat(outputFormatClass) // Doesn't work in Scala 2.9 due to what may be a generics bug
     conf.set("mapred.output.format.class", outputFormatClass.getName)
+    codec match {
```
ryanlecompte: Don't need to explicitly match on the option here, you can just:

``` scala
  codec.foreach { c =>
    conf.setCompressedMapOutput(true)
    conf.set("mapred.output.compress", true)
    conf.setMapOutputCompressorClass(c)
    conf.set("mapred.output.compression.codec", c.getCanonicalName)
    conf.set("mapred.output.compression.type", CompressionType.BLOCK.toString)
  }
```

ryanlecompte:
Nice!

mateiz:
Patrick, this is great, but can you also add it to the Java API?

core/src/main/scala/spark/PairRDDFunctions.scala
mateiz: Or another way is `for (c <- codec)`

mridulm:
This is pretty neat, thanks !
This helps us quite a lot :-)

core/src/main/scala/spark/PairRDDFunctions.scala
pwendell: Is this the canonical way to handle a binary option like this? I just don't like it beacuse it only saves 1 line of code, but it's much less explicit (i.e. it is using list semantics on something that really isn't a list).

rxin: It is a bit weird, but it is the standard Scala way in which you treat the Option as a monad. I don't like this either, but it is sort of the community "standard".

pwendell:
I added a Java-API version. In general this is less of a "win" for the Java-API, because you already have to pass several arguments to the Hadoop saveAsX functions.

I also cleaned up up the Option thing to use Scala style.

pwendell:
Jenkins, test this please

mateiz:
Thanks Patrick; this looks great.

pwendell:
I had to abort this test because it was running for 3 days.

PR #646, opened 2013-06-13
Fixed jvmArgs in maven build.

markhamstra:


mateiz:
Didn't realize this can break on some JVMs. Thanks.

PR #647, opened 2013-06-13
Reduce ZippedPartitionsRDD's getPreferredLocations complexity from O(2^2n) to O(2^n)

jerryshao:
ZippedPartitionsRDD's `getPreferredLocations()` complexity is now O(2^2n), which will become a efficiency problem when running iterative job or streaming job that uses previous calculated RDD, for example:

```
  var zipRdd = ...
  for (i  <-  1 to 100)
    val newRdd = ...
    zipRdd = zipRdd.zipPartitions(f, newRdd)
  }
```

This patch reduce complexity from O(2^2n) to originally O(2^n) while keep algorithm the same as before.

### DIFF
--- a/core/src/main/scala/spark/rdd/ZippedPartitionsRDD.scala
+++ b/core/src/main/scala/spark/rdd/ZippedPartitionsRDD.scala
@@ -53,14 +53,10 @@ abstract class ZippedPartitionsBaseRDD[V: ClassManifest](
     val exactMatchLocations = exactMatchPreferredLocations.reduce((x, y) => x.intersect(y))
 
     // Remove exact match and then do host local match.
-    val otherNodePreferredLocations = rddSplitZip.map(x => {
-      x._1.preferredLocations(x._2).map(hostPort => {
-        val host = Utils.parseHostPort(hostPort)._1
-
-        if (exactMatchLocations.contains(host)) null else host
-      }).filter(_ != null)
-    })
-    val otherNodeLocalLocations = otherNodePreferredLocations.reduce((x, y) => x.intersect(y))
+    val exactMatchHosts = exactMatchLocations.map(Utils.parseHostPort(_)._1)
+    val matchPreferredHosts = exactMatchPreferredLocations.map(locs => locs.map(Utils.parseHostPort(_)._1))
+      .reduce((x, y) => x.intersect(y))
+    val otherNodeLocalLocations = matchPreferredHosts.filter { s => !exactMatchHosts.contains(s) }
 
     otherNodeLocalLocations ++ exactMatchLocations
   }


jerryshao:
Test example is like this, it can reflect this issue though not so meaningful :

``` scala
package spark.streaming.examples

import java.util.Properties
import kafka.message.Message
import kafka.producer.SyncProducerConfig
import kafka.producer._
import spark.SparkContext
import spark.RDD
import spark.streaming._
import spark.streaming.StreamingContext._
import spark.streaming.util.RawTextHelper._

object KafkaWordCountTest {

    if (args.length < 5) {
      System.err.println("Usage: KafkaWordCount <master> <zkQuorum> <group> <topics> <numThreads>")
      System.exit(1)
    }

    val cached = new scala.collection.mutable.HashMap[String, RDD[(java.lang.String, Long)]]

    val Array(master, zkQuorum, group, topics, numThreads) = args

    val ssc =  new StreamingContext(master, "KafkaWordCount", Seconds(10),
      System.getenv("SPARK_HOME"), Seq(System.getenv("SPARK_EXAMPLES_JAR")))
    ssc.checkpoint("checkpoint")

    val topicpMap = topics.split(",").map((_,numThreads.toInt)).toMap
    val lines = ssc.kafkaStream[String](zkQuorum, group, topicpMap)
    val words = lines.window(Seconds(30), Seconds(10)).flatMap(_.split(" "))
    val wordCounts = words.countByValue()
    wordCounts.foreach(rdd => {
      cached.get("test") match {
        case None =>
          rdd.foreach(_ => Unit)
          cached("test") = rdd
        case Some(r) => 
          val newRdd = r.zipPartitions(
              (i1: Iterator[(java.lang.String, Long)], i2: Iterator[(java.lang.String, Long)]) => {
            i2
          }, rdd)
          newRdd.foreach(_ => Unit)
          cached("test") = newRdd
      }
    })

    ssc.start()
  }
}
```

core/src/main/scala/spark/rdd/ZippedPartitionsRDD.scala
```diff
@@ -53,14 +53,11 @@ abstract class ZippedPartitionsBaseRDD[V: ClassManifest](
...
-        val host = Utils.parseHostPort(hostPort)._1
-
-        if (exactMatchLocations.contains(host)) null else host
-      }).filter(_ != null)
-    })
-    val otherNodeLocalLocations = otherNodePreferredLocations.reduce((x, y) => x.intersect(y))
+    val exactMatchHosts = exactMatchLocations.map(Utils.parseHostPort(_)._1)
+    val matchPreferredHosts = exactMatchPreferredLocations.map(locs => locs.map(Utils.parseHostPort(_)._1))
+      .reduce((x, y) => x.intersect(y))
+    val otherNodeLocalLocations = matchPreferredHosts.map(s => if (exactMatchHosts.contains(s)) null else s)
```
ryanlecompte:
``` scala
  val otherNodeLocations = matchPreferredHosts.filter { s => exactMatchHosts.contains(s) }
```

mateiz:
Jenkins, this is ok to test

mateiz:
Thanks for the fix. I like Jerry's suggestion to simplify that -- please just add another commit with that to your branch.

core/src/main/scala/spark/rdd/ZippedPartitionsRDD.scala
```diff
@@ -53,14 +53,11 @@ abstract class ZippedPartitionsBaseRDD[V: ClassManifest](
...
     // Remove exact match and then do host local match.
-    val otherNodePreferredLocations = rddSplitZip.map(x => {
-      x._1.preferredLocations(x._2).map(hostPort => {
-        val host = Utils.parseHostPort(hostPort)._1
-
-        if (exactMatchLocations.contains(host)) null else host
-      }).filter(_ != null)
-    })
-    val otherNodeLocalLocations = otherNodePreferredLocations.reduce((x, y) => x.intersect(y))
+    val exactMatchHosts = exactMatchLocations.map(Utils.parseHostPort(_)._1)
```
mridulm, line 56: Make this a Set - so the subsequent comparisons against it are O(1).
Ofcourse, considering the really small size of all of these datastructures, the optimizations might be moot.

The recomputation of preferredLocations is obviously something which could have been avoided in first place.

mridulm:
I am not sure of the 2^2n vs 2^n number :-)
It would be closer to 2*n^2 vs n^2 ?

While I should obviously have avoided the braindead recomputation of preferredLocations, in general though, how severe are the performance issues ?
If we are finding methods like this as a performance bottleneck (computation and memory pressure), we would need to relook at a bunch of idioms we use for simplicity. Like for, zip, etc.

jerryshao:
@mridulm  I think the complexity is O(n^2) as you said in normal condition, which means zipPartitionsRDD are zipped by two calculated RDD, like this:

``` scala
  val rdd1 = ...
  val rdd2 = ...
  val zipRdd = rdd1.zipPartitions(f, rdd2)
```

`rdd1` and `rdd2` are got from input data or transformed from others. Here `rdd1` and `rdd2`'s dependency chain is constant, so no matter how many steps of transformation, the complexity of getPreferredLocations can be expected.

But in my test, I have a scenario: RDD is iteratively updated, which means previous zip-partitioned RDD will again be zip-partitioned with new RDD in next iteration, like this:

``` scala
  var zipRdd = ...
  for (i  <-  1 to 100)
    val newRdd = ...
    zipRdd = zipRdd.zipPartitions(f, newRdd)
  }
```

So the dependency chain of this `zipRdd` is accumulated by `newRdd` in each iteration. And  getPreferredLocations of this `zipRdd` will walk through the whole dependency tree like this:

```
               zipRdd
                 | |
               |     |
          zipRdd  newRdd
             | |
           |     |
      zipRdd  newRdd
         | |
       |     |
   zipRdd  newRdd
      | |
    |     |
 newRdd newRdd
```

Here `newRdd` is newly calculated RDD in each iteration, which has a constant dependency chain. Tree depth is increased each iteration.

I think previous code of getPreferredLocations seems more or less like recursive calculating of Fibonacci number, I don't know if it is correct :-), I guess there exists many useless calculation.

Also I tested using above `KafkaWordCountTest` example, result shows below

comparison

In this example, I use streaming job to instead iterative job. In each batch duration, received data is transformed to a word count DStream, this word count DStream is zip-partitioned with cached data per RDD.

As you can see each iteration's running time comparison between patch with or without, most of the time spend on getPreferredLocations according to my investigation.

Maybe my calculation of complexity is not correct, but I'm sure it's not O(n^2) in my scenario :). Normally use of Spark will not expose bottleneck, but in my application I have to get a new DStream and zip-partitioned with existed data per RDD, so this would be the problem :-)

Thanks
Jerry Shao

jerryshao:
Hi, I've updated this patch, please review. Quite strange all the tests are passed on my computer but failed on Amplab Jenkins...

Thanks
Jerry Shao

mridulm:
The complexity of the method does not change with the fix - though for practical purposes : it has an impact due to reusing results of preferredLocations : should at best improve by twice (if every other cost being zero) or less : but definitely an improvement - and so, this is indeed a welcome change !

The increasing time you observe is due to how rdd's are getting used within the user code : this fix will slightly alleviate the issue, but wont be a solution for it unfortunately - as number of slices or partitions increase, you will continue to see similar behaviour.

My hope is that we can improve this method to something better than n^2 - unfortunately, we did not come up with something when I submitted the patch initially ...

mridulm:
Btw, I am not very sure of that graph - the only different between both pieces of code is one additional invocation of preferredLocations from what I see.
I hope I am not misinterpreting the patch !

core/src/main/scala/spark/rdd/ZippedPartitionsRDD.scala
```diff
@@ -53,14 +53,10 @@ abstract class ZippedPartitionsBaseRDD[V: ClassManifest](
...
+    val matchPreferredHosts = exactMatchPreferredLocations.map(locs => locs.map(Utils.parseHostPort(_)._1))
+      .reduce((x, y) => x.intersect(y))
+    val otherNodeLocalLocations = matchPreferredHosts.filter { s => exactMatchHosts.contains(s) }
```
mridulm: The condition should be "! exactMatchHosts.contains(s) "

jerryshao:
Sorry for the miss, I've updated this, thanks @mridulm . I agree with you that this patch can only alleviate this issue but won't solve it :-)

mateiz:
I think the change makes sense, because it turns a potential multiple re-visiting of each parent RDD into a single one. That could definitely take an exponential cost into a linear one. It may be possible to do this even better by adding memoization to the method (i.e. caching the preferred locations in each RDD), but for now I'll merge this fix.

PR #649, opened 2013-06-13
Add top K method to RDD using a bounded priority queue

ryanlecompte:
This pull request contains an implementation for a new top K operation for an RDD[T] based on a specific Ordering[T]. This operation should be a more performant alternative to the usual first `sortByKey` and then `take(k)` for large data sets to avoid shuffling. The implementation currently relies on `mapPartitions`. I am not sure if there is a better approach, so it would be great to have other people look at it. Note that I am using Java's PriorityQueue because Scala's PriorityQueue currently has serialization issues (a bug has been filed already on Scala's version).

### DIFF
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -35,6 +35,7 @@ import spark.rdd.ZippedPartitionsRDD2
 import spark.rdd.ZippedPartitionsRDD3
 import spark.rdd.ZippedPartitionsRDD4
 import spark.storage.StorageLevel
+import spark.util.BoundedPriorityQueue
 
 import SparkContext._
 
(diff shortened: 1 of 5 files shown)


mateiz:
Jenkins, this is ok to test

core/src/main/scala/spark/RDD.scala
```diff
@@ -723,6 +724,29 @@ abstract class RDD[T: ClassManifest](
...
+      Iterator.single(queue)
+    }.reduce { (queue1, queue2) =>
+      queue1 ++= queue2
+      queue1
+    }
+
+    val builder = Array.newBuilder[T]
+    builder.sizeHint(topK.size)
+    builder ++= topK
+    builder.result()
```
mateiz: I'd add a `toArray` method in `BoundedPriorityQueue` instead of this, since it might be used elsewhere.

mateiz:
Hey Ryan, this looks good, but can you also add it to the Java API? You'll need a version that takes a Comparator and one that tries to use `com.google.common.collect.Ordering.natural()`, as we do in `JavaPairRDD.sortByKey`.

Also, if you'd like to use Guava instead of the Java collection library, that's fine. I'm not sure whether Guava has some better priority queue implementations but it might be possible.

ryanlecompte:
Hey @mateiz, I just made a commit that should address your comments. I refactored the `BoundedPriorityQueue` so that it properly uses delegation and can therefore act as a true Scala collection. This allows us to remove the custom `toArray` method that I previously had. I also added a corresponding Java API. I also looked into using Guava's `MinMaxPriorityQueue`. It looked promising, however it's unfortunately not `Serializable` out of the box. We'd have to come up with some custom serialization for it if we wanted to use. I'm not sure it's worth the effort at this point.

mateiz:
Looks good, thanks!

PR #650, opened 2013-06-14
Use a consistent port range for the BlockManager UI

pwendell:
This is a short-term fix which launches the BlockManager UI
at a consistent port. If there is contention, the port is incremented
until a free port is found.

This will eventually be subsumed by the upcoming web UI changes.


pwendell:
Jenkins, test this please.

core/src/main/scala/spark/Utils.scala
```diff
@@ -562,6 +562,32 @@ private object Utils extends Logging {
   }
 
   /**
+   * Test whether a specific port is in use.
+   */
+  def portIsFree(port: Int): Boolean = {
```
ryanlecompte: Just FYI, a bit more Scala-ish way to do this would be:

``` scala
  import scala.util.Try
  def isPortFree(port: Int): Boolean = Try { val ss = new ServerSocket(port); ss.close() }.isSuccess
```

core/src/main/scala/spark/Utils.scala
```diff
@@ -562,6 +562,32 @@ private object Utils extends Logging {
...
+      case e: IOException => free = false
+    }
+    free
+  }
+
+  /**
+   * Find a free port by first checking startingPort and then incrementing until a free
+   * port is found.
+   */
+  def findFreePort(startingPort: Int): Int = {
```
ryanlecompte: You can add a `@tailrec` annotation here since this method is tail-recursive. So it would be:

``` scala
  import scala.annotation.tailrec

  @tailrec
  def findFreePort(startingPort: Int): Int = { ... }
```

core/src/test/scala/spark/DistributedSuite.scala
```diff
@@ -51,6 +51,20 @@ class DistributedSuite extends FunSuite with ShouldMatchers with BeforeAndAfter
     resetSparkContext()
   }
 
+  /*
```
rxin: Why are they commented out?

pwendell: whops, let me update. The Spark test suite is somehow broken on my machine... was hoping Jenkins would test it.

core/src/main/scala/spark/Utils.scala
pwendell: Is the purpose of doing that for readability or to help the compiler optimize? If it's for the latter case, seems like it's not super necessary here since this is called a negligible amount of times.

ryanlecompte: The compiler will still optimize it to be a tail-recursive method, just adding the `@tailrec` bit will cause the compiler to generate an error if it's ever changed to be non-tail-recursive in the future. It's more of a future-proofing thing.

pwendell:
Jenkins, test this please.

pwendell:
Reynold pointed out a race condition which was already in the code but became much more likely with this commit. I've re-written the code to avoid this race condition (unfortunately also introducing a larger re-factoring I was hoping to avoid).

The issue is that we need to atomically decide if the port is free and try to grab the port. This does that using an optimistic approach where you just try to bind and fallback on failure.

Jenkins, test this please.

andyk:
Jenkins, retest this please

core/src/main/scala/spark/util/AkkaUtils.scala
```diff
@@ -71,21 +73,37 @@ private[spark] object AkkaUtils {
...
+      val timeout = 2.seconds
+      val future = server.ask(HttpServer.Bind(ip, bindPort))(timeout)
+      try {
+        Await.result(future, timeout) match {
+          case bound: HttpServer.Bound =>
+            connected = true
+            result = Some(server)
+          case other: Any =>
+            actorSystem.stop(server)
+            logWarning("Failed to bind web UI to port " + bindPort + ": " + other)
```
rxin: For the two warning messages here, since they are "normal", can we change the level from warning to info, and in the log message, add a bit explanation to calm the user down. We don't want this to freak people out because it is expected behavior. e.g. saying "will retry with port = ..."

core/src/main/scala/spark/util/AkkaUtils.scala
```diff
@@ -71,21 +73,37 @@ private[spark] object AkkaUtils {
...
+        * naming conflicts, we have to assign unique names for each attempt. */
+      val actorName = "%s_%s".format(name, attempt)
+      attempt = attempt + 1
+      val server = actorSystem.actorOf(
+        Props(new HttpServer(ioWorker, SingletonHandler(rootService))), name = actorName)
+      actorSystem.registerOnTermination { ioWorker.stop() }
+      val timeout = 2.seconds
+      val future = server.ask(HttpServer.Bind(ip, bindPort))(timeout)
+      try {
+        Await.result(future, timeout) match {
```
rxin: And there might be a more scala idiomatic way to handle the 3 code paths here. @ryanlecompte ?

ryanlecompte: Hey! Here's a quick take on a more idiomatic way for this. We basically get rid of the var's and use a `tailrec` method to walk through the ports and keep track of the attempts. We also collapse the 3 special cases down to 1:

``` scala
  def startSprayServer(
    actorSystem: ActorSystem,
    ip: String,
    port: Int,
    route: Route,
    name: String = "HttpServer"): ActorRef = {
    val ioWorker = new IoWorker(actorSystem).start()
    val httpService = actorSystem.actorOf(Props(new HttpService(route)))
    val rootService = actorSystem.actorOf(Props(new SprayCanRootService(httpService)))

    @tailrec
    def connect(attempt: Int = 0, currentPort: Int = port): ActorRef = {
      // We may try to bind the actor multiple times if there is port
      // contention. To avoid naming conflicts, we have to assign
      // unique names for each attempt.
      val actorName = "%s_%s".format(name, attempt)
      val server = actorSystem.actorOf(
        Props(new HttpServer(ioWorker, SingletonHandler(rootService))), name = actorName)
      actorSystem.registerOnTermination { ioWorker.stop() }
      val timeout = 2.seconds
      val future = server.ask(HttpServer.Bind(ip, currentPort))(timeout)
      Try(Await.result(future, timeout)) match {
        case Success(_: HttpServer.Bound) => server
        case Failure(other) =>
          actorSystem.stop(server)
          logWarning("Failed to bind web UI to port %d: %s".format(currentPort, other))
          connect(attempt + 1, (currentPort + 1) % 65535)
      }
    }

    val boundActor = connect()
    logInfo("Created %s webUI at port %s".format(name, boundActor))
    boundActor
  }
```

pwendell:
Hey @ryanlecompte,

Thanks for the advice! Could you take a look at what's there now? I couldn't exactly implement what you wrote because there is another failure case that isn't handled by your code. The case is when no exception is thrown by the Bind call returns a status other than HttpServer.Bound. Is there a better way to handle this than what's there now?
- Patrick

ryanlecompte:
That looks a lot cleaner, @pwendell. Nice job on catching that other failure condition. One suggestion would be try to add the `@tailrec` annotation (just import scala.annotation.tailrec) to your `connect` method. We want to ensure that the compiler can optimize this out to a tail-recursive method so that we know that we don't blow the stack for those cases where we end up cycling through many ports. Nice work.

pwendell:
Hey @ryanlecompte,

Because of the slight re-factoring I had to do, I don't think it's technically tail recursive. At least, my compiler throws an error if I add the annotation. The issue is that two different case statements now create the recursive call. I couldn't figure out how to coalesce these into a single case statement, since they have different types.
- Patrick

ryanlecompte:
@pwendell - I took a look at it again and you should be able to use the original code that I wrote with a slight modification here:

``` scala
      Try(Await.result(future, timeout)) match {
        case Success(_: HttpServer.Bound) => server
        case other @ Success(_) | Failure(_) =>
          actorSystem.stop(server)
          logWarning("Failed to bind web UI to port %d: %s".format(currentPort, other))
          connect(attempt + 1, (currentPort + 1) % 65535)
      }
```

Pattern matching allows us to collapse the two failure cases down to a single handler. This will also let us keep it tail-recursive. WDYT?

ryanlecompte:
You can actually just replace `case other @ Success(_) | Failure(_)` with `case other =>` because at this point we treat all other cases as a failure. Even simpler. :-)

pwendell:
Ah I see. Ya I just wasn't sure whether Failure.toString prints the internal error message. But it looks like it does, so this should work.

pwendell:
Jenkins, test this please.

pwendell:
Jenkins, test this please. (trying again, there weren't actually any failures in the previous test, so a bit confused).

andyk:
Jenkins, retest this please

andyk:
Jenkins, ok to test

mateiz:
Patrick, it seems that the new test might have failed, though maybe the test runner crashed for another reason. Did you try it by hand?

pwendell:
We're trying to fix some bugs in Jenkins right now, so not clear whether it's the environment or the test. In any case, hold off on committing this until we have something more stable.

mateiz:
Alright. FYI, it does seem to pass on my machine.

pwendell:
It also passes on my machine which makes this jenkins thing hard to debug - the error from jenkins is pretty opaque. If you also have it passing maybe we should just merge it. Up to you...

mateiz:
Jenkins, retest this please

mateiz:
We shouldn't commit it if it breaks on Jenkins; we need Jenkins to be consistent. But let's see whether it always happens.. there's definitely been some weird stuff there lately.

andyk:
Jenkins, retest this please. I'm issuing this command (again) to double-check the AMPLab Jenkins Github Pull Request Builder plugin is functioning as we expect.

pwendell:
Hm, okay I think this test is sometimes failing if another test fails to shutdown the server (and free up the port) quickly enough. This wasn't a problem before because we used random ports, but now that we use the same port there could be contention at small time scales within the test suite. Let me dig into this more...

mateiz:
That also means that this might not be a great test. Maybe it would be better if you made it search forward from another port.

pwendell:
Jenkins, test this please.

pwendell:
Jenkins, retest this please.

pwendell:
This is pending stabilizing the jenkins build... the last run had a test fail which was unrelated (in the Driver suite).

andyk:
Jenkins, retest this please.

pwendell:
Jenkins, retest this please.

pwendell:
Jenkins, test this please.

pwendell:
Okay it seems like the issue is actually a bit more complicated than just port contention. For some reason the exception that is thrown within Akka when this fails to bind causes subsequent bind requests to timeout. When the worker Web UI starts (in the same JVM here due to the unit tests) it fails sometimes with an exception. Understanding what is exactly happening will require digging around in Akka and Spray.

I might try just to move the test to not use the distributed suite for this reason. In a real cluster the worker UI and BlockManager UI don't live in the same JVM.

Also, this entire codepath is getting replaced in a forthcoming PR (probably ready in the next 48 hours), so it might make sense to just ignore it until then. One thing that PR does is replace the Akka/Spray code with much simpler Jetty code in the Web UI's, which makes debugging issues like this much less painful.

PR #652, opened 2013-06-14
Fixed maven build without netty fix

ScrapCodes:
It was a quick fix though. Excluding netty everywhere in poms will make them look uglier. 

Maybe there exists a smarter solution to that.


mateiz:
Thanks for adding this. We might consider just dropping the Maven build in Spark 0.8, and using SBT's `make-pom` command to generate a POM. We'll have to see how well that works.

ScrapCodes:
Sounds cool! Its difficult working with maven for scala atleast now anyway.

PR #653, opened 2013-06-17
SPARK-781: Log the temp directory path when Spark says "Failed to create temp directory."

rxin:


core/src/main/scala/spark/Utils.scala
```diff
@@ -116,8 +116,8 @@ private object Utils extends Logging {
     while (dir == null) {
       attempts += 1
       if (attempts > maxAttempts) {
-        throw new IOException("Failed to create a temp directory after " + maxAttempts +
-            " attempts!")
+        throw new IOException("Failed to create a temp directory under (" + root + ") after " +
```
mateiz: Shouldn't this say (under DIR) instead of under (DIR)?

rxin:
Good find - PTAL.

mateiz:
It looks good, but let's wait for Jenkins to test it. Jenkins, this is ok to test.

mateiz:
Looks good, thanks.

PR #655, opened 2013-06-19
Add support for running Spark on Yarn on a secure Hadoop Cluster

tgravescs:
I tested this on both Hadoop 0.23 and Hadoop 2.0.2 on both secure and unsecure clusters.  This allows Spark on Yarn to get all the necessary tokens to start containers and run on a secure yarn cluster. It also forwards on the HDFS tokens to allow the containers to access HDFS files.  It currently only supports accessing HDFS file on the same cluster as being run.


mateiz:
Jenkins, this is ok to test

mateiz:
Hey Thomas, so from the unit test output above, it seems like the changes in PairRDDFunctions don't work on the default version of Hadoop we link with (1.0.4). I'm okay linking to a later version of Hadoop 1 by default, but it would be nice to support it in addition to Hadoop 2. If it's not possible to get that API in a later version, can you factor that code into a conditionally compiled class (similar to our current hadoop1 vs hadoop2 vs hadoop2-yarn setup)?

tgravescs:
Thanks for the review.  It doesn't looks like that was ever pulled back into the Hadoop 1 api.   I just discovered that it doesn't build with the current version of cdh 4.1.2 that is default right now either. I was doing my testing with the apache Hadoop 2.0.2-alpha release rather then cdh so I missed that.  The api is present in the 4.2.0 CDH version of hadoop 2 though. Is it ok to change the version for the hadoop2 profile to use CDH4.2.0 or would you rather keep that at 4.1.2?   I'll look at conditionally compiling it for hadoop1.

mateiz:
If we'll do conditional compiles, it might be good to make it a no-op in hadoop2 without yarn too, in case people are still using older versions. I guess eventually we'll merge the hadoop2-yarn and nonyarn packages.

core/src/hadoop2-yarn/scala/spark/deploy/SparkHadoopUtil.scala
```diff
@@ -27,18 +27,7 @@ object SparkHadoopUtil {
...
-    val appMasterUgi: UserGroupInformation = UserGroupInformation.createRemoteUser(user)
-    appMasterUgi.doAs(new PrivilegedExceptionAction[AnyRef] {
-      def run: AnyRef = {
-        func(args)
-        // no return value ...
-        null
-      }
-    })
+    func(args)
   }
```
mridulm, line 32: Any reason for removing this ?
This is to ensure that the job gets run as the user via ugi.

core/src/hadoop2-yarn/scala/spark/deploy/yarn/ApplicationMaster.scala
```diff
@@ -30,23 +30,6 @@ class ApplicationMaster(args: ApplicationMasterArguments, conf: Configuration) e
...
-    appMasterUgi.doAs(new PrivilegedExceptionAction[AnyRef] {
-      def run: AnyRef = {
-        runImpl()
-        return null
-      }
-    })
-  }
-
-  private def runImpl() {
-
```
mridulm, line 49: Same as above ... is this handled elsewhere ?
This is to ensure that the job gets run as the user via ugi.

Note that there are two different codepaths which this happens through - one for master and one for workers

core/src/hadoop2-yarn/scala/spark/deploy/yarn/Client.scala
```diff
@@ -62,14 +67,21 @@ class Client(conf: Configuration, args: ClientArguments) extends YarnClientImpl
...
+    // if we have requested more then the clusters max for a single resource then exit.
+    if (args.workerMemory > maxMem) {
+      logError("the worker size is to large to run on this cluster " + args.workerMemory);
+      System.exit(1)
+    }
+    val amMem = args.amMemory + YarnAllocationHandler.MEMORY_OVERHEAD
+    if (amMem > maxMem) {
+      logError("AM size is to large to run on this cluster "  + amMem)
       System.exit(1)
     }
```
mridulm, line 81: We are trying to see if the cluster has sufficient memory to satisfy the memory requirements of the spark job.
Just to clarify - app.getMaximumResourceCapability().getMemory()  does not give maximum resource capability of the cluster ?

I thought that was the value which was returned when we had tested it quite a while back ... I hope it was not incorrect (or some bug in earlier version we had hit upon !).

core/src/hadoop2-yarn/scala/spark/deploy/yarn/WorkerRunnable.scala
```diff
@@ -76,7 +79,13 @@ class WorkerRunnable(container: Container, conf: Configuration, masterAddress: S
...
 
     ctx.setUser(UserGroupInformation.getCurrentUser().getShortUserName())
-    val commands = List[String]("java " +
+
+    val credentials = UserGroupInformation.getCurrentUser().getCredentials()
+    val dob = new DataOutputBuffer()
+    credentials.writeTokenStorageToStream(dob)
+    ctx.setContainerTokens(ByteBuffer.wrap(dob.getData()))
+
+    val commands = List[String](Environment.JAVA_HOME.$() + "/bin/java " +
```
mridulm: We should not be hardcoding the path to java (even to java_home) - but rely on system to dereference it appropriately.
That seems to be the general idiom in spark, and also seems to work fine in some of our test deployment ..

core/src/main/scala/spark/PairRDDFunctions.scala
```diff
@@ -613,6 +614,10 @@ class PairRDDFunctions[K: ClassManifest, V: ClassManifest](
       outputFormatClass: Class[_ <: OutputFormat[_, _]],
       conf: JobConf = new JobConf(self.context.hadoopConfiguration),
       codec: Option[Class[_ <: CompressionCodec]] = None) {
+    // make sure to propogate any credentials from the current user to the jobConf
+    // for Hadoop security
+    val jobCreds = conf.getCredentials();
+    jobCreds.mergeAll(UserGroupInformation.getCurrentUser().getCredentials())
     conf.setOutputKeyClass(keyClass)
```
mridulm, line 617: Do we need to do this for each rdd that is created ? Or is it done once ?
If former, then maybe move it to RDD.scala ? If latter, then move it to SparkContext ?
If just for hadoop case, we could modify only relevant methods in SparkContext .. (which you have already added to imo ?)

Also, is it the same for yarn and non-yarn codepaths ?
What about non hadoop deployment (mesos, standalone) ?

If just for yarn, we can add a method to SparkHadoopUtil - and no-op it for other maven profiles.

examples/pom.xml
```diff
@@ -193,6 +193,11 @@
           <artifactId>hadoop-yarn-common</artifactId>
           <scope>provided</scope>
         </dependency>
+        <dependency>
+          <groupId>org.apache.hbase</groupId>
+          <artifactId>hbase</artifactId>
+          <version>0.94.6</version>
+        </dependency>
       </dependencies>
```
mridulm, line 201: I am not sure if I missed it, but are we depending on hbase for this patch ?
Or was it part of testing something else which crept in ?

streaming/src/main/scala/spark/streaming/PairDStreamFunctions.scala
```diff
@@ -470,7 +471,11 @@ extends Serializable {
       valueClass: Class[_],
       outputFormatClass: Class[_ <: OutputFormat[_, _]],
       conf: JobConf = new JobConf
-    ) {
+    ) {  
+    // make sure to propogate any credentials from the current user to the jobConf    
+    // for Hadoop security
+    val jobCreds = conf.getCredentials();
+    jobCreds.mergeAll(UserGroupInformation.getCurrentUser().getCredentials())
     val saveFunc = (rdd: RDD[(K, V)], time: Time) => {
```
mridulm, line 475: Same comment as for PairRDDFunctions.scala ...

mridulm:
This is great Tom !
I have added a few comments where I was not clear about some aspects of it.

Since spark is used in different modes, wondering how this interacts with it.
Namely :

a) yarn mode with and without security.
b) hadoop v1 or v2 dfs + standalone cluster.
c) hadoop v1 or v2 dfs + mesos cluster.
d) local mode, etc.

Obviously we cant test all modes I guess :-) But wondering how the code interacts (dependencies, runtime behavior) when not used in the tested codepaths.

core/src/hadoop2-yarn/scala/spark/deploy/SparkHadoopUtil.scala
tgravescs, line 32: This is handled for you by hadoop simply by setting this:
-    // set this so that UGI set to correct user in unsecure mode
-    Apps.addToEnvironment(env, "HADOOP_USER_NAME", Utils.getUserNameFromEnvironment())

core/src/hadoop2-yarn/scala/spark/deploy/yarn/ApplicationMaster.scala
tgravescs, line 49: its unneeded as its handled by:
-    // set this so that UGI set to correct user in unsecure mode
-    Apps.addToEnvironment(env, "HADOOP_USER_NAME", args.amUser)

core/src/hadoop2-yarn/scala/spark/deploy/yarn/Client.scala
tgravescs, line 81: The app.getMaximumResourceCapability doesn't give you the resources of the entire cluster it gives you the maximum for a single resource (a resource is basically one container) request.  It returns whatever is set in the config yarn.scheduler.maximum-allocation-mb    (see http://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/ClusterSetup.html).  Perhaps you had a very large value for this.  You'll see there is also a minimum which corresponds to yarn.scheduler.minimum-allocation-mb.

The only way I saw to get the resource of the entire cluster would be to iterate through the entire nodes list, which could get expensive on a large cluster so I left it off for now. I'll see if I can come up with better way but figured it could be done separately.

core/src/hadoop2-yarn/scala/spark/deploy/yarn/WorkerRunnable.scala
tgravescs: hmm,this is how mapreduce application master handle it (see YARNRunner.java).  In the performance testing environment I was testing on, it didn't have it in the path on all the nodes. Ops could choose to install java in a different location from the normal system install and use that to run Hadoop.   I'll have to check our other environments and production. Having this (using JAVA_HOME) there is actually also beneficial for other reasons. For instance you could have multiple java installs and the user could chose which one to use if ops allowed it. For instance one 32 bit one 64 bit. I'll investigate some more.

core/src/main/scala/spark/PairRDDFunctions.scala
tgravescs, line 617: It needs to be included whenever a JobConf that is going to be passed to something that would need the tokens (ie tries to access HDFS from a hadoop container - like the AM or workers). In some of the cases it creates a new JobConf right there, in other cases like the hadoopRDD ithe jobConf is passed in by the user so we just need to make sure its added. I'll see if I can move to a common place. 

 It doesn't hurt being there for hadoop1 or hadoop2 and the deployment shouldn't matter, generally in those cases the credentials will just be empty, but since some versions don't have the api I'll conditionalize that out and I'll see if I can make this cleaner.

examples/pom.xml
tgravescs, line 201: I needed to add this when I upmerged to the lastest master version. Someone added in a sample HBASE program. The dependencies were added to the hadoop1 and hadoop2 profiles but not the hadoop2-yarn profile, so when I upmerged in order for it to compile I had to add this.

tgravescs:
Thanks for reviewing Mridul.  I did not test on mesos and I did not test hadoop v1 as I don't have a v1 or mesos install.  All the hadoop v2 testing was done on the apache hadoop 2.0.2-alpha, not on CDH which is why I missed the api issue.  I'll make sure to compile on all the versions in the future and since hadoop v1 doesn't have the api I used they will be conditionalized.

examples/pom.xml
tgravescs, line 201: I can split this into separate pull request if that is better?

core/src/hadoop2-yarn/scala/spark/deploy/SparkHadoopUtil.scala
mridulm, line 32: So this is applicable for both master and slaves ? That is real handy !

core/src/hadoop2-yarn/scala/spark/deploy/yarn/Client.scala
mridulm, line 81: Ok, this makes sense.
Looks like we misinterpreted the api and the config when we had configured it initially.

core/src/main/scala/spark/PairRDDFunctions.scala
mridulm, line 617: SparkHadoopUtil.newConfiguration (in each of the profiles) is used to create a new configuration : unless I missed some place.
You could add this there ?

If that is not sufficient, you could add a method to SparkHadoopUtil to create job conf and add this to it - and ensure that all attempts to create job conf go to that method.

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -294,6 +295,10 @@ class SparkContext(
       valueClass: Class[V],
       minSplits: Int = defaultMinSplits
       ): RDD[(K, V)] = {
+    // make sure to propogate any credentials from the current user to the jobConf 
+    // for Hadoop security
+    val jobCreds = conf.getCredentials();
+    jobCreds.mergeAll(UserGroupInformation.getCurrentUser().getCredentials())
     new HadoopRDD(this, conf, inputFormatClass, keyClass, valueClass, minSplits)
```
mridulm, line 298: We could encapsulate this into a method in SparkHadoopUtil ? And use that everywhere (and possibly extend that method for future uses).

examples/pom.xml
mridulm, line 201: I dont think that is required - I was just curious what the hbase dependency was !
Sounds logical.

core/src/hadoop2-yarn/scala/spark/deploy/SparkHadoopUtil.scala
tgravescs, line 32: Its for both the application master and then for any worker containers launched and it works for both.  Based on what we talked about I will just remove these and the --user option.  The job will run as the user that launched it.  There shouldn't be an issues with oozie launching the job on either secure or insecure cluster either as oozie uses the hadoop proxy user setup.   I'll try to get an oozie setup and try to test that too.

core/src/main/scala/spark/PairRDDFunctions.scala
tgravescs, line 617: Unfortunately the SparkHadoopUtil.newConfiguration gives you a YarnConfiguration not a jobConf. The jobConf is used for mapreduce specific things like FileInputFormat where as the YarnConfiguration is generic for Yarn. Both are based on Configuration though.    I'll look at exactly how that is used to see if it really needs to be YarnConfiguration. 

otherwise as you mentioned I was looking at added a function to SparkHadoopUtil to add in the credentials only for the hadoop2-yarn profile.

mridulm, line 617: We could just add another method to handle creation of job conf specifically ...
But we will need to ensure that all creation of job conf within spark goes through it.

mateiz:
Hey Thomas, just wondering, are you working on letting this compile with Hadoop1 as well? You don't need to install Hadoop1, just run SBT with that setting in the build file.

tgravescs:
Yes I plan on making sure it builds on hadoop1, hadoop2,and hadoop2-yarn.  Sorry for the delay I'm at the hadoop summit conference this week.

tgravescs:
I built this patch with all 3 profiles hadoop1,hadoop2, hadoop2-yarn.  I tested  on both secure and insecure hadoop 2.0.2 and 0.23 clusters.  I removed the --user option as the intent was to make sure the spark on yarn works with oozie.  Oozie uses the hadoop proxy user functionality in hadoop so we don't need the --user option.

mridulm:
Great, Thanks Tom !
Sounds good to me.

Regards
Mridul

mateiz:
Thanks Thomas and Mridul. I've merged this in.

mridulm:
Thanks Tom - this is great !

PR #657, opened 2013-06-21
Twitter Authorizations other than username/password

mrpotes:
Twitter supports other authorization methods than plain-text username/password (OAuth) - these should also be supported.

### DIFF
--- a/streaming/src/main/scala/spark/streaming/dstream/TwitterInputDStream.scala
+++ b/streaming/src/main/scala/spark/streaming/dstream/TwitterInputDStream.scala
@@ -6,6 +6,7 @@ import storage.StorageLevel
 
 import twitter4j._
 import twitter4j.auth.BasicAuthorization
+import twitter4j.auth.Authorization
 
 /* A stream of Twitter statuses, potentially filtered by one or more keywords.
 *
@@ -16,21 +17,19 @@ import twitter4j.auth.BasicAuthorization
 private[streaming]
 class TwitterInputDStream(
     @transient ssc_ : StreamingContext,
-    username: String,
-    password: String,
+    twitterAuth: Authorization,
     filters: Seq[String],
     storageLevel: StorageLevel
   ) extends NetworkInputDStream[Status](ssc_)  {
-
+  
   override def getReceiver(): NetworkReceiver[Status] = {
-    new TwitterReceiver(username, password, filters, storageLevel)
+    new TwitterReceiver(twitterAuth, filters, storageLevel)
   }
 }
 
 private[streaming]
 class TwitterReceiver(
-    username: String,
-    password: String,
+    twitterAuth: Authorization,
     filters: Seq[String],
     storageLevel: StorageLevel
   ) extends NetworkReceiver[Status] {
@@ -40,8 +39,7 @@ class TwitterReceiver(
 
   protected override def onStart() {
     blockGenerator.start()
-    twitterStream = new TwitterStreamFactory()
-      .getInstance(new BasicAuthorization(username, password))
+    twitterStream = new TwitterStreamFactory().getInstance(twitterAuth)
     twitterStream.addListener(new StatusListener {
       def onStatus(status: Status) = {
         blockGenerator += status
--- a/streaming/src/main/scala/spark/streaming/dstream/TwitterInputDStream.scala
+++ b/streaming/src/main/scala/spark/streaming/dstream/TwitterInputDStream.scala
@@ -3,27 +3,53 @@ package spark.streaming.dstream
 import spark._
 import spark.streaming._
 import storage.StorageLevel
-
 import twitter4j._
 import twitter4j.auth.BasicAuthorization
 import twitter4j.auth.Authorization
+import java.util.prefs.Preferences
+import twitter4j.conf.PropertyConfiguration
+import twitter4j.auth.OAuthAuthorization
+import twitter4j.auth.AccessToken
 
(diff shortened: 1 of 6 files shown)


JoshRosen:
Possibly related: [STREAMING-54](https://spark-project.atlassian.net/browse/STREAMING-54): "Update Twitter examples due to Twitter API retirement"

pwendell:
Ah I see - @mrpotes, is this username/password authentication just entirely not usable anymore? If so, it might be okay to change the API, but I want to run it by @tdas since he looks at all the streaming code. If the old API is simply impossible to support anymore, it might be better to remove it as you've done here.

mrpotes:
According to [this blog post](https://dev.twitter.com/blog/api-v1-is-retired), v1 API is retired now, and so basic auth is [no longer supported](https://dev.twitter.com/docs/streaming-apis/connecting#Authentication)

pwendell:
Okay thanks Phil - my instinct is then we should just remove it as you did initially. Would you mind updating the examples that use this? I'm not sure they will build with the changes to StreamingContext you've made.

pwendell:
Hey @mrpotes - any idea when you might be able to do this? We are thinking of trying to get this into 0.7.3 - we may  be able to update the examples on our side too, but not sure how their OAuth stuff works.

mrpotes:
I'll have a go tonight, I was having trouble getting the maven build to
compile earlier, hopefully I'll get it sorted at home.

mrpotes:
Ok, I think this should all work now. The Authorization is now an Option[Authorization], and if none is supplied a noddy OAuth is used that uses java.util.prefs.Preferences to store the authentication tokens, and requires the consumer key and secret to be provided as system properties.

I've updated the examples - hope they're ok.

streaming/src/main/scala/spark/streaming/dstream/TwitterInputDStream.scala
```diff
@@ -3,34 +3,59 @@ package spark.streaming.dstream
...
+	  println("Enter PIN: ")
+      val pin = Console.readLine
+      val accessToken = if (pin.length() > 0) oAuth.getOAuthAccessToken(requestToken, pin) else oAuth.getOAuthAccessToken()
+      userRoot.put(PropertyConfiguration.OAUTH_ACCESS_TOKEN, accessToken.getToken())
+      userRoot.put(PropertyConfiguration.OAUTH_ACCESS_TOKEN_SECRET, accessToken.getTokenSecret())
+      userRoot.flush()
+    } else {
+      oAuth.setOAuthAccessToken(new AccessToken(token.get, tokenSecret.get));
+    }
+    oAuth
```
mateiz, line 48: Hi James,

I don't like having our API create an OAuth token and prompting the user on stdin to enter a PIN. That risks being confusing. Can you change this method to just:

```
new OAuthAuthorization(new ConfigurationBuilder().build())
```

Then it will look at the system properties and require all of twitter4j.oauth.consumerKey, twitter4j.oauth.consumerSecret, twitter4j.oauth.accessToken and twitter4j.oauth.accessTokenSecret to be set. Please update the docs in StreamingContext to mention the last two as well. The user can create the token on the Twitter API webpage.

streaming/src/main/scala/spark/streaming/dstream/TwitterInputDStream.scala
```diff
@@ -3,34 +3,59 @@ package spark.streaming.dstream
...
+  }
+  
   override def getReceiver(): NetworkReceiver[Status] = {
-    new TwitterReceiver(username, password, filters, storageLevel)
+    new TwitterReceiver(if (twitterAuth.isEmpty) createOAuthAuthorization else twitterAuth.get, filters, storageLevel)
```
mateiz, line 52: You can change this to `twitterAuth.getOrElse(createOAuthAuthorization())`

streaming/src/main/scala/spark/streaming/dstream/TwitterInputDStream.scala
```diff
@@ -3,34 +3,59 @@ package spark.streaming.dstream
...
     @transient ssc_ : StreamingContext,
-    username: String,
-    password: String,
+    twitterAuth: Option[Authorization],
     filters: Seq[String],
     storageLevel: StorageLevel
   ) extends NetworkInputDStream[Status](ssc_)  {
-
+  
+  lazy val createOAuthAuthorization: Authorization = {
```
mateiz, line 31: Also a small style point.. make this a `def` instead of a `lazy val` and then add `private val authorization = twitterAuth.getOrElse(createOAuthAuthorization())` as a variable within the class; we don't want it to be created only on the first call to `getReceiver`.

mateiz:
Thanks for submitting this! I tried this out and made a few comments on the "default authorization" route to make it easier to use. Let me know if you have time to fix them, otherwise I can implement the fixes too. I'd like to have this in Spark 0.7.3.

mrpotes:
@mateiz thanks for the feedback. I should have time next week to do the suggested changes. If you need it sooner by all means change it yourself. I wonder if the OAuth implementation should be moved to an example?

mrpotes:
PS sorry dor the crappy scala, I was porting straight from a java app and didn't put as much effort into changing the code as I should have!

mateiz:
No worries. I'll tackle these fixes for now because it would be good to put this into 0.7.3, but thanks for submitting it!

mateiz:
Thanks again for submitting this. I've now merged it in.

PR #659, opened 2013-06-25
Refactoring, Consolidation, and Updates to the Web UI

pwendell:
This patch performs an internal re-factoring of the Spark Web UI and adds a new UI component that exposes stage progress and metrics.

The re-factoring removes the dependency on Spray and instead uses Scala's built in XML support and a lightweight server based on Jetty. The changes also consolidate code related to the spark ui in a "ui" package which can server as a point for future expansion. Certain very common rendering operations (such as enumerating a list of items) are factored out into utility functions.

The expansion adds a stage-level monitoring feature which exposes the progress of ongoing stages and statistical metrics for both ongoing and historical stages. It is meant to to be a modest first step towards a more comprehensive workflow UI. Nonetheless, it exposes some very useful previously hidden metrics, such as distribution data about tasks within a job, and an adjacent view of the service time and locality properties of given tasks.

So far this has been tested using a local standalone cluster with a variety of job conditions, including failures. I'm in the process of doing more rigorous tests, but wanted to put this up for review.


core/pom.xml
```diff
@@ -56,6 +56,10 @@
       <artifactId>akka-slf4j</artifactId>
     </dependency>
     <dependency>
+      <groupId>net.liftweb</groupId>
+      <artifactId>lift-json_2.9.2</artifactId>
```
mateiz, line 60: How self-contained is this, i.e., does it bring in a lot of other dependencies from Lift? You can use the sbt dependency-tree plugin to check this.

core/src/main/scala/spark/deploy/master/ui/ApplicationPage.scala
```diff
@@ -0,0 +1,100 @@
...
+
+class ApplicationPage(parent: MasterWebUI) {
+  val master = parent.master
+  implicit val timeout = parent.timeout
+
+  /** Executor details for a particular application */
+  def renderJson(request: HttpServletRequest): JValue = {
+    val appId = request.getParameter("appId")
+    val stateFuture = (master ? RequestMasterState)(timeout).mapTo[MasterState]
+    val state = Await.result(stateFuture, 3 seconds)
```
mateiz: These 3-second timeouts are probably too low if there is GC; would be better to do 20-30s

core/src/main/scala/spark/deploy/master/ui/MasterWebUI.scala
```diff
@@ -0,0 +1,59 @@
...
+    ("*", (request: HttpServletRequest) => indexPage.render(request))
+  )
+
+  def stop() {
+    server.foreach(_.stop())
+  }
+}
+
+object MasterWebUI {
+  val STATIC_RESOURCE_DIR = "spark/webui/static"
```
mateiz: Wouldn't this be better as spark/ui/static, since we don't have a spark.webui package anymore?

core/src/main/scala/spark/ui/SparkUI.scala
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  def stop() {
+    server.foreach(_.stop())
+  }
+
+  private[spark] def appUIAddress = "http://" + host + ":" + boundPort.getOrElse("-1")
+}
+
+object SparkUI {
```
mateiz: This and all the new objects/classes you made should probably be private[spark]; run sbt doc and look at the generated docs to see whether anything in this package shows up.

core/src/main/scala/spark/ui/jobs/JobProgressUI.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  def getHandlers = Seq[(String, Handler)](
+    ("/stages/stage", (request: HttpServletRequest) => stagePage.render(request)),
+    ("/stages", (request: HttpServletRequest) => indexPage.render(request))
+  )
+}
+
+private[spark] class JobProgressListener extends SparkListener {
+  // How many stages to remember
+  val RETAINED_STAGES = 1000
```
mateiz: Probably should make this configurable

mateiz:
Thanks Patrick, this is really exciting! I've made a few quick comments on it but will continue looking.

core/pom.xml
pwendell, line 60: One thing that's nice about this is it has exactly one dependency: the paranamer utility. This only conflicts with one other library we use (avro), where it causes a slight version bump that should be fine.

pwendell:
Hey Matei,

Just added commits to address all your comments. Re: The lift-json library, it's very self-contained and is decoupled from the overall lift framework, that's why I chose it.

mateiz:
Hey Patrick,

I tried it out a bit more and I think it looks great! A few observations though:
- If I click on an active stage, I get the following exception:
  
  13/06/27 10:22:00.668 WARN AbstractHttpConnection: /stages/stage/?id=6
  java.util.NoSuchElementException: key not found: 6
      at spark.ui.jobs.StagePage.render(StagePage.scala:22)
      at spark.ui.jobs.JobProgressUI$$anonfun$getHandlers$1.apply(JobProgressUI.scala:38)
      at spark.ui.jobs.JobProgressUI$$anonfun$getHandlers$1.apply(JobProgressUI.scala:38)
      at spark.ui.JettyUtils$$anon$1.handle(JettyUtils.scala:44)
- For some reason, the formatting of the body seems to be different from the previous UI; the font in tables is much larger than the body font size for Bootstrap should be, or the way it appears in other UI elements. (This is on Safari.)
- Tasks are showing a very low service time even when I make them sleep 1+ seconds each.
- Replace the name "service time" with "duration" to make it easier to understand
- For times displayed on the UI, it might be nice to have a user-understandable number such as "50 ms" or "52 s" instead of hh:mm:ss.sss. For JSON you should definitely standardize on one unit.
- It would be cool to render some progress bars where we show things like 0/10. Shouldn't be too hard with some nested divs or SVG.

mateiz:
Forgot to add one more cosmetic thing: the menu at the top with "Jobs" and "Storage" is kind of rough the way it is now.. would be better with some kind of tabs.

pwendell:
Hey Matei,

I've addressed everything you suggested in the review. The service time glitch was a bug in the local scheduler, that should be working now.

In order to add fancier headers, I had to create different codepaths for the Master/Worker UI generation and the Spark UI, this adds a bit more XML code in UIUtils. The new headers are based on the ones we use in the docs.

pwendell:
Jenkins, test this please.

pwendell:
Jenkins, test this please.

mateiz:
Hey Patrick, the header still looks wrong on the SparkUI in Safari: there's an invalid image URL. Is that meant to be the logo?

core/src/main/scala/spark/ui/jobs/IndexPage.scala
```diff
@@ -0,0 +1,99 @@
...
+import spark.storage.StorageLevel
+
+/** Page showing list of all ongoing and recently finished stages */
+private[spark] class IndexPage(parent: JobProgressUI) {
+  def listener = parent.listener
+  val dateFmt = parent.dateFmt
+
+  def render(request: HttpServletRequest): Seq[Node] = {
+    val stageHeaders = Seq("Stage ID", "Origin", "Submitted", "Duration", "Progress",
+                           "Tasks: Complete/Total", "Shuffle Activity", "Stored RDD")
```
mateiz: I'd merge the "progress" and "tasks: complete/total" into a single column here, just "progress". Then we can display a progress bar with the numbers next to it.

core/src/main/scala/spark/ui/jobs/JobProgressUI.scala
```diff
@@ -0,0 +1,125 @@
...
+import spark.scheduler.cluster.TaskInfo
+import spark.executor.TaskMetrics
+import spark.Success
+import spark.Utils
+
+/** Web UI showing progress status of all jobs in the given SparkContext. */
+private[spark] class JobProgressUI(val sc: SparkContext) {
+  private var _listener: Option[JobProgressListener] = None
+  def listener = _listener.get
+  val dateFmt = new SimpleDateFormat("EEE, MMM d yyyy HH:mm:ss")
```
mateiz: Make this just "MMM d yyyy HH:mm:ss"

mateiz: Or even "yyyy/MM/dd HH:mm:ss" to match the deploy UI one

core/src/main/scala/spark/ui/jobs/IndexPage.scala
```diff
@@ -0,0 +1,99 @@
...
+    val activeStages = listener.activeStages.toSeq
+    val completedStages = listener.completedStages.reverse.toSeq
+    val failedStages = listener.failedStages.reverse.toSeq
+
+    val activeStageTable: NodeSeq = listingTable(stageHeaders, stageRow(), activeStages)
+    val completedStageTable = listingTable(stageHeaders, stageRow(), completedStages)
+    val failedStageTable: NodeSeq = listingTable(stageHeaders, stageRow(false), failedStages)
```
mateiz: Why don't we want click-through links to failed stages? It seems those are specifically some stages where you'd like more info.

core/src/main/scala/spark/ui/jobs/IndexPage.scala
```diff
@@ -0,0 +1,99 @@
...
+      case (false, true) => "Write"
+      case _ => ""
+    }
+    val completedTasks = listener.stageToTasksComplete.getOrElse(s.id, 0)
+    val totalTasks = s.numPartitions
+
+    <tr>
+      {if (showLink) {<td><a href={"/stages/stage?id=%s".format(s.id)}>{s.id}</a></td>}
+       else {<td>{s.id}</td>}}
+      <td>{s.origin}</td>
```
mateiz: IMO place the link around the origin, not the ID, so that it's easier to click.

core/src/main/scala/spark/ui/jobs/IndexPage.scala
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  def getElapsedTime(submitted: Option[Long], completed: Long): String = {
+    submitted match {
+      case Some(t) => parent.formatDuration(completed - t)
+      case _ => "Unknown"
+    }
+  }
+
+  def makeSlider(completed: Int, total: Int): Seq[Node] = {
```
mateiz: This should be called makeProgressBar actually since a slider usually means a draggable control in UI-speak

core/src/main/scala/spark/ui/jobs/IndexPage.scala
```diff
@@ -0,0 +1,99 @@
...
+    val width=130
+    val height=15
+    val completeWidth = (completed.toDouble / total) * width
+
+    <svg width={width.toString} height={height.toString}>
+      <rect width={width.toString} height={height.toString}
+            fill="white" stroke="black" stroke-width="1" />
+      <rect width={completeWidth.toString} height={height.toString}
+            fill="rgb(206,206,247)" stroke="black" stroke-width="1" />
```
mateiz: Cosmetic thing: switch this to one of the theme colors.. maybe the blue for links; and consider making the border slightly lighter, maybe rgb(51, 51, 51) to match the text color. Ultimately it would be good to put these in a CSS file but I guess we can leave them here for now.

core/src/main/scala/spark/ui/JettyUtils.scala
```diff
@@ -0,0 +1,227 @@
...
+                  <div class="container">
+                    <div class="brand"><img src="/static/spark-logo-77x50px-hd.png" /></div>
+                    <ul class="nav">
+                      {storage}
+                      {jobs}
+                      </ul>
+                    <ul id="infolist">
+                      <li>Master: <strong>{sc.master}</strong></li>
+                      <li>Application: <strong>{sc.appName}</strong></li>
+                      <li>Executors: <strong>{sc.getExecutorStorageStatus.size}</strong></li>
```
mateiz: Put the app name before the master URL here

core/src/main/scala/spark/ui/jobs/IndexPage.scala
pwendell: Okay will do - I felt like the origin was a bit worse because it will be the same for multiple phases, so it's not good at distinguishing what you are clicking on. But I'll change it.

pwendell:
Hey Matei,

These commits should address everything in your feedback. I'd still like to manually test this a bit more, but I'm going away for the weekend - so wanted to post for review.
- Patrick

pwendell:
Hey Matei - this is "good to go" from my end. I added a testing script that you can use to generate interesting jobs for the UI.

core/src/main/scala/spark/ui/JettyUtils.scala
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  /** Creates a handler for serving files from a static directory */
+  def createStaticHandler(resourceBase: String): ResourceHandler = {
+    val staticHandler = new ResourceHandler
+    Option(getClass.getClassLoader.getResource(resourceBase)) match {
+      case Some(res) =>
+        staticHandler.setResourceBase(res.toString)
+      case None =>
+        logError("Could not find resource path for Web UI: " + resourceBase)
```
mateiz: Shouldn't this throw an exception?

core/src/main/scala/spark/ui/JettyUtils.scala
```diff
@@ -0,0 +1,227 @@
...
+          (server, server.getConnectors.head.getLocalPort)
+        case f: Failure[_] =>
+          server.stop()
+          logInfo("Failed to create UI at port, %s. Trying again.".format(currentPort))
+          logInfo("Error was: " + f.toString)
+          connect((currentPort + 1) % 65536)
+      }
+    }
+
+
```
mateiz: Make this just one blank line

core/src/main/scala/spark/ui/JettyUtils.scala
```diff
@@ -0,0 +1,227 @@
...
+    connect(port)
+  }
+}
+
+object Page extends Enumeration { val Storage, Jobs = Value }
```
mateiz: Make this private[spark] and move it to a separate file, Page.scala

core/src/main/scala/spark/ui/JettyUtils.scala
```diff
@@ -0,0 +1,227 @@
...
+
+/** Utility functions for generating XML pages with spark content. */
+private[spark] object UIUtils {
```
mateiz: This should also be in a separate file

core/src/main/scala/spark/ui/UIWorkloadGenerator.scala
```diff
@@ -0,0 +1,61 @@
...
+
+import spark.SparkContext
+import spark.SparkContext._
+
+/**
+ * Continuously generates jobs that expose various features of the WebUI (internal testing tool).
+ *
+ * Usage: ./run spark.ui.UIWorkloadGenerator [master]
+ */
+private[spark] object UIWorkloadGenerator {
```
mateiz, line 13: This is great, very useful!

core/src/main/scala/spark/deploy/worker/ui/WorkerWebUI.scala
```diff
@@ -0,0 +1,74 @@
...
+      boundPort = Some(bPort)
+      logInfo("Started Worker web UI at http://%s:%d".format(host, bPort))
+    } catch {
+      case e: Exception =>
+        logError("Failed to create Worker JettyUtils", e)
+        System.exit(1)
+    }
+  }
+
+  def log(request: HttpServletRequest): String = {
```
mateiz, line 53: Viewing logs doesn't work right now -- I get

```
13/07/01 12:31:40.966 WARN AbstractHttpConnection: /log/?appId=app-20130701122932-0000&executorId=0&logType=stdout
java.lang.NullPointerException
    at spark.deploy.worker.ui.WorkerWebUI.log(WorkerWebUI.scala:59)
    at spark.deploy.worker.ui.WorkerWebUI$$anonfun$2.apply(WorkerWebUI.scala:37)
    at spark.deploy.worker.ui.WorkerWebUI$$anonfun$2.apply(WorkerWebUI.scala:37)
    at spark.ui.JettyUtils$$anon$1.handle(JettyUtils.scala:45)
    at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1040)
    at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:976)
    at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:135)
    at org.eclipse.jetty.server.handler.HandlerList.handle(HandlerList.java:52)
    at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:116)
    at org.eclipse.jetty.server.Server.handle(Server.java:363)
    at org.eclipse.jetty.server.AbstractHttpConnection.handleRequest(AbstractHttpConnection.java:483)
    at org.eclipse.jetty.server.AbstractHttpConnection.headerComplete(AbstractHttpConnection.java:920)
    at org.eclipse.jetty.server.AbstractHttpConnection$RequestHandler.headerComplete(AbstractHttpConnection.java:982)
    at org.eclipse.jetty.http.HttpParser.parseNext(HttpParser.java:635)
    at org.eclipse.jetty.http.HttpParser.parseAvailable(HttpParser.java:235)
    at org.eclipse.jetty.server.AsyncHttpConnection.handle(AsyncHttpConnection.java:82)
    at org.eclipse.jetty.io.nio.SelectChannelEndPoint.handle(SelectChannelEndPoint.java:628)
    at org.eclipse.jetty.io.nio.SelectChannelEndPoint$1.run(SelectChannelEndPoint.java:52)
    at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:608)
    at org.eclipse.jetty.util.thread.QueuedThreadPool$3.run(QueuedThreadPool.java:543)
    at java.lang.Thread.run(Thread.java:680)
```

core/src/main/scala/spark/ui/jobs/IndexPage.scala
```diff
@@ -0,0 +1,114 @@
...
+  def makeProgressBar(completed: Int, total: Int): Seq[Node] = {
+    val width=130
+    val height=15
+    val completeWidth = (completed.toDouble / total) * width
+
+    <svg width={width.toString} height={height.toString}>
+      <rect width={width.toString} height={height.toString}
+            fill="white" stroke="black" stroke-width="1" />
+      <rect width={completeWidth.toString} height={height.toString}
+            fill="rgb(51,51,51)" stroke="black" stroke-width="1" />
```
mateiz: Small cosmetic thing.. make the progress bar the same color as the links on the page (whatever blue that is), and make the border rgb(51,51,51)

core/src/main/scala/spark/Utils.scala
```diff
@@ -486,6 +486,26 @@ private object Utils extends Logging {
...
+   * Returns a human-readable string representing a duration such as "35ms"
+   */
+  def msDurationToString(ms: Long): String = {
+    val second = 1000
+    val minute = 60 * second
+    val hour = 60 * minute
+
+    ms match {
+      case t if t < second =>
+        "%dms".format(t)
```
mateiz: Add a space before the units, i.e. "%d ms" or "%d s"

core/src/main/scala/spark/Utils.scala
```diff
@@ -486,6 +486,26 @@ private object Utils extends Logging {
...
+      case t if t < minute =>
+        "%d.%03ds".format(t / second, t % second)
```
mateiz: Here I'd put just one decimal place

core/src/main/scala/spark/Utils.scala
```diff
@@ -486,6 +486,26 @@ private object Utils extends Logging {
...
+      case t if t < hour =>
+        "%d:%02d".format(t / minute, (t % minute) / second)
```
mateiz: It might be better to use "4.5 min" instead of "4:30", though I'm not sure what's standard; the reason is that the latter looks like a time of the day. Same for the hours -- "1.2 h" instead of "1:12:00" or whatever.

mateiz:
Hey Patrick, I took another look at this and made some comments. In particular, the log links on the standalone workers don't seem to be working.

One other weird thing I noticed is that failed tasks don't have a duration field. It would be good to put in the time they took until they failed, so the user knows whether it happened somewhere in the middle.

core/src/main/scala/spark/deploy/worker/ui/WorkerWebUI.scala
```diff
@@ -0,0 +1,74 @@
...
+    val appId = request.getParameter("appId")
+    val executorId = request.getParameter("executorId")
+    val logType = request.getParameter("logType")
+    val path = "%s/%s/%s/%s".format(workDir.getPath, appId, executorId, logType)
+    val source = Source.fromFile(path)
+    val lines = source.mkString
+    source.close()
```
mateiz: Also, reading the whole file is into a String is asking for trouble here. You should either use sendFile if Jetty has that functionality (basically let it stream the file to the client), or cut off the file and only return the last 4-8 KB.

JoshRosen: If you wanted to generalize this, you could allow the user to specify a length and offset into the log as part of the request (which would fix [SPARK-608](https://spark-project.atlassian.net/browse/SPARK-608)).  This would support the `tail` use case and also enable fancy things like long-polling-based Javascript log viewers.  To more easily support `tail`, you could have the start offset work like a Python array index so you could pass -1024 to start reading from the last 1024 bytes of the log.

pwendell:
Okay, these commits fix everything except the log issue. I'm going to look into that now, it just may be a bit tricky.

pwendell:
Okay now this adds truncation to the log files. I bumped it up to 100KB rather than 8, because the logs are pretty chatty and 8KB might not go back very far. This should address everything from the latest review.

mateiz:
Patrick, do you want to wait to include the duration for failed tasks too?

pwendell:
Hey Matei,

I've added collection of metrics for tasks which didn't complete. I also fixed a long-standing bug where the sort order in the UI is not correct for formatted outputs.
- Patrick

mateiz:
Thanks Patrick! I've merged this manually since recent PRs introduced some small conflicts.

PR #662, opened 2013-06-25
Create binary Spark distribution

velvia:
This creates a new packaging target, a binary Spark distribution.  The existing scripts should continue to work with existing packaging (ie sbt package; run.....)

1) A new script, `make-distribution.sh`, creates a binary distro in dist/
2) The dist/ folder contains no sources, just a single assembly jar (for now), along with bin/ and conf/, and the run and spark-shell scripts.
3) The run script has been modified to support both the existing distro and the new binary distro.  The existing classpath logic has been split out to a new file, and the run script size has been cut by half.

The binary-only distribution makes deploys much easier, as it already contains Scala and all other dependencies packaged together.   Once you create the distribution, one could simply rsync the distro folder to the cluster, configure spark, and run the start scripts.
- spark-shell runs on the cluster without Scala installed (and restores echo properly)
- the "ps" output is much cleaner since the classpath will be much shorter

I also sneaked in a small addition to spark-shell to add convenient flags.

Caveats:
- docs have not been updated yet
- currently bagel, streaming, and the examples are excluded from the distro.


mateiz:
Jenkins, this is ok to test

spark-shell
```diff
@@ -1,4 +1,64 @@
-#!/bin/sh
+#!/bin/bash --posix
+#
+# Shell script for starting the Spark Shell REPL
+# Options:
+#    -m            Set MASTER to spark://$SPARK_MASTER_IP:$SPARK_MASTER_PORT
```
mateiz: This is kind of a weird option to have IMO; a better behavior would be to make MASTER default to spark://$SPARK_MASTER_IP:$SPARK_MASTER_PORT if it hasn't been set but those two have been.

mateiz:
Hey, so unfortunately this PR no longer merges cleanly because we updated the run scripts to factor out the classpath computation. Mind merging our master branch back into it?

spark-shell
velvia: sounds good, will change the spark-shell script to set MASTER to those.

velvia:
@mateiz  I actually did merge from master before submitting pull request, and I'll try again.   Not sure if this kind of commit can merge cleanly, unfortunately....

velvia:
I see that in master a very similar change to mine has been made: the classpath computation is split out into a separate file, bin/compute-classpath.sh.    In mine, I moved the classpath computation plus the setting of the JAVA/SCALA runner stuff as well into a separate shell script (because for the binary distribution, including the scala library in the class path is not necessary).

What if we had two compute-classpath scripts?  One for dev environments (the default) as well as one for binary distros?   ie you have 
bin/compute-dev-classpath.sh/cmd
bin/compute-binary-classpath.sh

velvia:
All changes merged now.   By the way, we can get rid of SPARK_LAUNCH_WITH_SCALA (unless needed for backwards compatibility) since I got rid of the need for that with spark-shell mods.

andyk:
Jenkins retest this please

velvia:
Hey guys, tests pass.  Any issues with merging this in?

thanks.

mateiz:
Sorry for the delay, but I've merged this now, after making a few small fixes (I noticed that the version printed in "RELEASE" had some color codes). Thanks for submitting it!

velvia:
Many thanks!

-Evan
To be free is not merely to cast off one's chains, but to live in a way that respects & enhances the freedom of others. (#NelsonMandela)

PR #665, opened 2013-06-28
Support Cancellation of Spark Jobs

harsha2010:
Hi

This patch allows us to cancel runaway Spark queries by issuing a killJob command passing in the jobId. The approach taken here is to let the DAG scheduler clean up its state about a job that is currently executing, and propagate an interrupt to the running task. Each task runs within a Future to properly respond to interrupts and cancel execution. Every Iterator that iterates over data (either via CacheManager/ or BlockFetcher or the HadoopRDD) is wrapped by an InterruptibleIterator which checks for the interrupt and cleans up state accordingly and exits.
I have tested the performance of this patch against master on a bunch of internal queries and the performance is not impacted by this patch.
I am in the process of obtaining TPCH benchmarks with and without the patch which i will attach here.
In the meanwhile, please review and let me know if the design needs changes.

### DIFF
--- a/core/src/main/scala/spark/CacheManager.scala
+++ b/core/src/main/scala/spark/CacheManager.scala
@@ -19,7 +19,8 @@ private[spark] class CacheManager(blockManager: BlockManager) extends Logging {
       case Some(cachedValues) =>
         // Partition is in cache, so just return its values
         logInfo("Found partition in cache!")
-        return cachedValues.asInstanceOf[Iterator[T]]
+        val iter = cachedValues.asInstanceOf[Iterator[T]]
+        return new InterruptibleIteratorDecorator(iter)
 
       case None =>
         // Mark the split as loading (unless someone else marks it first)
@@ -37,7 +38,8 @@ private[spark] class CacheManager(blockManager: BlockManager) extends Logging {
             // downside of the current code is that threads wait serially if this does happen.
             blockManager.get(key) match {
               case Some(values) =>
-                return values.asInstanceOf[Iterator[T]]
+                val iter =  values.asInstanceOf[Iterator[T]]
+                return new InterruptibleIteratorDecorator(iter)
               case None =>
                 logInfo("Whoever was loading " + key + " failed; we'll try it ourselves")
                 loading.add(key)
@@ -53,7 +55,8 @@ private[spark] class CacheManager(blockManager: BlockManager) extends Logging {
           elements ++= rdd.computeOrReadCheckpoint(split, context)
           // Try to put this block in the blockManager
           blockManager.put(key, elements, storageLevel, true)
-          return elements.iterator.asInstanceOf[Iterator[T]]
+            val iter = elements.iterator.asInstanceOf[Iterator[T]]
+          return new InterruptibleIteratorDecorator(iter)
         } finally {
           loading.synchronized {
             loading.remove(key)
--- a/core/src/main/scala/spark/InterruptibleIterator.scala
+++ b/core/src/main/scala/spark/InterruptibleIterator.scala
@@ -1,0 +1,26 @@
+package spark
+
+trait InterruptibleIterator[+T] extends  Iterator[T]{
+
+  override def hasNext(): Boolean = {
+    if (!Thread.currentThread().isInterrupted()) {
+      true
+    } else {
+      throw new InterruptedException ("Thread interrupted during RDD iteration")
+    }
+  }
+
+}
+
+class InterruptibleIteratorDecorator[T](delegate: Iterator[T])
+	extends AnyRef with InterruptibleIterator[T] {
+  
+  override def hasNext(): Boolean = {
+    super.hasNext
+    delegate.hasNext
+  }
+  
+  override def next(): T = {
+    delegate.next()
+  }
+}
--- a/core/src/main/scala/spark/executor/Executor.scala
+++ b/core/src/main/scala/spark/executor/Executor.scala
@@ -2,16 +2,16 @@ package spark.executor
 
 import java.io.{File, FileOutputStream}
 import java.net.{URI, URL, URLClassLoader}
+import java.nio.ByteBuffer
 import java.util.concurrent._
-
+import scala.collection.JavaConversions._
+import scala.collection.mutable.{ArrayBuffer,ConcurrentMap, HashMap, Map}
+import scala.concurrent.JavaConversions._
 import org.apache.hadoop.fs.FileUtil
-
-import scala.collection.mutable.{ArrayBuffer, Map, HashMap}
-
 import spark.broadcast._
 import spark.scheduler._
 import spark._
-import java.nio.ByteBuffer
+import spark.scheduler.cluster.TaskDescription
 
 /**
  * The Mesos executor for Spark.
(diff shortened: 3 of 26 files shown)


core/src/main/scala/spark/InterruptibleIterator.scala
```diff
@@ -0,0 +1,26 @@
...
+    }
+  }
+
+}
+
+class InterruptibleIteratorDecorator[T](delegate: Iterator[T])
+	extends AnyRef with InterruptibleIterator[T] {
+  
+  override def hasNext(): Boolean = {
+    super.hasNext
```
squito, line 19: It seems kinda weird to me that `InterruptibleIterator` is actually implementing `hasNext`, which then you override here.  Maybe it should have a method `exceptionIfThreadInterrupted`.  It seems like the trait is not actually implementing `hasNext` at all, its just supplying a utility methods for implementations.

core/src/main/scala/spark/scheduler/cluster/SchedulableBuilder.scala
```diff
@@ -112,4 +123,14 @@ private[spark] class FairSchedulableBuilder(val rootPool: Pool) extends Schedula
     parentPool.addSchedulable(manager)
     logInfo("Added task set " + manager.name + " tasks to pool "+poolName)
   }
+  
+  override def popTaskSetManagers(stageId: Int) = {
```
squito, line 127: can you avoid repeating this method, eg. move to a base class?

squito:
This looks great!  I have wanted this feature for a long time.  I made a couple of minor style comments.

I only have one question about the implementation -- it seems like you register the cores as freed as soon as the killTask request gets sent.  Do the tasks really die immediately?  Should it wait for some acknowledgement that the task really has been killed?  I guess its not horrible to have too many tasks running on an executor for a little while, so if it adds a lot of complication for a rare corner case, maybe we can forget it.

core/src/main/scala/spark/scheduler/cluster/SchedulableBuilder.scala
harsha2010, line 127: good point . thanks will do

harsha2010:
Thanks for your comments., I'll incorporate them in the update i will add shortly with TPCH performance #s.
The registering cores point is well taken: I couldn't find an easy way to deal with the acknowledgement that a task has been actually killed so implemented the optimistic solution. I'll take a look once more to see if i can do something better.

core/src/main/scala/spark/scheduler/Task.scala
```diff
@@ -1,23 +1,48 @@
...
+    f = new FutureTask(new Callable[T] {
+      def call(): T = {
+        runInterruptibly(attemptId)
+      }
+    })
+    try {
+      f.run()
+      f.get()
+    } catch {
+      case e: Exception => throw e.getCause()
```
rxin, line 29: Should we be catching Throwable here?

core/src/main/scala/spark/executor/Executor.scala
```diff
@@ -79,14 +79,26 @@ private[spark] class Executor(executorId: String, slaveHostname: String, propert
...
+    val runner = new TaskRunner(context, taskId, serializedTask)
+    val task = threadPool.submit(runner).asInstanceOf[FutureTask[_]]
+    tasks.put(taskId, task)
+    
+  }
+
+  def killTask(context: ExecutorBackend, taskId: Long, executorId: String) {
+    val task = tasks.get(taskId)
+    task match {
+      case Some(t) => t.cancel(true)
```
mateiz, line 93: It may be good to do tasks.remove(taskId) here too in case the task thread had never started and so we never got to the "finally" in its run() method.

harsha2010, line 93: good point, will do thanks

core/src/main/scala/spark/scheduler/Task.scala
harsha2010, line 29: yes, you are right., we should be catching throwable here.

shivaram:
There was one major issue I ran into when I tried to do this before. I am not sure if this still applies:

When canceling tasks that were reading files from HDFS, I noticed that the sockets open in `DFSClient` would be interrupted and this would lead to datanodes being marked as dead. I had to add a bunch of `catch InterruptedException` in DFSClient to avoid this before in Hadoop 0.20. I am not sure if this is the case in later versions of Hadoop and this would be good to test. 

(As a generalization, I guess we are relying on user-defined classes behaving appropriately when we interrupt a thread  ?)

mateiz:
Hey Ram, as I'm looking at this more closely, one question on the Iterator design: why do we need CompletionIterator to also extend InterruptibleIterator? Can't we just catch the InterruptedException in the code that's running the task and run the cleanup procedure there?

mateiz:
(Or more generally, catch it on a call to next or hasNext; basically the on-complete callbacks should be called even if the task throws an exception, so InterruptedException doesn't need to be different).

harsha2010:
Hey Matei,
Now that i think about it i think you are right: the runInterruptibly methods in all tasks have a finally block that calls oncomplete callbacks. So the interrupted exception does not need to be different.

core/src/main/scala/spark/CacheManager.scala
```diff
@@ -53,7 +55,8 @@ private[spark] class CacheManager(blockManager: BlockManager) extends Logging {
           elements ++= rdd.computeOrReadCheckpoint(split, context)
           // Try to put this block in the blockManager
           blockManager.put(key, elements, storageLevel, true)
-          return elements.iterator.asInstanceOf[Iterator[T]]
+            val iter = elements.iterator.asInstanceOf[Iterator[T]]
```
markhamstra, line 58: excess indentation

core/src/main/scala/spark/InterruptibleIterator.scala
markhamstra, line 19: Yes, this seems odd to me, too.  Is there some reason why `InterruptibleIterator` can't be implemented more like `CompletionIterator`?  In other words, `class InterruptibleIteratorDecorator` just becomes `class InterruptibleIterator`, which extends `Iterator`; and your `InterruptibleIterator` trait becomes `object InterruptibleIterator`, which defs a function called something like `notInterrupted` to replace all of the calls to `InterruptibleIterator.hasNext` -- i.e. the peculiar `super.hasNext` calls become something like `InterruptibleIterator.notInterrupted`.

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -278,6 +278,41 @@ class DAGScheduler(
...
   }
 
+  def killJob(jobId: Int, reason: String)
+  {
+    logInfo("Killing Job %s".format(jobId))
+    val j = activeJobs.find(j => j.runId.equals(jobId))
+    j match {
+      case Some(job) => killJob(job, reason)
+      case None => Unit
+    }
```
markhamstra, line 288: `j.map(killJob(_, reason))` instead of pattern match boilerplate with a noop `None` case.  Similarly for other instances where you only want to do something when you have `Some(thing)`.

markhamstra:
After merging this PR into master @ 7dcda9ae74818f17b57448acab8bd1dc40d78c19, CancellationSuite "Cancel Task" is not completing for me.

harsha2010:
Oh ok will take a look. The test might not have been as well thought out as I imagined, it seems to work on my machine but possibly has a race condition.
Thanks will fix it

core/src/main/scala/spark/scheduler/DAGScheduler.scala
harsha2010, line 288: nice, thanks...did not know this pattern

markhamstra, line 288: http://ymasory.github.io/error-handling-in-scala/#slide-52

Not everything that Yuvi presents there can be used in pre-2.10 Scala.

markhamstra, line 288: Oh, and in this particular case where `killJob` evaluates to `Unit`, `j.foreach(killJob(_, reason))` is a little clearer than the `map`.  Use `map` when you actually want to produce `Some(result)`.

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -278,6 +278,41 @@ class DAGScheduler(
...
+  }
+
+  private def killJob(job: ActiveJob, reason: String) {
```
markhamstra, line 291: @mateiz Doesn't this have the same problem discussed in https://github.com/mesos/spark/pull/414 where more than one ActiveJob can share a stage?

mateiz, line 291: Yes, that is actually true. To do this properly we'll need to do some kind of reference-counting on the stages (keep a list of which jobs currently want to run this stage). One difference here is that killJob is called by the user and for the first use case, of Shark, it's probably going to be fine. But it would be good to either track this properly or send a warning.

markhamstra, line 291: That's pretty much the conclusion that I was arriving at.  I'll work on the reference-counting refactoring.  Should be doable independently of this PR and only require a minimal change here once it is done.

mateiz, line 291: Cool, that would be great to have.

rxin:
Ram - I am closing this one because it is going to be subsumed by #935.

PR #666, opened 2013-07-02
hbase dependency is missed in hadoop2-yarn profile of examples module

ghost:
Adding the hbase dependency to the profile.
Perhaps the better way is to move all the common deps out of the profile, but it might need to be done separately.


mateiz:
Looks good, thanks!

PR #667, opened 2013-07-02
Added pairing to ec2 script.

itaibenari:
added utility for allowing access from local to master on the spark port, enabling shell connection to master.

### DIFF
--- a/ec2/spark_ec2.py
+++ b/ec2/spark_ec2.py
@@ -41,7 +41,7 @@ LATEST_AMI_URL = "https://s3.amazonaws.com/mesos-images/ids/latest-spark-0.7"
 # Configure and parse our command-line arguments
 def parse_args():
   parser = OptionParser(usage="spark-ec2 [options] <action> <cluster_name>"
-      + "\n\n<action> can be: launch, destroy, login, stop, start, get-master",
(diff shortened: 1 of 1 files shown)


JoshRosen:
Does this actually let you run jobs against the remote master?  I'm pretty sure that this won't work because Spark uses additional ephemeral ports for other components, and also because your driver might be behind a firewall that prevents the master from connecting back to it.  See https://groups.google.com/d/msg/spark-users/fI-873l2edY/Ds0snjezShEJ

itaibenari:
It wouldn't work for a client machine not accessible to the slaves. Another requirement seems to be that the directory structure on the client needs to mimic the one on the cluster nodes (/root/spark) which may be achieved by an additional step of soft linking. I agree these restrictions limit the usability of this feature.

JoshRosen:
Since that's the case, I'm against merging this pull request.  I think that it doesn't provide much value and will confuse users.  If you want a quick way to add the current machine to the master's EC2 security group, you can use `ec2-authorize` from the [EC2 API tools](http://aws.amazon.com/developertools/351):

```
ec2-authorize $CLUSTERNAME-master -P tcp -p 7077 -s `curl http://whatismyip.akamai.com/`/32
```

itaibenari:
understandable

PR #668, opened 2013-07-03
update guava version from 11.0.1 to 14.0.1

shimingfei:
update guava version from 11.0.1 to 14.0.1 for some application use

### DIFF
--- a/project/SparkBuild.scala
+++ b/project/SparkBuild.scala
@@ -142,7 +142,8 @@ object SparkBuild extends Build {
     ),
 
     libraryDependencies ++= Seq(
-      "com.google.guava" % "guava" % "11.0.1",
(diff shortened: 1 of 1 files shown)


mateiz:
Why did you also add a dependency on findbugs as part of this?

JoshRosen:
@mateiz See https://code.google.com/p/guava-libraries/wiki/UseGuavaInYourBuild

> Scala note: The Scala compiler, unlike the Java compiler, requires that annotations used by a library be available when compiling against that library. If you are compiling with Scala, you must [declare a dependency on JSR-305](http://stackoverflow.com/questions/10007994/why-do-i-need-jsr305-to-use-guava-in-scala/10013226#10013226).

mateiz:
Ah makes sense. I've merged this in then.

PR #669, opened 2013-07-03
Shuffle consolidation

jason-dai:
In Spark, it is common practice (and usually preferred) to launch a large number of small tasks, which unfortunately can create an even larger number of very small shuffle files – one of the major shuffle performance issues. To address this issue, this change will combine multiple such files (for the same reduce partition or bucket) into one single large shuffle file. Please see [SPARK-751](https://spark-project.atlassian.net/browse/SPARK-751) for the detailed design.


mateiz:
Hey Jason, I read the design document and this looks very good! Thanks for writing that up. I'm going to go over the code in detail in the next few days to see if I have comments.

mateiz:
Jenkins, this is ok to test

mateiz:
Jenkins, retest this please

rxin:
Matei - you can review this now but don't merge it. I talked to @jason-dai the other day and he said there were a couple more changes he wanted to add.

jason-dai:
I have actually made all the changes I want :-)

core/src/main/scala/spark/BlockStoreShuffleFetcher.scala
```diff
@@ -18,33 +18,34 @@ private[spark] class BlockStoreShuffleFetcher extends ShuffleFetcher with Loggin
     val blockManager = SparkEnv.get.blockManager
 
     val startTime = System.currentTimeMillis
-    val statuses = SparkEnv.get.mapOutputTracker.getServerStatuses(shuffleId, reduceId)
+    val (mapLocations, blockSizes)= SparkEnv.get.mapOutputTracker.getServerStatuses(shuffleId, reduceId)
```
mateiz: Put a space after the =

core/src/main/scala/spark/BlockStoreShuffleFetcher.scala
```diff
@@ -18,33 +18,34 @@ private[spark] class BlockStoreShuffleFetcher extends ShuffleFetcher with Loggin
...
-      splitsByAddress.getOrElseUpdate(address, ArrayBuffer()) += ((index, size))
-    }
-
-    val blocksByAddress: Seq[(BlockManagerId, Seq[(String, Long)])] = splitsByAddress.toSeq.map {
-      case (address, splits) =>
-        (address, splits.map(s => ("shuffle_%d_%d_%d".format(shuffleId, s._1, reduceId), s._2)))
-    }
+    val blocksByAddress: Seq[(BlockManagerId, Seq[(String, Long)])] = blockSizes.map { s =>
+      val bm = s._1
+      val groups = s._2
```
mateiz: Just FYI you could also write this as `val (bm, groups) = s` or write `blockSizes.map { case (bm, groups) =>`.

core/src/main/scala/spark/BlockStoreShuffleFetcher.scala
```diff
@@ -18,33 +18,34 @@ private[spark] class BlockStoreShuffleFetcher extends ShuffleFetcher with Loggin
...
+      (bm, blockIds)
+    }  
 
-    def unpackBlock(blockPair: (String, Option[Iterator[Any]])) : Iterator[(K, V)] = {
-      val blockId = blockPair._1
-      val blockOption = blockPair._2
+    def unpackBlock(blockTuple: (BlockManagerId, String, Option[Iterator[Any]])) : Iterator[(K, V)] = {
+      val address = blockTuple._1
+      val blockId = blockTuple._2
+      val blockOption = blockTuple._3
```
mateiz: Same here with the unpacking, you can do `val (address, blockId, blockOption) = blockTuple`

core/src/main/scala/spark/MapOutputTracker.scala
```diff
@@ -277,18 +303,19 @@ private[spark] object MapOutputTracker {
...
+        sizes: HashMap[BlockManagerId, ShuffleBlockGroupSizeArray]): Seq[(BlockManagerId, Seq[(Int, Array[Long])])] = {
+    assert (sizes != null)
+    sizes.toSeq.map {
+      case (bmId, groups) => 
+        if (groups == null) {
           throw new FetchFailedException(null, shuffleId, -1, reduceId,
             new Exception("Missing an output location for shuffle " + shuffleId))
         } else {
-          (status.location, decompressSize(status.compressedSizes(reduceId)))
+          val seq = for (i <- 0 until groups.groupNum) yield (i, groups(i).bucketSizes(reduceId).map(decompressSize(_)))
```
mateiz: This line is longer than 100 characters; split into two

core/src/main/scala/spark/MapOutputTracker.scala
```diff
@@ -319,3 +346,58 @@ private[spark] object MapOutputTracker {
     }
   }
 }
+
+private[spark] class MapOutputLocation (val location: BlockManagerId, val sequence: Int) extends Serializable {
+  def this (status: MapStatus) = this (status.location, status.sequence)
+  def debugString = "MapOutputLocation(location=" + location + ", sequence=" + sequence +")"
+}
+
+private[spark] class GroupBucketSizes (var sequence: Int, var bucketSizes: Array[Array[Byte]]) extends Serializable {
```
mateiz: This line is longer than 100 characters

core/src/main/scala/spark/MapOutputTracker.scala
```diff
@@ -319,3 +346,58 @@ private[spark] object MapOutputTracker {
...
+        str += (if (x._2 != 0) ", " else "") + x._1
+      }
+      str += ")"
+    }
+    str += ")"
+    str
+  }
+}
+
+private[spark] class ShuffleBlockGroupSizeArray () extends Serializable {
```
mateiz: You don't need the empty parens `()` if there are no constructor arguments; just do `class X extends Y`.

core/src/main/scala/spark/executor/Executor.scala
```diff
@@ -16,7 +16,8 @@ import java.nio.ByteBuffer
 /**
  * The Mesos executor for Spark.
  */
-private[spark] class Executor(executorId: String, slaveHostname: String, properties: Seq[(String, String)]) extends Logging {
+private[spark] class Executor(executorId: String, slaveHostname: String, properties: Seq[(String, String)], cores: Int = 128)
```
mateiz: This line is also too long and should be split

core/src/main/scala/spark/executor/Executor.scala
```diff
@@ -39,7 +40,8 @@ private[spark] class Executor(executorId: String, slaveHostname: String, propert
   for ((key, value) <- properties) {
     System.setProperty(key, value)
   }
-
+  System.setProperty("spark.slaveCores", cores.toString)
```
mateiz: It's a bit unfortunate that we have to pass this to the block manager through a system property. Can you add it to the constructor of BlockManager / ShuffleBlocksPool instead? I believe that gets created in SparkEnv so you could pass it to SparkEnv.

core/src/main/scala/spark/storage/ShuffleBlockManager.scala
```diff
@@ -2,49 +2,358 @@ package spark.storage
...
+
+  private def bytesToInt (buffer: Array[Byte]) = {
+    var result = buffer(0).toInt & 0xFF
+    result |= (buffer(1).toInt & 0xFF) << 8
+    result |= (buffer(2).toInt & 0xFF) << 16
+    result |= (buffer(3).toInt & 0xFF) << 24
+    result    
+  }
+  
+  def writeChunkHeader(id: Int, seq: Int, bs: OutputStream, buffer: Array[Byte]) {
```
mateiz: You could probably pass around DataOutputStreams instead of having to do the intToBytes stuff here, though it's not a major issue

mateiz:
Hey Jason,

I've looked through this in more detail now and made some comments, mostly about style. Overall it looks very good. Before merging it though I also want to test it on a cluster in various scenarios -- we're working on a performance test suite to validate these kinds of changes. But thanks for putting this together!

core/src/main/scala/spark/executor/Executor.scala
jason-dai: Yes, we can do that if Spark is running is cluster (including local cluster) mode, as each executor will construct its SparkEnv. However, if Spark in running in local mode, it has no executors and will just use BlockManager in SparkContext, which is ready constructed and currently we just set the system property in LocalScheduler.

An alternative is, instead of using a system property, we can pass it as a variable in BlockManager (by setting it as appropriate in Executor and LocalScheduler).

mateiz: Yeah, let's make it a variable in the BlockManager then.

mateiz:
Hey Jason, so FYI, some unit tests are legitimately failing with this (https://amplab.cs.berkeley.edu/jenkins/job/SparkPullRequestBuilder/246/) -- it seems to be because the tests compare some arrays of Longs by equality instead of dealing with the individual members.

pwendell:
I just tried to run this on a cluster and jobs failed with a fetch failure issue:

```
13/07/29 05:12:33 INFO cluster.ClusterTaskSetManager: Lost TID 415 (task 1.0:15)
13/07/29 05:12:33 INFO cluster.ClusterTaskSetManager: Loss was due to fetch failure from null
```

I can try and dig into this more tomorrow.

jason-dai:
We are setting up a cluster for large scale testing as well.

mateiz:
Hey Jason, just curious, have you had a chance to look into this? We'd like to code-freeze Spark 0.8 soon (ideally at the end of this week), and it would be nice to get this into it. There can be some time for bug fixing after but we should make sure it's working well before we add it. Of course, if it's too early, we'll just push it to a future release.

jason-dai:
Hi Matei - sorry we haven't had enough time to look into this yet. Maybe we should push it to a future release, as we'll be working on the graphx performance in the next couple of weeks.

mateiz:
No worries if it can't be done now, but I'm curious, have you guys been running with this code for a while, or are you still using the old version internally? Basically I'm wondering whether it requires a lot of testing or just a little. We can help with the testing too.

jason-dai:
Small scale testing works fine, but we ran into some wired failures in large scale testing and had not had enough time to look into it.

PR #670, opened 2013-07-04
EC2 SSH improvements

jey:
This patch makes two changes to the EC2 scripts:
1. Generates an ephemeral SSH key to be used within each cluster instead of copying the user's private key everywhere. This also has the side-effect of making the "--identity-file" parameter optional.
2. Replaces most of the string manipulation to construct shell commands with argument lists for safety and composability.

### DIFF
--- a/ec2/spark_ec2.py
+++ b/ec2/spark_ec2.py
@@ -23,6 +23,7 @@ from __future__ import with_statement
 
 import logging
 import os
+import pipes
 import random
 import shutil
 import subprocess
@@ -536,18 +537,41 @@ def deploy_files(conn, root_dir, opts, master_nodes, slave_nodes, modules):
               dest.write(text)
               dest.close()
   # rsync the whole directory over to the master machine
-  command = (("rsync -rv -e 'ssh -o StrictHostKeyChecking=no -i %s' " +
-      "'%s/' '%s@%s:/'") % (opts.identity_file, tmp_dir, opts.user, active_master))
-  subprocess.check_call(command, shell=True)
+  command = [
+      'rsync', '-rv',
+      '-e', stringify_command(ssh_command(opts)),
+      "%s/" % tmp_dir,
+      "%s@%s:/" % (opts.user, active_master)
+    ]
+  subprocess.check_call(command)
   # Remove the temp directory we created above
   shutil.rmtree(tmp_dir)
+
+
+def stringify_command(parts):
+  if isinstance(parts, str):
+    return parts
+  else:
+    return ' '.join(map(pipes.quote, parts))
+
+
+def ssh_args(opts):
+  parts = ['-o', 'StrictHostKeyChecking=no', '-i', opts.identity_file]
+  return parts
+
+
+def ssh_command(opts):
+  return ['ssh'] + ssh_args(opts)
+
+
+def scp_command(opts):
+  return ['scp', '-q'] + ssh_args(opts)
 
 
 # Copy a file to a given host through scp, throwing an exception if scp fails
 def scp(host, opts, local_file, dest_file):
   subprocess.check_call(
-      "scp -q -o StrictHostKeyChecking=no -i %s '%s' '%s@%s:%s'" %
-      (opts.identity_file, local_file, opts.user, host, dest_file), shell=True)
+      scp_command(opts) + [local_file, "%s@%s:%s" % (opts.user, host, dest_file)])
 
 
 # Run a command on a host through ssh, retrying up to two times
@@ -557,8 +581,7 @@ def ssh(host, opts, command):
   while True:
     try:
       return subprocess.check_call(
-        "ssh -t -o StrictHostKeyChecking=no -i %s %s@%s '%s'" %
-        (opts.identity_file, opts.user, host, command), shell=True)
+        ssh_command(opts) + ['-t', '%s@%s' % (opts.user, host), stringify_command(command)])
     except subprocess.CalledProcessError as e:
       if (tries > 2):
         raise e
@@ -670,11 +693,11 @@ def main():
         conn, opts, cluster_name)
     master = master_nodes[0].public_dns_name
     print "Logging into master " + master + "..."
-    proxy_opt = ""
+    proxy_opt = []
     if opts.proxy_port != None:
-      proxy_opt = "-D " + opts.proxy_port
-    subprocess.check_call("ssh -o StrictHostKeyChecking=no -i %s %s %s@%s" %
-        (opts.identity_file, proxy_opt, opts.user, master), shell=True)
+      proxy_opt = ['-D', opts.proxy_port]
+    subprocess.check_call(
+        ssh_command(opts) + proxy_opt + ['-t', "%s@%s" % (opts.user, master)])
 
   elif action == "get-master":
     (master_nodes, slave_nodes) = get_existing_cluster(conn, opts, cluster_name)
--- a/ec2/spark_ec2.py
+++ b/ec2/spark_ec2.py
@@ -37,6 +37,9 @@ import boto
 from boto.ec2.blockdevicemapping import BlockDeviceMapping, EBSBlockDeviceType
 from boto import ec2
 
+class UsageError(Exception):
+  pass
+
 # A URL prefix from which to fetch AMI information
 AMI_PREFIX = "https://raw.github.com/mesos/spark-ec2/v2/ami-list"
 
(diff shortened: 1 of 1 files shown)


jey:
To clarify the motivation: it's not good practice to assume that the user wants his own personal private key to become the private key that's used by the cluster. (For example, this cluster may be shared with other users who shouldn't have access to this user's private key.) This patch fixes this by generating a new key for each cluster that's only for use within that cluster.

mateiz:
Yeah, this makes sense, though the one thing I'm wary of is including a lot of commands in the script here, e.g. this:

```
(ssh-keygen -q -t rsa -N '' -f ~/.ssh/id_rsa && cat ~/.ssh/id_rsa.pub >> ~/.ssh/authorized_keys)
```

Would it be possible to put more of these in the scripts we run in the AMI (http://github.com/mesos/spark-ec2)?

shivaram:
Yes - that should be possible and you can also avoid the tar rsync process. You can just add this to `setup.sh`  in spark-ec2 I guess

jey:
If I remove the rsync call, what should replace it so the members of the cluster can agree on the key(s) to use? The reason I did it this way is that the only key in the `authorized_keys` file when the AMI is launched is the user's own public key, so we have to log in from the user's key to authorize any other key.

I can look into moving the key generation to `spark-ec2/setup.sh`, but we'd probably only want to do it once per cluster (e.g. on the master), otherwise with N members in the cluster we'd have to tell each member of the cluster about N-1 additional `authorized_keys`. We'd also probably want to be careful to keep backward compatibility so that old `spark_ec2.py` scripts don't conflict with new revisions of the `spark-ec2` repository.

shivaram:
Ah - I see your point. Its a chicken and egg problem in some sense. Here is another approach -- We can generate the key on the client's machine and then copy both the private and public key to all the machines (master + slave). After that we can run a command on all machines from spark_ec2.py to append the public key to authorized keys.
This should be a spark_ec2.py only change and the set of commands run on the machines is either just `scp` or `cat`

jey:
I like that approach, but I think it's too strong an assumption that the client machine will have the appropriate `ssh-keygen` available; seems like it'd be more reliable to do that in the AMI where we have a more controlled/predictable environment.

The reason for using `tar` instead of `scp` or `cat` is that it needs to transfer a minimum two files, `id_rsa` and `authorized_keys` (or three, if you include the optional `id_rsa.pub`) while preserving file modes etc. This could have been done with `cat` or `scp`, but I was aiming to minimize the total number of round trips (this is also why the key generation step is one big multi-part command).

mateiz:
Maybe in `spark-ec2`, we could have the key go into some named file? That way if we see a file with the same name, we don't attempt to re-generate it.

jey:
@mateiz, as we discussed a while ago, we have to do this on the host that is launching the cluster because that's the only one that is authorized to SSH when the cluster is first launched.

Could this be re-reviewed for merging? I've rebased this patch to the latest `master`, fixed a bug, and cleaned up the formatting a bit.

jey:
N.B.: One change that I made is that I got rid of the uses of `shell=True`, as it hinders composability and can be unsafe. I audited all the callers of methods like `ssh()` and did not see anywhere that made use of `shell=True` after I switched the commands to be constructed as sequences rather than strings. (Previously we relied on `shell=True` to parse the command string into `argv`, but with this patch we construct `argv` in way suitable for `execv()`.)

mateiz:
Hey Jey, this is a good idea but @pwendell and @shivaram are currently working on some AMI-side changes to the EC2 scripts. Mind waiting for that to be done?

jey:
No problem, though I'm hopeful it will be merged in time for v0.8

jey:
Should I update this patch so it will apply cleanly on top of #603?

mateiz:
Yeah, though maybe you can wait until that's merged since Patrick and Shivaram might still be editing it. CC @pwendell

jey:
Bump. Can this be merged now that #603 has landed? It's kind of embarrassing that we copy people's private keys around willy-nilly.

The patch needs updates to apply atop #603 but I doubt I have time to do it before I leave for vacation, though I'll take a look if I have time

ec2/spark_ec2.py
```diff
@@ -550,18 +570,37 @@ def ssh(host, opts, command):
...
+        ssh_command(opts) + ['%s@%s' % (opts.user, host), stringify_command(command)],
+        stdin=subprocess.PIPE)
+    proc.stdin.write(input)
+    proc.stdin.close()
+    if proc.wait() == 0:
+      break
+    elif (tries > 2):
+      raise RuntimeError("ssh_write error %s" % proc.returncode)
+    else:
+      print "Error connecting to host, sleeping 30"
```
pwendell: Is this necessarily a connection error? What if the command on the remote node returns nonzero... wouldn't you also end up in this branch?

jey: @pwendell ah yeah, true. Do you think it's OK to just change the error message? I suppose there's a chance that if `ssh_write` is being called with a non-idempotent operation that fails partially, we wouldn't want to retry it. However, since we have the `--resume` flag, I think we currently expect all operations to be repeatable?

jey:
I have rebased this to latest `master`.

ec2/spark_ec2.py
pwendell: I think changing the message is fine. I can do this myself when I merge, but it would be great if you could submit a patch.

ec2/spark_ec2.py
```diff
@@ -103,11 +104,7 @@ def parse_args():
     parser.print_help()
     sys.exit(1)
   (action, cluster_name) = args
-  if opts.identity_file == None and action in ['launch', 'login', 'start']:
```
pwendell, line 106: Isn't `identity-file` still required for logging in?

pwendell:
Hey @jey - I put down a few small comments. I know this has been sitting around for a while and my apologies for my more expediently reviewing once we did the larger merge. That said, I'd like to merge this into master just after we cut 0.8.0. This is a nontrivial change to the way that the clusters operate and I want some time to let people play with it and test it before we include it in a release.

We've never had someone request this before (though I agree it's a change we need to do) so I think it can wait until 0.8.1.

ec2/spark_ec2.py
jey, line 106: Nope, just like `ssh` login works without `-i` when the remote side has added your public key to its `authorized_keys` file. There's also some kind of mechanism to tell ssh(1) about additional keys than just the one in `~/.ssh/id_[rd]sa`.

pwendell, line 106: I'm still a little confused, are you saying that we just rely on ssh to detect the keys, or is there somewhere else in this script that I'm missing where you tell ssh to use the specific key? The thing I'm familiar with is that ssh will search through the keys in ~/.ssh and try them all when authenticating. The reason why I'd prefer not to rely on this is because of two failure modes:
1. The user has stored their key in a location not in the default ssh search path.
2. Some instances of the ssh protocol limit the number of keys that can be tried before failing to authenticate. This is really hard to debug when it happens... basically if you have 20 keys, it will try 6 of them and then fail, even if you have the necessary key correctly in ~/.ssh.

Are you referring to this type of discovery (where the ssh client scans the keys in ~/.ssh) or something else?

jey, line 106: Yeah, exactly. The proposed changes allow the user to still specify the `--identity-file` option if needed, but makes it optional for the sake of those who don't need it. We should continue to support the `-i` flag, and use the `-i` flag in our examples (because most people are using a key generated by AWS and saved as a `.pem` file somewhere in their `~/Downloads` directory). However, I think it's overkill to make the `-i` flag mandatory, and just increases the amount of redundant boilerplate that the user has to type out every single time.

jey:
No worries about the delay. I'm personally in favor of merging this for 0.8.0, but I'm happy to go with whatever the overall consensus is.

> We've never had someone request this before

True, but it may be that people don't know their private key is being used in this way. It's not the way people would expect things to work by default, and it takes a fairly close read of the operations being performed to notice the `scp` of the keyfile.

jey:
Jenkins, retest this please.

ec2/spark_ec2.py
```diff
@@ -536,18 +541,33 @@ def deploy_files(conn, root_dir, opts, master_nodes, slave_nodes, modules):
...
+  if isinstance(parts, str):
+    return parts
+  else:
+    return ' '.join(map(pipes.quote, parts))
+
+
+def ssh_args(opts):
+  parts = ['-o', 'StrictHostKeyChecking=no']
+  if opts.identity_file is not None:
+    parts += ['-i', opts.identity_file]
```
jey, line 568: This is where the keyfile is passed to `ssh`, conditional on the `--identity-file` flag being present.

shivaram, line 568: Hmm - I would still recommend making `-i` mandatory. As you mention above most of our users download the `.pem` file and it is in `~/Downloads` or something like that. Now if they fail to give a `-i` flag, things will fail with some cryptic error message that might be hard to debug. The trade-off of letting power-users (or users who use keys from `~/.ssh`) not provide this flag seems less important to me.

jey, line 568: Hi Shivaram, I agree that's a valid concern, but I also believe we can provide a good experience for both novices and experts at the expense of neither. I have added commit e86d1d4 which provides a detailed error message with usage instructions in case of connection problems with `ssh`.

rxin:
Thanks, Jey.

PR #671, opened 2013-07-04
Add metrics system for Spark

jerryshao:
This patch creates a metrics system for Spark based on Codahale's metrics library. There are some design considerations:
- Same configure format with Hadoop metrics2 configuration.
- Pluggable metric source and sink, also user defined source and sink can be configured and added.
  - JMX sink is enabled by default.
  - Default metrics source are:
    - Master, Worker (standalone mode)
    - Client Driver (DAGScheduler, BlockManager)
    - JVM instrumentation
- Less intrusion to original code.

We've already monitored by Nagios using this patch internally, but some more metrics are needed. We put this up for review and will continue to improve it. Any comments would be appreciated :-)

Thanks
@xiajunluan , Jerry Shao


mateiz:
Very cool! I'll look over it in the next few days.

mateiz:
Jenkins, this is ok to test

xiajunluan:
@mateiz and @pwendell , I have filed new feature request on https://spark-project.atlassian.net/browse/SPARK-788, and we implemented this feature almost based on discussion thread https://groups.google.com/forum/#!topic/spark-developers/lskhD_-61pI

core/src/main/scala/spark/metrics/sink/ConsoleSink.scala
```diff
@@ -0,0 +1,40 @@
...
+
+import com.codahale.metrics.{ConsoleReporter, MetricRegistry}
+
+import java.util.Properties
+import java.util.concurrent.TimeUnit
+
+import spark.metrics.MetricsSystem
+
+class ConsoleSink(val property: Properties, val registry: MetricRegistry) extends Sink {
+  val CONSOLE_DEFAULT_PERIOD = "10"
```
pwendell: Could you just make this an integer? I slightly prefer that style because then all default params are typechecked by the compiler.

core/src/main/scala/spark/metrics/MetricsSystem.scala
```diff
@@ -0,0 +1,92 @@
...
+        case e: Exception => logError("sink class " + classPath + " cannot be instantialized", e)
+      }
+    }
+  }
+}
+
+private[spark] object MetricsSystem {
+  val SINK_REGEX = "^sink\\.(.+)\\.(.+)".r
+  val SOURCE_REGEX = "^source\\.(.+)\\.(.+)".r
+  val timeUnits = Map(
```
pwendell: Is there a reason you don't just re-use the string enumerations already in the TimeUnit class? You wouldn't need this at all if instead you just created time units from strings like this:

```
TimeUnit.valueOf(myString.toUpperCase)
```

You would have to change to the plural representation (e.g. "seconds" not "second") but it would be good to just be consistent with how TimeUnit already does this.

core/src/main/scala/spark/metrics/MetricsSystem.scala
```diff
@@ -0,0 +1,92 @@
...
+      "illisecond" -> TimeUnit.MILLISECONDS,
```
pwendell: Typo: "illiseconds"

core/src/main/scala/spark/metrics/MetricsSystem.scala
```diff
@@ -0,0 +1,92 @@
...
+import java.util.Properties
+import java.util.concurrent.TimeUnit
+
+import scala.collection.mutable
+
+import spark.Logging
+import spark.metrics.sink.Sink
+import spark.metrics.source.Source
+
+private[spark] class MetricsSystem private (val instance: String) extends Logging {
```
pwendell, line 48: It would be good to add a docstring to this class since it's the main entry point for the entire metrics system. It should succinctly explain the overall architecture of the metrics system (i.e. sources and sinks are specified in config files, exactly what an instance corresponds to, etc).

core/src/main/scala/spark/metrics/MetricsSystem.scala
```diff
@@ -0,0 +1,92 @@
...
+    sinks.foreach(_.start)
+  }
+
+  def stop() {
+    sinks.foreach(_.stop)
+  }
+
+  def registerSource(source: Source) {
+    sources += source
+    registry.register(source.sourceName, source.metricRegistry)
```
pwendell: Just so I understand, what is the benefit of having a unique MetricsRegistry per-source? What isolation does this provide vs. having a single registry for the entire MetricsSystem?

core/src/main/scala/spark/metrics/MetricsSystem.scala
```diff
@@ -0,0 +1,92 @@
...
+import scala.collection.mutable
+
+import spark.Logging
+import spark.metrics.sink.Sink
+import spark.metrics.source.Source
+
+private[spark] class MetricsSystem private (val instance: String) extends Logging {
+  initLogging()
+
+  val confFile = System.getProperty("spark.metrics.conf.file", "unsupported")
```
pwendell: Passing a bogus filename here is a somewhat brittle way of handling no metrics file (and if they happen to have a file named "unsupported" it could cause weird failures). Why not instead have `MetricsConfig` take an `Option[String]` as a filename, and if the config option is not set, then pass `None`. Of course, you still need to check whether the file exists.

core/src/main/scala/spark/metrics/MetricsConfig.scala
```diff
@@ -0,0 +1,65 @@
...
+      } finally {
+        fis.close()
+      }
+    }
+
+    propertyCategories = subProperties(properties, INSTANCE_REGEX)
+    if (propertyCategories.contains(DEFAULT_PREFIX)) {
+      import scala.collection.JavaConversions._
+      val defaultProperty = propertyCategories(DEFAULT_PREFIX)
+      for ((inst, prop) <- propertyCategories; p <- defaultProperty
```
pwendell: Break out the lines here rather than using semi-colons. It's required by the style guide and also this code path is very dense.

core/src/main/scala/spark/metrics/sink/CsvSink.scala
```diff
@@ -0,0 +1,49 @@
...
+import java.util.concurrent.TimeUnit
+
+import spark.metrics.MetricsSystem
+
+class CsvSink(val property: Properties, val registry: MetricRegistry) extends Sink {
+  val CSV_KEY_PERIOD = "period"
+  val CSV_KEY_UNIT = "unit"
+  val CSV_KEY_DIR = "directory"
+
+  val CSV_DEFAULT_PERIOD = "10"
```
pwendell: "10" --> 10

core/src/main/scala/spark/scheduler/DAGSchedulerSource.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import com.codahale.metrics.{Gauge,MetricRegistry}
+
+import spark.metrics.source.Source
+
+private[spark] class DAGSchedulerSource(val dagScheduler: DAGScheduler) extends Source {
+  val metricRegistry = new MetricRegistry()
+  val sourceName = "DAGScheduler"
+
+  metricRegistry.register(MetricRegistry.name("stage", "failedStage"), new  Gauge[Int] {
```
pwendell: Extra space here between new and Gauge

core/src/main/scala/spark/storage/BlockManagerSource.scala
```diff
@@ -0,0 +1,35 @@
...
+      val storageStatusList = blockManager.master.getStorageStatus
+      val remainingMem = storageStatusList.map(_.memRemaining).reduce(_+_)
+      remainingMem
+    }
+  })
+
+  metricRegistry.register(MetricRegistry.name("disk", "diskSpaceUsed"), new Gauge[Long] {
+    override def getValue: Long = {
+      val storageStatusList = blockManager.master.getStorageStatus
+      val diskSpaceUsed = storageStatusList.flatMap(_.blocks.values.map(_.diskSize)).reduceOption(_+_).getOrElse(0L)
```
pwendell: This line is over 100 characters, please cut down all lines that are > 100 characters.

core/src/main/scala/spark/metrics/MetricsSystem.scala
pwendell: Also, please add this new option to the Spark documentation.

pwendell:
Hey @jerryshao ,

Thanks for getting this started - this is a great first step! Overall, I like the approach. There are a few important things to address though before this can go in, since this code could be performance sensitive:
1. It's important to verify that all the gauges are thread-safe, since many of them access important shared data structures that could be concurrently modified without synchronization. The "right way" to do this is to synchronize those data-structures or use concurrent versions, but that will have a performance penalty. An alternative is to just ensure that `size` is stored as an integer internally so updates will be atomic. This is somewhat brittle because if the underlying implementation changes we are in trouble, but I think it's the way to go. Maybe @mateiz has more thoughts.
2. We need to prevent people from setting an overly aggressive polling frequency for any of the sources. Otherwise, people may naively throttle the frequency to be very high and it will hurt performance.

Other things that you should add:
1. The semantics of the configuration language are a little vague. For instance, are wildcards supported for sources and sinks, or only for instances? I think it's only the latter, but this isn't specified anywhere. It would be good to add a documentation page about how to use the metrics system, along with a few complete examples of configuration files. 
2. You also should add the new config option to the documentation.
3. Please try to look through and cleanup style where possible (e.g. long lines, extra spaces, etc).

Once you've addressed these things I can take another look. Thanks again for doing this, I'm sure people are excited to see this in Spark!

jerryshao:
Hi @pwendell 

Thanks for your review, I've made changes according to your comments.
1. For documentation of metrics configuration, I've added detailed instructions about how to configure custom metrics system, but I don't know where to add the new config option to the documentation :-).
2. About thread safe issue, I go through the code and trace the internals of `size` in `HashSet` and `HashMap`, seems Scala use integer to store size variable and no other class variable is modified. For `ArrayBuffer`, `size` function is reentrant. I think they're safe without locking. Currently I have no better idea to solve this perfectly.
3. For each source has a unique `MetricsRegistry`, what my consideration is just for simple and less dependent. There's no difference between only on `MetricsRegistry` or per-source.

Thanks
Jerryshao

core/src/main/scala/spark/storage/BlockManagerSource.scala
```diff
@@ -0,0 +1,39 @@
+package spark.storage
+
+import com.codahale.metrics.{Gauge,MetricRegistry}
+
+import spark.metrics.source.Source
+import spark.storage._
+
+private[spark] class BlockManagerSource(val blockManager: BlockManager) extends Source {
```
pwendell, line 8: For this, could you also add a metric for `memUsed`. I realize it can be determined with (maxMem, remainingMem), but it's still nice for people to see.

core/src/main/scala/spark/metrics/sink/JmxSink.scala
```diff
@@ -0,0 +1,18 @@
+package spark.metrics.sink
+
+import com.codahale.metrics.{JmxReporter, MetricRegistry}
+
+import java.util.Properties
+
+class JmxSink(val property: Properties, val registry: MetricRegistry) extends Sink {
```
pwendell, line 7: I'm wondering, does this do it's own polling? Or does it just calculate the metric on-demand when the MXBean is accessed?

Also, on the Metrics website, there is a fairly large warning saying not to use this in production:
http://metrics.codahale.com/manual/core/#jmx

Unless we have a firm understanding of what this does internally, it might be good to not enable it by default. Do you know what that warning is about?

pwendell:
Hey @jerryshao ,

Thanks for the update! This addressed all of the issues from the previous review. There are a few minor comments inline and then I can pass this onto Matei to look over and/or merge.

One bigger, thing. I'd actually prefer to have no sinks enabled by default initially. For one thing, the `JMXReporter` seems to be explicitly discouraged from use in production. Second, I'd actually like to add a `MetricsServlet` to our existing Jetty UI as the initial "built-in" way to view metrics. BTW - if you are interested in doing this in a subsequent PR, let me know! It would be useful. For now thought lets just have these disabled by default unless users want to start tracking them.

So there are just a few small things and otherwise I think this looks good. Thanks again for taking the time to do this!
- Patrick

core/src/main/scala/spark/metrics/MetricsSystem.scala
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  def registerSinks() {
+    val instConfig = metricsConfig.getInstance(instance)
+    val sinkConfigs = metricsConfig.subProperties(instConfig, MetricsSystem.SINK_REGEX)
+
+    sinkConfigs.foreach { kv =>
+      val classPath = kv._2.getProperty("class")
+      try {
+        val sink = Class.forName(classPath).getConstructor(classOf[Properties], classOf[MetricRegistry])
```
pwendell: I think this line is > 100 characters.

core/src/main/scala/spark/deploy/worker/WorkerSource.scala
```diff
@@ -0,0 +1,34 @@
...
+  metricRegistry.register(MetricRegistry.name("mem_used", "MBytes"), new Gauge[Int] {
+    override def getValue: Int = worker.memoryUsed
+  })
+
+  // Gauge for cores free of this worker
+  metricRegistry.register(MetricRegistry.name("core_free", "number"), new Gauge[Int] {
+    override def getValue: Int = worker.coresFree
+  })
+
+  // Gauge for memory used of this worker
```
anfeng: Type in comment. It should be "memory free".

anfeng:
Comparing with Spark UI, your implementation of Master and Worker source seems to expose a subset of metrics only. Can you document all metrics that you intend to support, and ensure that API expose major master/worker metrics?

You don't have metrics related to applications. Any plan to add them? We could do a separated pull request if prefer.

pwendell:
Hey @afeng. Adding per-job metrics using this CodaHale metrics library might be trickier. I'd prefer to isolate that to a future pull request. The reason is that this library is typically used to report values on a fairly constant set of metrics. If we wanted to add, say, a bunch of per-job metrics in this as well, we'll need to figure out how to garbage collect these metrics over time and what scalability limitations this has in terms of the number of metrics it can simultaneously track. These issues are nontrivial and require separate design considerations.

mateiz:
I think @anfeng was actually saying to let applications register custom metrics, at least as far as I understood, and also to make sure that ExecutorSource, BlockManagerSource, etc put in all the metrics you get through the web UI. It wasn't to do with distinguishing "jobs" within the same application. How hard is it for apps to add a custom metric with Coda Hale's library? Can they just call the library directly or do we need to give them some kind of context?

pwendell:
Hey Matei,

I see - there are two issues being addressed, I was only commenting on the
first one.

(1) Bringing this to parity with what is in the new Web UI.
Matei - I assumed "Spark UI" means all of the per-job metrics we now report
having to do with, e.g. shuffle bytes, response time, etc of a job. I'm
just saying that there is some other stuff to think about before adding
these, because we have to deal with, e.g. removing metrics over time in the
same way that we garbage collect stages and RDD's in the existing UI's. I'd
like to add that stuff, but I think this patch is still useful as-is and it
would be in scope for a subsequent PR to address this.

Are there other things I'm missing here? What UI metrics are you referring
to if not these?

(2) Allowing custom per-application metrics:
The CodaHale library is designed for capturing metrics in a single JVM? If
application-specific metrics means having hadoop-like counters where you
can update things from within tasks, then that requires an extra
abstraction layer where we collect those things ourselves, do some
aggregation, then register them with metrics context at the driver. This is
what I envisioned as the next step, but requires some design discussion.

jerryshao:
Hi @pwendell @anfeng, Thanks very much for your reviewing, I've updated the code.

I think there are several reasons why I build this metrics system.
- Hadoop exists two layers monitoring the whole system. 
  - One is metrics system, which is aim for system admin who cares about the whole level of cluster, like worker state, free slots, total jobs, not the condition of one job. 
  - The other is counter (with built-in or user custom), which is used to monitor the condition of one job and aim for developers. 

This Spark metrics system is designed for the first purpose, so detailed application specific metrics is not added to the metrics system in my consideration.
- My intention is to build a metrics system and add some common metrics to this system. I think different people have different requirement of monitoring metrics, I cannot cover the whole stuff, My consideration is to add some common metrics and improve it according to feedback in future, in the other way customized metrics source can be added through configuration.
- Metrics system is different from Spark UI that different metric sink can be added to the system, like Ganglia, Nagios, what is easy and expressive in Spark UI may seem to be odd in Ganglia, so I think what metrics system collected may be different from Spark UI.

Thanks for all your suggestions, I will keep improving this metrics system like adding sevelet as @pwendell mentioned previously.

jerryshao

mateiz:
@jerryshao that makes sense about the custom metrics -- we can add them later. Just FYI, I'm going to performance-test the current patch and merge it in if everything is okay, so we can do the servlet in another pull request later. I'll let you know if I see any problems.

mateiz:
(The reason for this by the way is that folks at Yahoo! want to start working on some new metrics so they want a framework to start from.)

core/src/main/scala/spark/metrics/MetricsConfig.scala
```diff
@@ -0,0 +1,75 @@
...
+  val properties = new Properties()
+  val DEFAULT_PREFIX = "*"
+  val INSTANCE_REGEX = "^(\\*|[a-zA-Z]+)\\.(.+)".r
+  var propertyCategories: mutable.HashMap[String, Properties] = null
+
+  private def setDefaultProperties(prop: Properties) {
+    // empty function, any default property can be set here
+  }
+
+  def initilize() {
```
mateiz: Typo: this should be called initialize()

core/src/main/scala/spark/metrics/MetricsConfig.scala
```diff
@@ -0,0 +1,75 @@
...
+      if (confFile.exists()) {
+        var fis: FileInputStream = null
+        try {
+          fis = new FileInputStream(confFile)
+          properties.load(fis)
+        } finally {
+          fis.close()
+        }
+      } else {
+        logError("Metrics configure file " + f + " is not found, using default metrics")
```
mateiz: Here we should throw an exception instead of using the default metrics -- we want the user to notice this

conf/metrics.properties.template
```diff
@@ -0,0 +1,86 @@
...
+#    full name should be specified by class property.
+#    2. Some sinks can specify polling period, like console sink, which is 10 seconds,
+#    it should be attention minimal polling period is 1 seconds, any period
+#    below than 1s is illegal.
+#    3. Wild card property can be overlapped by specific instance property, for
+#    example, *.sink.console.period can be overlapped by master.sink.console.period.
+#    4. A metrics specific configuration
+#    "spark.metrics.conf.file={SPARK_HOME}/conf/metrics.properties" should be
+#    added to Java property using -Dspark.metrics.conf.file=xxx if you want to
+#    customize metrics system.
```
mateiz: Having a metrics.properties.template in our config directory but requiring the user to specify a path to it might be confusing, because people will think that they can just copy metrics.properties.template into metrics.properties and have it be read. How about, if spark.metrics.conf.file isn't set, you use [ClassLoader.findResource()](http://docs.oracle.com/javase/6/docs/api/java/lang/ClassLoader.html#findResource%28java.lang.String%29) to search for "/metrics.properties" on the classpath, and if it exists, use that?

Doing this will also help us when people want to use metrics on a cluster. Otherwise, they'd have to make sure that the file referenced by `spark.metrics.conf.file` is available on all nodes.

core/src/main/scala/spark/scheduler/DAGSchedulerSource.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  metricRegistry.register(MetricRegistry.name("stage", "waitingStage"), new Gauge[Int] {
+    override def getValue: Int = dagScheduler.waiting.size
+  })
+
+  metricRegistry.register(MetricRegistry.name("job", "allJobs"), new Gauge[Int] {
+    override def getValue: Int = dagScheduler.nextRunId.get()
+  })
+
+  metricRegistry.register(MetricRegistry.name("job", "ActiveJobs"), new Gauge[Int] {
```
mateiz: rename this to activeJobs

core/src/main/scala/spark/scheduler/DAGSchedulerSource.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import com.codahale.metrics.{Gauge,MetricRegistry}
+
+import spark.metrics.source.Source
+
+private[spark] class DAGSchedulerSource(val dagScheduler: DAGScheduler) extends Source {
+  val metricRegistry = new MetricRegistry()
+  val sourceName = "DAGScheduler"
+
+  metricRegistry.register(MetricRegistry.name("stage", "failedStage"), new Gauge[Int] {
```
mateiz: rename these stage metrics to failedStages, runningStages, waitingStages to be consistent with other names that are plural

conf/metrics.properties.template
mateiz: Also, let's maybe rename the property to just `spark.metrics.conf`

mateiz:
Hey Jerry, so I tried this out on a 20-node cluster with 1-second polling interval and two different sinks (console and CSV), and it seems to perform fine. I made a few small comments on the code though -- the biggest one is finding metrics.properties on the classpath if no spark.metrics.conf is set. Do you mind making these changes? Then we can merge this pull request and do further changes in other PRs.

jerryshao:
Hi @mateiz Thanks very much for review, I strongly agree with your suggestion to change configuration file searching way, I've already changed the code. Also I changed other small issues according to your comments.

Thanks
Jerry

core/src/main/scala/spark/metrics/MetricsConfig.scala
```diff
@@ -0,0 +1,88 @@
...
+    // If spark.metrics.conf is not set, try to get file in class path
+    val filePath = configFile.getOrElse {
+      Option(getClass.getClassLoader.getResource(METRICS_CONF)) match {
+        case Some(r) => r.getFile
+        case None => ""
+      }
+    }
+
+    try {
+      val confFile = new File(filePath)
```
mateiz: Shouldn't we do this only if `filePath != ""`? Otherwise we'll see that it doesn't exist and throw an exception below. You might want to format this so that the whole try block only happens if configFile is set or findResource returns a resource. Let me know if I'm missing something.. basically I'm worried about what happens if you don't set spark.metrics.conf and don't create a conf/metrics.properties either.

core/src/main/scala/spark/metrics/MetricsConfig.scala
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  def initialize() {
+    //Add default properties in case there's no properties file
+    setDefaultProperties(properties)
+
+    // If spark.metrics.conf is not set, try to get file in class path
+    val filePath = configFile.getOrElse {
+      Option(getClass.getClassLoader.getResource(METRICS_CONF)) match {
+        case Some(r) => r.getFile
```
mateiz: Here ideally instead of doing getFile you would just open the resource with `getResourceAsStream`. That way it will work if `metrics.properties` is inside a JAR as well. You'll have to refactor the code a bit to work with an InputStream instead of a File.

mateiz:
Made a couple more comments on the resource stuff.. other than that, it looks good!

jerryshao:
Hi @mateiz , I just complicated it, I'm now refactoring the code to make it simple.  BTW, if spark.metrics.conf is not set and metrics.properties is not in the classpath, metrics system will start with default property, currently default property is empty, so no sink is added to metrics system, and nothing flush out.

mateiz:
Okay, let me know when you finish refactoring.

jerryshao:
I've already done and pushed to Github, you can review the code :-)

mateiz:
Awesome, thanks! I've merged this in. We can make further additions based on this patch.

PR #673, opened 2013-07-04
Add config template file for fair scheduler feature

xiajunluan:


mateiz:
Thanks for adding this!

PR #675, opened 2013-07-04
Building spark assembly for further consumption of the Spark project with a deployed cluster

ghost:
Introducing new module 'dist' that allows to create unmodified set of the dependencies for spark artifacts.
The assembly doesn't include any of the hadoop transitives, as they are expected to be available from a deployed cluster.
This assembly should ease the further packaging of the project in Bigtop and alike.


ghost:
Rebased the commit on top of the current master so it stays up to date.

velvia:
You might want to check out https://github.com/mesos/spark/pull/662 .....  I took a slightly different approach to creating a binary dist/ distribution.

mateiz:
Yeah sorry Cos but how much of this is needed after #662? We can also modify that one if necessary. I guess one difference is that you're excluding the Hadoop dependencies here.

ghost:
Matei, 

Lemme look closer at #662 and get back to you here. From what I see there at the first glance is that Hadoop's libs are getting repackaged and that will force the integrators to rebuild Spark for every slight variance of the Hadoop, instead of simply using what's is set up on a cluster already.

mateiz:
So how would Spark find the Hadoop libraries with this patch -- is there a default place where they get installed under BigTop?

ghost:
Bigtop uses standard linux path layout for the libs of Hadoop components, e.g.
- /usr/lib/hadoop/[lib]
- /usr/lib/hadoop-hdfs/[lib]
- /usr/lib/hadoop-yarn/[lib]
- /usr/lib/hadoop-mapreduce/[lib]
  [lib] directories contain all the dependencies that the component require, which the top level component directory holds only the component jars, i.e hadoop-hdfs, hadoop-common, etc.

Essentially, this is the job of the package to make sure that all required libs are in the classpath before the Spark's daemons are kicked-off. The latest patch for https://issues.apache.org/jira/browse/BIGTOP-715 is doing just that.

ghost:
Here's my reporting back.
- make-distribution.sh didn't work for me because of 
  Error extracting zip entry 'scala/tools/nsc/ast/parser/Parsers$Parser$PatternContextSensitive$$anonfun$scala$tools$nsc$ast$parser$Parsers$Parser$PatternContextSensitive$$makeExistentialTypeTree$1.class' to '/home/cos/work/spark/repl/target/scala-2.9.3/cache/repl/global/assembly/0130126867e25546e17599aaf67f53e204460e13/scala/tools/nsc/ast/parser/Parsers$Parser$PatternContextSensitive$$anonfun$scala$tools$nsc$ast$parser$Parsers$Parser$PatternContextSensitive$$makeExistentialTypeTree$1.class': java.io.FileNotFoundException: /home/cos/work/spark/repl/target/scala-2.9.3/cache/repl/global/assembly/0130126867e25546e17599aaf67f53e204460e13/scala/tools/nsc/ast/parser/Parsers$Parser$PatternContextSensitive$$anonfun$scala$tools$nsc$ast$parser$Parsers$Parser$PatternContextSensitive$$makeExistentialTypeTree$1.class (File name too long
- secondly, as far as I see this new distribution functionality still make something similar to far shader's jar file where all the dependencies are slapped together, adding potentials for jar-hell situation. Maven assembly, on the other hand, adds the transitives unaltered, making it easier to manipulate the classpath if needed.

So, answering your original question Matei - the amount of work left hasn't been changed by adding #662 into the master.

velvia:
Cos, when you say "Maven assembly adds transitives unaltered", what does
that mean?  Does it add the original jars within the assembly jar?   If so,
how does the jars get loaded properly?

The above is what "sbt-onejar" does, which we can possibly try, but I
believe any assemblies that do jar inside jar requires some classloader to
load properly.

If maven assembly (hard to tell from its documentation) expands jars inside
the assembly jar itself (ie all the *class files are flattened) then it's
hard not to merge conflicts.

On Fri, Jul 19, 2013 at 3:20 PM, Cos notifications@github.com wrote:

> Here's my reporting back.
> - make-distribution.sh didn't work for me because of Error extracting
>   zip entry
>   'scala/tools/nsc/ast/parser/Parsers$Parser$PatternContextSensitive$$anonfun$scala$tools$nsc$ast$parser$Parsers$Parser$PatternContextSensitive$$makeExistentialTypeTree$1.class'
>   to
>   '/home/cos/work/spark/repl/target/scala-2.9.3/cache/repl/global/assembly/0130126867e25546e17599aaf67f53e204460e13/scala/tools/nsc/ast/parser/Parsers$Parser$PatternContextSensitive$$anonfun$scala$tools$nsc$ast$parser$Parsers$Parser$PatternContextSensitive$$makeExistentialTypeTree$1.class':
>   java.io.FileNotFoundException:
>   /home/cos/work/spark/repl/target/scala-2.9.3/cache/repl/global/assembly/0130126867e25546e17599aaf67f53e204460e13/scala/tools/nsc/ast/parser/Parsers$Parser$PatternContextSensitive$$anonfun$scala$tools$nsc$ast$parser$Parsers$Parser$PatternContextSensitive$$makeExistentialTypeTree$1.class
>   (File name too long
> - secondly, as far as I see this new distribution functionality still
>   make something similar to far shader's jar file where all the dependencies
>   are slapped together, adding potentials for jar-hell situation. Maven
>   assembly, on the other hand, adds the transitives unaltered, making it
>   easier to manipulate the classpath if needed.
> 
> So, answering your original question Matei - the amount of work left
> hasn't been changed by adding #662https://github.com/mesos/spark/issues/662into the master.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/mesos/spark/pull/675#issuecomment-21281006
> .

## 

ghost:
"adds transitives unaltered" means that the jars are simply - depending on the format of assembly - either packaged into tar.gz or zip, or copied over into a separate directory. What you refer as 'sbt-onejar' done in Maven with predefined 'jar-with-dependencies' assembly profile. 

Maven assembly has a control over how it flattens class files or leaves them inside of their jar. There's a configuration option "<unpack>" for every dependency set. Just take a look into this commit here - it is a couple of dozen of lines really.

mateiz:
Okay, makes sense. One problem with this though is that the `dist` directory will also be used by `make_distribution.sh`. Do you mind either editing that to use another directory, or renaming this subproject?

Also, am I right that the output of this "dist" module won't contain the run scripts? Do you want to eventually add that too (possibly replacing the functionality of `make_distribution.sh` or is the goal of this just to facilitate integration into BigTop?

ghost:
Matei, thanks for the review. I will include all run scripts into the assembly, but they need a separate rework as they have hard-coded assumptions about the development workspace structure, e.g. core/target directory, etc. I think it can be addressed in a consequent improvements. 

Meanwhile, the latest version with all comments addressed is pushed for your attention.

mateiz:
Alright, I'm going to merge this in then, but one issue is that we should try to pick between this and `make-distribution.sh` for 0.8, or make them somehow use each other. I'm going to look at it more later to see whether there's a good way to do it. Basically, if you could create something that returns the same kind of "dist" file as make-distribution.sh, but with the better dependency resolution of Maven, that would be ideal.

mateiz:
Jenkins, this is ok to test

ghost:
Matei, this makes sense. I will start a conversation with the Evan to converge the two more closely together. Thanks for the review.

mateiz:
Alright, thanks. I've merged this in now.

PR #676, opened 2013-07-04
Use standard ASF published avro module instead of a proprietory built one

ghost:
We can safely use non-proprietary build of Avro artifact


mateiz:
Looks good, thanks.

PR #677, opened 2013-07-05
Clean StageToInfos periodically when spark.cleaner.ttl is enabled

jerryshao:
StageToInfos in DAGScheduler should be cleaned periodically when running Spark Streaming or spark.cleaner.ttl is enabled, otherwise this structure will be accumulated and become very large after a long period.


mateiz:
Jenkins, this is ok to test

mateiz:
Thanks for the fix! I've tested this manually.

PR #678, opened 2013-07-05
Start of ML package

mateiz:


ryanlecompte:
Matei, this is super exciting! Will definitely review over the weekend. Great work.

markhamstra:
Woohoo, the beginnings of a standard ML library!  Certainly a welcome addition, but I do have a meta-question: Are we certain that we want the standard library to be part of Spark proper and not be a separate project?  I guess I'm asking whether we expect the library to be tightly linked to particular versions of Spark or whether the benefits of allowing the library to develop in a separate project outweigh the costs of having to deal with another set of version dependencies?

mateiz:
This is an effort with quite a few people in the lab behind it (the MLbase project) -- expect to see more soon. We decided we do want to make it a standard library because it will ensure that it stays high-quality and it gets tested and optimized along with changes in Spark. There will also be other, more experimental stuff outside. This is of course a bit of an experiment and might change in the future, but I think having good standard libraries is one thing we can do better than Hadoop in the long-term. Let's see how it goes!

mateiz:
(By the way, a related thing is that we do want to move to slightly more frequent releases, which will help get this out in a timely fashion. We're making progress towards that with more automated testing infrastructure.)

samklr:
Great job guys. I saw Ameet Talwalkar talk about mlbase, and I can't wait to see how it goes. This with the algorithms on top of GraphX will drive a lot of projects into spark. And I noticed that the guys on the mahout mailing list are talking about some adaptation to make mahout matrix and SSVD available on Spark. I think Nick Pentreath did some work about it.
Good days ahead. Thanks you guys for your hard work.

adelbertc:
This is awesome! Keep up the great work, very excited looking forward to see where this goes.

mllib/src/main/scala/spark/ml/clustering/KMeans.scala
```diff
@@ -0,0 +1,319 @@
...
+    var initializationSteps: Int,
+    var epsilon: Double)
+  extends Serializable with Logging
+{
+  private type ClusterCenters = Array[Array[Double]]
+
+  def this() = this(2, 20, 1, KMeans.K_MEANS_PARALLEL, 5, 1e-4)
+
+  /** Set the number of clusters to create (k). Default: 2. */
+  def setK(k: Int): KMeans = {
```
ryanlecompte: Just a minor nit: If you're going for a builder pattern here, usually the methods are named `withXYZ`, so it would be `withK`, `withMaxIterations`, etc. Usually setters in Scala take the form `def maxIterations_=(maxIterations: Int) { _maxIterations = maxIterations }` and don't return `this`.

mllib/src/main/scala/spark/ml/clustering/KMeans.scala
```diff
@@ -0,0 +1,319 @@
...
+  /**
+   * Train a K-means model on the given set of points; `data` should be cached for high
+   * performance, because this is an iterative algorithm.
+   */
+  def train(data: RDD[Array[Double]]): KMeansModel = {
+    // TODO: check whether data is persistent; this needs RDD.storageLevel to be publicly readable
+
+    val sc = data.sparkContext
+
+    var centers = if (initializationMode == KMeans.RANDOM) {
```
ryanlecompte: I believe `centers` can be a val here instead of a `var`.

mllib/src/main/scala/spark/ml/clustering/KMeans.scala
```diff
@@ -0,0 +1,319 @@
...
+      // Find the sum and count of points mapping to each center
+      val totalContribs = data.mapPartitions { points =>
+        val runs = activeCenters.length
+        val k = activeCenters(0).length
+        val dims = activeCenters(0)(0).length
+
+        val sums = Array.fill(runs, k)(new DoubleMatrix(dims))
+        val counts = Array.fill(runs, k)(0L)
+
+        for (point <- points) {
```
ryanlecompte: This could be written as a single for expression instead of two nested ones:

``` scala
  for {
    point <- points
    (centers, runIndex) <- activeCenters.zipWithIndex
  } // ...
```

mllib/src/main/scala/spark/ml/clustering/KMeans.scala
```diff
@@ -0,0 +1,319 @@
...
+  }
+
+  def train(data: RDD[Array[Double]], k: Int, maxIterations: Int): KMeansModel = {
+    train(data, k, maxIterations, 1, K_MEANS_PARALLEL)
+  }
+
+  /**
+   * Return the index of the closest point in `centers` to `point`, as well as its distance.
+   */
+  private[mllib] def findClosest(centers: Array[Array[Double]], point: Array[Double])
```
ryanlecompte: This could be written a bit more idiomatically as:

``` scala
  def findClosest(centers: Array[Array[Double]], point: Array[Double]): (Int, Double) = {
    val (bestDistance, bestIndex) = centers.zipWithIndex.minBy { case (points, _) =>
      MLUtils.squaredDistance(point, points)
    }
    (bestIndex, bestDistance)
  }
```

mllib/src/main/scala/spark/ml/clustering/KMeans.scala
```diff
@@ -0,0 +1,319 @@
...
+        bestIndex = i
+      }
+    }
+    (bestIndex, bestDistance)
+  }
+
+  /**
+   * Return the K-means cost of a given point against the given cluster centers.
+   */
+  private[mllib] def pointCost(centers: Array[Array[Double]], point: Array[Double]): Double = {
```
ryanlecompte: Could also be written a bit more idiomatically following a similar pattern as previous comment for `findClosest()`.

mllib/src/main/scala/spark/ml/recommendation/ALS.scala
```diff
@@ -0,0 +1,387 @@
...
+class ALS private (var numBlocks: Int, var rank: Int, var iterations: Int, var lambda: Double)
+  extends Serializable
+{
+  def this() = this(-1, 10, 10, 0.01)
+
+  /**
+   * Set the number of blocks to parallelize the computation into; pass -1 for an auto-configured
+   * number of blocks. Default: -1.
+   */
+  def setBlocks(numBlocks: Int): ALS = {
```
ryanlecompte: Same comment regarding the pattern of naming builder methods as `withBlocks`, `witchRank`, etc vs. `setXYZ`.

mateiz: Ryan, this is probably not super clear from the commit but the purpose of the builder API is to be available from Java. Do people commonly use `with` there too?

mllib/src/main/scala/spark/ml/recommendation/ALS.scala
```diff
@@ -0,0 +1,387 @@
...
+   * the users (or (blockId, (p, u, r)) for the products). We create these simultaneously to avoid
+   * having to shuffle the (blockId, (u, p, r)) RDD twice, or to cache it.
+   */
+  private def makeLinkRDDs(numBlocks: Int, ratings: RDD[(Int, (Int, Int, Double))])
+    : (RDD[(Int, InLinkBlock)], RDD[(Int, OutLinkBlock)]) =
+  {
+    val grouped = ratings.partitionBy(new HashPartitioner(numBlocks))
+    val links = grouped.mapPartitionsWithIndex((blockId, elements) => {
+      val ratings = elements.map(_._2).toArray
+      Iterator((blockId, (makeInLinkBlock(numBlocks, ratings), makeOutLinkBlock(numBlocks, ratings))))
```
ryanlecompte: If you just have a single element, it can usually be more performant to use Scala's `Iterator.single` vs. `Iterator(...)`

mllib/src/main/scala/spark/ml/recommendation/ALS.scala
ryanlecompte: Ah, that's a good point Matei. I think this would work fine in Java too. In
Java I mostly see a mixture of "setXYZ" or just "xyz" with a final
"build()" at the end. I guess it's not such a big deal in the end, it just
may look a bit Java-ish to Scala developers.

mateiz: In Scala, I think the right way might be to add a factory method with default arguments as well, on the singleton object (KMeans.apply). That will allow us to have a case class-like pattern. For Java, the chained `set()` naming pattern is used in both Avro and Protobuf, so I think we should go with that.

By the way, unfortunately I think it's impossible to add default arguments to the constructor and still call an argument-less version from Java. Otherwise that would've been the nicest.

mllib/src/main/scala/spark/ml/regression/RidgeRegression.scala
```diff
@@ -0,0 +1,183 @@
...
+          val H_ii = x.transpose().mmul(invXtX).mmul(x).get(0)
+          val residual = (y - yhat) / (1.0 - H_ii)
+          residual * residual
+      }.reduce(_ + _) / nexamples
+
+      (lambda, cverror, w)
+    }
+
+    // Binary search for the best assignment to lambda.
+    def binSearch(low: Double, high: Double): List[(Double, Double, DoubleMatrix)] = {
```
ryanlecompte: We should annotate this with `scala.annotation.tailrec`

mllib/src/main/scala/spark/ml/recommendation/ALS.scala
ryanlecompte: Yes, you're definitely right. The set() pattern sounds good to me. I had
forgotten about our Java users. :-)

ryanlecompte:
Just finished a quick pass at this pull request, mostly just minor style-related comments. Looks great!

mllib/src/main/scala/spark/ml/clustering/KMeans.scala
mateiz: This is true but we wanted to avoid the object allocation done by zipWithIndex

ryanlecompte: Makes sense! That extra collection can be costly.

rxin: Can create a "view" to do that to skip the extra collection - but I think the current way is fine as is.

mllib/src/main/scala/spark/ml/regression/RidgeRegression.scala
mateiz: Unfortunately this couldn't be tail-recursive because we are constructing a list from the front.

mllib/src/main/scala/spark/ml/clustering/KMeans.scala
ryanlecompte: Agreed.

mllib/src/test/scala/spark/ml/clustering/KMeansSuite.scala
```diff
@@ -0,0 +1,150 @@
...
+    model = KMeans.train(data, k=1, maxIterations=5)
+    assertSetsEqual(model.clusterCenters, Array(Array(1.0, 3.0, 4.0)))
+
+    model = KMeans.train(data, k=1, maxIterations=1, runs=5)
+    assertSetsEqual(model.clusterCenters, Array(Array(1.0, 3.0, 4.0)))
+
+    model = KMeans.train(data, k=1, maxIterations=1, runs=5)
+    assertSetsEqual(model.clusterCenters, Array(Array(1.0, 3.0, 4.0)))
+
+    model = KMeans.train(data, k=1, maxIterations=1, runs=1, initializationMode="random")
```
markhamstra: ... initializationMode = KMeans.RANDOM ?

Similarly in next test and for K_MEANS_PARALLEL.

Incidentally, I really don't like using raw strings for these, but I don't know the best way to expose a typesafe enumeration that is easy to use across Scala, Java and Python.

mateiz:
Thanks for your comments Ryan; I've addressed most of them except a couple that couldn't be done without some cost.

mllib/src/test/scala/spark/ml/clustering/KMeansSuite.scala
mateiz: Good catch, fixed that. The Python API is exactly why we allowed strings too.

ryanlecompte:
Sweet! Thanks, Matei. This is a nice addition to Spark!

shivaram:
Ryan - Thanks for taking a look and thanks Matei for the fixes. I am merging this now as there are other commits planned to add more features and tests. Any comments / suggestions are welcome.

adelbertc:
Hurrah! Today is a good day.

mllib/src/main/scala/spark/ml/clustering/KMeans.scala
ryanlecompte: BTW Reynold,

I just wanted to do a quick micro-benchmark with Thyme (
https://github.com/Ichoran/thyme) to see if using a view would actually
help in this case. It looks like the view slows things down quite a bit and
generates more garbage:

scala> val nums = Array.fill[Double](100000)(scala.util.Random.nextDouble)

scala> th.pbench {\* nums.zipWithIndex.minBy\* { case (n, _) => n } }
Benchmark (2540 calls in 15.48 s)
- Time:    3.032 ms\*   95% CI 2.931 ms - 3.133 ms   (n=20)
  _Garbage: 2.814 ms_   (n=566 sweeps measured)
  res4: (Double, Int) = (1.044753784418262E-5,48030)

scala> th.pbench { _nums.view.zipWithIndex.minBy_ { case (n, _) => n } }
Benchmark (300 calls in 9.499 s)
  _Time:    9.796 ms \*  95% CI 9.229 ms - 10.36 ms   (n=20)
  *Garbage: 21.21 ms_   (n=130 sweeps measured)
res5: (Double, Int) = (1.044753784418262E-5,48030)

That was something I was always curious about.

PR #680, opened 2013-07-06
Fixed major performance bug in Network Receiver

tdas:
The data received with NetworkReceiver.BlockGenerator into an ArrayBuffer which was then copied twice into 2 new ArrayBuffer unnecessarily, before being stored in the BlockManager. This was partially due to the use of BlockManager.put with iterator method, even though the BlockManager.put with ArrayBuffer exists. This bug affects almost all the receivers. I havent tested how much performance improvement in data ingestion rate would one get with this fix, but since its such an obvious fix and I want the fix go into 0.7.3 immediately, I am requesting this to be accepted.


mateiz:
Looks good, thanks!

PR #682, opened 2013-07-06
Fix occasional failure in UI listener.

pwendell:
If a task fails before the metrics are initialized, it remains possible
that the metrics field will be `None`. This patch accounts for that possbility
by keeping metrics as an `Option` at all times.

### DIFF
--- a/core/src/main/scala/spark/ui/jobs/StagePage.scala
+++ b/core/src/main/scala/spark/ui/jobs/StagePage.scala
@@ -52,7 +52,7 @@ private[spark] class StagePage(parent: JobProgressUI) {
       }
       else {
         val serviceTimes = validTasks.map{case (info, metrics, exception) =>
-          metrics.executorRunTime.toDouble}
+          metrics.get.executorRunTime.toDouble}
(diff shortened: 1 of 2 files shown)


core/src/main/scala/spark/ui/jobs/StagePage.scala
```diff
@@ -52,7 +52,7 @@ private[spark] class StagePage(parent: JobProgressUI) {
       }
       else {
         val serviceTimes = validTasks.map{case (info, metrics, exception) =>
-          metrics.executorRunTime.toDouble}
+          metrics.get.executorRunTime.toDouble}
```
ryanlecompte, line 55: Is it safe to always call `metrics.get` in these cases? Could it ever be `None` ?

pwendell, line 55: Yes, all `validTasks` have metrics, that's the definition of `validTasks`...

ryanlecompte, line 55: Sweet!

mateiz:
I've tested this manually and merged it in. Thanks Patrick!

PR #683, opened 2013-07-07
Remove some stack traces from sbt test output

shivaram:
This change gets rid of some of the output printed to stdout while running `sbt test`. 
1. Stack trace of the form

```
13/07/07 11:23:13.827 WARN StaticChannelPipeline: An exception was thrown by an exception handler.
java.util.concurrent.RejectedExecutionException
```

are now sent to SLF4J by initializing the `InternalLoggerFactory` used by StaticChannelPipeline in Netty . http://grepcode.com/file/repo1.maven.org/maven2/org.jboss.netty/netty/3.2.5.Final/org/jboss/netty/channel/StaticChannelPipeline.java#StaticChannelPipeline
1. Minor changes to make `CheckpointSuite` and `DriverSuite` not print anything to stdout as well.

### DIFF
--- a/core/src/test/scala/spark/PipedRDDSuite.scala
+++ b/core/src/test/scala/spark/PipedRDDSuite.scala
@@ -67,7 +67,7 @@ class PipedRDDSuite extends FunSuite with SharedSparkContext {
 
   test("pipe with non-zero exit status") {
     val nums = sc.makeRDD(Array(1, 2, 3, 4), 2)
-    val piped = nums.pipe("cat nonexistent_file")
+    val piped = nums.pipe(Seq("cat nonexistent_file", "2>", "/dev/null"))
     intercept[SparkException] {
       piped.collect()
     }
--- a/core/src/main/scala/spark/SparkEnv.scala
+++ b/core/src/main/scala/spark/SparkEnv.scala
@@ -5,7 +5,6 @@ import serializer.Serializer
 
 import akka.actor.{Actor, ActorRef, Props, ActorSystemImpl, ActorSystem}
 import akka.remote.RemoteActorRefProvider
-import akka.event.{Logging => AkkaLogging}
 
 import spark.broadcast.BroadcastManager
 import spark.storage.BlockManager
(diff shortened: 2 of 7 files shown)


pwendell:
I just "tested" this (ran tests) and it does indeed quiet the logging a great deal. 

I'm having some issues with my environment causing some tests to fail, but this patch isn't causing them.

Looks good to me, pending someone else running the unit tests and verifying they all work.

shivaram:
Thanks Patrick. In my environment I don't see anything extra -- The output from sbt/sbt test is at https://gist.github.com/shivaram/5950949

core/src/test/scala/spark/PipedRDDSuite.scala
```diff
@@ -67,7 +67,7 @@ class PipedRDDSuite extends FunSuite with SharedSparkContext {
 
   test("pipe with non-zero exit status") {
     val nums = sc.makeRDD(Array(1, 2, 3, 4), 2)
-    val piped = nums.pipe("cat nonexistent_file")
+    val piped = nums.pipe(Seq("cat nonexistent_file", "2>", "/dev/null"))
```
mateiz, line 70: I guess this won't work on Windows, but the previous version wouldn't work either.

rxin:
I haven't looked into the change, but does it show an error if the error is legit (e.g. null pointer exception)?

pwendell:
@shivaram hey - is that output stdout only or stdout and stderr. In my env I get a bunch of errors on stderr from Jetty (again, totally unrelated to this PR, but just wondering)

shivaram:
@rxin - Yes. This doesn't affect anything other than netty's logger. It just sets that logger to slf4j instead of java logging. So exceptions printed for errors will still remain the same

@pwendell - Actually its both. The test results were on stdout in my env and stderr just had the warning messages from DriverSuite about hostname. Is there a specific branch / patch that I should use to run Jetty tests or are these just on master ?

pwendell:
@shivaram  - nope I get output on stderr using master. I think it's something specific to my environment... we can take it offline as it really doesn't directly relate to this patch. I'm +1 on merging this!

pwendell:
Jenkins, test this please.

mateiz:
Cool, thanks Shivaram!

PR #684, opened 2013-07-08
Explicitly set class loader for MesosSchedulerDriver callbacks.

charlesreiss:
I think this fixes the issue where people saw ClassNotFoundExceptions from MesosSchedulerBackend.statusUpdate when running Spark master against Mesos. MesosSchedulerBackend callbacks are called from a thread which uses the default ClassLoader rather than inheriting the ClassLoader from the thread from which the scheduler backend is started.

### DIFF
--- a/core/src/main/scala/spark/scheduler/mesos/MesosSchedulerBackend.scala
+++ b/core/src/main/scala/spark/scheduler/mesos/MesosSchedulerBackend.scala
@@ -118,20 +118,28 @@ private[spark] class MesosSchedulerBackend(
     return execArgs
   }
 
-  private def setClassLoader() {
-    // Since native code starts the thread our callbacks run in, it may not correctly
-    // inherit and custom class loaders. Therefore, set the class loader manually.
+  private def setClassLoader(): ClassLoader = {
+    val oldClassLoader = Thread.currentThread.getContextClassLoader
     Thread.currentThread.setContextClassLoader(classLoader)
+    return oldClassLoader
+  }
+
+  private def restoreClassLoader(oldClassLoader: ClassLoader) {
+    Thread.currentThread.setContextClassLoader(oldClassLoader)
   }
 
   override def offerRescinded(d: SchedulerDriver, o: OfferID) {}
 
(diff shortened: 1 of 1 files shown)


mateiz:
I see, interesting. Is there any way to test this? I guess it might be tough except when users are putting their own code on the classpath.

charlesreiss:
I managed to get the ClassNotFoundException (and make it apparently go away with this patch) from 'MASTER=mesos://127.0.0.1:5050 ./spark-shell' (running mesos locally and pointing MESOS_NATIVE_LIBRARY to a trunk libmesos.so).

mridulm:
Changing the classloader in the callbacks, and then letting it escape back to the pool can be unintended sideeffects.
Not to mention, other pools which might be used (and not hitting mesos driver) will 'miss' on the classloader change.
Probably the reason why someone else mentioned a failure in user group after this patch too ...

Either the classloader must be set 'properly' elsewhere before the pools are bootstrapped : so that all threads share the right loader : or, if we are continuing with this hack, at the bare minimum we should reset the classloader back to what it was before exiting the callback. It wont solve the problem as such based on comment above, but atleast wont cause other issues.

charlesreiss:
Because Mesos just uses the JNI call AttachCurrentThread(), the context class loader in the callbacks should always be the bootstrap class loader (this is what I meant by "default class loader", but that was unclear), and it is currently not possible to have Mesos callbacks start with any other class loader. I didn't think about resetting the class loader because I knew the Mesos callback would just DetachCurrentThread(), but we shouldn't depend on that happening going forward, so I added restoring the old class loaders, too.

mridulm:
I have submitted a PR as a hack to attempt to circumvent this issue
entirely - assuming it works, we would not need to do the set/unset stuff.
My worry is about cases where mesos might be used (if not now, then later)
directly/indirectly from other threads which do not necessarily go through
the same codepaths we have patched it for.

Please see https://github.com/mesos/spark/pull/690

mateiz:
Thanks for the fix Charles.

PR #687, opened 2013-07-08
Added "Labeled" to util functions for labeled data

atalwalkar:


shivaram:
Thanks Ameet for the change!

PR #688, opened 2013-07-08
Fixed SPARK-795 with explicit dependencies

markhamstra:
Without explicit dependencies for 2.9.3 versions of scala-library and scalap, packages built against Scala 2.9.2 can pull in 2.9.2 versions of scala-compiler, scala-library, scalap, etc. which get resolved by maven ahead of the 2.9.3 versions.  This PR is sufficient to make sure that lift-json_2.9.2, akka 2.0.3, etc. don't get a chance to bring in these 2.9.2 transitive dependencies.

The specific problem of SPARK-795 was a result of Patrick's new UI code using scala.util.Try, which was first introduced in Scala 2.9.3 -- and obviously could not be found when scala-library 2.9.2 was resolved.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -58,12 +58,10 @@
     <dependency>
       <groupId>org.scala-lang</groupId>
       <artifactId>scalap</artifactId>
-      <version>2.9.3</version>
     </dependency>
     <dependency>
       <groupId>org.scala-lang</groupId>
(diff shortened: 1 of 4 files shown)


andyk:
Jenkins test this please.

mateiz:
Thanks for the fix, Mark!

PR #689, opened 2013-07-09
Bug fix: SPARK-796

BlackNiuza:
[SPARK-796](https://spark-project.atlassian.net/browse/SPARK-796): Jobs are always marked as SUCCEEDED even it's actually failed on Yarn.

### DIFF
--- a/core/src/hadoop2-yarn/scala/spark/deploy/yarn/ApplicationMaster.scala
+++ b/core/src/hadoop2-yarn/scala/spark/deploy/yarn/ApplicationMaster.scala
@@ -135,6 +135,9 @@ class ApplicationMaster(args: ApplicationMasterArguments, conf: Configuration) e
           var mainArgs: Array[String] = new Array[String](args.userArgs.size())
           args.userArgs.copyToArray(mainArgs, 0, args.userArgs.size())
           mainMethod.invoke(null, mainArgs)
+          // some job script has "System.exit(0)" at the end, for example SparkPi, SparkLR
+          // userThread will stop here unless it has uncaught exception thrown out
+          // It need shutdown hook to set SUCCEEDED
           successed = true
         } finally {
           if(successed){
@@ -308,11 +311,8 @@ object ApplicationMaster {
           logInfo("Invoking sc stop from shutdown hook") 
           sc.stop() 
           // best case ...
-          // due to the sparkContext is stopped and ApplicationMaster is down,
-          // the status of registered masters should be set KILLED better than FAILED.
-          // need discussion
           for (master <- applicationMasters) {
-            master.finishApplicationMaster(FinalApplicationStatus.KILLED)
+            master.finishApplicationMaster(FinalApplicationStatus.SUCCEEDED)
           }
         } 
       } )
--- a/core/src/hadoop2-yarn/scala/spark/deploy/yarn/ApplicationMaster.scala
+++ b/core/src/hadoop2-yarn/scala/spark/deploy/yarn/ApplicationMaster.scala
@@ -122,23 +122,26 @@ class ApplicationMaster(args: ApplicationMasterArguments, conf: Configuration) e
       }
     }
   }
-  
+
   private def startUserClass(): Thread  = {
     logInfo("Starting the user JAR in a separate Thread")
     val mainMethod = Class.forName(args.userClass, false, Thread.currentThread.getContextClassLoader)
       .getMethod("main", classOf[Array[String]])
     val t = new Thread {
       override def run() {
+        var successed = false
         try{
           // Copy
           var mainArgs: Array[String] = new Array[String](args.userArgs.size())
           args.userArgs.copyToArray(mainArgs, 0, args.userArgs.size())
           mainMethod.invoke(null, mainArgs)
-          ApplicationMaster.this.finishApplicationMaster(FinalApplicationStatus.SUCCEEDED)
-        } catch {
-          case th: Throwable =>
+          successed = true
+        } finally {
+          if(successed){
+            ApplicationMaster.this.finishApplicationMaster(FinalApplicationStatus.SUCCEEDED)
+          }else{
             ApplicationMaster.this.finishApplicationMaster(FinalApplicationStatus.FAILED)
-            logError("Finish ApplicationMaster with ",th)
+          }
         }
       }
     }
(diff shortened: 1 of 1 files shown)


BlackNiuza:
status of jobs after this code change
qq20130709-1

core/src/hadoop2-yarn/scala/spark/deploy/yarn/ApplicationMaster.scala
```diff
@@ -235,14 +240,22 @@ class ApplicationMaster(args: ApplicationMasterArguments, conf: Configuration) e
...
-    resourceManager.finishApplicationMaster(finishReq)
+
+  def finishApplicationMaster(status: FinalApplicationStatus) {
+
+    synchronized {
+      if(isFinished){
+        return
+      }
+      isFinished = true
+
```
mridulm: You can move the rest out of the sync block.

core/src/hadoop2-yarn/scala/spark/deploy/yarn/ApplicationMaster.scala
```diff
@@ -291,7 +304,9 @@ object ApplicationMaster {
           logInfo("Invoking sc stop from shutdown hook") 
           sc.stop() 
           // best case ...
-          for (master <- applicationMasters) master.finishApplicationMaster
+          for (master <- applicationMasters) {
+            master.finishApplicationMaster(FinalApplicationStatus.KILLED)
+          }
```
mridulm, line 316: I will let @tgravescs comment on this more - whether KILLED is the right param for this invocation or not.

core/src/hadoop2-yarn/scala/spark/deploy/yarn/ApplicationMaster.scala
```diff
@@ -131,10 +129,17 @@ class ApplicationMaster(args: ApplicationMasterArguments, conf: Configuration) e
...
+          // Copy
+          var mainArgs: Array[String] = new Array[String](args.userArgs.size())
+          args.userArgs.copyToArray(mainArgs, 0, args.userArgs.size())
+          mainMethod.invoke(null, mainArgs)
+          ApplicationMaster.this.finishApplicationMaster(FinalApplicationStatus.SUCCEEDED)
+        } catch {
+          case th: Throwable =>
+            ApplicationMaster.this.finishApplicationMaster(FinalApplicationStatus.FAILED)
+            logError("Finish ApplicationMaster with ",th)
+        }
```
mridulm, line 148: Unfortunately, exceptions are not the only reason why a job will fail : we left it unsolved to fix it at a later point due to the ambiguity there.
This is definitely a welcome step in the right direction though !

A better approach would be .
a) Set a status variable 'succeeded = false' outside try block.
b) set it to true after main method finishes 'normally.
c) remove the catch block (catching throwable is not a good idea) and replace with finally block - depending on succeeded invoke the appropriate finishApplicationMaster method as has been done here in this PR.

mridulm:
I added a few comments.
Also would like Tom Graves to look at this fix - I think he might already be looking at this.

BlackNiuza:
hi, mrldulm
thank you for your advise, I fix it

andyk:
Jenkins, it is ok to test this.

core/src/hadoop2-yarn/scala/spark/deploy/yarn/ApplicationMaster.scala
BlackNiuza, line 316: Thank you for your comment.  In my opinion, due to the sparkContext is stopped and ApplicationMaster is down,  I think the status should be set KILLED better than FAILED.

tgravescs:
I'm not currently working on fixing this.  I think using KILLED in the shutdownhook is ok for now. Its no worse then saying SUCCEEDED even though there is possibility of showing killed when it was really FAILED. The issue with this though is there is a race condition between when the shutdown hook runs and when finishApplicationMaster gets called after the mainMethod returns. You can test by simply putting a sleep or log message in after the mainMethod.invoke call and you will see it comes out with status KILLED even on successful programs - like SparkPi for example.

What conditions did you use to generate the image you have under "status of jobs after this code change"?  I'm surprised to see the application state in FAILED unless it failed to submit or something in the framework went wrong starting or finishing it.

BlackNiuza:
Thanks for testing my code. I have made a mistake that I forgot to test SparkPi and SparkLR in the example. What I test is my job without System.exit() at the end, so my job's final status is SUCCEEDED and FAILED. 

If job run without any exception, it will reaches System.exit() and stops the world (userThread will not enter finally block) and runs shutdown hook. You're right to set SUCCEEDED there. However, If it throw exception or error in mainMethod,  the finally block can catch it and set FAILED in userThread. After this,  mainThread will reaches System.exit and runs shutdown hook. I think it won't lead to a race condition between when the shutdown hook runs and when finishApplicationMaster gets called after the mainMethod returns.

I add a commit later, I test it on my yarn cluster. Much Thanks!!

tgravescs:
thanks, I'm good with leaving it SUCCEEDED and we can improve more later.

mateiz:
@mridulm and @tgravescs, is this OK to merge or do you have further comments?

core/src/hadoop2-yarn/scala/spark/deploy/yarn/ApplicationMaster.scala
```diff
@@ -235,14 +246,23 @@ class ApplicationMaster(args: ApplicationMasterArguments, conf: Configuration) e
     }
   }
   */
-  
-  def finishApplicationMaster() { 
+
+  def finishApplicationMaster(status: FinalApplicationStatus) {
+
+    synchronized {
+      if(isFinished){
```
mateiz: Add a space before the {

core/src/hadoop2-yarn/scala/spark/deploy/yarn/ApplicationMaster.scala
```diff
@@ -124,17 +122,30 @@ class ApplicationMaster(args: ApplicationMasterArguments, conf: Configuration) e
...
     val mainMethod = Class.forName(args.userClass, false, Thread.currentThread.getContextClassLoader)
       .getMethod("main", classOf[Array[String]])
     val t = new Thread {
       override def run() {
-        // Copy
-        var mainArgs: Array[String] = new Array[String](args.userArgs.size())
-        args.userArgs.copyToArray(mainArgs, 0, args.userArgs.size())
-        mainMethod.invoke(null, mainArgs)
+        var successed = false
+        try{
```
mateiz: Add a a space before the { here too

core/src/hadoop2-yarn/scala/spark/deploy/yarn/ApplicationMaster.scala
```diff
@@ -124,17 +122,30 @@ class ApplicationMaster(args: ApplicationMasterArguments, conf: Configuration) e
...
+          // Copy
+          var mainArgs: Array[String] = new Array[String](args.userArgs.size())
+          args.userArgs.copyToArray(mainArgs, 0, args.userArgs.size())
+          mainMethod.invoke(null, mainArgs)
+          // some job script has "System.exit(0)" at the end, for example SparkPi, SparkLR
+          // userThread will stop here unless it has uncaught exception thrown out
+          // It need shutdown hook to set SUCCEEDED
+          successed = true
+        } finally {
+          if(successed){
```
mateiz: Make this `if (succeeded) {`

mateiz:
Actually I noticed a few small style issues -- basically, put spaces before/after curly braces and for if conditions. Sorry to be nitpicky but it's good to fix these early.

mridulm:
I dont have any further comments.

Thanks
Mridul

mateiz:
Great, thanks for the fix!

PR #691, opened 2013-07-09
Create log pages

karenfeng:
The creation of log pages should fix [SPARK-608](https://spark-project.atlassian.net/browse/SPARK-608).
The following changes have been made:
- Add new "/logPage" path to give visual log exploration via pages
- Move some unit tests to correct package
- Update existing "/log" path to use newer options (offset, byteLength)
  log page


karenfeng:
Just realized that the log page names don't match... Should I go with "stdout-page" or "stdout"?

rxin:
The new web ui @pwendell is building might actually contain this as well.

rxin:
Ah never mind looks like it doesn't.

core/src/main/scala/spark/deploy/worker/ui/WorkerWebUI.scala
```diff
@@ -65,6 +70,62 @@ class WorkerWebUI(val worker: ActorRef, val workDir: File, requestedPort: Option
...
+    val offset = {
+      if (getOffset == null) 0
+      else if (getOffset.toInt < 0) 0
+      else getOffset.toInt
+    }
+    val lineLength = {
+      if (getLineLength == null) 0
+      else getLineLength.toInt
+    }
+    val logText = "<node>" + lines.slice(offset, offset+lineLength).mkString("\n") + "</node>"
```
pwendell: Would it work here to just do:

```
val logText = <node> {lines.slice(offset, offset+lineLength).mkString("\n")} </node>
```

core/src/main/scala/spark/deploy/worker/ui/WorkerWebUI.scala
```diff
@@ -13,6 +13,10 @@ import spark.{Utils, Logging}
 import spark.ui.JettyUtils
 import spark.ui.JettyUtils._
 
+import scala.xml._
```
pwendell: The import style we use is that we group imports from each package together, then we add a newline between each group. We also list the packages alphabetically. This means it would be good to group the scala imports together and move them before the spark ones. Then add the spark.ui.UIUtils import to be in the existing spark group.

core/src/main/scala/spark/deploy/worker/ui/WorkerWebUI.scala
```diff
@@ -65,6 +70,62 @@ class WorkerWebUI(val worker: ActorRef, val workDir: File, requestedPort: Option
     pre + Utils.lastNBytes(path, math.min(numBytes, maxBytes))
   }
 
+  def logPage(request: HttpServletRequest): Seq[scala.xml.Node] = {
+    val appId = request.getParameter("appId")
+    val executorId = request.getParameter("executorId")
+    val logType = request.getParameter("logType")
+    val getOffset = request.getParameter("offset")
+    val getLineLength = request.getParameter("lineLength")
```
pwendell: There is a slightly more scala-centric way to deal with potentially null variables here. You could use this for both offset and lineLength.

```
val getLineLength = Option(request.getParameter("lineLength")).map(_.toInt).getOrElse(0)
```

core/src/main/scala/spark/deploy/worker/ui/WorkerWebUI.scala
```diff
@@ -65,6 +70,62 @@ class WorkerWebUI(val worker: ActorRef, val workDir: File, requestedPort: Option
...
+    val lineLength = {
+      if (getLineLength == null) 0
+      else getLineLength.toInt
+    }
+    val logText = "<node>" + lines.slice(offset, offset+lineLength).mkString("\n") + "</node>"
+    val logXML = XML.loadString(logText)
+    val backButton =
+      if (offset > 0) {
+        if (offset-lineLength < 0) {
+          <a href={"?appId=%s&executorId=%s&logType=%s&offset=0&lineLength=%s".format(appId, executorId, logType, lineLength)}> <button style="float:left">back</button> </a>
```
pwendell: These lines are longer than 100 characters (which we enforce for style purposes). You should do a sweep of the commit and add breaks wherever they exceed 100 characters.

core/src/main/scala/spark/deploy/worker/ui/WorkerWebUI.scala
```diff
@@ -65,6 +70,62 @@ class WorkerWebUI(val worker: ActorRef, val workDir: File, requestedPort: Option
...
 
+  def logPage(request: HttpServletRequest): Seq[scala.xml.Node] = {
+    val appId = request.getParameter("appId")
+    val executorId = request.getParameter("executorId")
+    val logType = request.getParameter("logType")
+    val getOffset = request.getParameter("offset")
+    val getLineLength = request.getParameter("lineLength")
+    val path = "%s/%s/%s/%s".format(workDir.getPath, appId, executorId, logType)
+    val source = fromFile(path)
+    val lines = source.getLines().toArray
```
pwendell: This will read the entire log file into memory, which isn't such a good idea. Sometimes these log files can get as big as ~2GB. I'm going to add some overall comments below about how to consider doing this without reading the entire file... because it will be a bit of work.

pwendell:
Hey Karen,

This is really neat! I added some notes in the code review. Most of them were minor style things.

The one bigger thing is that reading the entire file into memory is not a great idea, because we will sometimes have log files that are very large (hence, the need to paginate them!). Also, it would be nice to try and have the `logPage` share as much code as possible with the `log` view that already directly exposes logs to users.

One way to do this is as follows:
1. Add a function in Utils that, given any offset pair (a, b) and a file, will read the parts of the file between a and b and return them in an array (look at Utils.lastNBytes for inspiration, much of the work is done already).
2. Refactor Utils.lastNBytes in terms of this new function.
3. Also implement logPage in terms of this function. 

It might also be easier to switch to using byte offsets instead of line offsets when you do this... since reading parts of the file becomes much trickier if you force it to parse on line boundaries.
- Patrick

karenfeng:
Made another commit based on suggestions from @pwendell
log page
application page
worker page

pwendell:
Hey Karen,

A couple more things:
1. It would be good to refactor lastNBytes to use offsetBytes. Basically it just needs to call offsetBytes(file length - n, file length).
2. Could you remove the old "Logs" column from the UI page? I want to keep that feature for people who write code which accesses it programatically, but no need to have it in the UI now that you have something strictly better. Then you can change "stdout-page" to just "stdout".

I'm going to download this and take a closer look later... this is a great start!

core/src/main/scala/spark/deploy/worker/ui/WorkerWebUI.scala
```diff
@@ -65,6 +70,58 @@ class WorkerWebUI(val worker: ActorRef, val workDir: File, requestedPort: Option
...
 
+  def logPage(request: HttpServletRequest): Seq[scala.xml.Node] = {
+    val appId = request.getParameter("appId")
+    val executorId = request.getParameter("executorId")
+    val logType = request.getParameter("logType")
+    val offset = Option(request.getParameter("offset")).map(_.toInt).getOrElse(0)
+
+    val maxBytes = 1024 * 1024
+    val defaultBytes = 100 * 1024
+    val byteLength = Option(request.getParameter("byteLength")).flatMap(s => Some(s.toInt)).getOrElse(defaultBytes)
```
pwendell: Did you try doing it like this?

```
Option(request.getParameter("byteLength")).map(_.toInt).getOrElse(defaultBytes)
```

I think that should work and its a bit nicer. There is other code in this file that uses flatMap, but it should be changed to use map.

core/src/main/scala/spark/deploy/worker/ui/WorkerWebUI.scala
```diff
@@ -9,9 +9,13 @@ import javax.servlet.http.HttpServletRequest
 
 import org.eclipse.jetty.server.{Handler, Server}
 
+import scala.io.Source._
```
rxin: Hi Karen,

Thanks for submitting this. Can you try use explicit imports and avoid wildcard imports (_). The reason is it makes it more clearer in the code which package a class is from, and in general it is a good practice.

karenfeng:
Thanks @pwendell and @rxin! I implemented your changes.
log page
application page
worker page

core/src/main/scala/spark/deploy/worker/ui/WorkerWebUI.scala
```diff
@@ -65,6 +67,58 @@ class WorkerWebUI(val worker: ActorRef, val workDir: File, requestedPort: Option
     pre + Utils.lastNBytes(path, math.min(numBytes, maxBytes))
   }
 
+  def logPage(request: HttpServletRequest): Seq[scala.xml.Node] = {
+    val appId = request.getParameter("appId")
+    val executorId = request.getParameter("executorId")
+    val logType = request.getParameter("logType")
+    val offset = Option(request.getParameter("offset")).map(_.toLong).getOrElse(0).asInstanceOf[Long]
```
pwendell: Here, it's better to avoid casting. Try:

val offset = Option(request.getParameter("offset")).map(_.toLong).getOrElse(0L)

core/src/main/scala/spark/deploy/worker/ui/IndexPage.scala
```diff
@@ -29,7 +29,7 @@ private[spark] class IndexPage(parent: WorkerWebUI) {
     val stateFuture = (worker ? RequestWorkerState)(timeout).mapTo[WorkerState]
     val workerState = Await.result(stateFuture, 30 seconds)
 
-    val executorHeaders = Seq("ExecutorID", "Cores", "Memory", "Job Details", "Logs")
+    val executorHeaders = Seq("ExecutorID", "Cores", "Memory", "Job Details", "Log Pages")
```
pwendell: Let's just call this page "Logs" in the UI link so it will be the same as before.

pwendell: Also fix this in ApplicationPage

karenfeng:
Made changes based on @pwendell's suggestions and fixed for nonexistent bytes.
application page
worker page

core/src/test/scala/spark/ui/UISuite.scala
```diff
@@ -74,4 +74,32 @@ class UISuite extends FunSuite {
...
+    // Read first few bytes
+    assert(Utils.offsetBytes(f1Path, 0, 5) === "1\n2\n3")
+
+    // Read some middle bytes
+    assert(Utils.offsetBytes(f1Path, 4, 11) === "3\n4\n5\n6")
+
+    // Read last few bytes
+    assert(Utils.offsetBytes(f1Path, 12, 18) === "7\n8\n9\n")
+
+    //Read some nonexistent bytes in the beginning
```
pwendell: Add spaces after this and all subsequent comment slashes.

core/src/main/scala/spark/Utils.scala
```diff
@@ -621,12 +621,14 @@ private object Utils extends Logging {
                          callSiteInfo.firstUserLine)
   }
 
-  /** Return a string containing the last `n` bytes of a file. */
-  def lastNBytes(path: String, n: Int): String = {
+  /** Return a string containing part of a file from byte 'a' to 'b'. */
+  def offsetBytes(path: String, a: Long, b: Long): String = {
```
pwendell: The style guide doesn't allow variables that start with capital letters. One idea:
1. Use 'start' and 'end' instead of 'a' and 'b'
2. Use 'effectiveStart' and 'effectiveEnd' instead of 'A' and 'B'
3. Remove skip, and just have stream.skip(effectiveStart)

karenfeng:
Thanks @pwendell, made the changes you suggested.

pwendell:
Hey Karen,

I just played with this a bit locally (and also chatted with Reynold and Matei) and a few things came up which would be nice feature enhancements. Ideally, these would go in before we merge it.
1. It would be better if there was a larger buffer, e.g. maybe 10,000 bytes, and there was a scrollbar that let you scroll up and down within the log. The log would then be a fixed height in the page. This is because often people want to look at large sections of logs without having to paginate. The main reason we have pagination is to just avoid fetching an enormous amount of log data (e.g. several MB) into the browser at a time, ideally we'd like most use cases to not involve excessive pagination.
2. At the top it should give the by range being looked at and also the total size of the file: e.g.
   "Bytes XXX-YYY of ZZZ".
3. It would be good to have a link that returns back to the master, because otherwise you have to hit "back" a bunch of times if you have a lot of data you are paging through and then you want to go back.
4. The page title should say whether you are looking at stderr or stdout. 
5. It would be better if the "Back" and "Next" buttons said "Previous XXX Bytes" and "Next XXX Bytes". Also if we fix the height of the log window, it's probably only necessary to have one set of these buttons. 
6. It would be nice if this initialized to the end of the log rather than the beginning. Ping me in person when you get to this one though because it might be a bit tricky.

JoshRosen:
This should fix [SPARK-608](https://spark-project.atlassian.net/browse/SPARK-608).

As I commented on that issue, maybe we can use an approach where there's a URL to return a portion of the log as text (not HTML).  This API would allow us to build Javascript features like "infinite scrolling", a real-time log viewer using long polling, or other features that require programatic remote access to the logs.

To be concrete, I'm suggesting that we also add the offset and length logic to `/log` and load the output of `/log` inside of `/logPage` (either on the server-side, or by embedding an iframe).

pwendell:
Hey @JoshRosen - I think that's solved by the existing code actually. Karen is working on the visual component, but  the code in master already supports adding a `numBytes` parameter to `/log`. Do you also want byte ranges in that?

JoshRosen:
@pwendell I want a `/log` with parameters that are sufficient to implement pagination, so that requires an offset.

pwendell:
Ah OK - so basically the proposal is that `/log/` just matches the semantics and parameters as `/logPage` with the difference being whether we render the page as text vs HTML in the nicer UI... correct @JoshRosen?

I support this.

JoshRosen:
@pwendell Yeah, that's exactly what I'm suggesting.  Good summary.

karenfeng:
@pwendell: Is this what you meant by a bigger buffer?
log page

karenfeng:
@pwendell, I added most of the feature enhancements you suggested. However, the "back to master" links on the worker and log pages still aren't working after making worker a Worker instead of an ActionRef (as we discussed earlier). This is the landing page (I removed the offset variable and the log page loads to the end of the log by default).
log page

karenfeng:
log page

core/src/main/scala/spark/deploy/worker/ui/WorkerWebUI.scala
```diff
@@ -65,6 +68,75 @@ class WorkerWebUI(val worker: ActorRef, val workDir: File, requestedPort: Option
...
+    val executorId = request.getParameter("executorId")
+    val logType = request.getParameter("logType")
+
+    val maxBytes = 1024 * 1024
+    val defaultBytes = 10000
+    val byteLength = Option(request.getParameter("byteLength")).map(_.toInt).getOrElse(defaultBytes)
+
+    val path = "%s/%s/%s/%s".format(workDir.getPath, appId, executorId, logType)
+    val logLength = new File(path).length()
+    val offset = Option(request.getParameter("offset")).map(_.toLong).getOrElse(logLength-10000)
```
pwendell: This should use defaultBytes rather than 10000

pwendell:
Hey Karen, looks good but could you change the `/log` view to match these same parameters now like @JoshRosen  suggested? Basically it should accept an offset and a number of bytes, and if the offset isn't specified it should assume that you want the bytes from the end of the file.

One nice way to do this would be to factor the logic of determining the actual byte range into a function. This function would accept an Option[Long] as an offset and an Int byte length.

Also I think this would allow you to remove the lastNBytes function.

karenfeng:
log page

karenfeng:
log page

pwendell:
Hey @mateiz, I'm finished looking at this. Want to take a look?

Karen implemented almost all of the things we discussed when I showed you in person. The one thing not included is using line offsets rather than byte offsets. Personally I prefer byte offsets because you control exactly how much data is passed in each page (which might be useful if this is used programatically to tail logs, e.g.).

I've tested this using UIWorkloadGenerator and things worked well.
- Patrick

pwendell:
Jenkins, test this please.

pwendell:
Jenkins, test this please.

---

sent from my phone

pwendell:
Hm... Jenkins, test this please.

mateiz:
Thanks Karen, this looks really great! I'm going to merge it in.

PR #692, opened 2013-07-10
adding takeOrdered() to RDD

mcnamaras:
We have been liking RDD.top(), but when used with ordering it feels awkward.

For instance- given this list: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10

`top(5).sorted` result: 6, 7, 8, 9, 10

Normally when you take the top values, the values would be sorted descending.  So to do this one ends up doing something like:

`top(5)(order).sorted(order.reverse)` result: 10, 9, 8, 7, 6

We were doing this everywhere in our code and decided to create a takeOrdered() so that this is cleaner, more clear, and comes out with order:

`takeOrdered(5)` result: 1, 2, 3, 4, 5
`takeOrdered(5)(descendingOrder)` result: 10, 9, 8, 7, 6

### DIFF
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -782,7 +782,7 @@ abstract class RDD[T: ClassManifest](
   }
 
   /**
-   * Returns the top K elements from this RDD as defined by
+   * Returns the first K elements from this RDD as defined by
    * the specified implicit Ordering[T] and maintains the
    * ordering.
    * @param num the number of top elements to return
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -778,7 +778,7 @@ abstract class RDD[T: ClassManifest](
     }.reduce { (queue1, queue2) =>
       queue1 ++= queue2
       queue1
-    }.toArray
+    }.toArray.sorted(ord.reverse)
   }
 
   /**
@@ -789,9 +789,7 @@ abstract class RDD[T: ClassManifest](
    * @param ord the implicit ordering for T
    * @return an array of top elements
    */
-  def takeOrdered(num: Int)(implicit ord: Ordering[T]): Array[T] = {
-    top(num)(ord.reverse).sorted(ord)
-  }
+  def takeOrdered(num: Int)(implicit ord: Ordering[T]): Array[T] = top(num)(ord.reverse)
 
   /**
    * Save this RDD as a text file, using string representations of elements.
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -782,6 +782,18 @@ abstract class RDD[T: ClassManifest](
   }
 
   /**
+   * Returns the top K elements from this RDD as defined by
+   * the specified implicit Ordering[T] and maintains the
+   * ordering.
+   * @param num the number of top elements to return
+   * @param ord the implicit ordering for T
+   * @return an array of top elements
+   */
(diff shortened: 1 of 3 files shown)


ryanlecompte:
Since this hasn't been released yet, why don't we just modify top itself to do this by default instead of introducing a new top-level method on RDD?

mcnamaras:
@ryanlecompte Yeah, this was something I debated.  My thinking for not modifying top was because in my mind top implies always returning the largest values.

Plus cases like this:

```
implicit val ord = ...
rdd.top(5)  //  <-- it may not be always be as apparent, but depending on ord, top could actually be returning the bottom.
```

ryanlecompte:
Nice justification!

markhamstra:
Couldn't this be addressed by changing `top` to take two `Ordering` parameters: one for the order used to select the top-n elements, the second for the order in which they are returned/reported?  Set the default values to do what `takeOrdered` does; override one or both to achieve more unusual results, such as bottom-n.

mcnamaras:
@markhamstra yes, I think that is very doable.  Using `top` you still run into cases where top may unknowingly return the bottom.  So I think if we went this route it would make more sense to add it to `takeOrdered`, and not top.  I will think about this tonight, I may go this path.  Would love to hear if others have opinions on this as well.

mateiz:
I like having a separate `takeOrdered`. But question about `top`: does it currently return values in a definite order? I think we should make sure it does that. In other words, if you do [1, 2, 3, 4, 5].top(3), you should get [5, 4, 3].

mcnamaras:
@mateiz re: `top`: no, it does not come back ordered (I agree, it should. This would be easy to change).  When there are multiple tasks you'll sometimes see the same data in different positions between runs, too.  ie, you'd possibly see:
run1: [5, 4, 3]
run2: [3, 5, 4]
run3: [4, 3, 5]
...

So, should this be the path forward then?:
1. Incorporate @markhamstra's idea (two ordering implicits into `takeOrdered`)
2. keep `top`'s interface the same, but have it return results with the correct order

mateiz:
I was actually thinking that takeOrdered would not take two Orderings, just one, so it would be similar to top(order.reverse). The reason is that as long as takeOrdered's results were guaranteed to be ordered, you could easily call .reverse() on them.

mcnamaras:
ok perfect, I think that sounds good.

mcnamaras:
Let me know how this looks.  Thanks!

core/src/main/scala/spark/RDD.scala
```diff
@@ -778,10 +778,20 @@ abstract class RDD[T: ClassManifest](
     }.reduce { (queue1, queue2) =>
       queue1 ++= queue2
       queue1
-    }.toArray
+    }.toArray.sorted(ord.reverse)
   }
 
   /**
+   * Returns the top K elements from this RDD as defined by
```
mateiz: Instead of saying "top" in this doc comment, say "first" to make it unambiguous.

core/src/main/scala/spark/api/java/JavaRDDLike.scala
```diff
@@ -384,4 +384,29 @@ trait JavaRDDLike[T, This <: JavaRDDLike[T, This]] extends Serializable {
     val comp = com.google.common.collect.Ordering.natural().asInstanceOf[Comparator[T]]
     top(num, comp)
   }
+
+  /**
+   * Returns the top K elements from this RDD as defined by
```
mateiz: Ditto here on the doc comment

mcnamaras:
Good catch, this is fixed now

mateiz:
Looks good, thanks!

PR #695, opened 2013-07-12
Enhance job ui in spark ui system with adding pool information

xiajunluan:
Enhance job ui in spark, now it will support following three scenarios
1. For FIFO scheduling, job page will similar with original one with adding some summary info and empty pool table for FIFO scheduling
fifo
2. For Fair scheduling, job page will show all pools info with minishare, weight, active stage number and all active/completed/failed stage info with pool name link
fair
3. In Fair shceduling mode, pool name in pool table and stage table will link to pool detail page, pool detail page will show detail info with its active stages for specific pool.
pooldetail


mateiz:
Jenkins, this is ok to test

pwendell:
Hey @xiajunluan - thanks for contributing this! I'm going to do a full review of this tomorrow but I need to dig into the scheduler code a bit more first.

One thing I wanted to mention, we've been using the `UIWorkloadGenerator.scala` in Spark as the sort of de-facto test suite for UI changes. It would be great if you could amend that to test some of the functionality introduced here. A simple version that assigns tasks to one of a few newly named pools (rather than requiring an XML file) would be really useful in testing out the behavior. You'd also want to set the scheduling mode system property to FAIR before initializing the SparkContext.

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -43,13 +44,14 @@ import org.apache.mesos.MesosNativeLibrary
...
-import spark.scheduler.{DAGScheduler, ResultTask, ShuffleMapTask, SparkListener, SplitInfo, Stage, StageInfo, TaskScheduler}
-import spark.scheduler.cluster.{StandaloneSchedulerBackend, SparkDeploySchedulerBackend, ClusterScheduler}
+import spark.scheduler.{DAGScheduler, ResultTask, ShuffleMapTask, SparkListener, SplitInfo, Stage, StageInfo, TaskScheduler, ActiveJob}
+import spark.scheduler.cluster.{StandaloneSchedulerBackend, SparkDeploySchedulerBackend, ClusterScheduler, Schedulable}
 import spark.scheduler.local.LocalScheduler
 import spark.scheduler.mesos.{CoarseMesosSchedulerBackend, MesosSchedulerBackend}
 import spark.storage.{StorageStatus, StorageUtils, RDDInfo}
 import spark.util.{MetadataCleaner, TimeStampedHashMap}
 import ui.{SparkUI}
+import spark.scheduler.cluster.SchedulingMode.SchedulingMode
```
pwendell: Group this with the other spark imports.

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -540,6 +542,17 @@ class SparkContext(
...
 
+  def getPoolsInfo: ArrayBuffer[Schedulable] = {
+    taskScheduler.rootPool.schedulableQueue
+  }
+
+  def getSchedulingMode: SchedulingMode = {
+    taskScheduler.schedulingMode
+  }
+
+  def getPoolNameToPool: HashMap[String, Schedulable] = {
```
pwendell, line 595: This seems a bit redundant with `getPoolsInfo`. What if you remove getPoolsInfo and just call this getPools with a docstring that explains what it returns. Then if people just need the list of pools they can call `val pools = sc.getPools.values`

core/src/main/scala/spark/scheduler/SparkListener.scala
```diff
@@ -8,7 +8,7 @@ import spark.executor.TaskMetrics
 
 sealed trait SparkListenerEvents
 
-case class SparkListenerStageSubmitted(stage: Stage, taskSize: Int) extends SparkListenerEvents
+case class SparkListenerStageSubmitted(stage: Stage, taskSize: Int, properties: Properties = null) extends SparkListenerEvents
```
pwendell: Is there a reason to set Properties to `null` as a default? It looks like we only use this when we actually have properties. We should avoid having null defaults when they aren't needed, because it makes the behavior harder to reason about for subclasses.

core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
```diff
@@ -12,6 +12,7 @@ import spark.scheduler._
 import java.nio.ByteBuffer
 import java.util.concurrent.atomic.AtomicLong
 import java.util.{TimerTask, Timer}
+import spark.scheduler.cluster.SchedulingMode.SchedulingMode
```
pwendell: Group this with the other Spark imports

core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
```diff
@@ -97,6 +98,8 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
 
   var schedulableBuilder: SchedulableBuilder = null
   var rootPool: Pool = null
+  //default scheduler is FIFO
```
pwendell: Our official comment style is to have a space after the `//`. If you could adjust all your comments accordingly that would be great.

core/src/main/scala/spark/scheduler/cluster/ClusterTaskSetManager.scala
```diff
@@ -90,8 +90,8 @@ private[spark] class ClusterTaskSetManager(
   var priority = taskSet.priority
   var stageId = taskSet.stageId
   var name = "TaskSet_"+taskSet.stageId.toString
-  var parent:Schedulable = null
-
+  var parent: Schedulable = null
+  var schedulableQueue :ArrayBuffer[Schedulable] = null
```
pwendell: the colon here should be before the space, not after

core/src/main/scala/spark/scheduler/cluster/Schedulable.scala
```diff
@@ -1,13 +1,17 @@
...
+import spark.scheduler.cluster.SchedulingMode.SchedulingMode
 
+import scala.collection.mutable.ArrayBuffer
 /**
  * An interface for schedulable entities.
  * there are two type of Schedulable entities(Pools and TaskSetManagers)
  */
 private[spark] trait Schedulable {
   var parent: Schedulable
+  //childrens
```
pwendell: This should say "children" or maybe "child queues"

core/src/main/scala/spark/scheduler/cluster/SchedulingMode.scala
```diff
@@ -1,7 +1,7 @@
 package spark.scheduler.cluster
 
-object SchedulingMode extends Enumeration("FAIR","FIFO"){
+object SchedulingMode extends Enumeration("FAIR", "FIFO", "NONE"){
```
pwendell: It would be good to have some brief documentation here describing what this enumeration means. In particular, I found it a little confusing what `None` is for and had to go look around a bunch to understand it. One idea:

"A `SchedulingMode` determines which policy is used to order tasks amongst a `Schedulable`'s sub-queues. NONE is used when the a `Schedulable` has no sub-queues."

core/src/main/scala/spark/scheduler/local/LocalTaskSetManager.scala
```diff
@@ -14,6 +14,7 @@ import spark.scheduler.cluster._
 
 private[spark] class LocalTaskSetManager(sched: LocalScheduler, val taskSet: TaskSet) extends TaskSetManager with Logging {
   var parent: Schedulable = null
+  var schedulableQueue :ArrayBuffer[Schedulable] = null
```
pwendell: Colon should come before space

core/src/main/scala/spark/ui/jobs/IndexPage.scala
```diff
@@ -6,107 +6,52 @@ import javax.servlet.http.HttpServletRequest
 
 import scala.Some
 import scala.xml.{NodeSeq, Node}
+import scala.collection.mutable.HashMap
```
pwendell: These imports should be alphabetized.

core/src/main/scala/spark/ui/jobs/IndexPage.scala
```diff
@@ -6,107 +6,52 @@ import javax.servlet.http.HttpServletRequest
...
+import scala.collection.mutable.HashSet
 
 import spark.scheduler.Stage
 import spark.ui.UIUtils._
 import spark.ui.Page._
 import spark.storage.StorageLevel
+import spark.scheduler.cluster.Schedulable
```
pwendell: These should come before spark.storage...

core/src/main/scala/spark/ui/jobs/JobProgressListener.scala
```diff
@@ -0,0 +1,140 @@
...
+import scala.Seq
+import scala.collection.mutable.{HashSet, ListBuffer, HashMap, ArrayBuffer}
+
+import spark.{ExceptionFailure, SparkContext, Success, Utils}
+import spark.scheduler._
+import spark.scheduler.cluster.TaskInfo
+import spark.executor.TaskMetrics
+import collection.mutable
+
+private[spark] class FairJobProgressListener(val sparkContext: SparkContext)
```
pwendell: It would be cleaner to just have a single JobProgressListener. When FIFO scheduling is in use, it's fine if there is just a single pool called "default" that every stage appears to belong to.

pwendell: So basically I'm proposing merging this functionality into the main JobProgressListener.

core/src/main/scala/spark/ui/jobs/JobProgressListener.scala
```diff
@@ -0,0 +1,140 @@
...
+    }
+  }
+}
+
+private[spark] class JobProgressListener(val sc: SparkContext) extends SparkListener {
+  // How many stages to remember
+  val RETAINED_STAGES = System.getProperty("spark.ui.retained_stages", "1000").toInt
+
+  def stageToPool: HashMap[Stage, String] = null
+  def poolToActiveStages: HashMap[String, HashSet[Stage]] =null
```
pwendell: space after `=` (though you will probably remove this line anyways).

core/src/main/scala/spark/ui/jobs/JobProgressUI.scala
```diff
@@ -24,104 +24,38 @@ private[spark] class JobProgressUI(val sc: SparkContext) {
   def listener = _listener.get
   val dateFmt = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss")
 
+
   private val indexPage = new IndexPage(this)
   private val stagePage = new StagePage(this)
+  private val poolPage = new PoolPage(this)
+
+  var stageTable: StageTable = null
```
pwendell: Having these table and datastructures stored inside of the JobProgressUI here breaks the modularity. Instead, it would be better if they were only referenced from within the page they are contained on.

To deal with initializing them properly,  you can just add a `start()` method to each of the child pages and then the JobProgressUI `start()` calls the `start()` of each child.

core/src/main/scala/spark/ui/jobs/JobProgressUI.scala
```diff
@@ -24,104 +24,38 @@ private[spark] class JobProgressUI(val sc: SparkContext) {
...
   def formatDuration(ms: Long) = Utils.msDurationToString(ms)
 
   def getHandlers = Seq[(String, Handler)](
     ("/stages/stage", (request: HttpServletRequest) => stagePage.render(request)),
+    ("/stages/pool", (request: HttpServletRequest) => poolPage.render(request)),
     ("/stages", (request: HttpServletRequest) => indexPage.render(request))
   )
 }
-
-private[spark] class JobProgressListener extends SparkListener {
```
pwendell, line 60: Good idea factoring this out into a new file - it was getting pretty long.

core/src/main/scala/spark/ui/jobs/StageTable.scala
```diff
@@ -0,0 +1,143 @@
...
+private[spark] class FairStagePoolInfo(listener: JobProgressListener) extends StagePoolInfo {
+  def getStagePoolName(s: Stage): String = {
+    listener.stageToPool(s)
+  }
+
+  def hasHerf: Boolean = true
+}
+
+/** Page showing list of all ongoing and recently finished stages */
+private[spark] class StageTable(val dateFmt: SimpleDateFormat, val formatDuration: Long => String, val listener: JobProgressListener) {
```
pwendell: Instead of having a single StageTable and then mutating its pool information and set of stages, I'd prefer you pass a sequence of stages and pool information as part of the constructor. Then you can create a new StageTable each time you need to plot a list of stages. This would be much cleaner. Also, remove dateFmt and formatDuration and just put them in the class... we'll want those to be the same every time we render a table.

`private[spark] class StageTable(val stages: Seq[Stage], val listener, JobProgressListener, val poolInfo: StagePoolInfo)`

core/src/main/scala/spark/ui/jobs/JobProgressListener.scala
```diff
@@ -0,0 +1,140 @@
...
+import spark.{ExceptionFailure, SparkContext, Success, Utils}
+import spark.scheduler._
+import spark.scheduler.cluster.TaskInfo
+import spark.executor.TaskMetrics
+import collection.mutable
+
+private[spark] class FairJobProgressListener(val sparkContext: SparkContext)
+  extends JobProgressListener(sparkContext) {
+
+  val FAIR_SCHEDULER_PROPERTIES = "spark.scheduler.cluster.fair.pool"
```
pwendell: Having this declared twice in the codebase is asking for trouble. You should expose this as a public constant where it is defined in the scheduler, and then reference that constant here.

core/src/main/scala/spark/ui/jobs/PoolTable.scala
```diff
@@ -0,0 +1,98 @@
...
+private[spark] class PoolDetailSource(sc: SparkContext, poolName: String) extends PoolSource{
+  def getPools: Seq[Schedulable] = {
+    val pools = HashSet[Schedulable]()
+    pools += sc.getPoolNameToPool(poolName)
+    pools.toSeq
+  }
+}
+
+/** Table showing list of pools */
+private[spark] class PoolTable(listener: JobProgressListener) {
```
pwendell: As I suggested with `StageTable`, here you should just pass a list of pools to the constructor rather than having a singleton PoolTable that where you mutate the state using external function calls. You can create a new PoolTable each time you need to render the PoolPage.

pwendell:
Hi @xiajunluan,

Thanks for the contribution! I've done a much more thorough review now. To summarize at a high level, here are the main things to address.
1. Treating PoolPage and StagePage as singletons is a bit clunky because you then have to mutate their internal state each time a new page is rendered. It would be cleaner to create a new instance of these for each rendering, and pass it all the relevant information in the constructor. 
2. It would be good to add code which tests this in the `UIWorkloadGenerator.scala`... see my earlier comment for more about that.
3. There are several style errors. Mostly lines over 100 characters, comment strings which don't have a space after the `//`, or import ordering problems. Please go through and check the code for those. For imports, some of the files you touch already have incorrect style, you don't have to fix those necessarily but you should at least try to bundle your imports with those of the same packages. 
4. There are various small re-factoring and typo suggestions.

Feel free to discuss these suggestions further here or to push new code whenever it's ready. It would be great to get this into 0.8 with the rest of the new UI code.

xiajunluan:
Hi @pwendell 
   Thanks for your comments, I mainly refactor two areas based on your comments
1. Remove single stagetable and pooltable for global parameters and create each inside page classes, fix code style  errors, such annotation format, code lines more than 100 characters, etc.
2. Add another parameter(FAIR|FIFO) for UIWorkloadGenerator test application, now we could make spark job run on Fair or FIFO mode manually
   Any comments is appreciated!

core/src/main/scala/spark/ui/jobs/JobProgressUI.scala
```diff
@@ -24,11 +24,25 @@ private[spark] class JobProgressUI(val sc: SparkContext) {
   def listener = _listener.get
   val dateFmt = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss")
 
+
```
pwendell, line 44: This looks like an extra blank line (the line above is also blank).

core/src/main/scala/spark/ui/jobs/JobProgressUI.scala
```diff
@@ -24,11 +24,25 @@ private[spark] class JobProgressUI(val sc: SparkContext) {
...
+  var stagePoolInfo: StagePoolInfo = null
+  var stagePagePoolSource: PoolSource = null
 
   def start() {
-    _listener = Some(new JobProgressListener)
+    _listener = Some(new JobProgressListener(sc))
+    sc.getSchedulingMode match {
+      case SchedulingMode.FIFO =>
+        stagePoolInfo = new FIFOStagePoolInfo()
```
pwendell, line 56: These two classes `FIFOStagePoolInfo` and `FIFOSource` contain just a very small amount of logic. Since `StageTable.scala` also has access to the listener, could you just remove these classes and have `StageTable.scala` itself check the scheduling mode and figure out that way how to render the stage? 

Then you could eliminate the `stagePoolInfo` and `stagePoolSource` variables in this class, you could eliminate this codeblock, and you could eliminate both of those classes. In return you have a few lines in `StageTable.scala` that check the scheduling mode.

core/src/main/scala/spark/ui/jobs/StageTable.scala
```diff
@@ -0,0 +1,143 @@
...
+
+import spark.scheduler.Stage
+import spark.ui.UIUtils._
+import spark.ui.Page._
+import spark.storage.StorageLevel
+
+/*
+ * Interface to get stage's pool name
+ */
+private[spark] trait StagePoolInfo {
```
pwendell, line 23: Having this trait is a bit heavyweight... see my earlier comment. I was suggesting removing these and instead just directly doing this calculation in the StageTable.

core/src/main/scala/spark/ui/jobs/StageTable.scala
```diff
@@ -0,0 +1,143 @@
...
+  def getStagePoolName(s: Stage): String
+  
+  def hasHerf: Boolean
```
pwendell: Just FYI - the correct term his href not herf.

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -541,6 +544,25 @@ class SparkContext(
   }
 
   /**
+   *  Return pools for fair scheduler
+   *  TODO:now, we have not taken nested pools into account
```
pwendell: Add a space after the colon.

Also, we should have a tentative name by all TODO's. E.g. `TODO(xiajunluan)`.

core/src/main/scala/spark/ui/UIWorkloadGenerator.scala
```diff
@@ -56,14 +66,21 @@ private[spark] object UIWorkloadGenerator {
...
-          job()
-          println("Job funished: " + desc)
-        } catch {
-          case e: Exception =>
-            println("Job Failed: " + desc)
-        }
+        new Thread {
+          override def run() {
+            if(schedulingMode == SchedulingMode.FAIR) {
+              sc.addLocalProperties("spark.scheduler.cluster.fair.pool",desc)
```
pwendell: Could we just always set this property and put it inside of `setName`... is it harmful to set this if the mode is FIFO or will it just be ignored?

core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
```diff
@@ -253,10 +255,12 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
       }
       var launchedTask = false
       val sortedTaskSetQueue = rootPool.getSortedTaskSetQueue()
-      for (manager <- sortedTaskSetQueue)
-      {
-        logInfo("parentName:%s,name:%s,runningTasks:%s".format(manager.parent.name, manager.name, manager.runningTasks))
+
+      for (manager <- sortedTaskSetQueue) {
```
pwendell, line 276: Could you remove this log statement or set it to debug? It is extremely verbose and generates a ton of output.

core/src/main/scala/spark/ui/jobs/IndexPage.scala
```diff
@@ -4,109 +4,49 @@ import java.util.Date
...
+                  </div> ++
+                  <h3>Pools </h3> ++ poolTable.toNodeSeq ++
+                  <h3>Active Stages : {activeStages.size}</h3> ++
+                  activeStagesTable.toNodeSeq++
+                  <h3>Completed Stages : {completedStages.size}</h3> ++
+                  completedStagesTable.toNodeSeq++
+                  <h3>Failed Stages : {failedStages.size}</h3> ++
+                  failedStagesTable.toNodeSeq
+
+    headerSparkPage(content, parent.sc, "Spark Stages/Pools", Jobs)
```
pwendell: It's fine to keep the title as just "Spark Stages"

core/src/main/scala/spark/ui/jobs/PoolTable.scala
```diff
@@ -0,0 +1,95 @@
...
+import spark.ui.UIUtils._
+import spark.ui.Page._
+import spark.storage.StorageLevel
+import spark.scheduler.cluster.Schedulable
+
+/*
+ * Interface for get pools seq showing on Index or pool detail page
+ */
+
+private[spark] trait PoolSource {
```
pwendell, line 23: I'd also suggest removing this trait. The abstraction layer of a trait/classes seems unnecessary. Just have stagePage and indexPage directly compute which pools they need look at.

pwendell:
Hey @xiajunluan,

Thanks for updating this! The workload generator is particularly helpful. I did another pass at review and here are the high level comments.
1. I suggest removing the traits/classes you added for various aspects of the pool data. This adds quite a lot of code, and there is only a tiny amount of logic abstracted away as a benefit. I suggest just directly computing the necessary information in the Table classes.
2. Unfortunately, there was a fairly large merge to this part of the code yesterday and you'll need to bring it up to date. I'll try to avoid merging anything else in during the next 48 hours since you are actively working on this.
3. When I actually ran the UIWorkloadGenerator, it seemed like the sliders showing active tasks weren't always updating correctly. Rather than dig into this now, I'll revisit this after you've taken care of (1) and (2), since these will cause major changes, and see if there is still a problem.

core/src/main/scala/spark/ui/jobs/JobProgressUI.scala
xiajunluan, line 56: Yes, I agree with you, StagePoolInfo and PoolSource  just contain a very tiny logic, but I would like that scheduling mode is transparent for stageTable/pool table/stage page/pool page, scheduling mode is only aware within JobProgressUI class. I have even tried to place logic you mentioned into table/page class, but it looks like somewhat tiny dirty, so my final decision is to move these logic up to JobProgressUI class, anyway, it is result of designation tradeoff.

pwendell, line 56: Hey @xiajunluan - OK don't worry about this then. When I merge this in, I might play with it a bit to see if we can find something more elegant. If not I'll just leave it as-is.

xiajunluan:
Hi @pwendell pls review my codes if you are free:)

pwendell:
Okay @xiajunluan - thanks for your work on this! I'm going to merge this and then merge in a few small changes.

PR #696, opened 2013-07-12
Pass executor env vars (e.g. SPARK_CLASSPATH) to compute-classpath.sh

charlesreiss:
ExecutorRunner passes environment variables (specified at the driver) to the executed java program, but not to bin/compute-classpath. This means that env. vars. like SPARK_CLASSPATH aren't passed to workers from the driver, making them not work as per-job configuration.


mateiz:
Good catch, thanks.

PR #697, opened 2013-07-13
Show block locations in Web UI.

pwendell:
This fixes SPARK-769. Support is added for enumerating the locations of blocks
in the UI. There is also some minor cleanup in StorageUtils.

### DIFF
--- a/core/src/main/scala/spark/ui/storage/RDDPage.scala
+++ b/core/src/main/scala/spark/ui/storage/RDDPage.scala
@@ -26,8 +26,14 @@ private[spark] class RDDPage(parent: BlockManagerUI) {
     val workers = filteredStorageStatusList.map((prefix, _))
     val workerTable = listingTable(workerHeaders, workerRow, workers)
 
-    val blockHeaders = Seq("Block Name", "Storage Level", "Size in Memory", "Size on Disk")
-    val blocks = filteredStorageStatusList.flatMap(_.blocks).toArray.sortWith(_._1 < _._1)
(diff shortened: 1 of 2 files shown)


pwendell:
Jenkins, test this please.

core/src/main/scala/spark/ui/storage/RDDPage.scala
```diff
@@ -87,6 +93,9 @@ private[spark] class RDDPage(parent: BlockManagerUI) {
       <td sorttable_customkey={block.diskSize.toString}>
         {Utils.memoryBytesToString(block.diskSize)}
       </td>
+      <td>
+        {locations.map(l => <span>{l}<br/></span>)}
```
mateiz, line 97: I haven't tried displaying this, but do we really want each location on a separate line? It would seem better to do a comma-separated list.

pwendell, line 97: I think a vertical list is much easier to parse visually. Typically we expect this list to be of length one, occasionally it will be two or three if replication is used.

Many clusters have long hostnames and if they are in a list the part that is different lines up well, e.g.:

```
mycluster.node.03.some.long.suffix
mycluster.node.04.some.long.suffix
mycluster.node.33.some.long.suffix
```

If we do comma-separated, it's likely to get wrapped anyways, we just have less control over where it decides to wrap.

mateiz, line 97: Okay, fair enough, let's try this.

PR #698, opened 2013-07-13
changing com.google.code.findbugs maven coordinates

mcnamaras:
"1.3.+" isn't in the public maven repository. Spark compiles fine as it finds it in the resolver list.  However, I did a local-publish and noticed that my drivers would no longer compile. This is the error I would see:

Access denied to: https://oss.sonatype.org/service/local/staging/deploy/maven2/com/google/code/findbugs/jsr305/1.3.+/jsr305-1.3.+.pom

Changing it to 1.3.9 fixes the driver compile issue.


mateiz:
I see, good catch.

PR #699, opened 2013-07-13
Add `Environment` tab to SparkUI.

pwendell:
This adds a tab which displays system property and classpath information. This
can be useful in debugging various types of issues such as:
1. Extra/incorrect Hadoop jars being included in the classpath
2. Spark launching with a different JRE version than intended
3. Spark system properties not being set to intended values
4. User added jars that conflict with Spark jars

### DIFF
--- a/core/src/main/scala/spark/ui/Page.scala
+++ b/core/src/main/scala/spark/ui/Page.scala
@@ -1,3 +1,3 @@
 package spark.ui
 
-private[spark] object Page extends Enumeration { val Storage, Jobs = Value }
+private[spark] object Page extends Enumeration { val Storage, Jobs, Environment = Value }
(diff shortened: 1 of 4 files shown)


pwendell:
Jenkins, test this please.

mateiz:
Patrick, this looks good, but for the properties, do you mind separating the system properties into two tables? So we have Spark Properties and Other System Properties. Also, Spark properties should be defined as `_.startsWith("spark.")`, not `_.contains("spark")`.

One other thought -- it may also be useful to add a section on Java and Scala versions. You can use scala.util.Properties.versionString to get the Scala version, and various standard Java properties to get the Java one.

pwendell:
Hey Matei, I took care of the things you mentioned. I used the scala `Properties` for both the Java and Scala information because it actually contains both.

mateiz:
Great, thanks!

PR #700, opened 2013-07-15
Scala 2.10 

ScrapCodes:
This updates to latest master with scala 2.10 support.


mateiz:
Thanks so much for keeping this up to date! I've merged this in, after running the unit tests, but I also noticed a couple of things that would be good to fix:

1) There are some new compile warnings in 2.10, such as these:

```
[warn]    /scratch/matei/spark/streaming/src/main/scala/spark/streaming/dstream/KafkaInputDStream.scala:121: This catches all Throwables. If this is really intended, use `case _ : Throwable` to clear this warning.
[warn]       case _ => // swallow
[warn]            ^
[info] Resolving com.typesafe.akka#akka-zeromq_2.10;2.1.4 ...
[warn] /scratch/matei/spark/streaming/src/main/scala/spark/streaming/dstream/KafkaInputDStream.scala:88: method erasure in trait ClassManifestDeprecatedApis is deprecated: Use runtimeClass instead
[warn]     val decoder = manifest[D].erasure.newInstance.asInstanceOf[Decoder[T]]
[warn]                               ^
[warn] /scratch/matei/spark/streaming/src/main/scala/spark/streaming/dstream/KafkaInputDStream.scala:98: type ClassManifest in object Predef is deprecated: Use scala.reflect.ClassTag instead
[warn]   private class MessageHandler[T: ClassManifest](stream: KafkaStream[T]) extends Runnable {
[warn]                                 ^
```

2) One of the DAGScheduler tests fails as follows:

```
[info] DAGSchedulerSuite:
[info] - cache location preferences w/ dependency *** FAILED ***
[info]   ArrayBuffer(hostB, hostA) did not equal List(hostA, hostB) (DAGSchedulerSuite.scala:393)
```

(This is probably a bug with the test in general that may also be possible in master, though I don't think I've seen it there.)

It would be good to fix these in a future pull request.

ScrapCodes:
Matei, I have already spent time on that test but could not make sense why does it fail. It does pass in master branch. Do you have suggestion on what should I look into specifically.

mateiz:
I think it's failing simply because the order of the strings in the ArrayBuffer is different from the list. We have (hostB, hostA) versus (hostA, hostB). My guess is that we use a Set class somewhere that sorts things differently in 2.10. We should compare these lists as sets.

PR #703, opened 2013-07-15
Link to job UI from standalone deploy cluster web UI

karenfeng:
This creates a link from the master to the app UI and should fix [SPARK-802](https://spark-project.atlassian.net/browse/SPARK-802).
master log

### DIFF
--- a/core/src/main/scala/spark/deploy/master/Master.scala
+++ b/core/src/main/scala/spark/deploy/master/Master.scala
@@ -80,6 +80,7 @@ private[spark] class Master(host: String, port: Int, webUiPort: Int) extends Act
       logInfo("Registering app " + description.name)
       val app = addApplication(description, sender)
       logInfo("Registered app " + description.name + " with ID " + app.id)
+      logInfo("Started App web UI at " + description.appUiUrl)
       waitingApps += app
       context.watch(sender)  // This doesn't work with remote actors but helps for testing
       sender ! RegisteredApplication(app.id)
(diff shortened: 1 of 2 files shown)


mateiz:
Hey Karen, this is not quite what the issue meant -- what I meant is that in the spark.deploy.master.Master web UI, where we list a table of running applications, there should be a link directly to the application's job UI page (this URL printed here). This will require the application to give the Master this URL when it registers. Take a look at Master.scala to see the code that deals with application registrations, and then search for where that message gets sent in the client to add in another field with the job UI URL.

karenfeng:
Hey @mateiz, is commit 3955711 what it meant?
master page

mateiz:
Yup, this looks good, but please also do the following:
- Instead of adding a separate "job UI" column, make the text in "description" link directly the required URL
- On the application page (what you get by clicking app-2013...), add a link the job UI URL as well.

karenfeng:
Hey @mateiz, I implemented the first point but it looks like the job UI URL on the application page already exists (as Application Detail UI).
master page
application page

mateiz:
Ah okay, even better. I didn't realize we were passing this URL before either. So I guess this was a lot less work than expected! Thanks.

JoshRosen:
On EC2, these links use the private EC2 hostnames instead of the public ones.  I've reported this as a new bug: https://spark-project.atlassian.net/browse/SPARK-833

PR #704, opened 2013-07-16
Add tgz option to make-distribution.sh

mcnamaras:
This pull request builds on #662 by adding a tgz flag.  ie- `./make-distribution.sh tgz`  When tgz is given a spark-$VERSION-bin.tar.gz file will be generated from the dist/ directory.

Use cases for this functionality:
- We have many spark environments and use chef to manage deploys and configs.  Using this method we can easily deploy/undeploy/re-deploy spark from a .tar.gz in a few seconds.
- Should make generating binary releases easy

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,5 @@ dependency-reduced-pom.xml
 .ensime_lucene
 checkpoint
 derby.log
+dist/
+spark-*-bin.tar.gz
--- a/make-distribution.sh
+++ b/make-distribution.sh
@@ -1,10 +1,14 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Script to create a binary distribution for easy deploys of Spark.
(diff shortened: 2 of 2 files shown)


mcnamaras:
@ScrapCodes I agree, this pull request is a short term solution. When/if the project locks on sbt something like sbt-native-packager would definitely be the path forward.

mateiz:
Looks good, thanks. I agree that looking into sbt-native-packager would be good though. Have either of you used it before?

ScrapCodes:
Not yet. But looking at [this](https://github.com/sbt/sbt-launcher-package/blob/full-packaging/project/packaging.scala) and [that](http://www.scala-sbt.org/sbt-native-packager/gettingstarted.html) seemed impressive. I am not sure if it works!

mcnamaras:
I am in the same boat, have not used it before.

PR #705, opened 2013-07-16
Throw a more meaningful message when runJob is called to launch tasks on non-existent partitions.

rxin:
Let me know if you like this error message...


mateiz:
Looks good, thanks.

PR #706, opened 2013-07-16
Spark Streaming stop application mechanism

ScrapCodes:
This is to continue the discussion, that started at mailing list regarding idiomatic way to stop spark streaming app inside a receiver. 

Following is the example that shows how can a streaming application can be stopped using stopApp function. The example will stop streaming app once an end of stream is reached on receiver. I am not sure if its a great thing to do. But can be useful incase we want to detect a situation like connect exception and we would not want to continue. Currently streaming continues even if receiver did not start at all ! 

``` scala
package spark.examples

import spark.streaming.dstream.NetworkReceiver
import java.net.Socket
import java.io.{BufferedReader, InputStreamReader}
import spark.streaming.{Seconds, StreamingContext}
import spark.streaming.StreamingContext._

class SocketTextStreamReceiver(host: String,
                               port: Int
                                ) extends NetworkReceiver[String] {

  protected lazy val blocksGenerator: BlockGenerator =
    new BlockGenerator(StorageLevel.MEMORY_ONLY_SER_2)

  protected def onStart() = {
    blocksGenerator.start()
    val socket = new Socket(host, port)
    val dataInputStream = new BufferedReader(new InputStreamReader(socket.getInputStream(), "UTF-8"))
    var data: String = ""
    while (data != null) {
      data = dataInputStream.readLine();
      blocksGenerator += data
    }
    println("Reached end of stream!!")
    stopApp() //<-- this can be called to stop the streaming application.
  }

  protected def onStop() {
    blocksGenerator.stop()
  }

}

object DocSample {

    val ssc = new StreamingContext("local[2]", "WordCountCustomStreamSource",
      Seconds(1))
    val lines = ssc.networkStream[String](new SocketTextStreamReceiver(
      "localhost", 8445))
    val words = lines.flatMap(_.split(" "))
    val wordCounts = words.map(x => (x, 1)).reduceByKey(_ + _)

    wordCounts.print()
    ssc.start()

  }
}
```


tdas:
This is good start! This is precisely what I had in mind when I had referred to using the existing communication channel between receiver and tracker. However, this is a stop gap solution and is not very pluggable. What if there are multiple receivers and one wants to stop the application only when all the receivers have reported that they have received all the data? In the bigger picture, one may want to send arbitrary messages from the receiver to the driver / app. So instead of making a specific stopApp, I think its better to generalize it. And on the driver side, I think  it may be nice for InputDStream a listener interface (much like SparkListener) whose functions will get called whenever a message (failure or end-of-data or custom message) occurs on the receiver. There may be a default listener that just prints those messages, but the user can specify any listener. Then using the listener, the developer can implement arbitrary functionality - stop when one or all the streams have reported to have ended. 

What do you think? Does this make sense?

ScrapCodes:
AFAIU Lets provide one more feature, where we say if no receiver is running
we stop the app. So user will have option to either call stopApp or stop
where later will just stop that receiver. Does it sound synonymous ?

On Wed, Jul 17, 2013 at 11:06 AM, Tathagata Das notifications@github.comwrote:

> This is good start! This is precisely what I had in mind when I had
> referred to using the existing communication channel between receiver and
> tracker. However, this is a stop gap solution and is not very pluggable.
> What if there are multiple receivers and one wants to stop the application
> only when all the receivers have reported that they have received all the
> data? In the bigger picture, one may want to send arbitrary messages from
> the receiver to the driver / app. So instead of making a specific stopApp,
> I think its better to generalize it. And on the driver side, I think it may
> be nice for InputDStream a listener interface (much like SparkListener)
> whose functions will get called whenever a message (failure or end-of-data
> or custom message) occurs on the receiver. There may be a default listener
> that just prints those messages, but the user can specify any listener.
> Then using the listener, the developer can implement arbitrary
> functionality - stop when one or all the streams have reported to have
> ended.
> 
> What do you think? Does this make sense?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/mesos/spark/pull/706#issuecomment-21093135
> .

## 

s

tdas:
I am not fully convinced by the idea of stopping the app when no receiver is running. While this is a good solution for now, in future, we may use Spark Streaming in a persistent server mode. Kind of like Spark and Shark, they are still running even if there are no Spark / Shark jobs on it. That's why I am thinking twice before introducing this. However, having a pluggable listerner interface, with default policies of shutting down the app when all receivers are stopped, is an acceptable solution, as it can be changed by the app developer as necessary.

ScrapCodes:
hm.. I think I agree with your opinion and lets not arrive at this stop gap solution then. Actually, back of my mind I had this thought of having spark streaming as server, where receivers could be started or shut down in many ways(both internally and externally). I will start thinking about it, if you think it makes sense !

tdas:
Sure, go for it! Happy to discuss this further.

PR #707, opened 2013-07-16
Fixed hadoop related deprecation warnings.

ScrapCodes:
http://grepcode.com/file/repository.cloudera.com/content/repositories/releases/com.cloudera.hadoop/hadoop-core/0.20.2-737/org/apache/hadoop/mapred/OutputCommitter.java#OutputCommitter.cleanupJob%28org.apache.hadoop.mapred.JobContext%29

Hadoop 20.2 onwards commitJob internally calls cleanJob


mateiz:
Jenkins, add to whitelist

mateiz:
Hmm, are you sure `cleanup()` won't be called on failures too? In that case we should not commit the job.

ScrapCodes:
Hm.. commitJob although calls cleanJob , but on failures also may mark it as SUCCESS as in put a file named _SUCCESS there, which should not be done in case of a failed job. Then shall we replace it with abortJob ? as it also provides a way for saying if its failed or killed etc..

mateiz:
Yeah, it sounds like that's what we have to do. Can you maybe also design a test for it to check that jobs are aborted?

ScrapCodes:
Not sure how to test that, since we are simply calling a hadoop api
functions, where from 20.2 to current trunk abortJob only calls cleanupJob.
My confusion is should we test this, since the called function is 
from hadoop API and is not used anywhere(apart from a redundant use!).

On Wed, Jul 17, 2013 at 11:09 PM, Matei Zaharia notifications@github.comwrote:

> Yeah, it sounds like that's what we have to do. Can you maybe also design
> a test for it to check that jobs are aborted?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/mesos/spark/pull/707#issuecomment-21130209
> .

## 

s

mateiz:
Sorry, what I meant is, if a job doing saveAsTextFile fails (e.g. because one of the tasks throws an exception), we should test that no output directory is created. Calling commitJob would probably still cause the output directory to be created, which would be bad.

ScrapCodes:
@mateiz Take a look if it makes sense !

mateiz:
So sorry if I'm misunderstanding this, but does anything call `cleanup` now? Or even `abortJob`? We do need to call that if a task fails.

ScrapCodes:
How do we call cleanup when a task fails ? Do you think this PR should
address that ?

On Tue, Aug 13, 2013 at 11:42 PM, Matei Zaharia notifications@github.comwrote:

> So sorry if I'm misunderstanding this, but does anything call cleanupnow? Or even
> abortJob? We do need to call that if a task fails.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/mesos/spark/pull/707#issuecomment-22585639
> .

## 

s

ScrapCodes:
Did you mean something like the above commit ?, IMHO it won't be needed.

garyaiki:
I grabbed this as a starter task before seeing this pull request. I'm commenting so I can watch the thread.

PR #710, opened 2013-07-17
Updates to LogisticRegression

shivaram:
Set of updates to LogisticRegression based on comments from @etrain and @atalwalkar. Major change is to add another `train` method which allows an initial weight vector to be passed to gradient descent. This should allow higher-level users to switch between different step sizes or miniBatchFractions.

### DIFF
--- a/mllib/src/main/scala/spark/mllib/optimization/Gradient.scala
+++ b/mllib/src/main/scala/spark/mllib/optimization/Gradient.scala
@@ -30,21 +30,3 @@ abstract class Gradient extends Serializable {
   def compute(data: DoubleMatrix, label: Double, weights: DoubleMatrix): 
       (DoubleMatrix, Double)
 }
-
-class LogisticGradient extends Gradient {
-  override def compute(data: DoubleMatrix, label: Double, weights: DoubleMatrix): 
-      (DoubleMatrix, Double) = {
-    val margin: Double = -1.0 * data.dot(weights)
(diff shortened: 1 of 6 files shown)


etrain:
It seems a little weird for initial weights to come in as an Array[Double] and result weights to come out as a DoubleMatrix. Should probably be consistent - particularly if people want to chain together different Optimizers (5-passes of SGD followed by 2-passes of some other optimizer, for example). I think so far our convention has been public interfaces are Array[Double] inside MLLib, so I'm in favor of output weights being an Array[Double].

Separately, we should think about the shape of the weight vector internally. While we're doing point-wise dot products with the weight vector currently, if we get into the minibatch case, we probably want to make sure the weight vector is generally d x 1, that way we can do a single matrix-vector multiply. Not a big deal here, but thinking forward a bit it might be nice to be consistent.

shivaram:
Thanks for taking a look. I agree that the public interfaces should all have `Array[Double]`. This will also mean that we can change from JBLAS to any other library in the future without affecting higher-level code. The only thing which is not ideal is that we go from `DoubleMatrix` to `Array` and back to `DoubleMatrix` inside the model. But I guess its not that expensive to do that.

Regd. the weight vector, the convention is that it is always a column vector -- And the matrix-vector multiply is a great idea ! We could look at the performance impact of that in a future change.

mateiz:
Looks good to me. Merging it in.

PR #711, opened 2013-07-17
Move ML lib data generator files to util/

shivaram:
The generator classes are out of place in the regression directory --- I've moved them to util to avoid creating another directory (say data or something like that ?) 

@mateiz, @etrain, @atalwalkar -- Any other ideas ?


etrain:
I think this is a fine place to put this - but can we call it "LogisticRegressionDataGenerator"? or something like that? 

Also, can we factor the part that produces the RDD into it's own function within the LogisticRegression[Data]Generator object? I think having Data Generation functions available externally that don't rely on writing things to disk will be extremely useful for e.g. unit tests.

Separately, I think it should be pretty easy to set up such a function with an arbitrary input distribution: maybe it could take something like this:
def generateSamples(nsamples: Long, nfeatures: Int, probOne: Double = 0.5, distFun: Random => Double = _.nextGaussian): RDD[Array[Double]]

The last part is probably overkill, though.

shivaram:
Thanks for the comments. Renamed files to `DataGenerator` and I also refactored the classes to have methods which can be reused.

shivaram:
FYI - the Jenkins build failed because it couldn't reach github.com.

etrain:
The refactored code looks good to me, thanks Shivavaram!

mateiz:
Thanks guys, merged this in.

PR #714, opened 2013-07-18
[BUGFIX]  Fix for sbt/sbt script SPARK_HOME setting

ctn:
In some environments, this command generates incorrect result. This PR fixes it. More details in the fork's commit.

```
export SPARK_HOME=$(cd "$(dirname $0)/.."; pwd)
```

Tested/works on Ubuntu 12/13, MacOS 10.7/10.8, Amazon Linux AMI 2013.03.


mateiz:
Weird, I didn't realize cd could print stuff, but thanks for the fix.

PR #715, opened 2013-07-18
fix a bug in build process that pulls in two versions of ASM.

viirya:
Current build process pulls in two versions of ASM (asm-3.2 and asm-4.0). It causes java.lang.IncompatibleClassChangeError. This PB fixes the bug. Tested with hadoop 2.0.5-alpha.


mateiz:
Looks good, thanks! I didn't realize Hadoop brought in ASM.

PR #716, opened 2013-07-19
Regression: default webui-port can't be set via command line "--webui-port" anymore

ghost:
This fixes the old behaviour broken during the transition to the new UI code

### DIFF
--- a/core/src/main/scala/spark/deploy/master/Master.scala
+++ b/core/src/main/scala/spark/deploy/master/Master.scala
@@ -53,7 +53,7 @@ private[spark] class Master(host: String, port: Int, webUiPort: Int) extends Act
 
   var firstApp: Option[ApplicationInfo] = None
 
-  val webUi = new MasterWebUI(self)
+  val webUi = new MasterWebUI(self, webUiPort)
 
   Utils.checkHost(host, "Expected hostname")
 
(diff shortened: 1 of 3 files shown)


ghost:
Thanks to Michael Parkin for the fix's idea!

pwendell:
Hey @cos - thanks for catching this! The underlying problem is that both MasterArguments and MasterWebUI try to calculate the correct port, deal with defaults, etc.

This patch does fix the issue, but could you amend it to actually remove the redundancy?

I'm suggesting that you make the ui port no longer optional in the MasterWebUI constructor and remove the DEFAULT_PORT value in that class. Leave that logic of deciding the port number entirely to MasterArguments.

ghost:
Makes sense. I will re-do the patch tomorrow. I think the whole issue is caused by a somewhat over-complication of the data flow in this part and the simplification you just proposed would be very welcome.

ghost:
Updated patch with simplified logic. I still am not sure why there are three different way to specify the alternative port, but I think it can be left along for now.

pwendell:
Hey @cos - once your patch goes in we'll have the system property `master.ui.port` as the single way to do this, and it is handled by MasterArguments and nowhere else.

You said there are three ways to do this - what are the other two? We should definitely simplify this as much as possible.

ghost:
command line is the second way
shell environment variable is yet another one ;)

And to be honest - these two above look valid both for quick cli runs and using in the linux packaging. The sys.prop seems like an excess to me, to be honest.

pwendell:
Ah, point taken. It's a bit cluttered to have a system property and a command line argument (the env variable is just an intermediate config option that gets translated into a CL arg). 

In any case, this looks good to me. Maybe @mateiz could take a look and merge it?

pwendell:
Jenkins, test this please.

mateiz:
Looks good, thanks Cos!

PR #717, opened 2013-07-19
Do not copy local jars given to SparkContext in yarn mode

viirya:
Do not copy local jars given to SparkContext in yarn mode since the Context is not running on local. This bug causes failure when jars can not be found. Those jars should be given in other ways (HDFS, HTTP, etc.). Example codes (such as spark.examples.SparkPi) can not work under yarn mode without this PR.

### DIFF
--- a/core/src/hadoop1/scala/spark/deploy/SparkHadoopUtil.scala
+++ b/core/src/hadoop1/scala/spark/deploy/SparkHadoopUtil.scala
@@ -42,6 +42,6 @@ object SparkHadoopUtil {
   // add any user credentials to the job conf which are necessary for running on a secure Hadoop cluster
   def addCredentials(conf: JobConf) {}
 
-  def isYarnMode(): Boolean = { False }
+  def isYarnMode(): Boolean = { false }
 
 }
--- a/core/src/hadoop1/scala/spark/deploy/SparkHadoopUtil.scala
+++ b/core/src/hadoop1/scala/spark/deploy/SparkHadoopUtil.scala
@@ -41,4 +41,7 @@ object SparkHadoopUtil {
 
   // add any user credentials to the job conf which are necessary for running on a secure Hadoop cluster
   def addCredentials(conf: JobConf) {}
+
+  def isYarnMode(): Boolean = { False }
+
 }
--- a/core/src/hadoop2/scala/spark/deploy/SparkHadoopUtil.scala
+++ b/core/src/hadoop2/scala/spark/deploy/SparkHadoopUtil.scala
@@ -42,6 +42,6 @@ object SparkHadoopUtil {
   // add any user credentials to the job conf which are necessary for running on a secure Hadoop cluster
   def addCredentials(conf: JobConf) {}
 
-  def isYarnMode(): Boolean = { False }
+  def isYarnMode(): Boolean = { false }
 
 }
--- a/core/src/hadoop2/scala/spark/deploy/SparkHadoopUtil.scala
+++ b/core/src/hadoop2/scala/spark/deploy/SparkHadoopUtil.scala
@@ -41,4 +41,7 @@ object SparkHadoopUtil {
 
   // add any user credentials to the job conf which are necessary for running on a secure Hadoop cluster
   def addCredentials(conf: JobConf) {}
+
+  def isYarnMode(): Boolean = { False }
+
 }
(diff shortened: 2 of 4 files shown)


mridulm:
addJar is a noop in the case of yarn mode.
You should specify the user jar via the command line flag : --jar
yarn does resource management along with resource localization : so all artifacts must be specified apriori.
I dont think this has changed yet, but will defer to Tom in case it has @tgravescs

The documentation has an example of how to run SparkPi btw, and that works for me.

mridulm:
btw, just curious - is this PR motivated by the need to add more than one user jar ?
Currently we support only one - and expect user to build a consolidated user jar : which might not be the most friendly approach.

If this is the issue - then I could add a hack to workaround the limitation with app.jar.<count>  : allowing us to support multiple user jar's ... unless @tgravescs knows a cleaner yarn solution

viirya:
Using spark.examples.SparkPi as an example, this line,

val spark = new SparkContext(args(0), "SparkPi", System.getenv("SPARK_HOME"), Seq(System.getenv("SPARK_EXAMPLES_JAR")))

fills jars parameter of SparkContext so addJar is not a noop as I saw. Since the local file SPARK_EXAMPLES_JAR will not be available on the host running SparkContext, file copying action causes the program to throw exception and stop. That is tested on a yarn hadoop cluster running hadoop-2.0.5-alpha.

I did specify the user jar via the command line flag : --jar as the documentation shows.

Do not know if it works for you because the jar (SPARK_EXAMPLES_JAR) is available on all your hadoop nodes? If there is not the jar file on the nodes, assigning SPARK_EXAMPLES_JAR to SparkContext as SparkPi does should cause failure because the file cannot be found.

Btw, this PR is not motivated by the need to add more than one user jar. Thanks.

mridulm:
Ah, I see the disconnect : in my test, there is no SPARK_EXAMPLES_JAR env
variable.
A better fix (until addJar is properly fixed from yarn point of view also)
would be to :

a) Modify all spark.deploy.SparkHadoopUtil files to have a isYarnMode()
method - core/src/hadoop2-yarn/scala/spark/deploy/SparkHadoopUtil.scala
already does the right thing.
We need to return false for
core/src/hadoop2/scala/spark/deploy/SparkHadoopUtil.scala and
core/src/hadoop1/scala/spark/deploy/SparkHadoopUtil.scala

b) Add a WARN message in addJar when
spark.deploy.SparkHadoopUtil.isYarnMode() returns true : and return without
doing anything.
This will inform user about incorrect expectation from addJar

If we want to handle the multiple user jar (and prioritizing it), we can
add support for that later.

Do you want to take a shot at it ? Either modify this PR or submit a new PR
to handle this ?
Else I can do it.

viirya:
OK. Let me modify this PR.

viirya:
Please see if new comment is better. Thanks.

mridulm:
Looks good, thanks for the changes !
@mateiz  can you also take a look please ? spark/scala style, etc constraint among other things ?

andyk:
Jenkins, test this please.

viirya:
Typo sorry. Fix now.

viirya:
Btw, add excludeAsm for hadoop2 that pulls asm-3.2 too.

tgravescs:
thanks, I like adding the isYarnMode to the other HadoopUtils, I was actually doing that in some other work but haven't pull requested it so this is great.

Note a workaround you could just unset SPARK_EXAMPLES_JAR.  I run the Pi example all the time on yarn but never had it set so didn't run into this problem.

mridulm:
In general, there could be other usages of addJar (explicitly or implicitly
in user code).
In yarn mode,  this would not work as user expects - hence why I prefer a
warning for the same.

Btw, is there a clean way to add an unbounded list of resources (jar files,
etc) ?

Regards,
Mridul

tgravescs:
You can use the distributed cache in Hadoop. You can specify archives or files which will get put on all the nodes and unarchived and put into the classpath as necessary.  we just need to expose that in the client.

mateiz:
The distributed cache does work for deploying an application, but unfortunately calling addJar on a JAR you added that way will still cause it to be fetched by workers through Spark's HTTP server. If you want to use the distributed cache to distribute your application code though, another way is to add the local (relative) path of the JARs to the SPARK_CLASSPATH environment variable. You guys can do that if this is a concern.

mateiz:
In the meantime, I'll merge this with the warning because it's a good thing to test for.

PR #718, opened 2013-07-19
Fix bug preventing defaultMinSplits from being set >2 (SPARK-822)

JoshRosen:
This fixes [SPARK-822](https://spark-project.atlassian.net/browse/SPARK-822), an issue where defaultMinSplits would be capped at 2 even if `spark.default.parallelism` was set higher.  This could cause jobs to not use enough parallelism if they used the default settings.

The problem was the use of `min` instead of `max`.

### DIFF
--- a/core/src/main/scala/spark/SparkContext.scala
+++ b/core/src/main/scala/spark/SparkContext.scala
@@ -772,7 +772,7 @@ class SparkContext(
   def defaultParallelism: Int = taskScheduler.defaultParallelism
 
   /** Default min number of partitions for Hadoop RDDs when not given by user */
-  def defaultMinSplits: Int = math.min(defaultParallelism, 2)
+  def defaultMinSplits: Int = math.max(defaultParallelism, 2)
 
   private var nextShuffleId = new AtomicInteger(0)
 
--- a/python/pyspark/context.py
+++ b/python/pyspark/context.py
@@ -122,6 +122,13 @@ class SparkContext(object):
         """
         return self._jsc.sc().defaultParallelism()
 
+    @property
+    def defaultMinSplits(self):
+        """
+        Default min number of partitions for Hadoop RDDs when not given by user
+        """
+        return self._jsc.sc().defaultMinSplits()
+
     def __del__(self):
         self.stop()
 
@@ -162,7 +169,7 @@ class SparkContext(object):
         nodes), or any Hadoop-supported file system URI, and return it as an
         RDD of Strings.
         """
-        minSplits = minSplits or min(self.defaultParallelism, 2)
+        minSplits = minSplits or self.defaultMinSplits
         jrdd = self._jsc.textFile(name, minSplits)
         return RDD(jrdd, self)
 


mateiz:
Hmm so this kind of makes sense but unfortunately I'm not 100% sure it's what we wanted. I think we might've actually wanted a min originally. The reason is that although this change will make small input files faster to process faster to process, it will also mean that when you join a small file with a large one, you get a lot of partitions and many more map tasks on a cluster than you normally would have. I'm not sure that's worth the tradeoff for quickly processing small files. (This is all assuming that when you run on a cluster, defaultParallelism will eventually give you the # of cores in the cluster.)

mateiz:
I guess one other option would be to clamp it within some range that will work better on a multicore (e.g. make the 2 a 4 or something). What's the reason you wanted to change this?

JoshRosen:
I noticed this when trying to answer [a question on spark-users](https://groups.google.com/d/msg/spark-users/d1bhul7r610/5DHntlhAllwJ).  In that post, the user noticed that PySpark's kmeans.py example only used two cores.  In the Scala, Java, and Python K means examples, the training points are loaded using `sc.textFile("path/to/points")` without specifying `minSplits`.  In practice, this means that if you run the K means example with `local[8]`, 2 map tasks and 8 reduce tasks will run.

I also noticed that `spark.default.parallelism`'s default is inconsistent across scheduler backends, but that seemed like a separate issue (I reported it as [SPARK-823](https://spark-project.atlassian.net/browse/SPARK-823)).

I find the current `min` confusing because it gives users this binary choice between 1 and 2 splits, but I see your point about wanting to limit the number of splits for small files.

mateiz:
Yeah, I'm not sure what to do. Maybe increasing the 2 to a 4 or 8 would be a better fix for now. The other thing is that we could create some join benchmarks where we try giving the small table more partitions. But I just don't see it working when you get to 100+ partitions, which can easily happen on a cluster.

JoshRosen:
This is a question about sensible default behavior when the user hasn't specified the degree of parallelism for an input.

There are two obvious approaches:
- Cap `defaultMinSplits` to a small value (2 or 4 or 8), optimizing for small input files.
- Use no cap, optimizing for large input files.

My gut feeling is that the absolute performance penalty for picking the wrong degree of parallelism will be greater for large input files, suggesting that we should assume large inputs by default and leave it up to the user to optimize for small inputs. If the inputs are _really_ small, the user may want to make other optimizations anyways (such as performing broadcast joins).

Note that we already use `spark.default.parallelism` as the default number of partitions for `parallelize()`, which we typically assume is called with small inputs.

How common is it to use textFile() or hadoopFile() with very small inputs?  If I pass a directory containing many small files to `textFile()`, do I get at least `defaultMinSplits` splits or `defaultMinSplits * numFiles` splits?  If it's the latter, then I could see this change posing bigger problems.

@mateiz I'm a bit confused about the join performance issue.  You said:

>  although this change will make small input files faster to process, it will also mean that when you join a small file with a large one, you get a lot of partitions and many more map tasks on a cluster than you normally would have.

If small inputs' processing benefits from a higher degree of parallelism, then how does this make joins more expensive?  The only extra map tasks should be from the small file and the number of reducers should be unaffected.  If you're joining a small input with a much larger input, the small input's processing cost usually won't dominate the total runtime.

mateiz:
Josh, just to be clear, the Hadoop InputFormat library will _always_ create more partitions for large files, and does so already. For example, if you have a HDFS block size of 64 MB, and you call textFile on a 640 MB file, you'll get 10 partitions, even though defaultMinSplits is 2. So this is only a question of how we deal with very small files, which may occur when a user is testing locally, and may be confusing if they see just one thread being used. Note that even for local files though, InputFormat treats them as having 32 MB blocks (I believe).

I don't think this is an issue if you pass a directory -- it will not create multiple splits per file there. For the joins, what I meant is that setting defaultMinSplits to a really large value, like the # of cores in the cluster, is probably a bad idea, because it could double the number of map tasks we run. It's true that processing the small file quickly can be good but usually when you join a small table with a big one you'll spend most of your time on the map part of the big one anyway.

JoshRosen:
Ah, good point about Hadoop InputFormat automatically splitting files.  I just tested and it splits files properly under local mode.

Do you me want to close this PR and [SPARK-822](https://spark-project.atlassian.net/browse/SPARK-822) as "won't fix / not a bug", since I no longer think it's an issue?

mateiz:
Well, unless you think we should increase the 2 to something bigger. Would it be better to use 4 since developers are starting to have more cores?

JoshRosen:
Maybe we could do something like `min(defaultParallelism, numberOfCoresOnDriver, 2)`?

mateiz:
Hmm, I think it's better to just close this for now and document this better. It doesn't look like increasing the default would add a ton -- for example, in the question you linked, people were running on a cluster but the file was small so they were seeing too few tasks per machine. Maybe we can add a section on this in the tuning guide.

zjffdu:
Find this issue from spark source code. I meet a case that I want to increase the parallelism of the root RDD which read input from hdfs, but there's no way for me to do that because I use MlLib. If it is math.max, then I can set spark.default.parallelism to increase parallelism.  The current implementation means the max parallelism is number of hdfs block, but can not split hdfs block further.

mcapuccini:
@mateiz I think it would be really nice to have a separated configuration for defaultMinPartitions, so that this can be quickly changed using --conf when submitting to the master. Right now it's very inconvenient, one actually have to hardcode it, or have an option, in the user program.

PR #719, opened 2013-07-19
Creates Executors tab for Jobs UI

karenfeng:
This create an Executors Tab for the Jobs UI and should solve [SPARK-808](https://spark-project.atlassian.net/browse/SPARK-808).
- Moved summary from Storage to Executors Tab
- Created executor tables for RDD blocks, disk and memory usage
  executors tab


mateiz:
Jenkins, add to whitelist

mateiz:
Hey Karen, this looks good, but instead of a separate table for each executor, can you make it have just one big table with ID, address, RDD blocks, memory and disk as columns? That would allow us to sort the executors by these things. Look at how it's done for the tasks on the Stage page, for example.

karenfeng:
Thanks @mateiz. Does this work? (On second thought, maybe I should separate the memory used from the maximum memory)
executors tab

mateiz:
I think it's fine to show it this way, but show it as memoryUsed / memoryTotal instead of the current version, which is remaining / used.

core/src/main/scala/spark/ui/exec/ExecutorsUI.scala
```diff
@@ -0,0 +1,85 @@
...
+      </div>;
+
+    headerSparkPage(content, sc, "Executors", Executors)
+  }
+
+  def getExecInfo(a: Int): Seq[String] = {
+    val execId = sc.getExecutorStorageStatus(a).blockManagerId.executorId
+    val hostPort = sc.getExecutorStorageStatus(a).blockManagerId.hostPort
+    val memUsed = Utils.memoryBytesToString(sc.getExecutorStorageStatus(a).maxMem)
+    val maxMem = Utils.memoryBytesToString(sc.getExecutorStorageStatus(a).memUsed())
```
mateiz: Maybe the problem is that you are using "used" for "max" here and the other way around above.

karenfeng:
@mateiz, the Executors page now displays the number of complete/total (failed) tasks.
executors page

mateiz:
Hey, tanks for the changes. A couple of other comments:
- For the tasks, split it into three columns: Total Tasks, Tasks Complete, and Tasks Failed. This is because we might want to sort by these different categories.
- For the memory column, add a `sorttable_customkey` attribute to the `<td>` element so that we will sort the column by the numeric value of the memory rather than the string that will be printed. For example, "2 MB" should show up before "1 GB", but if you compare strings instead of numbers, the "1 GB" would show up first. Look at `spark.deploy.worker.ui.IndexPage` on line 98 to see how we add this attribute.

Other than that, it's looking really good.

karenfeng:
executors tab

mateiz:
Thanks Karen. Merging this in.

PR #720, opened 2013-07-22
Add a public method getCachedRdds to SparkContext

velvia:
Pretty self explanatory - adds a method to SparkContext to return the immutable map of RDD ID's to RDDs themselves.   Amongst other uses, for our job server, this allows two different jobs submitted as separate JARs to obtain cached RDDs just by using a method from the SparkContext.

### DIFF
--- a/core/src/main/scala/spark/SparkContext.scala
+++ b/core/src/main/scala/spark/SparkContext.scala
@@ -550,7 +550,7 @@ class SparkContext(
    * Returns an immutable map of RDDs that have marked themselves as cached via cache() call.
    * Note that this does not necessarily mean the caching or computation was successful.
    */
-  def getCachedRDDs: Map[Int, RDD[_]] = persistentRdds.asInstanceOf[Map[Int, RDD[_]]]
+  def getCachedRDDs: Map[Int, RDD[_]] = persistentRdds.toMap
 
   def getStageInfo: Map[Stage,StageInfo] = {
     dagScheduler.stageToInfos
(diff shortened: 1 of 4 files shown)


mateiz:
Jenkins, this is ok to test

mateiz:
Hey, a couple of comments on this:
1. It would be better to clone the persistentRdds hashmap instead of just casting it to an immutable Map, because underneath it will still change over time as we remove old RDDs. I believe you can do immutable.Map() ++ persistentRdds.
2. For the unit tests, instead of adding them to RDDSuite and having them depend on the number of cached RDDs before, make a new test suite or find one that doesn't reuse the SparkContext across tests. We have a trait for that.

rxin:
Thanks. In addition to Matei's comment, it would also be better for this to be called getPersistentRdds instead of getCachedRdds. In general we are moving the terminology from "cache" to "persistent".

ryanlecompte:
@mateiz, nice catch for returning a proper `scala.collection.immutable.Map` instead of the casting. I'd also suggest perhaps adding a `toMap` method to `TimestampedHashMap` that returns a proper `scala.collection.immutable.Map` so that we can reuse it elsewhere. This is common for other maps in the Scala collections library, too, e.g.:

``` scala
scala> scala.collection.immutable.TreeMap(1 -> 2).toMap
res2: scala.collection.immutable.Map[Int,Int] = Map(1 -> 2)

scala> scala.collection.mutable.HashMap(1 -> 2).toMap
res3: scala.collection.immutable.Map[Int,Int] = Map(1 -> 2)
```

mateiz:
Yeah, a `toMap` is a good idea.

velvia:
Thanks for all the great feedback guys.  I'm almost done with the changes.

mateiz:
Jenkins, test this please

mateiz:
LGTM, thanks.

PR #721, opened 2013-07-22
Fix bug where DoubleRDDFunctions.sampleStdev() computed the non-sample stdev()

JoshRosen:
This fixes [SPARK-825](https://spark-project.atlassian.net/browse/SPARK-825), a bug where DoubleRDDFunctions.sampleStdev() computed the non-sample stdev() due to a typo.

I also added a commit to bring JavaDoubleRDD up-to-date with the current DoubleRDDFunctions.

After merging, please cherry-pick this into branch-0.7.

PR #722, opened 2013-07-22
Fix bug: DoubleRDDFunctions.sampleStdev() computed non-sample stdev()

JoshRosen:
This fixes [SPARK-825](https://spark-project.atlassian.net/browse/SPARK-825), a bug where DoubleRDDFunctions.sampleStdev() computed the non-sample stdev() due to a typo.

I also added a commit to bring JavaDoubleRDD up-to-date with the current DoubleRDDFunctions.

After merging, please cherry-pick this into branch-0.7.

(This is a copy of #721, which Jenkins didn't seem to want to test, perhaps because I rebased that branch after submitting the pull request).


mateiz:
Good catch, thanks!

mateiz:
I've merged this in 0.7 as well. Probably won't make another 0.6.x release at this point unless there are users who really want it and we find more serious bugs.

PR #723, opened 2013-07-23
Made RegressionModel serializable and added unit tests to make sure predict methods would work.

rxin:
Previously the RegressionModel's were not serializable, and thus the predict methods in them would not work.


shivaram:
Thanks for the fix. This looks good ! Can you also add a test where we use a new set of data points for prediction ? The array `testData` is actual the training set here (can rename this to `trainData`), so its better to have a separate test data set. One way to do that is we could generate more points in the first `generateLogisticInput` call and partition into training, test

rxin:
Updated - now using a generated validation dataset.

shivaram:
Looks good. Will merge this after Jenkins tests

PR #724, opened 2013-07-23
SPARK-826: fold(), reduce(), collect() always attempt to use java serialization

dlyubimov:
[SPARK-826](https://spark-project.atlassian.net/browse/SPARK-826): "fold(), reduce(), collect() always attempt to use java serialization"

Tested with local and standalone mode only at this point. 
But i think it should work since i patched all existing implementations of TaskSetManager, my guess is  ClusterTaskSetManager is used with Mesos backend as well. 

I will have to test it with mesos anyway soon since this is our deployment mode.


JoshRosen:
I renamed your pull request and edited the comment to link to the relevant JIRA issue.

mateiz:
Jenkins, this is ok to test

mateiz:
Dmitriy, did you try testing this locally? I think it might've been legitimately failing tests.

dlyubimov:
i think this is due to spark context leak or not stopping in my test. Still
looking.

dlyubimov:
yeah.
You guys have a base test class thing  SharedSparkContext, but i can't use
that for kryo tests since i need to re-define a serializer.

dlyubimov:
i guess this should be fine now.

mateiz:
I saw you included some changes to SparkBuild.scala to build against CDH2 by default -- can you undo those?

dlyubimov:
Oops. Sure.

dlyubimov:
I think i can just merge my dependent SPARK-835 twin fix into this pull request to unclog review a little bit. Does it make sense?

mateiz:
Sure, please go ahead.

core/src/main/scala/spark/scheduler/TaskResult.scala
```diff
@@ -30,7 +32,19 @@ class TaskResult[T](var value: T, var accumUpdates: Map[Long, Any], var metrics:
   def this() = this(null.asInstanceOf[T], null, null)
 
   override def writeExternal(out: ObjectOutput) {
-    out.writeObject(value)
+
+    val objectSer = SparkEnv.get.serializer.newInstance()
+    val bb = objectSer.serialize(value)
+
+    out.writeInt( bb.remaining())
```
mateiz: Small formatting issue here: remove the space after (

core/src/main/scala/spark/scheduler/TaskResult.scala
```diff
@@ -30,7 +32,19 @@ class TaskResult[T](var value: T, var accumUpdates: Map[Long, Any], var metrics:
...
+    if (bb.hasArray) {
+      out.write(bb.array(), bb.arrayOffset() + bb.position(), bb.remaining())
+    } else {
+      val bbval = new Array[Byte](bb.remaining())
+      bb.get(bbval)
+      out.write(bbval)
+    }
```
mateiz: Add this as a utility method in Spark.Utils (maybe writeByteBuffer)

core/src/test/scala/spark/KryoSerializerSuite.scala
```diff
@@ -126,6 +127,24 @@ class KryoSerializerSuite extends FunSuite {
     
     System.clearProperty("spark.kryo.registrator")
   }
+
+  test("kryo with collect") {
+    val control = 1 :: 2 :: Nil
+    val result = sc.parallelize(control, 2).map(new ClassWithoutNoArgConstructor(_)).collect().map(_.x)
+    assert(control == result.toSeq)
```
mateiz: Use === instead of == here -- it's a special operator in the ScalaTest library that throws a nicer error message if the two aren't equal

mateiz:
Hey, sorry for taking a while to review this, but it looks good overall. Just made two small comments. Feel free to also add the `parallelize` stuff into this.

dlyubimov:
- merged SPARK-835, addressed Matei's comments.

I guess my Idea did some reformatting in a few places which did not match its settings, don't know if it is viewed as a big deal.

core/src/main/scala/spark/rdd/ParallelCollectionRDD.scala
```diff
@@ -20,14 +20,21 @@ package spark.rdd
...
+import java.io._
+import java.nio.ByteBuffer
+import scala.Serializable
+import akka.serialization.JavaSerializer
 
 private[spark] class ParallelCollectionPartition[T: ClassManifest](
-    val rddId: Long,
-    val slice: Int,
-    values: Seq[T])
+                                                                    var rddId: Long,
```
rxin: Mind reseting the style here? The style we use in Spark is to have 4 space indent for function arguments.

dlyubimov: Hm. 

can't seem to find this thing in Idea. Is it an Eclipse style setting?

rxin:
Yea I also didn't find the style setting for indenting function parameters in IntelliJ. I just avoided using the auto formatter in this case.

mateiz:
Jenkins, retest this please

core/src/main/scala/spark/rdd/ParallelCollectionRDD.scala
```diff
@@ -37,15 +41,57 @@ private[spark] class ParallelCollectionPartition[T: ClassManifest](
...
-  override val index: Int = slice
+  override def index: Int = slice
+
+  @throws(classOf[IOException])
+  private def writeObject(out: ObjectOutputStream): Unit = {
+
+    val sfactory = SparkEnv.get.serializer
+    // treat java serializer with default action rather
+    // than going thru serialization,
+    // to avoid a separate serialization header.
```
mateiz: Format this comment and the one below to use 100 characters per line and start with a capital letter, similar to other comments in our code

core/src/main/scala/spark/rdd/ParallelCollectionRDD.scala
```diff
@@ -37,15 +41,57 @@ private[spark] class ParallelCollectionPartition[T: ClassManifest](
...
+      case _ => {
+        // for every other serializer, we
+        // assume that it would support Seq[T] and
+        // do so efficiently.
+        val ser = sfactory.newInstance()
+        out.writeLong(rddId)
+        out.writeInt(slice)
+        val bb = ser.serialize(values)
+        out.writeInt(bb.remaining())
+        Utils.writeByteBuffer(bb, out)
```
mateiz: In here, use the serializeStream method of spark.Serializer instead of creating a giant ByteBuffer. It will probably be more efficient for Seq types by serializing each object one by one.

mateiz: BTW, if determining the end of the stream is a problem, you can write out the Seq and then a null. The non the deserialization side, you read up to the null.

dlyubimov: not sure about SerializeStream(). it will break the ObjectOutput envelope
imo.

mateiz: Oh, I see, that's true. In that case I'd actually make an OutputStream subclass that wraps around ObjectOutput's write(byte[]) method. I still think it's worth it compared to materializing a big byte array in memory.

dlyubimov: yeah, i got rid of big array. 

Unfortunately object output doesn't support "varint" etc. encoding protobuf style, otherwise it would be as compact as it gets imo.

dlyubimov: also, growing buffer in this (f4b4b88) commit is probably ok for this (for now anyway) even though that growth is uncontrolled unlike with the serializeStream you probably guys do. 

We already have assumed we have enough memory to fit the entire "values" object tree with all the reference overheads into current memory, sure we  can admit a buffer that fits one more in a serialized form (albeit largest of them).

mateiz:
Hmm, not sure what's going on with Jenkins -- let me try running this a few times tomorrow to see whether it really can cause a test failure.

dlyubimov:
I think i may have clogged it by frequent commits. Perhaps will unclog by
tomorrow :)

dlyubimov:
it's just whenever it actually tries to use java serializer to serialize values one-by-one it obivously creates its own serialization header to every object, which is a huge overhead. Which is why it is important use default serialization logic in case of Java serializer. This overhead caused those 2 tests to fail. The test for java serializer was broken by IDEA being too helpful and importing a wrong class with the same name. commit 9666443 fixes that, should be fine now. 

Regarding specific stream implementation: i gave it a quick thought, looked at serializer instance implementations. I came to conclusion that such wrapper would have to have a buffer that fits at least one object at any time anyway, and would require additional chunked envelope protocol inside object stream in order not to over-read its own envelope and not break the ObjectInput's envelope. So there would be no memory or performance benefit to compared to the provided implementation, _and_ there actually would be small overhead for maintaining a envelope sync for such nested stream, _and_ it will require one extra class to code. So I honestly don't see a benefit of that approach over the provided one per KISS.

mateiz:
Serializer.serializeStream doesn't try to serialize objects one by one. What I was saying is the following. Say your Seq is an array of Int for example. If you serialize the whole Seq to one byte array, it will be roughly the same size as the original thing. On the other hand, if you created a Java ObjectOutputStream and tried to write the array to it, that would very likely write only one integer at a time, and never have to allocate a giant buffer and perform two expensive (non-cache-local) memory scans instead of one. The same is true for how Kryo serializes arrays.

If you want to make an OutputStream wrapper around ObjectOutput, you don't need any extra buffers. All you have to do to implement OutputStream is implement its write(byte[]) method, and that can just call ObjectOutput.write(byte[]). As long as the caller writes small buffers at a time, which is likely to be the case for good serialization libraries, you won't have any problem with big buffers.

mateiz:
BTW to see how Serializer.serializeStream is implemented, look at JavaSerializer for example. It just creates a standard Java ObjectOutputStream so if you write a Seq to it it's exactly the same thing as it would do in your code, but without the big temporary buffer.

dlyubimov:
On Wed, Jul 31, 2013 at 4:00 PM, Matei Zaharia notifications@github.comwrote:

> Serializer.serializeStream doesn't try to serialize objects one by one.
> What I was saying is the following. Say your Seq is an array of Int for
> example. If you serialize the whole Seq to one byte array, it will be
> roughly the same size as the original thing. On the other hand, if you
> created a Java ObjectOutputStream and tried to write the array to it, that
> would very likely write only one integer at a time, and never have to
> allocate a giant buffer and perform two expensive (non-cache-local) memory
> scans instead of one. The same is true for how Kryo serializes arrays.
> 
> Matei, that is all understood and that's not what i was saying. I was
> saying that this solution will still require buffer as large as the largest
> single element, just like current code does. There's no giant buffer. There
> barely have ever been (it lasted for one commit).

if you want to try something else than what it is now, i am not sure i have
resource to do other than what is done at the moment.

mateiz:
Oh, sorry, I missed some commits along the way and didn't see that you changed ParallelCollectionPartition's writeObject!

The new version is indeed good, but there is actually one subtle problem, which is that it won't be efficient when the collection is a Range. A Range object (like Range(1, 10)) is technically a collection of integers, but it is implemented efficiently by storing just a start and end number, which means that the old version of parallelize() works very efficiently on ranges. However, in the new code, you'll write each object separately, not to mention with a header. This is why I was suggesting to instead do the following:
- Create an OutputStream wrapper around ObjectOutput
- Create a Serializer.serializeStream() around that
- Write just one object to this, which is the Seq; this means it will be written efficiently if it's a Range (or other similarly implemented sequence), or it will be written one element at a time with a small buffer otherwise.

If you don't have time to do this, I would be happy to.

mateiz:
BTW this is also why I was saying you could write out a null after the Seq to identify the end of it on the reading side. I guess my comment got hidden by GitHub when you changed the code.

dlyubimov:
On Wed, Jul 31, 2013 at 4:12 PM, Matei Zaharia notifications@github.comwrote:

> Oh, sorry, I missed some commits along the way and didn't see that you
> changed ParallelCollectionPartition's writeObject!
> 
> The new version is indeed good, but there is actually one subtle problem,
> which is that it won't be efficient when the collection is a Range. A Range
> object (like Range(1, 10)) is technically a collection of integers, but it
> is implemented efficiently by storing just a start and end number, which
> means that the old version of parallelize() works very efficiently on
> ranges. However, in the new code, you'll write each object separately, not
> to mention with a header. This is why I was suggesting to instead do the
> following:
> - Create an OutputStream wrapper around ObjectOutput
> - Create a Serializer.serializeStream() around that
> - Write just one object to this, which is the Seq; this means it will
>   be written efficiently if it's a Range (or other similarly implemented
>   sequence), or it will be written one element at a time with a small buffer
>   otherwise.
> 
> i see. Support for serializing self-contained sequences.

one small question: why you think we want to write a NULL if we want to
write exactly one object , be it sequence or range? It would seem to me
NULL terminator is only useful if we actually unwind the sequence as it is
now.

The way it looks then the wrapper seems to be needed only to stop
propagation of close() to underlying stream. that should be enough it seems.

mateiz:
Yeah, true, it probably isn't necessary. Just reading or writing one object from the stream is enough.

dlyubimov:
Ok, so nested streaming seems to be working in (7c52ecc). 

reduce() works but fold() still doesn't work as well as a bunch of other methods i guess such as aggregate() that require active frontend involvement. I guess there's a whole host of such issues perhaps not coverable in one approach.

dlyubimov:
Strange. my local tests are passing. must be something to do with the fact
that i merged the master in.

On the other note, i see a number of cases such as fold() where an element
is sent in thru a closure, seems to be a pretty pervasive practice. i guess
it will take longer than i anticipated to identify and fix all such
occurrences.

mateiz:
Yeah, unfortunately for fold and friends we might have to do more work, so maybe you can scope this pull request to just task results and parallelize(). Would that be okay?

dlyubimov:
Yes lets split it into several installments.

shivaram:
FYI the build error was 

```
[error] /root/workspace/SparkPullRequestBuilder/core/src/test/scala/spark/KryoSerializerSuite.scala:25: object test is not a member of package spark
[error] import spark.test._
```

mateiz:
Hey Dmitriy, sorry for taking a while to get back to this, but I have a question: why is the test with fold ignored? Does it mean that fold will not work with Kryo serializer set (doesn't seem to be the case), or just that you didn't optimize it? If it does work, you should enable the test just to test for correctness.

dlyubimov:
On Tue, Aug 6, 2013 at 5:39 PM, Matei Zaharia notifications@github.comwrote:

> Hey Dmitriy, sorry for taking a while to get back to this,
> 
> no problem
> 
> but I have a question: why is the test with fold ignored? Does it mean
> that fold will not work with Kryo serializer set (doesn't seem to be the
> case), or just that you didn't optimize it? If it does work, you should
> enable the test just to test for correctness.
> 
> Yes, i believe as we just discussed a message away, fold and a few methods
> use closure to serialize object from front to back. For that reason, fold()
> and some other things like aggregate() still do not work and the fold test
> does not pass. Since we seem to have agreed just a message away to rewamp
> in more than one installment, i assumed i'd just put the fold() into ignore
> and keep around until we figure a better solution to in-closure
> transportation of objects.

On a side note, my needs for now are quite satisfied by this patch, i.e.
reduce(), collect() and parallelize(). I would be happy to work on enabling
proper serialization support in the rest of the api but i think I'd like to
have your or somebody else's opinion w.r.t how'd we want to generalize this
fix in case of closure-as-object-transportation-device.

mateiz:
Oh I see, so the test does not work because it uses this special class that is not Java serializable? I just want to make sure that code that worked for the user before continues to, and I tried fold() in the console with serializable types and it seems fine.

dlyubimov:
yes. the whole motivation of this patch was to work around objects that are
not java serializable (properly anyway, such as Mahout vectors) but are
easily serializable with another serializer.

As such, the kryo tests obviously target kryo-only objects. (actually they
have always been, even before i added new ones).

mateiz:
Alright, thanks. I've merged this in as is then and we can do fold and aggregate later.

rxin:
There are some problems with this change. 

For example, try run the following with kryoserializer: 

```
sc.parallelize(Array((4, 14), (1, 11), (7, 17), (0, 10))).count
```

I get the following error:

```
java.lang.ArrayIndexOutOfBoundsException: -2
at java.util.ArrayList.get(ArrayList.java:324)
at com.esotericsoftware.kryo.util.MapReferenceResolver.getReadObject(MapReferenceResolver.java:42)
at com.esotericsoftware.kryo.Kryo.readReferenceOrNull(Kryo.java:773)
at com.esotericsoftware.kryo.Kryo.readObject(Kryo.java:624)
at com.twitter.chill.WrappedArraySerializer$$anonfun$read$1.apply(WrappedArraySerializer.scala:44)
at com.twitter.chill.WrappedArraySerializer$$anonfun$read$1.apply(WrappedArraySerializer.scala:43)
at scala.collection.immutable.Range.foreach(Range.scala:78)
at com.twitter.chill.WrappedArraySerializer.read(WrappedArraySerializer.scala:43)
at com.twitter.chill.WrappedArraySerializer.read(WrappedArraySerializer.scala:21)
at com.esotericsoftware.kryo.Kryo.readClassAndObject(Kryo.java:729)
at spark.KryoDeserializationStream.readObject(KryoSerializer.scala:49)
at spark.rdd.ParallelCollectionPartition$$anonfun$readObject$1.apply(ParallelCollectionRDD.scala:74)
at spark.rdd.ParallelCollectionPartition$$anonfun$readObject$1.apply(ParallelCollectionRDD.scala:74)
at spark.Utils$.deserializeViaNestedStream(Utils.scala:94)
at spark.rdd.ParallelCollectionPartition.readObject(ParallelCollectionRDD.scala:74)
at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:39)
at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:25)
at java.lang.reflect.Method.invoke(Method.java:597)
at java.io.ObjectStreamClass.invokeReadObject(ObjectStreamClass.java:979)
at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:1871)
at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:1775)
at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1327)
at java.io.ObjectInputStream.readObject(ObjectInputStream.java:349)
at spark.scheduler.ResultTask.readExternal(ResultTask.scala:137)
at java.io.ObjectInputStream.readExternalData(ObjectInputStream.java:1814)
at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:1773)
at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1327)
at java.io.ObjectInputStream.readObject(ObjectInputStream.java:349)
at spark.JavaDeserializationStream.readObject(JavaSerializer.scala:40)
at spark.JavaSerializerInstance.deserialize(JavaSerializer.scala:62)
at spark.scheduler.local.LocalScheduler.runTask(LocalScheduler.scala:184)
at spark.scheduler.local.LocalActor$$anonfun$launchTask$1$$anon$1.run(LocalScheduler.scala:64)
at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:439)
at java.util.concurrent.FutureTask$Sync.innerRun(FutureTask.java:303)
at java.util.concurrent.FutureTask.run(FutureTask.java:138)
at java.util.concurrent.ThreadPoolExecutor$Worker.runTask(ThreadPoolExecutor.java:895)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:918)
at java.lang.Thread.run(Thread.java:680)
```

I am going to take a look at this tonight. If I can't figure out how to fix this quickly, I'd revert the merge and merge it again until we have a test.

dlyubimov:
well, i assume that was using java serialization for Array of tuples , and
now it does Kryo (as you have requested) and kryo turned out to have a
limited support for this. (my guess, in particular, for tuples seem to
cause problem).

mateiz:
Reynold's code was using Kryo, not Java. It's true that Kryo/Chill might have a bug for this, but it's worth looking into to make sure it's not this patch.

mateiz:
By the way, I think the problem might be with Chill's serializer for WrappedArrays (the result of Array.slice in ParallelCollectionRDD). For example look at this:

```
scala> val ser = new spark.KryoSerializer().newInstance
ser: spark.serializer.SerializerInstance = spark.KryoSerializerInstance@d64b7e3

scala> val a = Array((1,1), (2,2), (3,3))
a: Array[(Int, Int)] = Array((1,1), (2,2), (3,3))

scala> val b = a.slice(0, 1).toSeq
b: Seq[(Int, Int)] = WrappedArray((1,1))

scala> val bos = new ByteArrayOutputStream
bos: java.io.ByteArrayOutputStream =

scala> val sos = ser.serializeStream(bos)
sos: spark.serializer.SerializationStream = spark.KryoSerializationStream@30dfc977

scala> sos.writeObject(b)
res22: spark.serializer.SerializationStream = spark.KryoSerializationStream@30dfc977

scala> sos.close

scala> val buf = bos.toByteArray
buf: Array[Byte] = Array(1, 0, 115, 99, 97, 108, 97, 46, 99, 111, 108, 108, 101, 99, 116, 105, 111, 110, 46, 109, 117, 116, 97, 98, 108, 101, 46, 87, 114, 97, 112, 112, 101, 100, 65, 114, 114, 97, 121, 36, 111, 102, 82, 101, -26, 1, 1, 28, 1, 1, 0, 0, 0, 1, 0, 0, 0, 1)

scala> val in = ser.deserialize
deserialize         deserializeMany     deserializeStream

scala> val in = ser.deserializeStream(new ByteArrayInputStream
ByteArrayInputStream

scala> val in = ser.deserializeStream(new ByteArrayInputStream(buf))
in: spark.serializer.DeserializationStream = spark.KryoDeserializationStream@28878581

scala> in.readObject[Seq[(Int, Int)]]
res24: Seq[(Int, Int)] = WrappedArray((null,null))
```

Maybe it doesn't deal with specialized classes well. What we might want to do is to call toArray after the slice() in there so that everything gets written as a "real" array instead of WrappedArray.. though it's not 100% clear this will fix it (it might be that because parallelize is a generic method, the types get erased and we always get WrappedArray).

dlyubimov:
On Wed, Aug 7, 2013 at 6:18 PM, Matei Zaharia notifications@github.comwrote:

> Reynold's code was using Kryo, not Java. It's true that Kryo/Chill might
> have a bug for this, but it's worth looking into to make sure it's not this
> patch.
> 
> I meant it (parallelize())  _was_ using java even when asked for kryo,
> before. now it tries to do kryo.

I will try to reproduce but it looks like limited support for standard
scala types in chill so far.

mateiz:
I think the main question is whether we can make it work in this case despite the limited support in Kryo. For example, would using a collection other than a WrappedArray work there?

Also, if we determine for sure that it is a Chill bug, we should let those guys know.

rxin:
I posted the problem at https://github.com/twitter/chill/issues/114

mateiz:
OK, thanks. I guess we'll see what they reply -- we could switch to using ArrayBuffers instead, but it would also add a bit of space overhead that might not be worth it.

PR #725, opened 2013-07-23
Creates task start events

karenfeng:
This should be the final step to solving [SPARK-808](https://spark-project.atlassian.net/browse/SPARK-808) by creating a column for number of active tasks per executor.
- Creates task start events
- Creates column on Executors UI

### DIFF
--- a/core/src/main/scala/spark/scheduler/DAGScheduler.scala
+++ b/core/src/main/scala/spark/scheduler/DAGScheduler.scala
@@ -52,7 +52,7 @@ class DAGScheduler(
   }
   taskSched.setListener(this)
 
-  //Called by TaskScheduler to report task's starting.
+  // Called by TaskScheduler to report task's starting.
   override def taskStarted(task: Task[_], taskInfo: TaskInfo) {
     eventQueue.put(BeginEvent(task, taskInfo))
   }
--- a/core/src/main/scala/spark/scheduler/DAGScheduler.scala
+++ b/core/src/main/scala/spark/scheduler/DAGScheduler.scala
@@ -52,6 +52,11 @@ class DAGScheduler(
   }
   taskSched.setListener(this)
 
+  //Called by TaskScheduler to report task's starting.
+  override def taskStarted(task: Task[_], taskInfo: TaskInfo) {
+    eventQueue.put(BeginEvent(task, taskInfo))
+  }
+
   // Called by TaskScheduler to report task completions or failures.
   override def taskEnded(
       task: Task[_],
@@ -342,6 +347,9 @@ class DAGScheduler(
 
       case ExecutorLost(execId) =>
         handleExecutorLost(execId)
+
+      case begin: BeginEvent =>
+        sparkListeners.foreach(_.onTaskStart(SparkListenerTaskStart(begin.task, begin.taskInfo)))
 
       case completion: CompletionEvent =>
         sparkListeners.foreach(_.onTaskEnd(SparkListenerTaskEnd(completion.task,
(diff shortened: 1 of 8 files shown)


karenfeng:
@mateiz, I'm having this issue of the number of active tasks show up as negative... What gives?
screen shot 2013-07-23 at 1 38 04 pm

karenfeng:
screen shot 2013-07-23 at 3 11 36 pm

core/src/main/scala/spark/scheduler/SparkListener.scala
```diff
@@ -29,6 +29,8 @@ case class SparkListenerStageSubmitted(stage: Stage, taskSize: Int) extends Spar
 
 case class StageCompleted(val stageInfo: StageInfo) extends SparkListenerEvents
 
+case class SparkListenerTaskStart(task: Task[_], taskInfo: TaskInfo) extends SparkListenerEvents
```
squito, line 32: I think its weird that the name has `SparkListener` as a prefix.  I think that in those cases where its ambiguous, the fully qualified class name can disambiguate.  I realize I may be a little on this, since there is already `SparkListenerStageSubmitted`.  But  I think if we do really want the prefix, then we should at least be consistent and put it on all the events.  Considering that would break the API anyway, I'd vote we drop the prefix from everything.

core/src/main/scala/spark/ui/exec/ExecutorsUI.scala
```diff
@@ -111,13 +113,22 @@ private[spark] class ExecutorsUI(val sc: SparkContext) {
...
+    val executorToTasksActive = HashMap[String, HashSet[Long]]()
     val executorToTasksComplete = HashMap[String, Int]()
     val executorToTasksFailed = HashMap[String, Int]()
     val executorToTaskInfos =
       HashMap[String, ArrayBuffer[(TaskInfo, Option[TaskMetrics], Option[ExceptionFailure])]]()
 
+    override def onTaskStart(taskStart: SparkListenerTaskStart) {
+      val eid = taskStart.taskInfo.executorId
+      executorToTasksActive(eid) = executorToTasksActive.getOrElse(eid, HashSet[Long]()) +
+        taskStart.taskInfo.taskId
```
mateiz: Unfortunately this will copy the old HashSet if it exists. Instead, do something like this:

```
if (!executorToTasksActive.contains(eid)) {
  executorToTasksActive(eid) = HashSet[Long]()
}
executorToTasksActive(eid) += taskStart.taskInfo.taskId
```

The `+=` operator will modify the HashSet in place instead of copying it the way `+` would.

squito:
cool!  I've just got a minor quibble on naming, but this is a nice addition

core/src/main/scala/spark/ui/exec/ExecutorsUI.scala
```diff
@@ -111,13 +113,22 @@ private[spark] class ExecutorsUI(val sc: SparkContext) {
...
+    }
+
     override def onTaskEnd(taskEnd: SparkListenerTaskEnd) {
       val eid = taskEnd.taskInfo.executorId
+      executorToTasksActive(eid) = executorToTasksActive.getOrElse(eid, HashSet[Long]()) -
+        taskEnd.taskInfo.taskId
```
mateiz: Similar thing here, use `-=` instead of `-`

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -52,6 +52,11 @@ class DAGScheduler(
   }
   taskSched.setListener(this)
 
+  //Called by TaskScheduler to report task's starting.
```
mateiz: Small style thing: put a space after the //

core/src/main/scala/spark/scheduler/SparkListener.scala
mateiz, line 32: Yeah, we should fix this in a later patch, but let's keep the prefix for now to keep it consistent with the other events.

mateiz:
Looks great, thanks!

PR #726, opened 2013-07-23
SPARK-829: scheduler shouldn't hang if a task contains unserializable objects in its closure

rxin:


core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -466,6 +465,18 @@ class DAGScheduler(
...
+    // Preemptively serialize the stage RDD to make sure the tasks for this stage will be
+    // serializable. We are catching this exception here because it would be fairly hard to
+    // catch the non-serializable exception down the road, where we have several different
+    // implementations for local scheduler and cluster schedulers.
+    try {
+      SparkEnv.get.closureSerializer.newInstance().serialize(stage.rdd)
+    } catch {
+      case e: NotSerializableException => abortStage(stage, e.toString)
+      return
+    }
```
mateiz: Unfortunately this only checks whether the stage's RDD is serializable, but won't catch problems if the function to run on it in a result stage isn't serializable. Instead of modifying this, change `submitMissingTasks`, and check that the tasks are serializable (just serialize one of them).

Also, not sure why this is being passed a `waiter`.

rxin:
Matei - please take another look.

One thing I am not sure about is whether we need the line

```
running -= stage
```

in the case of failures.

mateiz:
Looks good, thanks. I'll keep the `running -= stage` because some code paths seem to add the stage to `running` before, though an even better thing might be to put it in `abortStage`. I'll look into that.

PR #727, opened 2013-07-23
Scheduler code style cleanup.

rxin:
Some cleanup I did while I was working on SPARK-829. Please merge SPARK-829 first.

 I gotta run now so let's use Jenkins for the unit test run.

### DIFF
--- a/core/src/main/scala/spark/scheduler/DAGScheduler.scala
+++ b/core/src/main/scala/spark/scheduler/DAGScheduler.scala
@@ -17,19 +17,17 @@
 
 package spark.scheduler
 
-import cluster.TaskInfo
-import java.util.concurrent.atomic.AtomicInteger
-import java.util.concurrent.LinkedBlockingQueue
-import java.util.concurrent.TimeUnit
+import java.io.NotSerializableException
 import java.util.Properties
+import java.util.concurrent.{LinkedBlockingQueue, TimeUnit}
+import java.util.concurrent.atomic.AtomicInteger
 
 import scala.collection.mutable.{ArrayBuffer, HashMap, HashSet, Map}
 
 import spark._
 import spark.executor.TaskMetrics
-import spark.partial.ApproximateActionListener
-import spark.partial.ApproximateEvaluator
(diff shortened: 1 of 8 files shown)


mridulm:
Have we changed line width to 80 or less ?
A lot of formatting changes seem to be splitting into smaller lines ...
would result in quite a lot of merge conflicts for other patches later on
if this was unintentional.

Regards
Mridul

rxin:
It is always been 100. 

Do you have other patches you are working on that would conflict with these?

mridulm:
Hmm, unless this mobile browser is messing up, I thought some of the line
splits in the PR were for lines < ~100 but >= ~ 80

There are a few things we are playing wrt sched - whether they work out is
still being investigated :-)

Regards
Mridul

rxin:
I mostly wrap lines > 100.

The cleanup also fixed some minor bugs. For example, there was a case when a format string was actually not formatting the proper string. Anyway, the merge shouldn't be too bad if you make future changes.

mateiz:
I'm going to merge this in too -- thanks. Hopefully it's not too hard to reconcile with scheduling changes.

@mridulm, what changes are you looking at? One thing I want to do is update the way we deal with locality when we've assigned all the node-local tasks to a node. The problem is that right now we look immediately to non-node-local tasks for the remaining cores, but in fact there might be _other_ nodes pending where we could launch those locally, so we get poorer locality than would be ideal. I haven't started working on this yet but I'll send a PR when I do and let you know so that you can test it in your environment. I'll also keep the feature for scheduling for rack locality first instead of node (in really large clusters).

mridulm:
Hmm, what you describe should not happen.
Given a set of available nodes, we would always optimally schedule - and go
to other tasks only in case tbere was no better schedule possible.

With two caveats :

a) Use of noPrefs in case of task failures (this is one of the things we
are expt with).
b) When there is only a single available node reported.

Given current impl, I am not sure how to alleviate (b).
Is there something else I am missing ?

Regards
Mridul

PR #728, opened 2013-07-24
Fix setting of SPARK_EXAMPLES_JAR

jey:
This PR moves the setting of the `SPARK_EXAMPLES_JAR` environment variable back to the `run` script because it has no effect when set in the `compute-classpath.sh` script (which is run in a subshell whose environment is then immediately discarded).

### DIFF
--- a/bin/compute-classpath.sh
+++ b/bin/compute-classpath.sh
@@ -76,17 +76,6 @@ function dev_classpath {
     CLASSPATH="$CLASSPATH:$jar"
   done
 
-  # Figure out the JAR file that our examples were packaged into. This includes a bit of a hack
-  # to avoid the -sources and -doc packages that are built by publish-local.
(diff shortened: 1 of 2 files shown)


mateiz:
Hey Jey, this is good, but you'll have to do it in the Windows version of the scripts as well. You can spin up a Windows instance on EC2 and log into it with remote desktop to try it, or maybe run a VM.

jey:
Matei, this bug isn't present in the Windows scripts as far as I can tell. The `run2.cmd` script seems like it sets `SPARK_EXAMPLES_JAR` correctly.

mateiz:
Ah, good point!

I noticed one other issue though, which is that $EXAMPLES_DIR won't be set if we're in a "release" version of the system (see the if statement above). Maybe you should move this stuff within the if statement?

jey:
Oops, yeah. Fixed

mateiz:
Thanks, merged this.

PR #729, opened 2013-07-24
Stage Page updates

karenfeng:
This shows running tasks and the task status on the Stage Page and should fix [SPARK-804](https://spark-project.atlassian.net/browse/SPARK-804) and [811](https://spark-project.atlassian.net/browse/SPARK-811).
screen shot 2013-07-24 at 10 51 34 am

### DIFF
--- a/core/src/main/scala/spark/scheduler/cluster/TaskInfo.scala
+++ b/core/src/main/scala/spark/scheduler/cluster/TaskInfo.scala
@@ -51,6 +51,18 @@ class TaskInfo(
 
   def running: Boolean = !finished
 
+  def status: String = {
+    if (running)
+      return "RUNNING"
+    if (failed)
+      return "FAILED"
+    if (successful)
+      return "SUCCESSFUL"
+    if (finished)
+      return "FINISHED"
+    "UNKNOWN"
+  }
+
   def duration: Long = {
     if (!finished) {
       throw new UnsupportedOperationException("duration() called on unfinished tasks")
--- a/core/src/main/scala/spark/scheduler/cluster/TaskInfo.scala
+++ b/core/src/main/scala/spark/scheduler/cluster/TaskInfo.scala
@@ -57,9 +57,7 @@ class TaskInfo(
     else if (failed)
       "FAILED"
     else if (successful)
-      "SUCCESSFUL"
-    else if (finished)
-      "FINISHED"
+      "SUCCESS"
     else
       "UNKNOWN"
   }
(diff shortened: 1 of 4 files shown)


core/src/main/scala/spark/scheduler/cluster/TaskInfo.scala
```diff
@@ -51,6 +51,18 @@ class TaskInfo(
 
   def running: Boolean = !finished
 
+  def status: String = {
+    if (running)
```
rxin, line 55: Hi Karen,

Instead of doing ifs with returns, it'd be better to do something like this in Scala:

```
if (running) {
  "RUNNING"
} else if (failed) {
  "FAILED"
} else if (...) {
  ...
} else {
  "UNKNOWN"
}
```

karenfeng, line 55: Thanks Reynold, I definitely should've done that.

core/src/main/scala/spark/scheduler/cluster/TaskInfo.scala
```diff
@@ -51,6 +51,19 @@ class TaskInfo(
...
+      "RUNNING"
+    else if (failed)
+      "FAILED"
+    else if (successful)
+      "SUCCESSFUL"
+    else if (finished)
+      "FINISHED"
+    else
+      "UNKNOWN"
+  }
```
mateiz, line 63: A few comments on the status strings:
- We shouldn't have a case when something is "finished" but not either "successful" or "failed", so just remove that if statement
- Instead of "successful", I would actually use the word "success", because it's shorter.

karenfeng:
stage page

core/src/main/scala/spark/ui/jobs/StagePage.scala
```diff
@@ -61,7 +61,7 @@ private[spark] class StagePage(parent: JobProgressUI) {
     val taskTable = listingTable(taskHeaders, taskRow, tasks)
 
     // Excludes tasks which failed and have incomplete metrics
-    val validTasks = tasks.filter(t => Option(t._2).isDefined)
+    val validTasks = tasks.filter(t => t._1.status == "SUCCESSFUL" && (Option(t._2).isDefined))
```
mateiz: Shouldn't this be changed to "SUCCESS"?

mateiz:
Thanks, merged this.

xiajunluan:
Hi @mateiz and @pwendell, I have submitted the PR 695 to add pool information of spark stages and refactor some design of job ui, recently @karenfeng also contributes some patches about executor and stages ui, so I think it may be necessary to review my codes if you are free to avoid resolving more conflicts later.

mateiz:
@xiajunluan Yeah, we're aware of it and we do want to merge your request too. Just need a bit of time to look over it.

PR #731, opened 2013-07-24
Adding SVM and Lasso

pxinghao:
Also, moving LogisticRegression to classification from regression

Also, add regularization parameter to SGD


mllib/src/main/scala/spark/mllib/classification/SVM.scala
```diff
@@ -0,0 +1,170 @@
...
+/**
+ * SVM using Stochastic Gradient Descent.
+ */
+class SVMModel(
+  val weights: DoubleMatrix,
+  val intercept: Double,
+  val losses: Array[Double]) extends ClassificationModel {
+
+  override def predict(testData: spark.RDD[Array[Double]]) = {
+    testData.map { x => {
```
rxin: The Spark style is (less curly brackets)

```
testData.map { x =>
  ...
}
```

mllib/src/main/scala/spark/mllib/classification/SVM.scala
```diff
@@ -0,0 +1,170 @@
...
+      println("Predicting " + x)
+      signum(new DoubleMatrix(1, x.length, x:_*).dot(this.weights) + this.intercept)
```
rxin: BTW this is actually not going to work because SVMModel is not serializable.

I had a pull request on this issue. You should update your branch and do the same thing for ClassificationModel.

https://github.com/mesos/spark/pull/723

mllib/src/main/scala/spark/mllib/classification/LogisticRegression.scala
```diff
@@ -29,14 +29,14 @@ class LogisticRegressionModel(
...
-class LogisticRegression private (var stepSize: Double, var miniBatchFraction: Double,
+class LogisticRegression private (var stepSize: Double, var regParam: Double, var miniBatchFraction: Double,
     var numIters: Int)
   extends Logging {
 
   /**
    * Construct a LogisticRegression object with default parameters
    */
-  def this() = this(1.0, 1.0, 100)
+  def this() = this(1.0, 1.0, 1.0, 100)
```
atalwalkar: it's a bit strange to be passing a regularization parameter for unregularized logistic regression.  if there's no easy way around this, then we should at least set the default value to zero (not 1).

mllib/src/main/scala/spark/mllib/classification/LogisticRegression.scala
```diff
@@ -141,17 +145,17 @@ object LogisticRegression {
       numIterations: Int)
     : LogisticRegressionModel =
   {
-    train(input, numIterations, 1.0, 1.0)
+    train(input, numIterations, 1.0, 1.0, 1.0)
```
atalwalkar: change regParam default value here as well.

mllib/src/main/scala/spark/mllib/classification/SVM.scala
```diff
@@ -0,0 +1,170 @@
...
+    }
+  }
+
+  override def predict(testData: Array[Double]): Double = {
+    val dataMat = new DoubleMatrix(1, testData.length, testData:_*)
+    signum(dataMat.dot(this.weights) + this.intercept)
+  }
+}
+
+class SVM private (var stepSize: Double, var regParam: Double, var miniBatchFraction: Double,
```
atalwalkar: This class looks nearly identical to the LogisticRegression class.  We should consider providing an abstraction here.

atalwalkar: I think the name of this class should be changed, e.g., 'SVMLocalRandomSGD', to make it clear that this is a particular implementation of SVMs (whereby at each iteration we run SGD locally on a random sample of each partition and then average).  Similarly, alternative implementations should also have descriptive names, e.g., 'SVMLocalDisjointSGD', 'SVMParallelGradient', 'SVMLBFGS', SVMADMM', etc. 'This same comment holds for Logistic Regression.

mllib/src/main/scala/spark/mllib/classification/SVM.scala
```diff
@@ -0,0 +1,170 @@
...
+   */
+  def setStepSize(step: Double) = {
+    this.stepSize = step
+    this
+  }
+
+  /**
+   * Set the regularization parameter. Default 1.0.
+   */
+  def setRegParam(param: Double) = {
```
atalwalkar, line 76: Change to 0.

mllib/src/main/scala/spark/mllib/classification/SVMGenerator.scala
```diff
@@ -0,0 +1,45 @@
...
+import org.jblas.DoubleMatrix
+
+import spark.{RDD, SparkContext}
+import spark.mllib.util.MLUtils
+
+object LassoGenerator {
+
+  def main(args: Array[String]) {
+    if (args.length != 5) {
+      println("Usage: LassoGenerator " +
```
atalwalkar: Typo: this should say SVMGenerator.

mllib/src/main/scala/spark/mllib/classification/SVMGenerator.scala
```diff
@@ -0,0 +1,45 @@
...
+
+    val sparkMaster: String = args(0)
+    val outputPath: String = args(1)
+    val nexamples: Int = if (args.length > 2) args(2).toInt else 1000
+    val nfeatures: Int = if (args.length > 3) args(3).toInt else 2
+    val parts: Int = if (args.length > 4) args(4).toInt else 2
+    val eps = 3
+
+    val sc = new SparkContext(sparkMaster, "LassoGenerator")
+
```
atalwalkar: seed should be 42 right (b/c we're geeks =)

mllib/src/main/scala/spark/mllib/classification/SVMGenerator.scala
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    val sparkMaster: String = args(0)
+    val outputPath: String = args(1)
+    val nexamples: Int = if (args.length > 2) args(2).toInt else 1000
+    val nfeatures: Int = if (args.length > 3) args(3).toInt else 2
+    val parts: Int = if (args.length > 4) args(4).toInt else 2
+    val eps = 3
+
+    val sc = new SparkContext(sparkMaster, "LassoGenerator")
```
atalwalkar: Change Losso to SVM

mllib/src/main/scala/spark/mllib/classification/SVMGenerator.scala
```diff
@@ -0,0 +1,45 @@
...
+    val globalRnd = new Random(94720)
+    val trueWeights = Array.fill[Double](nfeatures + 1) { globalRnd.nextGaussian() }
+
+    val data: RDD[(Double, Array[Double])] = sc.parallelize(0 until nexamples, parts).map { idx =>
+      val rnd = new Random(42 + idx)
+
+      val x = Array.fill[Double](nfeatures) {
+        rnd.nextDouble() * 2.0 - 1.0
+      }
+      val y = signum(((1.0 +: x) zip trueWeights).map{wx => wx._1 * wx._2}.reduceLeft(_+_) + rnd.nextGaussian() * 0.1)
```
atalwalkar: isn't the zip, map, reduce just a dot product?  If so, why not just write it explicitly as one?

mllib/src/main/scala/spark/mllib/optimization/GradientDescent.scala
```diff
@@ -51,10 +53,14 @@ object GradientDescent {
...
       }.reduce((a, b) => (a._1.addi(b._1), a._2 + b._2))
 
-      lossHistory.append(lossSum / miniBatchSize + reg_val)
-      val update = updater.compute(weights, gradientSum.div(miniBatchSize), stepSize, i)
+      val update = updater.compute(weights, gradientSum.div(miniBatchSize), stepSize, i, regParam)
       weights = update._1
       reg_val = update._2
+      lossHistory.append(lossSum / miniBatchSize + reg_val)
+      /***
+      Xinghao: The loss here is sum of lossSum computed using the weights before applying updater,
```
atalwalkar: Agreed, this needs to be fixed.  One simple fix would be to compute the loss for iteration i during iteration i+1.  To do so, we'd need to pass the gradient function both the current and previous weights.

mllib/src/main/scala/spark/mllib/optimization/Updater.scala
```diff
@@ -10,18 +11,44 @@ abstract class Updater extends Serializable {
...
+
+class L1Updater extends Updater {
+  override def compute(weightsOld: DoubleMatrix, gradient: DoubleMatrix,
+      stepSize: Double, iter: Int, regParam: Double): (DoubleMatrix, Double) = {
+    val thisIterStepSize = stepSize / math.sqrt(iter)
+    val normGradient = gradient.mul(thisIterStepSize)
+    val newWeights = weightsOld.sub(normGradient)
+    (0 until newWeights.length).foreach(i => {
+      val wi = newWeights.get(i)
+      newWeights.put(i, signum(wi) * max(0.0, abs(wi) - regParam * thisIterStepSize))
```
atalwalkar: Perhaps add a comment noting that this is soft-thresholding?

mllib/src/main/scala/spark/mllib/regression/Lasso.scala
```diff
@@ -0,0 +1,167 @@
...
+    }
+  }
+
+  override def predict(testData: Array[Double]): Double = {
+    val dataMat = new DoubleMatrix(1, testData.length, testData:_*)
+    dataMat.dot(this.weights) + this.intercept
+  }
+}
+
+class Lasso private (var stepSize: Double, var regParam: Double, var miniBatchFraction: Double,
```
atalwalkar: Again, very similar to the corresponding classes for SVM and LogisticRegression.  We should refactor the code.

atalwalkar:
Overall this looks great.  We might want to talk in person (with Evan/Shiv) to make some design decisions.

mllib/src/main/scala/spark/mllib/classification/SVM.scala
etrain: @rxin - Is it necessary or by convention that the trait ClassificationModel be Serializable? It's never going to be at the bottom of the type hierarchy.

rxin: All models with predict(RDD) function implemented should be serializable. 

For one, it makes it easier to save & distribute the model.

But more importantly, the predict function implemented here simply won't work because the map closure is referencing this.weights and this.intercept, which means "this" will be part of the closure. And since "this" is not serializable, the predict function can't even run.

mllib/src/main/scala/spark/mllib/classification/SVM.scala
```diff
@@ -0,0 +1,170 @@
...
+ * SVM using Stochastic Gradient Descent.
+ */
+class SVMModel(
+  val weights: DoubleMatrix,
+  val intercept: Double,
+  val losses: Array[Double]) extends ClassificationModel {
+
+  override def predict(testData: spark.RDD[Array[Double]]) = {
+    testData.map { x => {
+      println("Predicting " + x)
```
etrain: We should avoid println's in the prediction code - instead, let's opt for spark's logging facilities. In this case, though, I don't think we want to print "Predicting " x for every point that comes into the model.

mllib/src/main/scala/spark/mllib/classification/SVM.scala
etrain: I'm probably missing something, but I meant specifically for ClassificationModel - since it's a trait (or roughly abstract) - does the _trait_ ClassificationModel need to be Serializable if we enforce that any class that extends it is Serializable?

mllib/src/main/scala/spark/mllib/optimization/Gradient.scala
```diff
@@ -31,3 +31,31 @@ class LogisticGradient extends Gradient {
     (gradient, loss)
   }
 }
+
+
+class SquaredGradient extends Gradient {
```
etrain, line 53: We had talked about moving the Gradient functions into the file where they are used (organizationally keeping the gradients close to where they are used) - right @shivaram? We should probably do that and follow that convention with these gradients as well.

mllib/src/main/scala/spark/mllib/classification/SVM.scala
rxin: Yes, we should definitely enforce all subclasses of Model be serializable. The convention to "enforce" this in type hierarchy in Java is to make the top level interface/trait/abstract class to be Serializable.

mllib/src/main/scala/spark/mllib/optimization/Gradient.scala
atalwalkar, line 53: Yes, we did.  However, I now think it makes more sense to leave these gradient functions where they are b/c the same gradient functions will be used for multiple ML algorithms, e.g., Lasso, Ridge regression and Linear Regression all use the SquaredGradient.

mllib/src/test/scala/spark/mllib/classification/SVMSuite.scala
```diff
@@ -0,0 +1,61 @@
...
+  override def afterAll() {
+    sc.stop()
+    System.clearProperty("spark.driver.port")
+  }
+
+  test("SVM") {
+    val nPoints = 10000
+    val rnd = new Random(42)
+
+    val x1 = Array.fill[Double](nPoints)(rnd.nextGaussian())
```
etrain: We should follow the DataGenerator model here and create a SVMDataGenerator. If we can piggy back off the existing data generator for LR (or modify it to take in new parameters), even better.

mllib/src/test/scala/spark/mllib/regression/LassoSuite.scala
```diff
@@ -0,0 +1,51 @@
...
+  override def afterAll() {
+    sc.stop()
+    System.clearProperty("spark.driver.port")
+  }
+
+  test("Lasso") {
+    val nPoints = 10000
+    val rnd = new Random(42)
+
+    val x1 = Array.fill[Double](nPoints)(rnd.nextGaussian())
```
etrain: same here - test data should use a generator.

mllib/src/main/scala/spark/mllib/classification/SVMGenerator.scala
pxinghao: Yes, it is a dot product. I'm not sure if Array[Double] has a dot product operator though.. if not, I'll have to cast to DoubleMatrix. Would that be preferable?

mllib/src/main/scala/spark/mllib/classification/SVM.scala
pxinghao, line 76: Could you specify what should be changed to 0?

mllib/src/main/scala/spark/mllib/classification/SVM.scala
pxinghao: Creating an abstraction is a semi-major change. Do we want to come to a consensus on what the abstraction should be?

mllib/src/main/scala/spark/mllib/classification/SVMGenerator.scala
mateiz: FWIW, DoubleMatrix's dot will be much faster than this. I'd recommend designing your internal data structures to use DoubleMatrix.

mllib/src/main/scala/spark/mllib/classification/SVM.scala
atalwalkar, line 76: Sorry for the confusion.  I think you should change the regularization parameter to zero throughout the code.

mllib/src/main/scala/spark/mllib/classification/SVM.scala
atalwalkar: I agree.  We can leave this change to a separate pull-request.

etrain:
Once the "__LocalRandomSGD" rename to just "_LocalRandomSGD" bit is done, this looks good to me - thanks Xinghao!

mllib/src/main/scala/spark/mllib/classification/Classification.scala
```diff
@@ -0,0 +1,21 @@
...
+import spark.RDD
+
+trait ClassificationModel extends Serializable {
+  /**
+   * Predict values for the given data set using the model trained.
+   *
+   * @param testData RDD representing data points to be predicted
+   * @return RDD[Double] where each entry contains the corresponding prediction
+   */
+  def predict(testData: RDD[Array[Double]]): RDD[Double]
```
shivaram: Should we make the return type RDD[Boolean] for classification ? If not, this looks identical to the RegressionModel and we should unify them.

mllib/src/main/scala/spark/mllib/classification/LogisticRegression.scala
```diff
@@ -88,6 +70,14 @@ class LogisticRegression private (var stepSize: Double, var miniBatchFraction: D
     this
   }
 
+
```
shivaram: Remove these new lines ?

mllib/src/main/scala/spark/mllib/classification/LogisticRegression.scala
```diff
@@ -166,11 +158,12 @@ object LogisticRegression {
       input: RDD[(Double, Array[Double])],
       numIterations: Int,
       stepSize: Double,
+
```
shivaram, line 154: I see blank lines here (and 152L, 177L, 184L) etc. I believe these are the changes to take in the regularization parameter ? Maybe an issue with github or a merge.

mllib/src/main/scala/spark/mllib/classification/SVMGenerator.scala
shivaram: I think dot is more readable as well. Creating a new DoubleMatrix from an array is relatively cheap, so we should do that here

mllib/src/main/scala/spark/mllib/optimization/Gradient.scala
```diff
@@ -30,3 +30,48 @@ abstract class Gradient extends Serializable {
...
+
+
+class HingeGradient extends Gradient {
+  override def compute(data: DoubleMatrix, label: Double, weights: DoubleMatrix): 
+      (DoubleMatrix, Double) = {
+
+    val dotProduct = data.dot(weights)
+
+    if (1.0 > label * dotProduct)
+      (data.mul(-label),                        1.0 - label * dotProduct)
```
shivaram: I'd suggest removing the extra spacing between the two parts of the tuple

mllib/src/main/scala/spark/mllib/optimization/GradientDescent.scala
```diff
@@ -70,9 +72,14 @@ object GradientDescent {
       }.reduce((a, b) => (a._1.addi(b._1), a._2 + b._2))
 
       stochasticLossHistory.append(lossSum / miniBatchSize + reg_val)
-      val update = updater.compute(weights, gradientSum.div(miniBatchSize), stepSize, i)
+      val update = updater.compute(weights, gradientSum.div(miniBatchSize), stepSize, i, regParam)
       weights = update._1
       reg_val = update._2
+      stochasticLossHistory.append(lossSum / miniBatchSize + reg_val)
+      /***
```
shivaram: The multi-line comment format we use is:
/**
- NOTE(xinghao): 
  */

Or in this case, you could just use a single line comment format

// NOTE: The loss here ...
// and reg_val

mllib/src/main/scala/spark/mllib/optimization/Updater.scala
```diff
@@ -27,18 +28,47 @@ abstract class Updater extends Serializable {
...
 class SimpleUpdater extends Updater {
   override def compute(weightsOld: DoubleMatrix, gradient: DoubleMatrix,
-      stepSize: Double, iter: Int): (DoubleMatrix, Double) = {
+      stepSize: Double, iter: Int, regParam: Double): (DoubleMatrix, Double) = {
     val normGradient = gradient.mul(stepSize / math.sqrt(iter))
     (weightsOld.sub(normGradient), 0)
   }
 }
+
+/**
```
shivaram, line 48: Same comment as before regd. the multi-line comment format

mllib/src/main/scala/spark/mllib/optimization/Updater.scala
```diff
@@ -27,18 +28,47 @@ abstract class Updater extends Serializable {
...
+    val normGradient = gradient.mul(thisIterStepSize)
+    val newWeights = weightsOld.sub(normGradient)
+    (0 until newWeights.length).foreach(i => {
+      val wi = newWeights.get(i)
+      newWeights.put(i, signum(wi) * max(0.0, abs(wi) - regParam * thisIterStepSize))
+      })
+    (newWeights, newWeights.norm1 * regParam)
+  }
+}
+
```
shivaram, line 72: Perhaps add a comment here describing the updater

mllib/src/test/scala/spark/mllib/classification/SVMSuite.scala
```diff
@@ -0,0 +1,61 @@
...
+
+    val y = (0 until nPoints).map { i =>
+      signum(A + B * x1(i) + C * x2(i) + 0.0*rnd.nextGaussian())
+    }
+
+    val testData = (0 until nPoints).map(i => (y(i).toDouble, Array(x1(i),x2(i)))).toArray
+
+    val testRDD = sc.parallelize(testData, 2)
+    testRDD.cache()
+
```
shivaram, line 102: Is this file svmtest.dat useful for just checking if things look sane or am I missing something ? I guess we can remove this if it was just used for debugging.

mllib/src/test/scala/spark/mllib/classification/SVMSuite.scala
```diff
@@ -0,0 +1,61 @@
...
+    writer_data.close()
+
+    val svm = new SVMLocalRandomSGD().setStepSize(1.0)
+                      .setRegParam(1.0)
+                      .setNumIterations(100)
+
+    val model = svm.train(testRDD)
+
+    val yPredict = (0 until nPoints).map(i => model.predict(Array(x1(i),x2(i))))
+
```
shivaram, line 109: It'll also be cool to check if the weights are close to what we expect them to be (2.0, 1.5 etc.)

shivaram:
Hi Xinghao

This looks great ! Thanks also for merging this with the master branch. I had some minor style comments which I left inline. Couple of higher-level comments
- I moved the generator classes into the `util` directory in `mllib` last week. Could you also move the generator classes for Lasso, SVM to `util` ?
- Agree with @atalwalkar that we should refactor the classification, regression classes to reduce redundant code. We can chat about that offline and I can try to fix that later in a separate PR.
  - Finally, I am not sure if we want to name classes by the technique we use for them (like `LogisticRegressionLocalRandomSGD`). One is that its a mouthful and I think it'll also create too many top-level classes. We could instead consider having a enum which sets the computation method used for `LogisticRegression` in the constructor/builder. Users can then set this value to one of `LocalSGD` or `ParallelGradient`. But again I think we can change this later.

mllib/src/main/scala/spark/mllib/classification/SVM.scala
pxinghao, line 76: Did you mean to change all regularization parameters to 0? For some models, such as Lasso, the regularization is an essential part of the model. I would suggest leaving the default regularization parameter at something sensible. Particularly for MLlib, where we expect the models to used out-of-the-box.

mllib/src/main/scala/spark/mllib/classification/Classification.scala
pxinghao: I propose returning Int instead of Double, in order to accommodate multi-class classification. Just pushed that change to my fork.

pxinghao:
I think my commits have finally caught up with all the suggestions, other than the major refactoring that we probably do not want to do in this pull request. If anyone finds something else that needs fixing, send a comment my way and I'll get to fixing it.

mllib/src/main/scala/spark/mllib/classification/LogisticRegression.scala
```diff
@@ -30,48 +30,30 @@ import org.jblas.DoubleMatrix
...
-  override def predict(testData: spark.RDD[Array[Double]]) = {
+  override def predict(testData: spark.RDD[Array[Double]]): RDD[Int] = {
     // A small optimization to avoid serializing the entire model. Only the weightsMatrix
     // and intercept is needed.
     val localWeights = weightsMatrix
     val localIntercept = intercept
     testData.map { x =>
       val margin = new DoubleMatrix(1, x.length, x:_*).mmul(localWeights).get(0) + localIntercept
-      1.0/ (1.0 + math.exp(margin * -1))
+      (1.0/ (1.0 + math.exp(margin * -1))).toInt
```
shivaram: Hmm -- Is toInt the right choice here ? toInt will truncate down prediction of say 0.99 to 0.0

mllib/src/main/scala/spark/mllib/util/LassoDataGenerator.scala
```diff
@@ -0,0 +1,44 @@
...
+    val outputPath: String = args(1)
+    val nexamples: Int = if (args.length > 2) args(2).toInt else 1000
+    val nfeatures: Int = if (args.length > 3) args(3).toInt else 2
+    val parts: Int = if (args.length > 4) args(4).toInt else 2
+    val eps = 3
+
+    val sc = new SparkContext(sparkMaster, "LassoGenerator")
+
+    val globalRnd = new Random(94720)
+    val trueWeights = new DoubleMatrix(1, nfeatures+1, Array.fill[Double](nfeatures + 1) { globalRnd.nextGaussian() }:_*)
```
shivaram: This line looks longer than 100 characters. Limiting lines to 100 characters is a Spark style guideline.

mllib/src/main/scala/spark/mllib/util/SVMDataGenerator.scala
```diff
@@ -0,0 +1,47 @@
...
+    val outputPath: String = args(1)
+    val nexamples: Int = if (args.length > 2) args(2).toInt else 1000
+    val nfeatures: Int = if (args.length > 3) args(3).toInt else 2
+    val parts: Int = if (args.length > 4) args(4).toInt else 2
+    val eps = 3
+
+    val sc = new SparkContext(sparkMaster, "SVMGenerator")
+
+    val globalRnd = new Random(94720)
+    val trueWeights = new DoubleMatrix(1, nfeatures+1, Array.fill[Double](nfeatures + 1) { globalRnd.nextGaussian() }:_*)
```
shivaram: Another line which looks longer than 100 characters.

mllib/src/test/scala/spark/mllib/classification/LogisticRegressionSuite.scala
```diff
@@ -51,19 +51,19 @@ class LogisticRegressionSuite extends FunSuite with BeforeAndAfterAll {
...
     val testData = (0 until nPoints).map(i => (y(i), Array(x1(i))))
     testData
   }
 
-  def validatePrediction(predictions: Seq[Double], input: Seq[(Double, Array[Double])]) {
+  def validatePrediction(predictions: Seq[Int], input: Seq[(Int, Array[Double])]) {
     val numOffPredictions = predictions.zip(input).filter { case (prediction, (expected, _)) =>
       // A prediction is off if the prediction is more than 0.5 away from expected value.
-      math.abs(prediction - expected) > 0.5
+      math.abs(prediction.toDouble - expected.toDouble) > 0.5
```
shivaram: Since predicted and expected are integers we shouldn't really need the math.abs check -- we would just do predicted == expected ?

shivaram:
Thanks for the update -- The changes look good to me. I had some questions about the integer value predicted in LR that I left as an inline comment.

Also one more request regarding the generators: I guess a copy of the LR generator and Ridge generator were added back to `regression/` during the merge. You can just remove them from the regression directory using `git rm` and commit to fix things.

rxin:
Jenkins, test this please.

shivaram:
Thanks Xinghao ! Merging this now and lets talk about the refactoring sometime soon.

pxinghao:
My first successful pull request! =) Thanks to everyone who chipped in with comments and suggestions!

atalwalkar:
Xinghao,

Thanks for your hard work!

-Ameet

PR #733, opened 2013-07-25
Fix Chill pom dependency to use 2.10

tgravescs:
maven build wouldn't work.  I updated to use Chill 2.10.  I'm not positive if that is the right version of if the 2.93 is but it builds with this.


ryanlecompte:
Ah, sorry about that! The correct version to use is `2.9.3`, not `2.10`. Could you please update your pull req with that version and try again?

mateiz:
Sorry for not catching this. I've fixed it manually in 8eb8b52997da56acb3500fa48aac1ab9a3dfdddf and verified that the Maven build works.

mateiz:
Closing this PR since I fixed it.

PR #734, opened 2013-07-25
Get more env vars from driver rather than worker

charlesreiss:
ExecutorRunner interprets SPARK_LIBRARY_PATH/SPARK_JAVA_OPTS/etc. itself now (rather than the run script), but it doesn't honor the driver's settings for them. This patch fixes that, using the ApplicationDescription environment variables, with fallback to the Worker's environment.

### DIFF
--- a/core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
+++ b/core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
@@ -51,6 +51,9 @@ private[spark] class ExecutorRunner(
   var process: Process = null
   var shutdownHook: Thread = null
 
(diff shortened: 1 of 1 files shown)


mateiz:
Good catch, but it doesn't make sense to pass JAVA_HOME this way. For the others, it does. Does our standalone scheduler even attempt to pass JAVA_HOME?

charlesreiss:
JAVA_HOME isn't passed automatically, but it's possible to pass it explicitly in the environment in SparkContext's constructor.

mateiz:
Ah, okay, that makes sense. I'll merge this as is then.

PR #735, opened 2013-07-25
Totals for shuffle data and CPU time

karenfeng:
This shows totals for shuffle data and CPU time (including active tasks) on the homepage overview and Stage pages. It should fix [SPARK-807](https://spark-project.atlassian.net/browse/SPARK-807).
stages
overview

### DIFF
--- a/core/src/main/scala/spark/ui/exec/ExecutorsUI.scala
+++ b/core/src/main/scala/spark/ui/exec/ExecutorsUI.scala
@@ -113,7 +113,7 @@ private[spark] class ExecutorsUI(val sc: SparkContext) {
   }
 
   private[spark] class ExecutorsListener extends SparkListener with Logging {
-    val executorToTasksActive = HashMap[String, HashSet[Long]]()
+    val executorToTasksActive = HashMap[String, HashSet[TaskInfo]]()
     val executorToTasksComplete = HashMap[String, Int]()
     val executorToTasksFailed = HashMap[String, Int]()
     val executorToTaskInfos =
@@ -122,8 +122,8 @@ private[spark] class ExecutorsUI(val sc: SparkContext) {
     override def onTaskStart(taskStart: SparkListenerTaskStart) {
       val eid = taskStart.taskInfo.executorId
       if (!executorToTasksActive.contains(eid))
-        executorToTasksActive(eid) = HashSet[Long]()
-      executorToTasksActive(eid) += taskStart.taskInfo.taskId
+        executorToTasksActive(eid) = HashSet[TaskInfo]()
+      executorToTasksActive(eid) += taskStart.taskInfo
       val taskList = executorToTaskInfos.getOrElse(
         eid, ArrayBuffer[(TaskInfo, Option[TaskMetrics], Option[ExceptionFailure])]())
       taskList += ((taskStart.taskInfo, None, None))
@@ -133,8 +133,8 @@ private[spark] class ExecutorsUI(val sc: SparkContext) {
     override def onTaskEnd(taskEnd: SparkListenerTaskEnd) {
       val eid = taskEnd.taskInfo.executorId
       if (!executorToTasksActive.contains(eid))
-        executorToTasksActive(eid) = HashSet[Long]()
-      executorToTasksActive(eid) -= taskEnd.taskInfo.taskId
+        executorToTasksActive(eid) = HashSet[TaskInfo]()
+      executorToTasksActive(eid) -= taskEnd.taskInfo
       val (failureInfo, metrics): (Option[ExceptionFailure], Option[TaskMetrics]) =
         taskEnd.reason match {
           case e: ExceptionFailure =>
--- a/core/src/main/scala/spark/ui/exec/ExecutorsUI.scala
+++ b/core/src/main/scala/spark/ui/exec/ExecutorsUI.scala
@@ -121,8 +121,9 @@ private[spark] class ExecutorsUI(val sc: SparkContext) {
 
     override def onTaskStart(taskStart: SparkListenerTaskStart) {
       val eid = taskStart.taskInfo.executorId
-      if (!executorToTasksActive.contains(eid))
+      if (!executorToTasksActive.contains(eid)) {
         executorToTasksActive(eid) = HashSet[TaskInfo]()
+      }
       executorToTasksActive(eid) += taskStart.taskInfo
       val taskList = executorToTaskInfos.getOrElse(
         eid, ArrayBuffer[(TaskInfo, Option[TaskMetrics], Option[ExceptionFailure])]())
@@ -132,8 +133,9 @@ private[spark] class ExecutorsUI(val sc: SparkContext) {
 
     override def onTaskEnd(taskEnd: SparkListenerTaskEnd) {
       val eid = taskEnd.taskInfo.executorId
-      if (!executorToTasksActive.contains(eid))
+      if (!executorToTasksActive.contains(eid)) {
         executorToTasksActive(eid) = HashSet[TaskInfo]()
+      }
       executorToTasksActive(eid) -= taskEnd.taskInfo
       val (failureInfo, metrics): (Option[ExceptionFailure], Option[TaskMetrics]) =
         taskEnd.reason match {
--- a/core/src/main/scala/spark/ui/exec/ExecutorsUI.scala
+++ b/core/src/main/scala/spark/ui/exec/ExecutorsUI.scala
@@ -121,10 +121,8 @@ private[spark] class ExecutorsUI(val sc: SparkContext) {
 
     override def onTaskStart(taskStart: SparkListenerTaskStart) {
       val eid = taskStart.taskInfo.executorId
-      if (!executorToTasksActive.contains(eid)) {
-        executorToTasksActive(eid) = HashSet[TaskInfo]()
-      }
-      executorToTasksActive(eid) += taskStart.taskInfo
+      val activeTasks = executorToTasksActive.getOrElseUpdate(eid, new HashSet[TaskInfo]())
+      activeTasks += taskStart.taskInfo
       val taskList = executorToTaskInfos.getOrElse(
         eid, ArrayBuffer[(TaskInfo, Option[TaskMetrics], Option[ExceptionFailure])]())
       taskList += ((taskStart.taskInfo, None, None))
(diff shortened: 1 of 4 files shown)


karenfeng:
overview

core/src/main/scala/spark/ui/jobs/IndexPage.scala
```diff
@@ -39,6 +40,13 @@ private[spark] class IndexPage(parent: JobProgressUI) {
     val completedStages = listener.completedStages.reverse.toSeq
     val failedStages = listener.failedStages.reverse.toSeq
 
+    var activeTime = 0L
+    listener.stageToTasksActive.foreach {s =>
+      s._2.foreach { t =>
+        activeTime += t.timeRunning(System.currentTimeMillis())
+      }
+    }
```
mateiz, line 47: You can write this as

```
for (tasks <- stageToTaskActive.values; t <- tasks) {
  activeTime += t.timeRunning(System.currentTimeMillis())
}
```

Actually, to make it better, you should also grab `System.currentTimeMillis()` just once at the top (e.g. `val now = System.currentTimeMillis()`) instead of calling it repeatedly there.

core/src/main/scala/spark/ui/jobs/JobProgressUI.scala
```diff
@@ -65,7 +65,14 @@ private[spark] class JobProgressListener extends SparkListener {
...
 
-  val stageToTasksActive = HashMap[Int, HashSet[Long]]()
+  var totalTime = 0L
+  var totalShuffleRead = 0L
+  var totalShuffleWrite = 0L
+
+  val stageToTime = HashMap[Int, Long]()
+  val stageToShuffleRead = HashMap[Int, Long]()
+  val stageToShuffleWrite = HashMap[Int, Long]()
+  val stageToTasksActive = HashMap[Int, HashSet[TaskInfo]]()
```
mateiz, line 76: All these hashtables, and some of the ones we added earlier, look like they need to have old values removed in `trimIfNecessary()`. Can you add that there? Look at how `stageToTaskInfos` gets trimmed.

core/src/main/scala/spark/ui/exec/ExecutorsUI.scala
```diff
@@ -122,8 +122,8 @@ private[spark] class ExecutorsUI(val sc: SparkContext) {
     override def onTaskStart(taskStart: SparkListenerTaskStart) {
       val eid = taskStart.taskInfo.executorId
       if (!executorToTasksActive.contains(eid))
```
rxin: Even if the body of the if has only one statement, we should wrap it with  { }

```
if (!executorToTasksActive.contains(eid)) {
  executorToTasksActive(eid) = HashSet[TaskInfo]()
}
```

It makes the code much less error prone. 

One exception to this rule is for "tenary" operations that fit in a single line:

```
if (condition) valueA else valueB
```

mateiz: Yeah, I agree on this -- would be good to fix it elsewhere too.

core/src/main/scala/spark/ui/jobs/JobProgressUI.scala
```diff
@@ -119,6 +134,40 @@ private[spark] class JobProgressListener extends SparkListener {
...
+    stageToShuffleRead(sid) += shuffleRead
+    totalShuffleRead += shuffleRead
+
+    if (!stageToShuffleWrite.contains(sid)) {
+      stageToShuffleWrite(sid) = 0L
+    }
+    val shuffleWrite =
+      if (metrics.isDefined) {
+        metrics.flatMap(m => m.shuffleWriteMetrics).map(s => s.shuffleBytesWritten).getOrElse(0L)
+      }
```
rxin: Put the "else" on the same line as the preceding }. i.e.

```
} else {
```

core/src/main/scala/spark/ui/jobs/StagePage.scala
```diff
@@ -52,6 +52,33 @@ private[spark] class StagePage(parent: JobProgressUI) {
     val shuffleRead = listener.hasShuffleRead(stageId)
     val shuffleWrite = listener.hasShuffleWrite(stageId)
 
+    var activeTime = 0L
+    listener.stageToTasksActive(stageId).foreach { t =>
+      activeTime += t.timeRunning(System.currentTimeMillis())
```
mateiz: One more small thing -- save currentTimeMillis once at the top, like in the index page

core/src/main/scala/spark/ui/jobs/StagePage.scala
```diff
@@ -34,6 +34,7 @@ import spark.executor.TaskMetrics
 private[spark] class StagePage(parent: JobProgressUI) {
   def listener = parent.listener
   val dateFmt = parent.dateFmt
+  val now = System.currentTimeMillis()
```
mateiz: Hmm, this should be in render(), shouldn't it?

karenfeng: Yes, I'm not sure why I put it there...

core/src/main/scala/spark/ui/exec/ExecutorsUI.scala
```diff
@@ -121,9 +121,10 @@ private[spark] class ExecutorsUI(val sc: SparkContext) {
 
     override def onTaskStart(taskStart: SparkListenerTaskStart) {
       val eid = taskStart.taskInfo.executorId
-      if (!executorToTasksActive.contains(eid))
-        executorToTasksActive(eid) = HashSet[Long]()
-      executorToTasksActive(eid) += taskStart.taskInfo.taskId
+      if (!executorToTasksActive.contains(eid)) {
```
pwendell: There is a slightly cleaner way to write this:

```
val activeTasks = executorToTasksActive.getOrElseUpdate(eid, new HashSet[TaskInfo]())
activeTasks += taskStart.taskInfo
```

core/src/main/scala/spark/ui/exec/ExecutorsUI.scala
```diff
@@ -132,9 +133,10 @@ private[spark] class ExecutorsUI(val sc: SparkContext) {
 
     override def onTaskEnd(taskEnd: SparkListenerTaskEnd) {
       val eid = taskEnd.taskInfo.executorId
-      if (!executorToTasksActive.contains(eid))
-        executorToTasksActive(eid) = HashSet[Long]()
-      executorToTasksActive(eid) -= taskEnd.taskInfo.taskId
+      if (!executorToTasksActive.contains(eid)) {
```
pwendell: This can also be cleaned up in the same way.

core/src/main/scala/spark/ui/jobs/JobProgressUI.scala
```diff
@@ -65,7 +65,14 @@ private[spark] class JobProgressListener extends SparkListener {
   val completedStages = ListBuffer[Stage]()
   val failedStages = ListBuffer[Stage]()
 
-  val stageToTasksActive = HashMap[Int, HashSet[Long]]()
+  var totalTime = 0L
```
pwendell, line 69: Karen, it would be good to add a brief comment explaining that these three metrics reflect totals only for _completed_ tasks.

core/src/main/scala/spark/ui/jobs/IndexPage.scala
```diff
@@ -94,13 +124,18 @@ private[spark] class IndexPage(parent: JobProgressUI) {
...
-    val (read, write) = (listener.hasShuffleRead(s.id), listener.hasShuffleWrite(s.id))
-    val shuffleInfo = (read, write) match {
-      case (true, true) => "Read/Write"
-      case (true, false) => "Read"
-      case (false, true) => "Write"
-      case _ => ""
-    }
+
+    val shuffleRead =
+      if (!listener.hasShuffleRead(s.id))
```
pwendell: Now that you are tracking `stageToShuffleRead` and `stageToShuffleWrite` in the listener, the `hasShuffleRead` and `hasShuffleWrite` functions are no longer necessary. It would be good to remove those. Then you could write this as follows:

```
val shuffleRead =  listener.stageToShuffleRead(s.id) match {
  case 0 => "" 
  case b => Utils.memoryBytesToString(b)
}
```

core/src/main/scala/spark/ui/jobs/JobProgressUI.scala
```diff
@@ -96,9 +109,10 @@ private[spark] class JobProgressListener extends SparkListener {
 
   override def onTaskStart(taskStart: SparkListenerTaskStart) {
     val sid = taskStart.task.stageId
-    if (!stageToTasksActive.contains(sid))
-      stageToTasksActive(sid) = HashSet[Long]()
-    stageToTasksActive(sid) += taskStart.taskInfo.taskId
+    if (!stageToTasksActive.contains(sid)) {
```
pwendell: you can use the same pattern above using `getOrElseUpdate` here

core/src/main/scala/spark/ui/jobs/JobProgressUI.scala
```diff
@@ -107,9 +121,10 @@ private[spark] class JobProgressListener extends SparkListener {
 
   override def onTaskEnd(taskEnd: SparkListenerTaskEnd) {
     val sid = taskEnd.task.stageId
-    if (!stageToTasksActive.contains(sid))
-      stageToTasksActive(sid) = HashSet[Long]()
-    stageToTasksActive(sid) -= taskEnd.taskInfo.taskId
+    if (!stageToTasksActive.contains(sid)) {
```
pwendell: you can use the same pattern above using `getOrElseUpdate` here

core/src/main/scala/spark/ui/jobs/JobProgressUI.scala
```diff
@@ -117,8 +132,32 @@ private[spark] class JobProgressListener extends SparkListener {
           (Some(e), e.metrics)
         case _ =>
           stageToTasksComplete(sid) = stageToTasksComplete.getOrElse(sid, 0) + 1
-          (None, Some(taskEnd.taskMetrics))
+          (None, Option(taskEnd.taskMetrics))
       }
+
+    if (!stageToTime.contains(sid)) {
```
pwendell: these three lines could be replaced with stageToTime.getOrElseUpdate(sid, 0L)

pwendell: Same with the two times below.

pwendell:
Jenkins, test this please.

pwendell:
Thanks Karen! This looks good, I've merged it.

PR #737, opened 2013-07-26
Fixing memory leak issue exposed by SHARK-116

ghost:
The fix for major memory leak as discussed in https://groups.google.com/forum/#!topic/shark-users/XT1CexIuAKk
The idea to is provide a way for Shark users get around with version 0.7 until more fundamental solution comes around in 0.8 release.

I have ran patched version of Spark cluster with Shark client for a few hours constantly running different queries and can see that GC manages to clean up everything properly on every cycle; hence the issue with KryoSerialization as a ThreadLocal seems to be addressed.

### DIFF
--- a/core/src/main/scala/spark/KryoSerializer.scala
+++ b/core/src/main/scala/spark/KryoSerializer.scala
@@ -53,9 +53,9 @@ class KryoDeserializationStream(kryo: Kryo, inStream: InputStream) extends Deser
 
 private[spark] class KryoSerializerInstance(ks: KryoSerializer) extends SerializerInstance {
 
-  val kryo = ks.kryo.get()
-  val output = ks.output.get()
-  val input = ks.input.get()
+  val kryo = ks.newKryo()
+  val output = ks.newKryoOutput()
+  val input = ks.newKryoInput()
 
   def serialize[T](t: T): ByteBuffer = {
     output.clear()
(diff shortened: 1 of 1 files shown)


ryanlecompte:
Ah, this is for the 0.7 branch. Just an FYI, but the new `KryoSerializer` that was recently merged into master does not use `ThreadLocal` any more so this shouldn't be an issue.

rxin:
Jenkins, test this please.

ghost:
Yup, I am aware about the recent master development.

The reason for this patch is to fix certain memory issues on what is
considered to be a relatively stable release of Spark.

On Fri, Jul 26, 2013 at 03:42PM, Ryan LeCompte wrote:

> Ah, this is for the 0.7 branch. Just an FYI, but the new `KryoSerializer`
> that was recently merged into master does not use `ThreadLocal` any more so
> this shouldn't be an issue. 
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/mesos/spark/pull/737#issuecomment-21652576

## 

THIS MESSAGE AND ANY ATTACHMENTS ARE CONFIDENTIAL, PROPRIETARY, AND MAY BE 
PRIVILEGED.  If this message was misdirected, WANdisco, Inc. and its 
subsidiaries, ("WANdisco") does not waive any confidentiality or privilege. 
 If you are not the intended recipient, please notify us immediately and 
destroy the message without disclosing its contents to anyone.  Any 
distribution, use or copying of this e-mail or the information it contains 
by other than an intended recipient is unauthorized.  The views and 
opinions expressed in this e-mail message are the author's own and may not 
reflect the views and opinions of WANdisco, unless the author is authorized 
by WANdisco to express such views or opinions on its behalf.  All email 
sent to or from this address is subject to electronic storage and review by 
WANdisco.  Although WANdisco operates anti-virus programs, it does not 
accept responsibility for any damage whatsoever caused by viruses being 
passed.

rxin:
Thanks, cos!

PR #738, opened 2013-07-26
Fix bug in Partition Pruning.

harsha2010:
Partition Pruned RDD should inherit partition index from parent


core/src/main/scala/spark/rdd/PartitionPruningRDD.scala
```diff
@@ -16,8 +16,9 @@ class PruneDependency[T](rdd: RDD[T], @transient partitionFilterFunc: Int => Boo
   extends NarrowDependency[T](rdd) {
 
   @transient
-  val partitions: Array[Partition] = rdd.partitions.filter(s => partitionFilterFunc(s.index))
-    .zipWithIndex.map { case(split, idx) => new PartitionPruningRDDPartition(idx, split) : Partition }
+  val partitions: Array[Partition] = rdd.partitions.
```
rxin: Try it like this style:

```
val partitions: Array[Partition] = rdd.partitions
  .zipWithIndex.filter(s => partitionFilterFunc(s._2))
  .map { case(split, idx) => new PartitionPruningRDDPartition(idx, split) : Partition }
```

core/src/test/scala/spark/PartitionPruningRDDSuite.scala
```diff
@@ -0,0 +1,28 @@
...
+      override protected def getPartitions = {
+        Array[Partition](
+            new TestPartition(1),
+            new TestPartition(2), 
+            new TestPartition(3))
+      }
+      def compute(split: Partition, context: TaskContext) = {Iterator()}
+    }
+    val prunedRDD = PartitionPruningRDD.create(rdd, {x => if (x==2) true else false})
+    println(prunedRDD.partitions.length)
```
rxin: remove the println. and add an assert on total number of partitions?

rxin:
Jenkins, test this please.

Jenkins, add to whitelist.

rxin:
Jenkins, test this please.

rxin:
Thanks, Ram!

PR #740, opened 2013-07-27
SPARK-827: MAX_TASK_FAILURES not configurable

dlyubimov:
Add configuration for MAX_TASK_FAILURES. 

Sometimes there's a need to adjust number of retries (such as in real time situations when the request must rather be dropped on the floor than fulfilled much later than SLA).


mateiz:
Jenkins, this is ok to test

mateiz:
Hey Dmitriy, this looks good, but I have two suggestions:
- Change the property name to spark.task.maxFailures
- Add this parameter into the table of config parameters in docs/configuration.md

mateiz:
I've merged this manually due to a small conflict. Thanks Dmitriy!

dlyubimov:
Don't mention it, my goals are always selfish. Thanks for developing the
awesome machine cluster tool.

PR #741, opened 2013-07-27
Fix two small usability issues

pwendell:
This fixes two issues reported by @rxin.
1. We expand an exception catch scope to include an exception that can occur in Client.scala
2. We log the `host:port` of the executor when a tasks finishes.


pwendell:
Jenkins, test this please.

PR #742, opened 2013-07-29
PySpark should set worker PYTHONPATH from SPARK_HOME instead of inheriting it from the master (SPARK-832)

JoshRosen:
This fixes [SPARK-832](https://spark-project.atlassian.net/browse/SPARK-832), an issue where the workers inherited the master's PYTHONPATH instead of setting it from SPARK_HOME.  This prevented PySpark from working in environments where the master and workers used different SPARK_HOME paths.

I could have fixed this by prepending `$SPARK_HOME/python/:` to the workers' PYTHONPATH while retaining the PYTHONPATH inheritance from the master, but the inheritance is confusing and I don't see a compelling reason to keep it (PySpark PYTHONPATH customization can still be performed by in spark-env.sh, though).

Please merge this into the master branch, too.

### DIFF
--- a/core/src/main/scala/spark/api/python/PythonWorkerFactory.scala
+++ b/core/src/main/scala/spark/api/python/PythonWorkerFactory.scala
@@ -50,6 +50,8 @@ private[spark] class PythonWorkerFactory(pythonExec: String, envVars: Map[String
         val pb = new ProcessBuilder(Seq(pythonExec, sparkHome + "/python/pyspark/daemon.py"))
         val workerEnv = pb.environment()
         workerEnv.putAll(envVars)
+        val pythonPath = sparkHome + "/python/:" + workerEnv.get("PYTHONPATH")
+        workerEnv.put("PYTHONPATH", pythonPath)
         daemon = pb.start()
 
         // Redirect the stderr to ours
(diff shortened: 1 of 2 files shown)


core/src/main/scala/spark/api/python/PythonWorkerFactory.scala
```diff
@@ -50,6 +50,8 @@ private[spark] class PythonWorkerFactory(pythonExec: String, envVars: Map[String
         val pb = new ProcessBuilder(Seq(pythonExec, sparkHome + "/python/pyspark/daemon.py"))
         val workerEnv = pb.environment()
         workerEnv.putAll(envVars)
+        val pythonPath = sparkHome + "/python/:" + workerEnv.get("PYTHONPATH")
```
pwendell: Josh, would it be better to use `os.pathsep` instead of hard coding a colon here?

pwendell: I think on Windows it might be a semi-colon, e.g.

JoshRosen: I just updated the code to use `File.pathSeparator`, but I'm not sure whether PySpark ever worked on Windows ([SPARK-649](https://spark-project.atlassian.net/browse/SPARK-649)).

mateiz:
Looks good to me, thanks.

PR #743, opened 2013-07-29
Add application metrics to standalone master

pwendell:
This patch adds metrics for each application to the master. There is also a small bug fix in `JobProgressUI.scala` that I found when writing the patch.

### DIFF
--- a/core/src/main/scala/spark/deploy/master/Master.scala
+++ b/core/src/main/scala/spark/deploy/master/Master.scala
@@ -38,6 +38,7 @@ import spark.util.AkkaUtils
 private[spark] class Master(host: String, port: Int, webUiPort: Int) extends Actor with Logging {
   val DATE_FORMAT = new SimpleDateFormat("yyyyMMddHHmmss")  // For application IDs
   val WORKER_TIMEOUT = System.getProperty("spark.worker.timeout", "60").toLong * 1000
+  val RETAINED_APPLICATIONS = System.getProperty("spark.deploy.retained_applications", "1000").toInt
 
   var nextAppNumber = 0
   val workers = new HashSet[WorkerInfo]
@@ -59,7 +60,8 @@ private[spark] class Master(host: String, port: Int, webUiPort: Int) extends Act
 
   Utils.checkHost(host, "Expected hostname")
 
-  val metricsSystem = MetricsSystem.createMetricsSystem("master")
+  val masterMetricsSystem = MetricsSystem.createMetricsSystem("master")
+  val applicationMetricsSystem = MetricsSystem.createMetricsSystem("applications")
   val masterSource = new MasterSource(this)
 
   val masterPublicAddress = {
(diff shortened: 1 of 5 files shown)


core/src/main/scala/spark/deploy/master/Master.scala
```diff
@@ -37,6 +37,7 @@ import ui.MasterWebUI
 private[spark] class Master(host: String, port: Int, webUiPort: Int) extends Actor with Logging {
   val DATE_FORMAT = new SimpleDateFormat("yyyyMMddHHmmss")  // For application IDs
   val WORKER_TIMEOUT = System.getProperty("spark.worker.timeout", "60").toLong * 1000
+  val RETAINED_APPLICATIONS = System.getProperty("spark.deploy.retained_applications", "1000").toInt
```
jerryshao: Hi @pwendell , I would like to know when retained application up to 1000, is it OK to display in JConsole or others with 1000 apps, what about memory occupation and poll latency for so many metrics?

Thanks
Jerry

core/src/main/scala/spark/deploy/master/Master.scala
```diff
@@ -58,7 +59,8 @@ private[spark] class Master(host: String, port: Int, webUiPort: Int) extends Act
 
   Utils.checkHost(host, "Expected hostname")
 
-  val metricsSystem = MetricsSystem.createMetricsSystem("master")
+  val masterMetricsSystem = MetricsSystem.createMetricsSystem("master")
+  val applicationMetricsSystem = MetricsSystem.createMetricsSystem("applications")
```
jerryshao, line 64: Currently we have only master, worker, driver, executor 4 metrics systems (instances), a new metrics system **applications** (instance) added should also update the metrics.properties to let people know how to configure instance **applications** like others. :-)

core/src/main/scala/spark/deploy/master/Master.scala
mateiz: A note on this property name -- change it to camelCase (retainedApplications) to match or other properties.

mateiz: Actually we should consider changing the metric names to camelCase too in another PR.

mateiz:
Patrick, you should check with @anfeng's team too to see whether this is what they'd like. It might be that having a separate metrics source per app is too much, and it would be better to have aggregate counts of different app statuses, etc.

core/src/main/scala/spark/deploy/master/Master.scala
pwendell: Ah I didn't realize about camelCase... I'll also fix the similar option inside of the `JobProgressUI.scala`.

pwendell: @jerryshao Thanks for the note. I'm going to pare this down to 200 apps. The polling shouldn't be a big problem since we've capped it at once per second (there aren't any expensive operations in this source). JConsole might be clunky with a large number, but people can tune this to be lower if they want. But maybe 1000 was a little overboard...

core/src/main/scala/spark/deploy/master/Master.scala
pwendell, line 64: Hey @jerryshao - I added information about the new instance to the docs. I also took this chance to do some overall writing cleanup on the docs.

jerryshao, line 64: Thanks for the update

pwendell:
Jenkins, test this please.

PR #744, opened 2013-07-29
Use Bootstrap progress bars in web UI

karenfeng:
This uses the [Twitter Bootstrap progress bars](http://twitter.github.io/bootstrap/components/#progress) in the Web UI instead of SVG, solving [SPARK-820](https://spark-project.atlassian.net/browse/SPARK-820).
stage progress bars

### DIFF
--- a/core/src/main/scala/spark/ui/exec/ExecutorsUI.scala
+++ b/core/src/main/scala/spark/ui/exec/ExecutorsUI.scala
@@ -93,8 +93,8 @@ private[spark] class ExecutorsUI(val sc: SparkContext) {
     val memUsed = sc.getExecutorStorageStatus(a).memUsed().toString
     val maxMem = sc.getExecutorStorageStatus(a).maxMem.toString
     val diskUsed = sc.getExecutorStorageStatus(a).diskUsed().toString
-    val activeTasks = listener.executorToTasksActive.get(a.toString).map(l => l.size).
-      getOrElse(0).toString
+    val activeTasks = listener.executorToTasksActive.get(a.toString).map(l => l.size)
+      .getOrElse(0).toString
     val failedTasks = listener.executorToTasksFailed.getOrElse(a.toString, 0).toString
     val completedTasks = listener.executorToTasksComplete.getOrElse(a.toString, 0).toString
     val totalTasks = listener.executorToTaskInfos(a.toString).size.toString
--- a/core/src/main/scala/spark/ui/exec/ExecutorsUI.scala
+++ b/core/src/main/scala/spark/ui/exec/ExecutorsUI.scala
@@ -93,7 +93,8 @@ private[spark] class ExecutorsUI(val sc: SparkContext) {
     val memUsed = sc.getExecutorStorageStatus(a).memUsed().toString
     val maxMem = sc.getExecutorStorageStatus(a).maxMem.toString
     val diskUsed = sc.getExecutorStorageStatus(a).diskUsed().toString
-    val activeTasks = listener.executorToTasksActive.getOrElse(a.toString, HashSet[TaskInfo]()).size.toString
+    val activeTasks = listener.executorToTasksActive.get(a.toString).map(l => l.size).
+      getOrElse(0).toString
     val failedTasks = listener.executorToTasksFailed.getOrElse(a.toString, 0).toString
     val completedTasks = listener.executorToTasksComplete.getOrElse(a.toString, 0).toString
     val totalTasks = listener.executorToTaskInfos(a.toString).size.toString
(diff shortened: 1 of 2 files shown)


pwendell:
Hey Karen,

One thing that might be cool is to somehow indicate how far along the
_started  tasks are (now that you've added the code to track that) in
addition to the finished tasks. One way would be to add a much lighter bar
in the same element (maybe very light blue or gray) that grows with the
`started + finished` tasks instead of just the `finished` ones. Then this
bar would show you how many tasks are in progress.
- Patrick

karenfeng:
Is this what you meant?
stage progress bars

pwendell:
Hey Karen,

Yep, that's the idea. Any chance you could use different colors though? Maybe the "tasks finished" bar could be the same color as before and the "tasks running" bar could be a much lighter version of that color or a light grey.

karenfeng:
Without altering the Bootstrap CSS, this is the best I could come up with:
stage progress bars

mateiz:
Karen, any reason why the rows become twice as high in your picture? Before each progress bar fit neatly into a cell and you could see more rows per page.

core/src/main/scala/spark/ui/exec/ExecutorsUI.scala
```diff
@@ -93,7 +93,7 @@ private[spark] class ExecutorsUI(val sc: SparkContext) {
     val memUsed = sc.getExecutorStorageStatus(a).memUsed().toString
     val maxMem = sc.getExecutorStorageStatus(a).maxMem.toString
     val diskUsed = sc.getExecutorStorageStatus(a).diskUsed().toString
-    val activeTasks = listener.executorToTasksActive.getOrElse(a.toString, Seq[Long]()).size.toString
+    val activeTasks = listener.executorToTasksActive.getOrElse(a.toString, HashSet[TaskInfo]()).size.toString
```
pwendell: There is slightly nicer way to write this that avoids having to create an intermediate set:

```
listener.executorToTasksActive.get(a.toString).map(l => l.size).getOrElse(0).toString
```

Calling `map` on an `Option` simply returns `None` if the option is not defined, so this works in that case. The way you have it is fine also, just wanted to let you know.

karenfeng:
@mateiz, does this look better? The Bootstrap CSS defines `margin-bottom: 20px` for progress elements so I added inline style to the element.
stage progress bars

core/src/main/scala/spark/ui/jobs/IndexPage.scala
```diff
@@ -105,17 +107,14 @@ private[spark] class IndexPage(parent: JobProgressUI) {
...
-      <rect width={width.toString} height={height.toString}
-            fill="white" stroke="rgb(51,51,51)" stroke-width="1" />
-      <rect width={completeWidth.toString} height={height.toString}
-            fill="rgb(0,136,204)" stroke="black" stroke-width="1" />
-    </svg>
+  def makeProgressBar(started: Int, completed: Int, total: Int): Seq[Node] = {
+    val completeWidth = "width: %s%%".format((completed.toDouble/total)*100)
+    val startWidth = "width: %s%%".format((started.toDouble/total)*100)
+
+    <div class="progress" style ="margin-bottom: 0px">
```
pwendell: There should be no space after style

core/src/main/scala/spark/ui/exec/ExecutorsUI.scala
```diff
@@ -93,7 +93,8 @@ private[spark] class ExecutorsUI(val sc: SparkContext) {
     val memUsed = sc.getExecutorStorageStatus(a).memUsed().toString
     val maxMem = sc.getExecutorStorageStatus(a).maxMem.toString
     val diskUsed = sc.getExecutorStorageStatus(a).diskUsed().toString
-    val activeTasks = listener.executorToTasksActive.getOrElse(a.toString, Seq[Long]()).size.toString
+    val activeTasks = listener.executorToTasksActive.get(a.toString).map(l => l.size).
```
pwendell: When wrapping like this put the period on the second line.

mateiz:
@karenfeng the new line height does look good. One more thing on the looks -- can you also try lowering the height of the progress bars? You can look at the bootstrap CSS file to see how they set it, but my guess is that you can just set the "height" attribute to something lower. I'd made them maybe 2/3 or 3/4 the height they currently are.

karenfeng:
@mateiz, is this what you meant? 
stage progress bars

mateiz:
Yeah, this new style looks good.

pwendell:
Thanks Karen. This looks good!

PR #745, opened 2013-07-29
Remove duplicate loss history in Gradient Descent

shivaram:
Also clarify that the regularization is from the previous iteration. 
@pxinghao Could you take a look to ensure this makes sense ?


shivaram:
@atalwalkar @pxinghao Added clarification that the regVal is computed using the updated weights. Let me know if this looks okay.

atalwalkar:
LGTM.  Thanks Shiv (and Xinghao for your feedback).

PR #747, opened 2013-07-30
Update the Python logistic regression example

mateiz:
Updates the example to (1) read data from a file instead of generating it and (2) batch input records into one matrix for each block for more efficient use of NumPy.

When I tested this version it was substantially faster than the original one.

### DIFF
--- a/python/examples/logistic_regression.py
+++ b/python/examples/logistic_regression.py
@@ -39,7 +39,7 @@ def readPointBatch(iterator):
     strs = list(iterator)
     matrix = np.zeros((len(strs), D + 1))
     for i in xrange(len(strs)):
-        matrix[i] = np.fromstring(strs[i], dtype=np.float32, sep = ' ')
+        matrix[i] = np.fromstring(strs[i].replace(',', ' '), dtype=np.float32, sep=' ')
     return [matrix]
 
 if __name__ == "__main__":
--- a/python/examples/logistic_regression.py
+++ b/python/examples/logistic_regression.py
@@ -16,7 +16,8 @@
 #
 
 """
-This example requires numpy (http://www.numpy.org/)
+A logistic regression implementation that uses NumPy (http://www.numpy.org) to act on batches
+of input data using efficient matrix operations.
 """
 from collections import namedtuple
 from math import exp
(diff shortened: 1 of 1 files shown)


python/examples/logistic_regression.py
```diff
@@ -27,47 +28,45 @@
...
-
+# Read a batch of points from the input file into a NumPy matrix object. We operate on batches to
+# make further computations faster.
+# The data file contains lines of the form <label> <x1> <x2> ... <xD>. We load each block of these
+# into a NumPy array of size numLines * (D + 1) and pull out column 0 vs the others in gradient().
+def readPointBatch(iterator):
+    strs = list(iterator)
+    matrix = np.zeros((len(strs), D + 1))
+    for i in xrange(len(strs)):
+        matrix[i] = np.fromstring(strs[i], dtype=np.float32, sep = ' ')
```
JoshRosen: There shouldn't be spaces around the `=` in `sep = ' '`.

JoshRosen:
This patch looks good, but two minor issues:

If I pass input of the form `<label>, <feature1> <feature2>...` with a comma separating the label and features, then it looks like numpy stops reading input at the comma without logging any sort of warning:

``` python
>>> np.fromstring("0.0,1.234 2.345 4.567", dtype=np.float32, sep = ' ')
array([ 0.], dtype=float32)
>>> np.fromstring("0.0 1.234 2.345 4.567", dtype=np.float32, sep = ' ')
array([ 0.        ,  1.23399997,  2.34500003,  4.56699991], dtype=float32)
```

I noticed this when I tried to use `spark.mllib.util.LogisticRegressionDataGenerator` to generate larger test data.

Also, it looks this gives different results than SparkHdfsLR:

```
$ ./pyspark python/examples/logistic_regression.py local lr_data.txt 10
Final w: [ 525.71624098  447.20997975  485.55775022  390.53627368  557.50278781
  635.88975317  426.77816554  622.70467501  681.53666673  648.72731912]

$ ./run spark.examples.SparkHdfsLR local lr_data.txt 10
Final w: (237.54566705221552, 278.3171786307095, 372.1335281050186, 363.3331290743261, 280.156152727127, 214.93646964595595, 330.18294311121156, 382.1408847106401, 319.4176615732616, 258.5973162137713)

$ ./run spark.examples.JavaHdfsLR local lr_data.txt 10
Final w: [237.54566705221552, 278.3171786307095, 372.1335281050186, 363.3331290743261, 280.156152727127, 214.93646964595595, 330.18294311121156, 382.1408847106401, 319.4176615732616, 258.5973162137713]
```

I think this is related [SPARK-646](https://spark-project.atlassian.net/browse/SPARK-646), an issue where we experience floating point overflow/underflow in the LR examples.  Python and Java handle those overflows differently, which may explain this behavior.

mateiz:
Josh, the different results are not due to underflow, they're due to random initialization in Python as opposed to fixed in Scala. You can try making the seed random in the Scala version too and you'll see it gives values in the same range.

I'll fixed the = sign stuff. I think underflow is less of an issue now that I operate on a whole partition at a time with Numpy -- basically it can do its own rounding in exp and such (presumably with whatever the default in libc is) instead of relying on Python's as much.

mateiz:
I've updated the example to fix the code style and also allow commas as separators now. It also works fine with extra whitespace next to the comma.

mateiz:
Merged this for now since the underflow thing doesn't happen; let me know if you see any other issues.

PR #748, opened 2013-07-30
made SimpleUpdater consistent with other updaters

atalwalkar:
@shivaram, @etrain: It seemed more complicated (and more lines of code) to have L1 and L2 updaters extend SimpleUpdater, so I made this small change instead to make the code a bit more readable.


shivaram:
Agree. I think this is better than involving class hierarchies. Thanks Ameet, I am merging this.

PR #749, opened 2013-07-30
Added property 'spark.executor.uri' for launching on Mesos.

benh:
This enables launching Spark without requiring it already be "installed". Example:

$  ./make_distribution.sh
$ mv dist spark-x.y.z
$ tar czf spark-x.y.z.tar.gz spark-x.y.z
$ hadoop fs -put spark-x.y.z.tar.gz /path/to/spark-x.y.z.tar.gz
$ SPARK_EXECUTOR_URI=hdfs://host:port/path/to/spark-x.y.z.tar.gz MESOS_NATIVE_LIBRARY=/path/to/libmesos.dylib|so MASTER=mesos://host:port ./spark-shell

A job can set the 'spark.executor.uri' directly prior to creating the context. Note that for any version of Mesos before 0.12.1 you must 'export MESOS_NATIVE_LIBRARY' in the conf/spark-env.sh _within_ the spark-x.y.z.tar.gz they upload (and it must point to the MESOS_NATIVE_LIBRARY on the machine, which may be different than the laptop running a job or REPL). This limitation will be removed in Mesos 0.12.1 and on.

### DIFF
--- a/core/src/main/scala/spark/scheduler/mesos/CoarseMesosSchedulerBackend.scala
+++ b/core/src/main/scala/spark/scheduler/mesos/CoarseMesosSchedulerBackend.scala
@@ -131,7 +131,7 @@ private[spark] class CoarseMesosSchedulerBackend(
     } else {
       // Grab everything to the first '.'. We'll use that and '*' to
       // glob the directory "correctly".
-      val basename = new File(uri).getName().split('.')(0)
+      val basename = uri.split('/').last.split('.').head
       command.setValue("cd %s*; ./run spark.executor.StandaloneExecutorBackend %s %s %s %d".format(
         basename, driverUrl, offer.getSlaveId.getValue, offer.getHostname, numCores))
       command.addUris(CommandInfo.URI.newBuilder().setValue(uri))
--- a/core/src/main/scala/spark/scheduler/mesos/CoarseMesosSchedulerBackend.scala
+++ b/core/src/main/scala/spark/scheduler/mesos/CoarseMesosSchedulerBackend.scala
@@ -110,12 +110,6 @@ private[spark] class CoarseMesosSchedulerBackend(
   }
 
   def createCommand(offer: Offer, numCores: Int): CommandInfo = {
-    val runScript = new File(sparkHome, "run").getCanonicalPath
-    val driverUrl = "akka://spark@%s:%s/user/%s".format(
-      System.getProperty("spark.driver.host"), System.getProperty("spark.driver.port"),
-      StandaloneSchedulerBackend.ACTOR_NAME)
-    val command = "\"%s\" spark.executor.StandaloneExecutorBackend %s %s %s %d".format(
-      runScript, driverUrl, offer.getSlaveId.getValue, offer.getHostname, numCores)
     val environment = Environment.newBuilder()
     sc.executorEnvs.foreach { case (key, value) =>
       environment.addVariables(Environment.Variable.newBuilder()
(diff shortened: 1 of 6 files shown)


andyk:
Jenkins, add to whitelist and ok to test.

core/src/main/scala/spark/scheduler/mesos/CoarseMesosSchedulerBackend.scala
```diff
@@ -123,7 +117,26 @@ private[spark] class CoarseMesosSchedulerBackend(
...
+      StandaloneSchedulerBackend.ACTOR_NAME)
+    val uri = System.getProperty("spark.executor.uri")
+    if (uri == null) {
+      val runScript = new File(sparkHome, "run").getCanonicalPath
+      command.setValue("\"%s\" spark.executor.StandaloneExecutorBackend %s %s %s %d".format(
+        runScript, driverUrl, offer.getSlaveId.getValue, offer.getHostname, numCores))
+    } else {
+      // Grab everything to the first '.'. We'll use that and '*' to
+      // glob the directory "correctly".
+      val basename = new File(uri).getName().split('.')(0)
```
mateiz: It's a little weird to be creating a File from a URI to parse this string but I guess it works. Is this what you guys did before?

repl/src/main/scala/spark/repl/SparkILoop.scala
```diff
@@ -831,6 +831,8 @@ class SparkILoop(in0: Option[BufferedReader], val out: PrintWriter, val master:
   var sparkContext: SparkContext = null
 
   def createSparkContext(): SparkContext = {
+    val uri = System.getenv("SPARK_EXECUTOR_URI")
+    if (uri != null) System.setProperty("spark.executor.uri", uri)
```
mateiz: Small formatting thing: place the body of the if statement in a block

core/src/main/scala/spark/scheduler/mesos/CoarseMesosSchedulerBackend.scala
benh: Happy for suggestions here. Originally I had planned on using java.net.URI(uri) but it didn't have a method to get the "file". Then I thought about using URL but "hdfs" is not a valid protocol for URL (and URI.toURL doesn't work either).

I'm not sure what we did before ... we might have done the extracting ourselves so we didn't need to determine the directory here.

repl/src/main/scala/spark/repl/SparkILoop.scala
benh: Done.

core/src/main/scala/spark/scheduler/mesos/CoarseMesosSchedulerBackend.scala
mateiz: I think that in other places, like `Utils.fetchFile`, we just did `url.split("/").last`. So let's do that here too. Also, below this, why do you do `cd %s*` instead of `cd %s` -- is it that the folder might not be called the same thing as the executor without a .tgz?

benh: I'm happy to do `url.split("/").last`, but we still need a way to remove the extension (we can't `cd` into a tarball). I tried [Files.getNameWithoutExtension](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/io/Files.html#getNameWithoutExtension%28java.lang.String%29) from Guava but this doesn't remove all of `.tar.gz` (just the `.gz`). Thus, I was splitting up to the first `.` and then `cd`ing into a directory that started with whatever I grabbed + `*`. Make sense? Happy for suggestions.

Also, I'll assume you want me to update MesosSchedulerBackend.java to use `url.split("/").last` too.

tkroman: If I got you right, maybe less error-prone (at least more definite) should be iteratively removing last after-dot part and testing with `isDirectory`? I mean, if there are two directories with versions in their names, like `name.1.0` and `name.1.1`, you aren't completely sure where you aim with `*`. And since this isn't the bottleneck of computation, we can afford to check (at most 3-5 times, I suppose) `name.1.1.tar.gz`, then - `name.1.1.tar`, then, finally, `name.1.1`.

mateiz: Ben, I think it's fine to just remove up to the first dot as you did. So simply change it to

```
new uri.split('/').last.split('.').head
```

Same in the one below. I just think this is slightly easier to understand than making a file from a potentially-non-file:// URI.

benh: SGTM. Did you mean to have `new` there though?

mateiz:
Looks good; thanks Ben!

PR #751, opened 2013-07-30
Cleaned Partitioner & PythonPartitioner source by taking out non-related logic to Utils

tkroman:
1. Replaced implicit implementation of mod(x, m) (modulo operation) in default partitioner with call to freshly defined mod(x, m) if Partitioner.scala
2. Replaced asInstanceOf/isInstanceOf with pattern matching; added a call to mod(x, m) again, thus, removed duplicated code chunks.

P.S. I also tried to replace unnecessary if-else checks with pattern-matching. Is that appropriate?

P.P.S. I messed with commits a little, so it would be better to view changed files straight.

### DIFF
--- a/core/src/main/scala/spark/Partitioner.scala
+++ b/core/src/main/scala/spark/Partitioner.scala
@@ -65,17 +65,9 @@ object Partitioner {
 class HashPartitioner(partitions: Int) extends Partitioner {
   def numPartitions = partitions
 
-  def getPartition(key: Any): Int = {
-    if (key == null) {
-      return 0
-    } else {
-      val mod = key.hashCode % partitions
-      if (mod < 0) {
-        mod + partitions
-      } else {
-        mod // Guard against negative hash codes
-      }
-    }
+  def getPartition(key: Any): Int = key match {
+    case null => 0
+    case _ => Utils.nonNegativeMod(key.hashCode, numPartitions)
   }
   
   override def equals(other: Any): Boolean = other match {
--- a/core/src/main/scala/spark/Utils.scala
+++ b/core/src/main/scala/spark/Utils.scala
@@ -756,4 +756,13 @@ private object Utils extends Logging {
     }
     return buf
   }
+
+ /* Calculates 'x' modulo 'mod', takes to consideration sign of x,
+  * i.e. if 'x' is negative, than 'x' % 'mod' is negative too
+  * so function return (x % mod) + mod in that case.
+  */
+  def nonNegativeMod(x: Int, mod: Int): Int = {
+    val rawMod = x % mod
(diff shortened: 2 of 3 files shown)


core/src/main/scala/spark/Utils.scala
```diff
@@ -756,4 +756,10 @@ private object Utils extends Logging {
     }
     return buf
   }
+  /**
+   * Calculates 'x' modulo 'm', takes to consideration sign of x
+   */
+  def mod(x: Int, m: Int): Int = {
```
mateiz: This is a good idea, but call this something like nonNegativeMod to make clear that it's not just the usual '%'

tkroman: Thank you. I changed it to signAwareModulo() and tried to refactor it all with pattern-matching.

core/src/main/scala/spark/Utils.scala
```diff
@@ -756,4 +756,10 @@ private object Utils extends Logging {
     }
     return buf
   }
+  /**
+   * Calculates 'x' modulo 'm', takes to consideration sign of x
+   */
+  def signAwareModulo(x: Int, m: Int): Int = {
```
mateiz: Ah, okay. If you don't mind though call it positiveMod or nonNegativeMod just to make it clearer what it means. Otherwise "sign-aware" isn't super clear.

tkroman: Alright, I'm doing it now. nonNegativeMod seems to do the job.

core/src/main/scala/spark/Partitioner.scala
```diff
@@ -65,16 +67,9 @@ object Partitioner {
...
-      val mod = key.hashCode % partitions
-      if (mod < 0) {
-        mod + partitions
-      } else {
-        mod // Guard against negative hash codes
-      }
+  def getPartition(key: Any): Int = key match {
+    case null => 0
+    case _ => Utils.nonNegativeMod(key.hashCode, numPartitions)
     }
```
rxin: Isn't there an extra } here?

rxin:
Hi @cdshines - thanks for doing this. Do you mind merging the commits together (either just by reapplying the changes to a clean branch, or doing a squash) so we can do small change in a single commit?

mateiz:
Jenkins, this is ok to test

mateiz:
I also agree with Reynold -- try to squash these into one commit, using git rebase or by just checking out a new branch and applying this patch.

tkroman:
@rxin, @mateiz, I'm going to squash it as soon as I get to my PC.

tkroman:
@rxin, @mateiz, I made all commits into single one and eliminated logic error so tests are passing now.

mateiz:
Thanks, going to merge it.

PR #753, opened 2013-07-30
Build changes for ML lib

shivaram:
This addes a pom.xml for building ML lib with Maven. Also I found that ML lib and bagel were not included in the SBT assembly and this change fixes that.

A couple of questions that I ran into:
@matei -- Should examples be included in sbt assembly as well ? 
@velvia -- I also found that make-distribution.sh only builds repl/assembly and so streaming, mllib etc. are not included in the dist jar.  Is this the intended behavior ?

### DIFF
--- a/mllib/pom.xml
+++ b/mllib/pom.xml
@@ -1,0 +1,165 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one or more
+  ~ contributor license agreements.  See the NOTICE file distributed with
+  ~ this work for additional information regarding copyright ownership.
+  ~ The ASF licenses this file to You under the Apache License, Version 2.0
+  ~ (the "License"); you may not use this file except in compliance with
+  ~ the License.  You may obtain a copy of the License at
+  ~
+  ~    http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.spark-project</groupId>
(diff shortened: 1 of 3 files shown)


mateiz:
Shivaram, don't add the assembly target to mllib and bagel because that will cause SBT to build a _separate_ assembly for each one. People can just add those JARs to their projects if they'd like. Maybe one question is whether we want MLlib and Bagel to be part of the repl assembly and classpath -- so far we haven't done that, but it might not be a bad idea.

shivaram:
Ah - I misunderstood the meaning of assembly. Makes sense to not generate a new assembly jar. For the repl assembly, is it enough to just add `dependsOn(mllib)` to `replSettings` ? Or is there something else which controls the classpath ?

mateiz:
Yeah, just add dependsOn(mllib).

shivaram:
Added the dependency for bagel and mllib. Also this automatically answers my question about `make-distribution.sh` as repl/assembly now includes mllib, bagel.

velvia:
Hi Shivaram,

If you add `dependsOn(mllib)` to the replSettings, that would add the MLLib
project and its dependencies to the REPL project.  It accomplishes the goal
of including mllib in the fat jar, but at a cost of bloating the deps for
the repl project.

Does the mllib library have many other dependencies?  If they are few the
easiest path would be to add the jars for mllib directly to the dist/jars
directory.

I think perhaps the best thing is to create a new project called "assembly"
or "dist" or something like that, whose goal is to aggregate all the
projects that should be included in the one fat assembly jar.   I would
remove the assembly setting from the repl project then.   This is a pretty
simple pull request and I'd be happy to submit it if there's any interest
(but I think there's still some debate about build system to use etc going
forward).

The reason why only repl/assembly was created originally was because I
wasn't sure how much to include, and that seemed like a reasonable set for
starters (and for us I guess).

cheers,
Evan

On Tue, Jul 30, 2013 at 3:30 PM, shivaram notifications@github.com wrote:

> Ah - I misunderstood the meaning of assembly. Makes sense to not generate
> a new assembly jar. For the repl assembly, is it enough to just add
> dependsOn(mllib) to replSettings ? Or is there something else which
> controls the classpath ?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/mesos/spark/pull/753#issuecomment-21827996
> .

## 

shivaram:
Thanks Evan for the clarification. mllib only has one extra dependency and that is for jblas. So my guess is that it is pretty lightweight to include this in the repl project.  (bagel has no extra dependencies)

I see your point that having an assembly project might make sense -- but I guess we can get to that later once we figure out the build & distribution mechanisms going forward.

ghost:
Maven part looks good. I will update maven assembly to include mllib, once this is commited.

mateiz:
Alright, merged this in.

Evan -- it's true that REPL didn't depend on MLlib and Bagel initially, but we want to try making it do so since those libraries won't bring a lot of dependencies and it would be cool to use them from the shell.

PR #754, opened 2013-07-31
Compression codec change

rxin:
This is based on and subsumes @lyogavin's pull request #685.
- Cleaned up the code; moved the compression codec's into a new package spark.io
- Added unit tests
- Switched the default compression codec to Snappy, since it has lower memory overhead
- Added configuration options to the documentation file
- Properly resolved dependency conflicts in SBT for Snappy
- Added the Snappy dependency in Maven


lyogavin:
Reynold, thanks very much for this. Sorry, I've been working on deploying our first release on production, haven't got time to handle this. So thanks for the help, appreciate it!

mateiz:
Hey Reynold, one question: is there a native Snappy library this depends on? If so, where will it look for it?

rxin:
Yes. 

See http://xerial.org/snappy-java/

Portable across various operating systems; Snappy-java contains native libraries built for Window/Mac/Linux (32/64-bit). At runtime, snappy-java loads one of these libraries according to your machine environment (It looks system properties, os.name and os.arch).

lyogavin:
Yes, the jar published in Maven repository already contains native libraries for Win/Mac/Linux(32/64). While for users of other OS/CPU architecture, they may need to build the native libraries from source code as introduced in http://xerial.org/snappy-java/.

mateiz:
Ah okay, that's great.

core/src/main/scala/spark/io/CompressionCodec.scala
```diff
@@ -0,0 +1,82 @@
...
+import org.xerial.snappy.{SnappyInputStream, SnappyOutputStream}
+
+
+/**
+ * CompressionCodec allows the customization of choosing different compression implementations
+ * to be used in block storage.
+ */
+trait CompressionCodec {
+
+  def compressionOutputStream(s: OutputStream): OutputStream
```
mateiz: If you don't mind, rename these to compressedOutputStream and compressedInputStream -- just seems like a better naming convention for streams.

mateiz:
Hey Reynold, FYI, there are two other places where we should use the codec instead of LZF:
- HttpBroadcast.scala
- Checkpoint.scala

Basically do a search through the codebase for LZF and make sure we're replacing it.

mateiz:
BTW apart from this the patch looks good to me.

rxin:
changed according to feedback. ptal.

streaming/src/main/scala/spark/streaming/Checkpoint.scala
```diff
@@ -49,6 +50,11 @@ class Checkpoint(@transient ssc: StreamingContext, val checkpointTime: Time)
   }
 }
 
+private[streaming]
+object Checkpoint {
```
mateiz: Using singleton objects is kind of ugly for testing purposes; could we just create a new codec on each checkpoint? It doesn't seem like that big a deal.

rxin: But then we introduce the possibility that the writer could be using one codec, and the reader uses another one.

mateiz: Hmm, I don't really get it. The reader is going to be a different instance of the program anyway, after a failure. It will have to be configured with the same codec class either way.

rxin:
Ok I removed the singleton and squashed the last two commits. 

Jenkins, test this please.

mateiz:
Alright, merged this in. Thanks @rxin and @lyogavin .

streaming/src/main/scala/spark/streaming/Checkpoint.scala
jey: Can we add a header to the output that specifies the codec used? Seems
brittle to require the configuration to be correct for simply reading the
checkpoint. (Not that it's an unreasonable assumption in >90% of cases, but
robustness is nice.)

PR #756, opened 2013-07-31
Refactored Vector.apply(length, initializer) replacing excessive code with library method

tkroman:
(also removed unused variable `ans` as minor change)

Relying on library method rather than writing one's own replacement should result in better performance (or at least the same performance while more concise code).


andyk:
Jenkins, test this please.

rxin:
Thanks!

PR #759, opened 2013-07-31
Use the Char version of split() instead of the String one in MLUtils

mateiz:
The Char one is slightly more efficient since it doesn't create a regex from the string you pass in.


shivaram:
Looks good. Merging this.

PR #760, opened 2013-07-31
Clean up web UI page headers

karenfeng:
This removes the master URL from the job UI and reduces the heading size of basic Spark pages, fixing [SPARK-828](https://spark-project.atlassian.net/browse/SPARK-828).
navbar
heading after


mateiz:
Looks good, thanks.

PR #761, opened 2013-07-31
Add data generator for K-means

mateiz:
This PR adds a data generator for Gaussian clusters and also adds support for configuring the number of concurrent of runs in KMeans.main().


mateiz:
I added one other important fix, which is to call cache() in KMeans!

shivaram:
Looks good. Merging this.

PR #762, opened 2013-07-31
Refactor SGD options into a new class.

shivaram:
This refactoring pulls out code shared between SVM, Lasso, LR into a common GradientDescentOpts class.

The main motivation behind this is two-fold:
a. Share code among different algorithms which use SGD (like Lasso, SVM etc.)
b. Set things up so that we can add different optimizers for the same algorithm in the future (for example SVM using ADMM and not SGD).

@mateiz - Could you take a look to see if this makes sense ? 

@atalwalkar @etrain - This is the first part of the refactoring we talked about -- I'll get to refactoring the algorithms, models next. Also I didn't actually create a type hierarchy for the optimizer options in this PR. That should be simple to add once we have a second optimizer though.


mllib/src/main/scala/spark/mllib/classification/LogisticRegression.scala
```diff
@@ -109,11 +83,8 @@ class LogisticRegressionLocalRandomSGD private (var stepSize: Double, var miniBa
       data,
       new LogisticGradient(),
       new SimpleUpdater(),
-      stepSize,
```
rxin, line 112: There are some merits in keeping a method like this. It is very very clear what parameters need to be set, and the user don't need to worry about what GradientDescentOpts consists.

I like having a common GradientDescentOpts, but I think we should also keep one method in SVM / LR that can users can call by specifying the parameters.

mateiz:
To keep the API more consistent across MLlib algorithms, would it be possible to make LogisticRegression and such extend GradientDescentOpts instead of taking it as a parameter? The reason is that for some algorithms, like ALS, you set the options directly on the algorithm object (e.g. new ALS().setRank(x)), but here, you'd have to make a new GradientDescentOpts and then a new Lasso or whatever. I can see the benefits of refactoring this internally but I also think it's important to keep the user API extremely simple and consistent.

atalwalkar:
@mateiz, @rxin: As noted in Shiv's comment, one goal of this PR (and an upcoming PR that Shiv mentioned) is to allow different optimization primitives to train the Logistic Regression model (and other models).  Both of your suggestions would couple LogisticRegression with SGD.  In contrast, Shiv's strategy would allow us to more easily decouple the optimization primitive from LogisticRegression.  Alternatively, we could just rename LogisticRegression to LogisticRegressionSGD and then extend GradientDescentOpts as Matei suggested.  

Regarding the relationship with ALS, this is a good point that touches upon a distinction between model parameters and algorithm parameters.  Model parameters specify a particular model to be trained, e.g., the rank parameter for the biconvex matrix factorization model or the regularization parameter for Lasso.  In contrast, algorithm parameters are specific to the particular optimization algorithm used to train a model, e.g., step size for SGD or number of iterations for ALS.  I agree that it is very important to keep things simple and be consistent in how we deal with model and algorithm parameters throughout MLlib.

mateiz:
In that case, I think we need to decide whether the top-level objects in MLlib are algorithms or "problems". For the problem of logistic regression, there are many algorithms, but it's true that here we are only implementing SGD.

My suggestion then would be to call these SGDLogisticRegression, SGDLasso, etc to make it clear that they are the algorithms. (Or LogisticRegressionSGD and such.. not sure what is better but I think it sounds better with the algorithm name first). Optionally, we could also have a top-level LogisticRegression object that will call a "default" algorithm, but since the parameters (step size, etc) are so different for each one, I'm not sure that makes a ton of sense.

atalwalkar:
Making the top level objects algorithms and using a naming convention such as SGDLasso etc. sounds good to me (the algorithm isn't actually stochastic when miniBatchFraction=1.0, so we could alternatively call them GDLasso etc., but that's a minor detail and I'm OK with going either way).

MLnick:
FWIW, I've done something very similar with my little ML package: 

``` scala
abstract class LinearModel(override val modelSettings: Settings, lossFn: Loss)
  extends BaseModel(modelSettings) {

}

case class SGDModelSettings(config: Config) extends Settings(config)  {

  val alpha = config.getDouble("alpha")
  val lambda = config.getInt("lambda")
  val iterations = config.getInt("iterations")

}

class SGDLinearModel(override val modelSettings: SGDModelSettings, lossFn: Loss = SqLoss)
  extends LinearModel(modelSettings = modelSettings, lossFn = lossFn)
```

I really like scikit-learn's API as a template: http://scikit-learn.org/stable/modules/generated/sklearn.linear_model.SGDClassifier.html#sklearn.linear_model.SGDClassifier. Here there are a few higher-level estimator classes that can be called with sensible defaults, while still leaving the underlying engine available for users who want to use it directly.

I tend to think about GLMs in particular as just an optimization engine, into which is plugged in a Loss function and a Regularization function. The `SGDLasso` and `SGDLogisticRegression` could just be convenience wrappers around `SGDModel` with the appropriate settings called? I assume this is what you refer to as the "next refactoring" for algorithms?

shivaram:
Alright, I've re-factored the algorithms to use mixins now instead of composition. Along the way, I also refactored to create a GLM class which is re-used across the algorithms and named the algorithms as `WithSGD` (for example `SVMWithSGD`.  

@mateiz @etrain Let me know if this looks close to what we discussed.

shivaram:
This is ready for review as well. I have added Java examples and unit tests for all GLM algorithms to make sure the MLLib interface works from Java. Changes for that include:
- Introduce LabeledPoint and avoid using Doubles in train arguments
- Rename train to run in class methods
- Make the optimizer a member variable of GLM to make sure the builder pattern works

MLnick:
Hi

I like this refactoring. What do you think about making the GLM framework even more "general" by having the `gradient` and the `updater` being traits too?

So you could have something like `val model = new LinearModelWithSGD with LogisticGradient with SimpleUpdater`
And if you wanted to then try L1 regularization: `val model = new LinearModelWithSGD with LogisticGradient with L1Updater`

The `updater` could also be split up further into a regularizer and a learning rate approach etc? So then you could try SVM, with L1 regularization and an adaptive learning rate:
`val model = new LinearModelWithSGD with HingeGradient with L1Regularizer with AdaptiveLearningRate`

The `predict` method could possibly be part of the `Gradient` trait since to compute loss one needs to first `predict` in any case?

mllib/src/main/scala/spark/mllib/regression/LabeledPoint.scala
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @param label Label for this data point.
+ * @param features List of features for this data point.
+ */
+case class LabeledPoint(val label: Double, val features: Array[Double]) {
+
+  /**
+   * Construct a labeled point using java.lang.Double.
+   */
+  def this(label: java.lang.Double, features: Array[Double]) = this(label.doubleValue(), features)
```
mateiz: Like my comment on ALS, it would be better to not make this a case class and have a constructor that works with primitive types. That will work just fine from both Java and Scala.

shivaram: As before, I added this as I thought `parseDouble` need to use `java.lang.Double`. Removed this constructor now

shivaram:
Hi @MLnick,
Thanks for taking a look ! I agree that having the ability to change the gradient and updater would be useful. However to make it simpler to do this from Java and to keep things consistent with our builder pattern, I've added setters for Gradient, Updater now. 

And you have a good point about separating things inside the gradient and updater -- I think we'll need to think a little bit more on how to reorganize things for that and how to share a loss function with say other optimization routines. This might  not make it for the 0.8 release, but we should revisit this soon.

etrain:
This is looking very good. 

Main points:
1) We should add a class for unregularized Linear Regression. I'm happy to do that at the same time I make Ridge follow this format.
2) We should probably add a "Point" class that corresponds with the LabeledPoint class for the sake of consistency.
case class Point(val features: Array[Double])
2) The java support is really making this verbose.
3) Hmm... there's got to be a nice way to avoid rewriting the prediction method of GLM. Maybe each model provides a margin function that evaluates what to do with the dot product of the weights and the data? That way for the common case, they do nothing, but for the SVM and LR case they're just wrapped in the sigmoid and signum functions, respectively?
4) I noticed an inconsistency with SVM vs. the rest of the models in terms of input expectation. We should almost certainly change this.

Minor nitpick:
The JavaLassoSuite should go in src/test/java/spark/mllib/regression - particularly since we're going to have several more of those.

mllib/src/main/scala/spark/mllib/regression/GeneralizedLinearAlgorithm.scala
```diff
@@ -0,0 +1,120 @@
...
+    predictPoint(dataMat, weightsMatrix, intercept)
+  }
+}
+
+/**
+ * GeneralizedLinearAlgorithm abstracts out the training for all GLMs. 
+ * This class should be extended with an Optimizer to create a new GLM.
+ */
+abstract class GeneralizedLinearAlgorithm[M](implicit
+    methodEv: M <:< GeneralizedLinearModel)
```
mateiz: Are you sure you want to be using the view bound `<:<` here? It means that there's a conversion from M to GeneralizedLinearModel. If you just want M to be a subtype of GeneralizedLinearModel, you can do

```
abstract class GeneralizedLinearAlgorithm[M <: GeneralizedLinearModel] {
  ...
}
```

mllib/src/main/scala/spark/mllib/regression/GeneralizedLinearAlgorithm.scala
```diff
@@ -0,0 +1,120 @@
...
+  def run(input: RDD[LabeledPoint]) : M = {
+    val nfeatures: Int = input.first().features.length
+    val initialWeights = Array.fill(nfeatures)(1.0)
+    run(input, initialWeights)
+  }
+
+  def run(
+      input: RDD[LabeledPoint],
+      initialWeights: Array[Double])
+    : M = {
```
mateiz: This could probably fit on one line, but if it doesn't, do

```
def run(
    input: RDD[LabeledPoint],
    initialWeights: Array[Double])
  : M =
{
```

(that is, move the brace down onto a line by itself)

mateiz:
Regarding Evan's comments -- I don't think we should have `Point` instead of `Array[Double]`. It just makes it harder to call from other languages. For example the Python, R, etc libraries for binding to Java generally have support for converting an array in their language to a Java array.

mllib/src/main/scala/spark/mllib/regression/GeneralizedLinearAlgorithm.scala
shivaram: You are right, it fits in one line. Fixed now

mllib/src/main/scala/spark/mllib/regression/GeneralizedLinearAlgorithm.scala
shivaram: Right -- this should be just `M <: GenearlizedLinearModel`. This was an unnecessary remnant from things I was trying out while testing the scala / java inter-op.

shivaram:
Thanks @mateiz and @etrain for taking a look. I fixed the inline comments. Some higher level replies
-  Agree that we should have an unregularized LR. It'll be great if you add this in the RidgeRegression change.
-  I think we should postpone the `Point` class until we need it to see if it is required. My view is that we should create these case classes only when we want to avoid `Tuple2` or `Tuple3` from showing up in Java.
-  Regarding the predict method, I don't really see a huge benefit from further abstraction to a margin. I think for the next cut, we should change the internal structure a little bit so that we take in a `link` function that gets reused in the loss calculation and prediction. We can chat more about this some time.
-  Nice catch on the SVM label inconsistency. I'll fix this in a separate PR by adding an auto-conversion.

mateiz:
@etrain should we merge this then? It looks good from my side. Feel free to hit Merge.

etrain:
All your comments make sense - this looks good to me. Merging now.

PR #763, opened 2013-07-31
SPARK-842. Maven assembly is including examples libs and dependencies

ghost:
Also, this commit includes (and depends upon) mllib module as represented in
  https://github.com/mesos/spark/pull/753

### DIFF
--- a/assembly/pom.xml
+++ b/assembly/pom.xml
@@ -62,43 +62,31 @@
       <groupId>org.spark-project</groupId>
       <artifactId>spark-core</artifactId>
       <classifier>${classifier.name}</classifier>
-      <version>0.8.0-SNAPSHOT</version>
+      <version>${project.version}</version>
     </dependency>
     <dependency>
       <groupId>org.spark-project</groupId>
       <artifactId>spark-bagel</artifactId>
       <classifier>${classifier.name}</classifier>
-      <version>0.8.0-SNAPSHOT</version>
+      <version>${project.version}</version>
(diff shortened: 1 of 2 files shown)


assembly/pom.xml
```diff
@@ -62,43 +62,43 @@
...
       <groupId>org.spark-project</groupId>
-      <artifactId>spark-examples</artifactId>
+      <artifactId>spark-mllib</artifactId>
       <classifier>${classifier.name}</classifier>
-      <version>0.8.0-SNAPSHOT</version>
+      <version>${project.version}</version>
     </dependency>
     <dependency>
       <groupId>org.spark-project</groupId>
       <artifactId>spark-examples</artifactId>
```
mateiz: Looks like this and the one below are replicas of the spark-examples dependency for some reason. You should remove them.

ghost: One of them is for javadoc
Another is for sources

they are not dups.

mateiz: Oh, I see. Either way though we shouldn't include or depend on the spark-examples javadoc and sources.

mateiz: Maybe you meant to change those to mllib?

ghost:
Removed examples completely.

assembly/pom.xml
ghost: mllib is done separately. I have removed examples

mateiz:
Okay, thanks!

PR #767, opened 2013-08-01
Update Transformations title to More On RDD Operations to match content

stoneyv:
To match quick guide actions and transformation commands.


mateiz:
Looks  good, thanks!

PR #769, opened 2013-08-01
SPARK-847 + SPARK-845: Zombie workers and negative cores

markhamstra:
I can split these up if necessary, but they are closely related.

Also fixes the milliseconds instead of seconds logging buglet for timed-out workers.

### DIFF
--- a/core/src/main/scala/spark/deploy/master/ApplicationInfo.scala
+++ b/core/src/main/scala/spark/deploy/master/ApplicationInfo.scala
@@ -52,8 +52,10 @@ private[spark] class ApplicationInfo(
   }
 
   def removeExecutor(exec: ExecutorInfo) {
-    executors -= exec.id
-    coresGranted -= exec.cores
+    if (executors.contains(exec.id)) {
+      executors -= exec.id
+      coresGranted -= exec.cores
+    }
   }
 
   def coresLeft: Int = desc.maxCores - coresGranted
--- a/core/src/main/scala/spark/deploy/DeployMessage.scala
+++ b/core/src/main/scala/spark/deploy/DeployMessage.scala
@@ -109,6 +109,7 @@ private[deploy] object DeployMessages {
   }
 
   //  WorkerWebUI to Worker
+
   case object RequestWorkerState
 
   // Worker to WorkerWebUI
@@ -120,4 +121,9 @@ private[deploy] object DeployMessages {
     Utils.checkHost(host, "Required hostname")
     assert (port > 0)
   }
+
+  // Actor System to Master
+
+  case object CheckForWorkerTimeOut
+
 }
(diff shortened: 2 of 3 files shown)


core/src/main/scala/spark/deploy/master/ApplicationInfo.scala
```diff
@@ -51,8 +51,10 @@ private[spark] class ApplicationInfo(
   }
 
   def removeExecutor(exec: ExecutorInfo) {
-    executors -= exec.id
-    coresGranted -= exec.cores
+    if (executors.contains(exec.id)) {
+      executors -= exec.id
+      coresGranted -= exec.cores
+    }
   }
```
mridulm, line 59: This would need to be within a lock

markhamstra, line 59: I don't believe so, since it is only evaluated within the Master actor's receive.

markhamstra, line 59: After a closer look, I'll take that back.  This function gets called from `Master.removeWorker`, and `removeWorker` itself is mostly called safely and sequentially from within the `Master` actor's `receive` block (handling `Terminated`, `RemoteClientDisconnected` or `RemoteClientShutdown` messages); but `Master.removeWorker` is also called directly by the Actor System from `timeOutDeadWorkers`, which is scheduled in `Master.preStart`.  That means that there already were concurrency and race issues in `removeWorker`, and this pull request adds to them with the check-then-set in `ApplicationInfo.removeExecutor`.

@mateiz It looks to me like `Master.preStart` should schedule a new `WorkerTimedOut` message to be sent to self (i.e. `Master`) every `WORKER_TIMEOUT` interval, and that `timeOutDeadWorkers` should be called from within the `receive` block's `case` for that message.  That means that there could be messages queued in `Master`'s mailbox ahead of a `WorkerTimedOut` that were issued without the sender's knowledge that one or more workers have timed-out and are now `DEAD`, but I don't think that there was ever a guarantee or expectation of that information being present and synchronized at senders of messages to `Master`, so I don't think that queuing `WorkerTimedOut` messages instead of jumping the line with a separate `system` thread directly running `timeOutDeadWorkers` introduces any new problems.

mridulm, line 59: Locking is not just to prevent concurrent lookups/updates (which is required in use of timeOutDeadWorkers), but also to ensure visibility of updates between threads : this (second issue) is a general problem in other parts of spark too btw ...
We had done some work on a private branch to alleviate this, but abandoned it due to lack of time.

You are right that this is an existing issue - the compare and update just made me notice it here.

mateiz:
Mark, this looks good to me modulo the lock issue. For that, can you replace the scheduled call to timeoutDeadWorkers with a scheduled message to the Master actor? That way all updates to these data structures will happen within that actor, where they're guaranteed to be consistent. In general it's a bad idea to mix actor code with locks -- we're using actors specifically to avoid messing with locks.

mateiz:
Jenkins, add to whitelist

mateiz:
Jenkins, test this please

mateiz:
Looks good. Thanks Mark!

PR #770, opened 2013-08-02
Simple PageRank algorithm implementation in Java for SPARK-760

ghost:


mateiz:
Jenkins, this is ok to test

ghost:
code is compiled and tested on simple datasets with correct output.

ghost:
Can someone please take a look at this? Thanks a lot.

examples/src/main/java/spark/examples/JavaPageRank.java
```diff
@@ -0,0 +1,129 @@
...
+  //     URL         neighbor URL
+  //     URL         neighbor URL
+  //     ...
+  JavaRDD<String> lines = ctx.textFile(args[1], 1).cache();
+
+  // Loads all URLs from input file and initialize their neighbors.
+  JavaPairRDD<String, List<String>> links = lines.map(new PairFunction<String, String, String>() {
+    @Override
+    public Tuple2<String, String> call(String s) {
+      return new Tuple2<String, String>(s.split("\\s+")[0], s.split("\\s+")[1]);
```
mateiz: This is somewhat inefficient.. you should split only once and take both parts of the array then

examples/src/main/java/spark/examples/JavaPageRank.java
```diff
@@ -0,0 +1,129 @@
...
+
+  JavaSparkContext ctx = new JavaSparkContext(args[0], "JavaPageRank",
+    System.getenv("SPARK_HOME"), System.getenv("SPARK_EXAMPLES_JAR"));
+
+  // Loads in input file. It should be in format of:
+  //     URL         neighbor URL
+  //     URL         neighbor URL
+  //     URL         neighbor URL
+  //     ...
+  JavaRDD<String> lines = ctx.textFile(args[1], 1).cache();
```
mateiz: It would be better to cache the result of `groupByKey` below than to cache the text file

examples/src/main/java/spark/examples/JavaPageRank.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }).distinct().groupByKey();
+
+  // Loads all URLs with other URL(s) link to from input file and initialize ranks of them to one.
+  JavaPairRDD<String, Double> ranks = lines.map(new PairFunction<String, String, Double>() {
+    @Override
+    public Tuple2<String, Double> call(String s) {
+      return new Tuple2<String, Double>(s.split("\\s+")[1], 1.0);
+    }
+  }).distinct();
```
mateiz: It would be better to compute this as `links.mapValues( rank => 1.0 )`

mateiz: (The reason is that it would cause links and ranks to be co-partitioned and save on some joins -- see WikipediaPageRankStandalone.scala)

examples/src/main/java/spark/examples/JavaPageRank.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+      });
+
+      // Re-calculates URL ranks based on neighbor contributions.
+      ranks = contribs.groupByKey().mapValues(new Function<List<Double>, Double>() {
+        @Override
+        public Double call(List<Double> cs) throws Exception {
+          return 0.15 + sum(cs) * 0.85;
+        }
+        });
```
mateiz: Indentation here seems wrong -- do you have tabs by any chance?

examples/src/main/java/spark/examples/JavaPageRank.java
ghost: done

examples/src/main/java/spark/examples/JavaPageRank.java
ghost: done

examples/src/main/java/spark/examples/JavaPageRank.java
ghost: done

examples/src/main/java/spark/examples/JavaPageRank.java
ghost: fixed

ghost:
@mateiz , I updated the code with your suggestions and ran tests over it. Can you please take another look? Thank you.

ghost:
@mateiz , really appreciate if you can take another look at the updated code with changes based on your previous comments.

examples/src/main/java/spark/examples/JavaPageRank.java
```diff
@@ -0,0 +1,129 @@
...
+        public Double call(List<Double> cs) throws Exception {
+          return 0.15 + sum(cs) * 0.85;
+        }
+      });
+    }
+
+    // Collects all URL ranks and dump them to console.
+    List<Tuple2<String, Double>> output = ranks.collect();
+    for (Tuple2 tuple : output) {
+        System.out.println(tuple._1 + " has rank: " + round((Double)(tuple._2), 2) + ".");
```
mateiz: Instead of having a round() method and calling it, just use System.out.printf

examples/src/main/java/spark/examples/JavaPageRank.java
```diff
@@ -0,0 +1,129 @@
...
+            }
+
+            return results;
+          }
+      }).map(new PairFunction<Tuple2<String, Double>, String, Double>() {
+        @Override
+        public Tuple2<String, Double> call(Tuple2<String, Double> s) {
+          return s;
+        }
+      });
```
mateiz, line 97: What is the purpose of this extra map()?

mateiz:
Hey, sorry for the delay; made a couple of other comments.

examples/src/main/java/spark/examples/JavaPageRank.java
ghost, line 97: since the flatmap function generates JavaRDD[Tuple2<String, Double>] and I want them to be in JavaPairRDD[String, Double] before the next stage of the code, therefore, the map function is used. If there are better syntactical support for this, I am more than glad to know and change the code.

examples/src/main/java/spark/examples/JavaPageRank.java
ghost: Done.

ghost:
No worries, I just do not want the code to stay there too long. Thanks for your review.

examples/src/main/java/spark/examples/JavaPageRank.java
mateiz, line 97: Ah, got it. You can actually get around this by passing `new PairFlatMapFunction`; when seeing that function type, `flatMap` will return a `JavaPairRDD`.

mateiz:
Cool, thanks a lot! Merging this in.

PR #771, opened 2013-08-02
Allow user to remove broadcast variables when they are no longer used

RongGu:
In Spark, users can create broadcast variables to share read-only variables across tasks or operations,especially when they are large. However, the current Spark does not allow users to remove those variables in one SparkContext. This becomes a major issue for long running Shark servers which uses one SparkContext. To address this issue, this patch allows users to remove broadcast variables when they are no longer used. To remove a broadcast variable, users only need to call the Broadcast.rm(toClearSource:Boolean) methond, the broadcast variable across the slaves will be deleted. If toClearSource is set true, data source (e.g., file used by HttpServer) will be deleted too.

### DIFF
--- a/core/src/main/scala/spark/broadcast/BitTorrentBroadcast.scala
+++ b/core/src/main/scala/spark/broadcast/BitTorrentBroadcast.scala
@@ -21,7 +21,8 @@ private[spark] class BitTorrentBroadcast[T](@transient var value_ : T, isLocal: 
   def blockId: String = "broadcast_" + id
 
   MultiTracker.synchronized {
-    SparkEnv.get.blockManager.putSingle(blockId, value_, StorageLevel.MEMORY_AND_DISK, false)
+    //Let BlockManagerMaster know that we have the broadcast block for its latter notification us to remove.
+    SparkEnv.get.blockManager.putSingle(blockId, value_, StorageLevel.MEMORY_AND_DISK, true)
   }
 
   @transient var arrayOfBlocks: Array[BroadcastBlock] = null
@@ -57,6 +58,23 @@ private[spark] class BitTorrentBroadcast[T](@transient var value_ : T, isLocal: 
   // Must call this after all the variables have been created/initialized
   if (!isLocal) {
     sendBroadcast()
+  }
+  
+  override def rm(toClearSource: Boolean = false) {
+    logInfo("Remove broadcast variable " + blockId)
+    SparkEnv.get.blockManager.master.removeBlock(blockId)
+    SparkEnv.get.blockManager.removeBlock(blockId, false)
+    if(toClearSource)
+      clearBlockSource()
+  }
+  
+  def clearBlockSource(){
+    arrayOfBlocks = null
+    hasBlocksBitVector = null
+    numCopiesSent = null
+    listOfSources = null
+    serveMR = null
+    guideMR = null
   }
 
   def sendBroadcast() {
@@ -116,7 +134,7 @@ private[spark] class BitTorrentBroadcast[T](@transient var value_ : T, isLocal: 
   private def readObject(in: ObjectInputStream) {
     in.defaultReadObject()
     MultiTracker.synchronized {
-      SparkEnv.get.blockManager.getSingle(blockId) match {
+      SparkEnv.get.blockManager.getSingleLocal(blockId) match {
         case Some(x) =>
           value_ = x.asInstanceOf[T]
 
@@ -139,8 +157,9 @@ private[spark] class BitTorrentBroadcast[T](@transient var value_ : T, isLocal: 
           val receptionSucceeded = receiveBroadcast(id)
           if (receptionSucceeded) {
             value_ = MultiTracker.unBlockifyObject[T](arrayOfBlocks, totalBytes, totalBlocks)
+            //Let BlockManagerMaster know that we have the broadcast block for its latter notification us to remove.
             SparkEnv.get.blockManager.putSingle(
-              blockId, value_, StorageLevel.MEMORY_AND_DISK, false)
+              blockId, value_, StorageLevel.MEMORY_AND_DISK, true)
           }  else {
             logError("Reading broadcast variable " + id + " failed")
           }
--- a/core/src/main/scala/spark/broadcast/BitTorrentBroadcast.scala
+++ b/core/src/main/scala/spark/broadcast/BitTorrentBroadcast.scala
@@ -11,7 +11,7 @@ import scala.math
 import spark._
 import spark.storage.StorageLevel
 
-private[spark] class BitTorrentBroadcast[T](@transient var value_ : T, isLocal: Boolean, id: Long)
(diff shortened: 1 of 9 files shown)


core/src/main/scala/spark/broadcast/TreeBroadcast.scala
```diff
@@ -46,6 +47,21 @@ extends Broadcast[T](id) with Logging with Serializable {
   if (!isLocal) {
     sendBroadcast()
   }
+  
+  override def rm(toClearSource: Boolean = false) {
```
rxin: Can we rename this function to remove, and toClearSource to releaseSource?

core/src/main/scala/spark/broadcast/TreeBroadcast.scala
```diff
@@ -46,6 +47,21 @@ extends Broadcast[T](id) with Logging with Serializable {
...
+    logInfo("Remove broadcast variable " + blockId)
+    SparkEnv.get.blockManager.master.removeBlock(blockId)
+    SparkEnv.get.blockManager.removeBlock(blockId, false)
+    if(toClearSource)
```
rxin: Use 

```
if (toClearSource) {
  clearBlockSource()
}
```

core/src/main/scala/spark/broadcast/HttpBroadcast.scala
```diff
@@ -21,24 +21,38 @@ extends Broadcast[T](id) with Logging with Serializable {
...
 
   if (!isLocal) { 
     HttpBroadcast.write(id, value_)
   }
+  
+  override def rm(toClearSource: Boolean = false) {
+    logInfo("Remove broadcast variable " + blockId)
+    SparkEnv.get.blockManager.master.removeBlock(blockId)
+    SparkEnv.get.blockManager.removeBlock(blockId, false)
+    if(toClearSource){
```
rxin: Add a space after if

core/src/main/scala/spark/broadcast/BitTorrentBroadcast.scala
```diff
@@ -21,7 +21,8 @@ private[spark] class BitTorrentBroadcast[T](@transient var value_ : T, isLocal:
   def blockId: String = "broadcast_" + id
 
   MultiTracker.synchronized {
-    SparkEnv.get.blockManager.putSingle(blockId, value_, StorageLevel.MEMORY_AND_DISK, false)
+    //Let BlockManagerMaster know that we have the broadcast block for its latter notification us to remove.
+    SparkEnv.get.blockManager.putSingle(blockId, value_, StorageLevel.MEMORY_AND_DISK, true)
```
rxin: There might be a performance problem here.

Spark actually uses broadcast variable to broadcast the JobConf in HadoopRDD to avoid having that in the task closure (10KB). If every worker has to send a message back to the master, it might slow things down since every HadoopRDD we create will need to do that ... 

Ideally, we should track the memory usage of broadcast variables, but I am not sure what the best way to do this is.

RongGu: Thank you for reviewing my pull request, Reynold. I have followed your advice as blow.
1.Change the name of methods.
Answer: I have changed their names in the last commit as you suggested.

2.Performance issues for HadoopRDD.
Answer: This is a good concerning. In fact, the message size of a blockinfo transferred is no large. To make it flexible, I added a boolean parameter 'tellMaster (default is true)' for broadcast variables. When set false, the broadcast variable will not be reportted to master and not be removed at slave machines in a SparkContext, this is suitable for small-size broadcast variables. Users can choose the broadcast type at their wish.

3.Track the memory usages of broadcast. 
Answer: This a good idea. It can lead making a automatic memory cleaner for broadcast variables. Nevertheless, the purpose of this patch is providing a removing broadcast API to users. These two things do not conflict in essence. For memory cleanup tasks, the lesson I learned is that, whatever program-monitoring mechanisms seems not better than clear the memory explicitly by users if possible. GC can not always be in time and it has overhead costs. Moreover, in this case, it is hard to determine whether a broadcast needed be used by users any more. On the other side, it is a issue to leave large unused broadcast variables in memory, and users have no means to handle that. Therefore, we provide a explicit removing broadcast method to users. It solves my problem in practise. 

Thank you for commenting again and I will continue thinking about the automatic cleaner based on tracking memory usage you put forwarded here , it's interesting.

jerryshao:
Hi @RongGu , AFAIK Spark already has a time based automatic clean way in HttpBroadcast when spark.cleaner.ttl is enabled, this can mostly clean JobConf in HadoopRDD,  But this mechanism has a issue with Spark Streaming (https://spark-project.atlassian.net/browse/STREAMING-38?jql=project%20%3D%20STREAMING), it would be a great help to use a memory track way to clean the broadcast var automatically, not the time based way.

core/src/main/scala/spark/broadcast/BitTorrentBroadcast.scala
RongGu: Thank you for reviewing my pull request, Reynold @rxin . I have followed your advice as blow.
1.Change the name of methods.
Answer: I have changed their names in the last commit as you suggested.

2.Performance issues for HadoopRDD.
Answer: This is a good concerning. In fact, the message size of a blockinfo transferred is no large. To make it flexible, I added a boolean parameter 'tellMaster (default is true)' for broadcast variables. When set false, the broadcast variable will not be reportted to master and not be removed at slave machines in a SparkContext, this is suitable for small-size broadcast variables. Users can choose the broadcast type at their wish.

3.Track the memory usages of broadcast.
Answer: This a good idea. It can lead making a automatic memory cleaner for broadcast variables. Nevertheless, the purpose of this patch is providing a removing broadcast API to users. These two things do not conflict in essence. For memory cleanup tasks, the lesson I learned is that, whatever program-monitoring mechanisms seems not better than clear the memory explicitly by users if possible. GC can not always be in time and it has overhead costs. Moreover, in this case, it is hard to determine whether a broadcast needed be used by users any more. On the other side, it is a issue to leave large unused broadcast variables in memory, and users have no means to handle that. Therefore, we provide a explicit removing broadcast method to users. It solves my problem in practise.

Thank you for commenting again and I will continue thinking about the automatic cleaner based on tracking memory usage you put forwarded here , it's interesting.

RongGu:
Hi, @jerryshao , Thanks for your comment. It is nice to make a automatic memory cleaner for broadcast variables. Nevertheless, the purpose of this patch is providing a removing broadcast API to users. These two things do not conflict in essence. For memory cleanup tasks, the lesson I learned is that, whatever program-monitoring mechanisms seems not better than clear the memory explicitly by users if possible. GC can not always be in time and it has overhead costs. Moreover, in this case, it is hard to determine whether a broadcast needed be used by users any more，TTL may lead to error as the issue in the Spark Streaming said. On the other side, it is a problem to leave large unused broadcast variables in memory, and users have no means to handle that. Therefore, ,here we provide a explicit removing broadcast method to users.

PR #774, opened 2013-08-02
Show user-defined job name in UI

pwendell:
This patch displays the "spark.job.description" property in the web UI, to let users or higher level applications (such as Shark) annotate stages with job-level information.

It also does some code clean-up (`addLocalProperties` --> `addLocalProperty`).

image


mateiz:
LGTM, thanks.

PR #775, opened 2013-08-03
Log the launch command for Spark daemons

pwendell:
For debugging and analysis purposes, it's nice to have the exact command
used to launch Spark contained within the logs. This adds the necessary
hooks to make that possible.

### DIFF
--- a/core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
+++ b/core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
@@ -164,8 +164,7 @@ private[spark] class ExecutorRunner(
       env.put("SPARK_LAUNCH_WITH_SCALA", "0")
       process = builder.start()
 
-      val header = "Spark Executor Command: %s\n======================================\n\n"
-        .format(command.mkString(" "))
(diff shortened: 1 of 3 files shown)


rxin:
Is this only for Spark daemons?

When I filed the ticket, I think I might've been referring to the applications too.

pwendell:
@rxin Good idea. I added this now for the executors as well.

core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
```diff
@@ -161,9 +164,15 @@ private[spark] class ExecutorRunner(
       env.put("SPARK_LAUNCH_WITH_SCALA", "0")
       process = builder.start()
 
+      val header = "Spark Executor Command: %s\n%s\n\n".format(command.mkString(" "), "=" * 40)
```
rxin: To make the command re-runnable, let's wrap all parts of the command with quotes.

pwendell:
Hey @rxin this prints the arguments with quotes (and escapes any existing quotes in them).

pwendell:
Jenkins, test this please.

PR #776, opened 2013-08-03
adding matrix factorization data generator

gingsmith:


mllib/src/main/scala/spark/mllib/util/MFDataGenerator.scala
```diff
@@ -0,0 +1,105 @@
...
+   * @param sc SparkContext to use for creating the RDD.
+   * @param outputPath Directory to save output.
+   * @param m Number of rows in data matrix.
+   * @param n Number of columns in data matrix.
+   * @param rank Underlying rank of data matrix.
+   * @param tr_samp_fact Oversampling factor.
+   * @param noise Boolean value - whether to add gaussian noise to training data.
+   * @param sigma Standard deviation of added gaussian noise.
+   * @param test Boolean value - whether to create testing RDD.
+   * @param te_samp_fact Percentage of training data to use as test data.
```
mateiz: Move this doc comment to the class as a whole, not the main method. Also, since the parameters here are command-line arguments but not arguments to the class, it's probably better to turn them into a list instead of `@param` so as not to confuse tools.

mllib/src/main/scala/spark/mllib/util/MFDataGenerator.scala
```diff
@@ -0,0 +1,105 @@
...
+    		.toInt
+    	val test_omega = shuffled.slice(sampsize,sampsize+test_sampsize)
+    	val test_ordered = test_omega.sortWith(_ < _).toArray
+    	val testData: RDD[(Int, Int, Double)] = sc.parallelize(test_ordered)
+    		.map(x=> (fullData.indexRows(x-1),fullData.indexColumns(x-1),fullData.get(x-1)))
+      testData.map(x => x._1 + "," + x._2 + "," + x._3).saveAsTextFile(outputPath)
+    }
+        
+	sc.stop()
+  }
```
mateiz, line 112: Formatting seems wrong here.. are there tabs by any chance? You should only use spaces

mllib/src/main/scala/spark/mllib/util/MFDataGenerator.scala
```diff
@@ -0,0 +1,112 @@
...
+* 	sparkMaster 		 (String) The master URL.
+* 	outputPath  		 (String) Directory to save output.
+* 	m 					 		 (Int) Number of rows in data matrix.
+* 	n 							 (Int) Number of columns in data matrix.
+* 	rank 					 (Int) Underlying rank of data matrix.
+* 	tr_samp_fact 	 (Double) Oversampling factor.
+* 	noise 					 (Boolean) Whether to add gaussian noise to training data.
+* 	sigma 					 (Double) Standard deviation of added gaussian noise.
+* 	test 					 (Boolean) Whether to create testing RDD.
+* 	te_samp_fact 	 (Double) Percentage of training data to use as test data.
```
mateiz: Hey Ginger, instead of using tabs here, please use spaces. The whole file should have no tabs to make sure it displays the same way in any editor.

mllib/src/main/scala/spark/mllib/util/MFDataGenerator.scala
```diff
@@ -0,0 +1,112 @@
...
+*/
+
+object MFDataGenerator{
+
+  def main(args: Array[String]) {
+    if (args.length != 10) {
```
mateiz: Won't this exit when the user is providing fewer than 10 arguments, so that the default values below are never used?

mllib/src/main/scala/spark/mllib/util/MFDataGenerator.scala
```diff
@@ -0,0 +1,112 @@
...
+      println("Usage: MFGenerator " +
+        "<master> <output_dir> <m> <n> <rank> <tr_samp_fact> <noise> <sigma> <test> <te_samp_fact>")
```
mateiz: Put the last eight arguments in square brackets (`[` and `]`) to show they're optional.

mllib/src/main/scala/spark/mllib/util/MFDataGenerator.scala
```diff
@@ -0,0 +1,112 @@
...
+    		.min(scala.math.round(sampsize*te_samp_fact),scala.math.round(mn-sampsize))
+    		.toInt
+    	val test_omega = shuffled.slice(sampsize,sampsize+test_sampsize)
+    	val test_ordered = test_omega.sortWith(_ < _).toArray
+    	val testData: RDD[(Int, Int, Double)] = sc.parallelize(test_ordered)
+    		.map(x=> (fullData.indexRows(x-1),fullData.indexColumns(x-1),fullData.get(x-1)))
+      testData.map(x => x._1 + "," + x._2 + "," + x._3).saveAsTextFile(outputPath)
+    }
+        
+	sc.stop()
```
mateiz: This seems indented wrong, shouldn't it have two spaces before?

mllib/src/main/scala/spark/mllib/util/MFDataGenerator.scala
```diff
@@ -0,0 +1,112 @@
...
+    val mn = m*n
+    val shuffled = rand.shuffle(1 to mn toIterable)
+
+    val omega = shuffled.slice(0,sampsize)
+    val ordered = omega.sortWith(_ < _).toArray
+    val trainData: RDD[(Int, Int, Double)] = sc.parallelize(ordered)
+    		.map(x => (fullData.indexRows(x-1),fullData.indexColumns(x-1),fullData.get(x-1)))
+
+    // optionally add gaussian noise
+    if(noise){
```
mateiz: Should be `if (noise) {` (with spaces); same for the if statement below.

mateiz:
Made some other formatting comments. One other thing is that when you have commas, operators, etc, try to put a space around them -- in some places that's not being done. Basically try to follow the Scala style guide here: http://docs.scala-lang.org/style/overview.html to make sure your code matches the rest of the project.

gingsmith:
Thanks Matei. I'll fix the input and formatting issues. I was looking at the other data generators (KMeansDataGenerator, LassoDataGenerator, LogisticRegressionDataGenerator, RidgeRegressionDataGenerator, and SVMDataGenerator) as a reference for input format, and so we should probably update those as well.

mateiz:
Great, thanks for fixing it! It's quite possible that some of the other generators are out of sync; we might want to do a pass over them to make them consistent.

PR #777, opened 2013-08-05
Add properties and build sections to set java source version and UTF-8 e...

stoneyv:
...ncoding

Added properties and build sections to example pom.xml file to set the java version to 1.5 and encoding to UTF-8.  This eliminates an error and a warning.


mateiz:
What is the error and warning? It seems weird to have to specify these properties manually.

mateiz:
That's really weird, because that certainly hasn't happened to me. Maybe you have some local Maven configuration files on your machine that set the source level to something else?

I'm also using Maven 3.0.4 by the way, on a mac.

stoneyv:
I am switching back to the openjdk and double checking my environment before I use up more of your time

stoneyv:
I was able to eliminate the source level error at compile by installing maven from apache instead of the package distributed by Ubuntu 12.04 LTS repositories.  It might have been the /etc/settings.xml profile that was set to 1.4.  I am not sure if this information is used

mateiz:
Okay, got it. I'm inclined to close this PR then unless we see other people complain about this problem.. though if you want, maybe you can just add a paragraph to the doc saying "if you see errors about a wrong Java version, also set these properties". It's just a bit daunting to see a lot of properties in the POM that most people probably don't add to their projects.

mateiz:
Going to close the PR for now.

PR #778, opened 2013-08-05
Add properties and build sections to minimal pom.xml to set java version...

stoneyv:
... and UTF-8 encoding

Eliminate error and warning when packaging java / maven example.
Add properties and build sections to minimal pom.xml to set java version and UTF-8 encoding.


stoneyv:
I was able to eliminate the source level error at compile by installing maven from apache instead of the package distributed by Ubuntu 12.04 LTS repositories. It might have been the /etc/settings.xml profile that was set to 1.4. I am not sure if this information is used

PR #779, opened 2013-08-05
[HOTFIX] Extend thread safety for SparkEnv.get()

ctn:
A ThreadLocal SparkEnv.env is facing various situations leading to NullPointerExceptions, where SparkEnv.env set in one thread is not gettable in another thread, but often assumed to be available.

See, e.g., https://groups.google.com/forum/#!topic/spark-developers/GLx8yunSj0A

This hotfixes SparkEnv.env to return either (a) the ThreadLocal value if non-null, or (b) the previously set value in any thread.

This approach preserves SparkEnv.set() thread safety needed by RDD.compute() and possibly other places. A refactoring that parameterizes SparkEnv should be addressed subsequently.

### DIFF
--- a/core/src/main/scala/spark/SparkEnv.scala
+++ b/core/src/main/scala/spark/SparkEnv.scala
@@ -97,7 +97,7 @@ class SparkEnv (
 
 object SparkEnv extends Logging {
   private val env = new ThreadLocal[SparkEnv]
-  private var lastSetSparkEnv : SparkEnv = _
+  @volatile private var lastSetSparkEnv : SparkEnv = _
 
   def set(e: SparkEnv) {
 	  lastSetSparkEnv = e
--- a/core/src/main/scala/spark/SparkEnv.scala
+++ b/core/src/main/scala/spark/SparkEnv.scala
@@ -97,13 +97,26 @@ class SparkEnv (
 
 object SparkEnv extends Logging {
   private val env = new ThreadLocal[SparkEnv]
+  private var lastSetSparkEnv : SparkEnv = _
 
   def set(e: SparkEnv) {
+	  lastSetSparkEnv = e
     env.set(e)
   }
 
+  /**
+   * Returns the ThreadLocal SparkEnv, if non-null. Else returns the SparkEnv
+   * previously set in any thread.
+   */
   def get: SparkEnv = {
-    env.get()
+    Option(env.get()).getOrElse(lastSetSparkEnv)
+  }
+
+  /**
+   * Returns the ThreadLocal SparkEnv.
+   */
+  def getThreadLocal : SparkEnv = {
+	  env.get()
   }
 
   def createFromSystemProperties(


core/src/main/scala/spark/SparkEnv.scala
```diff
@@ -97,13 +97,26 @@ class SparkEnv (
 
 object SparkEnv extends Logging {
   private val env = new ThreadLocal[SparkEnv]
+  private var lastSetSparkEnv : SparkEnv = _
```
mridulm: make this volatile, or ensure get/set goes through a barrier.

mridulm:
I am not sure if this is a good idea, but commented strictly on the code bug.

core/src/main/scala/spark/SparkEnv.scala
ctn: `@volatile` (https://github.com/adatao/spark/commit/b1bbbe699cd23b937e6fa5c90d5ea0fcc9801681) may help in the unlikely event that `lastSetSparkEnv` gets hotspotted; `synchronized` is overkill in this case as we're already willing to "dumb down" to whatever `lastSetSparkEnv` happens to be.

As discussed in this context, this is to hotfix-avoid highly probable NPEs given the current ThreadLocal design, without taking it away completely, in case someone still wants it. The goal isn't to provide complete guarantees on happens-before relationships at the get/set level.

A proper refactoring would have `SparkEnv.env` passed around as a parameter to most places that need it. I/we may get around to doing that, time permitting.

mateiz:
Christopher, to submit Spark jobs from multiple threads, you should simply call SparkEnv.set with the correct SparkEnv value in each thread. Is this not possible in your application?

We could merge something like this, but in the future I'd like to get rid of this way of finding SparkEnv altogether, so that's why I'd rather not make changes to it now.

core/src/main/scala/spark/SparkEnv.scala
mridulm: add of synchronized is to ensure visibility of update across threads, not necessarily only to serialize execution ...
volatile would do too in this case.

mateiz:
Ah I see, you actually had a problem with it in a Spark thread in HttpBroadcast. In that case it's probably a good idea to merge it. Just out of curiosity, can you tell me the stack trace where you saw the problem happen, if you still have it?

mridulm:
The side effect of merging this is that it will become even more difficult to add support for multiple spark contexts in future : due to implicit failure alleviation done via this patch.

mateiz:
@mridulm yeah, that is a good point. Maybe it would be better to fix the HttpBroadcast problem instead.

ctn:
We're agreed that these are all treating symptoms---short of a refactoring, as discussed in the thread on spark-developers. That said, treating symptom here where the culprit resides is the least-unsatisfying of the alternatives.

Fixing `HttpBroadcast` is something along the lines of blaming the victim; there will be other victims, and they won't know what hit them.

I think it will be hard to forget that this central issue needs to be dealt with when you get around to supporting multiple Spark contexts. Even before that you would likely back off to a non-ThreadLocal SparkEnv.env (as @rxin suggests), along with the passing of SparkEnv as params to slaves. Meanwhile this approach saves the greatest number of probable headaches. There are going to be pros and cons; they are not equally weighted.

Fixing it elsewhere is easier to forget about.

ctn:
@mateiz I'll dig up the stack trace related to `HttpBroadcast` later tonight.

Others are having issues from other angles as discussed on the spark-developers thread.

ctn:
For stack trace related to `HttpBroadcast`, see https://groups.google.com/forum/#!topic/spark-users/73fA390Q0Fc. We eventually diagnosed this as an object being ser/deser from slave back to user, that has a reference to an `HttpBroadcast` inside it. Since there is no guarantee that the thread handling the deserialization has a valid ThreadLocal SparkEnv, the NPE has a decent probability of happening.

This is but one vector as suggested earlier by @JoshRosen in https://groups.google.com/forum/#!topic/spark-developers/GLx8yunSj0A. Others have encountered this issue through completely orthogonal vectors.

The root cause is an incomplete implementation of SparkEnv ThreadLocal isolation/sharing. As @mateiz has pointed out, the reason for using ThreadLocal today has more to do with, e.g., `RDD.compute()` than future support for multiple SparkContexts. So, short of going back to non-ThreadLocal before a refactoring (and risking greater thread safety issues), this proposed fix is the most attractive among the alternatives. When you do get around to supporting multiple SparkContexts, you will certainly re-architect the way SparkEnv is accessed and shared, and making it safe to do so, anyway.

mridulm:
SparkEnv, as it is currently used, leads to lot of uncomfortable situations
... and while there might be some merit in hacking this in right now to
alleviate immediate concerns, once done, a more principled way to fix it
might become more difficult.
There is a laundry list of hacks which is already making this an
interesting problem.

The major issues with multiple spark contexts are (add more as required) :
a) actor system seems to not be able to handle port 0 for bind as reported
by some users - might be a race condition somewhere ?
b) set of env/sys props in spark to convey info : we will need to replace
this somehow.
c) The code makes assumptions that SparkEnv is available as threadlocal -
which leads to all sorts of corner cases when you run deferred computations.
d) You cannot reliably mix different spark contexts in the same thread.
e) Yarn mode runs master in the cluster - making configuring other master
types 'interesting' (other than probably local).

But all of these are just implementation issues right now imo - not user
expectations.
This hack would surface this limitation to the user explicitly and attempts
to make it an expected behaviour.
The issues observed would anyway need to be handled while moving to
multiple spark contexts - so there is merit in fixing them.

Assuming we cannot live without this, at best, we should qualify it with
some config which user explicitly enables (some sys prop), log a warning
message (so that we know the cases where it fails) and we deprecate/remove
it later.

Regards,
Mridul
Principal Research Engineer
Yahoo Labs

ctn:
Mridul, I'm completely sympathetic to the analysis, and the problem
statement of "Would we be encouraging users to expect something that they
should not?" To be sure, the problem statement cannot be, "Would we forget
to deal with this when implementing multiple SparkContexts?"; this latter
one being highly improbable.

On the former, let's shift the frame of reference a bit. Today's users
understand that there is only support for one SparkContext per JVM. So in
that frame the behavior of a shared "static" SparkEnv is entirely
reasonable to users.

What isn't reasonable is to expect users to be aware of the very real
present-day issue that this single SparkEnv may not be quite single after
all, and work around it. Indeed for many users the issue crops up in
esoteric, highly unexpected ways (the thread switching is in the
infrastructure), with the chain of responsibility several levels removed.

So the problem statement above is better revised as, "Are we
(inadvertently) forcing users to deal with SparkEnv in terms of that
multiple-SparkContext future, when the current implementation is still
incompatible with that model?" I say "inadvertent" because the current
ThreadLocal usage intent has less to do with multiple-SparkContext, and a
lot more to do with being able to ship it around thread-safely.

Thus, in sum, something must be done to alleviate user pain today. The
practical thing can be done is either (a) make SparkEnv non-ThreadLocal, or
(b) make it fall back gracefully to a non-ThreadLocal value when needed.
Communication-wise, make sure users understand that the present
single-SparkContext design is a feature, not a bug.

No one should be surprised if they have to make some changes in a
multiple-SparkContext future.

## 

Christopher T. Nguyen
Co-founder & CEO, Adatao http://adatao.com
linkedin.com/in/ctnguyen

mateiz:
How about this as a solution that fixes the problem today, and still lets us switch away from a thread-local SparkEnv in the future?

1) We merge this PR as is to catch cases where the user hasn't set SparkEnv
2) When we want to get rid of this thread-local stuff, we remove the SparkEnv.get() method, and change it to passing a SparkEnv as a constructor parameter everywhere. Thus, existing calls to SparkEnv.get will be very visible as compile errors.

The only thing this won't directly fix is uses of SparkEnv inside deserialize functions, such as this HttpBroadcast one, because it's not possible to pass extra parameters there; for those though, we can use another thread-local variable accessed through a different method name, and maybe we can add some help to make sure this is always present (e.g. make the TaskResult class search for it).

mridulm:
On Tue, Aug 6, 2013 at 11:09 PM, Matei Zaharia notifications@github.comwrote:

> How about this as a solution that fixes the problem today, and still lets
> us switch away from a thread-local SparkEnv in the future?
> 
> 1) We merge this PR as is to catch cases where the user hasn't set SparkEnv
> 2) When we want to get rid of this thread-local stuff, we remove the
> SparkEnv.get() method, and change it to passing a SparkEnv as a constructor
> parameter everywhere. Thus, existing calls to SparkEnv.get will be very
> visible as compile errors.

There have been a bunch of cases where users are hitting issues related to
this - unfortunately, unless we fix multiple SparkContext properly, we
cannot resolve this other than with a hack like this.
The plan above will mean incompatible interface changes in spark - will we
be able to insulate user code completely from it ?
If yes, it would be a clean transition - else, a rocky one :-)
Either way, in retrospect, we will need to do something to resolve this
issue : if no one is planning to work on the principled way, atleast via
this patch.

> The only thing this won't directly fix is uses of SparkEnv inside
> deserialize functions, such as this HttpBroadcast one, because it's not
> possible to pass extra parameters there; for those though, we can use
> another thread-local variable accessed through a different method name, and
> maybe we can add some help to make sure this is always present (e.g. make
> the TaskResult class search for it).

We can pass that in the constructor, and require all Broadcast
constructor's to take a SparkEnv - since broadcast is tied to a SparkEnv :
it should be fine.
Or am I missing something ?

Regards,
Mridul

mateiz:
@mridulm I don't get it, what's incompatible about this change? Ideally, users wouldn't have to call SparkEnv.set or .get. The only reason they do now is because of this thread-local issue. But in the future, we would remove the need to do that, and users would just remove calls to SparkEnv in their code.

For deserialization methods like readObject, the problem is that we never construct the object; Java's ObjectInputStream does.

ctn:
I believe we're converging. A possible roadmap is as follows:
1. We apply this patch to address the various mysterious NPEs at hand.
2. We refactor code paths to `RDD.compute()` to relieve the need/desire for `ThreadLocal[SparkEnv]`, and likely at the same time remove `ThreadLocal`. I can commit to provide resources to help with this within the next 90 days with 60% probability.
3. Multiple-SparkContext is implemented and its impact explicitly ripples through to user code as needed.

@mateiz this patch does address the `HttpBroadcast` issue that we see, because by the time it's being deserialized (in our case as an unintentional side effect, but who knows with other users) at the master in `TaskSetManager.taskFinished()`, there is surely already a `SparkContext` set earlier to send the task to slaves---just not one visible on the thread that happens to handle the deser.

mridulm:
I was referring to how user initializes and uses SparkContext(s) within
main/etc method - as part of adding support for that, whether existing
exposed api/spi will need to change.
Please note that I have not looked at all spi/api's exposed to see if this
will be the case or not - I am hoping it wont have any end user impact.

About readObject - I am not sure I understand the usecase : are we
anticipating readObject (in class being deser) to require reference to
SparkEnv ?
I was expecting only the custom (de)-serializer init to require SparkEnv -
not the actual object ser/deser code.

Regards,
Mridul

mateiz:
Yup, the long-term plan is to not change the APIs for initializing SparkContext and such. The only thing that adding multi-SparkContext support would mean is that the user wouldn't be expected to call SparkEnv.set in worker threads anymore -- instead, the right SparkEnv will be used based on the RDD that operations are called on.

The `readObject` comment was indeed about objects whose `readObject` method needs a SparkEnv, like `HttpBroadcast`; I was just saying that these are cases when we can't pass it as a parameter. It might be possible to deserialize these objects differently at the output stream level to avoid this, but otherwise, they will have to receive the SparkEnv in another way.

Anyway, I do want to have multi-SparkContext support, and a switch away from system properties, on the roadmap for 1.0, so they'll be addressed soon. But before that I want to release 0.8 with the fairly extensive changes we've gotten so far (YARN support, rack locality, shuffle fixes, MLlib, monitoring/metrics, etc). I'm going to merge this patch as is for now.

mateiz:
Jenkins, test this please

ctn:
@mridulm re HttpBroadcast.readObject: https://github.com/mesos/spark/blob/master/core/src/main/scala/spark/broadcast/HttpBroadcast.scala#L51

An NPE happens when this code is executed on the wrong thread, on the master, as part of the deser of a task result that happens to contain an `HttpBroadcast` (which side effect is several levels removed from the user's logic): https://github.com/mesos/spark/blob/5e7b38fbb3264921931f5ef2dcba25e3233a191a/core/src/main/scala/spark/scheduler/cluster/ClusterTaskSetManager.scala#L570

You're right in that in a future multiple-SparkContext implementation, things like `HttpBroadcast` would have something like a "`sparkContextId`" field set at first instantiation, with which they should be able to reliably `SparkContext.get(sparkContextId)` anywhere in the cluster, security isolation notwithstanding.

mateiz:
Jenkins, this is ok to test

mridulm:
Interesting, should have looked more into that stacktrace, my apologies.
Yeah, I like the approach proposed - maintain a SparkContext/SparkEnv
global somewhere, and use the id - lookup based on the id.

Btw, just noticed that this patch has an issue with stop() - but I guess
since this is a temp workaround, we should punt on trying to resolve that.

On Wed, Aug 7, 2013 at 2:46 AM, Christopher Nguyen <notifications@github.com

mateiz:
You mean with SparkEnv.stop() not clearing it? That's true, that is not great, but I guess it will get reset if you create a new SparkContext.

mateiz:
Merged this in -- thanks Christopher.

PR #781, opened 2013-08-05
UI fixes: SPARK-848, 849

karenfeng:
This should fix the problems suggested in [SPARK-849](https://spark-project.atlassian.net/browse/SPARK-849) and [SPARK-848](https://spark-project.atlassian.net/browse/SPARK-848).
screen shot 2013-08-05 at 1 06 43 pm
screen shot 2013-08-05 at 1 06 28 pm

### DIFF
--- a/core/src/main/scala/spark/ui/env/EnvironmentUI.scala
+++ b/core/src/main/scala/spark/ui/env/EnvironmentUI.scala
@@ -46,7 +46,7 @@ private[spark] class EnvironmentUI(sc: SparkContext) {
       ("Scala Home", Properties.scalaHome)
     )
     def jvmRow(kv: (String, String)) = <tr><td>{kv._1}</td><td>{kv._2}</td></tr>
-    def jvmTable = UIUtils.listingTable(Seq("Name", "Value"), jvmRow, jvmInformation)
+    def jvmTable = UIUtils.listingTable(Seq("Name", "Value"), jvmRow, jvmInformation.sorted)
 
     val properties = System.getProperties.iterator.toSeq
     val classPathProperty = properties
@@ -58,8 +58,8 @@ private[spark] class EnvironmentUI(sc: SparkContext) {
 
     val propertyHeaders = Seq("Name", "Value")
     def propertyRow(kv: (String, String)) = <tr><td>{kv._1}</td><td>{kv._2}</td></tr>
-    val sparkPropertyTable = UIUtils.listingTable(propertyHeaders, propertyRow, sparkProperties)
-    val otherPropertyTable = UIUtils.listingTable(propertyHeaders, propertyRow, otherProperties)
+    val sparkPropertyTable = UIUtils.listingTable(propertyHeaders, propertyRow, sparkProperties.sorted)
+    val otherPropertyTable = UIUtils.listingTable(propertyHeaders, propertyRow, otherProperties.sorted)
 
     val classPathEntries = classPathProperty._2
         .split(System.getProperty("path.separator", ":"))
@@ -71,7 +71,7 @@ private[spark] class EnvironmentUI(sc: SparkContext) {
 
     val classPathHeaders = Seq("Resource", "Source")
     def classPathRow(data: (String, String)) = <tr><td>{data._1}</td><td>{data._2}</td></tr>
-    val classPathTable = UIUtils.listingTable(classPathHeaders, classPathRow, classPath)
+    val classPathTable = UIUtils.listingTable(classPathHeaders, classPathRow, classPath.sorted)
 
     val content =
       <span>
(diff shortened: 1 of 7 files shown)


karenfeng:
Regarding the table sort, it seems that we need to have `dean_addEvent(headrow[i],"click", sorttable.innerSortFunction = function(e) {` to sort the table before the page is loaded; however, that line seems to have been changed within the sorttable.js file. Should I sort the Seqs before they are passed then?

mateiz:
@karenfeng I don't quite understand -- is there an issue adding the JavaScript code that requires us to change sorttable.js? I thought it would have an API to sort the table.

karenfeng:
@mateiz, after searching for a while it seems like I have to add some clumsy code: http://www.kryogenix.org/code/browser/sorttable/#externalcall Is there a better way to do it?

mateiz:
I see, interesting. They really seem to discourage sorting the table on page load (see http://www.kryogenix.org/code/browser/sorttable/#sortonload). How much work is it to sort these on the server -- that is, what data type is usually in the first column? I'm just afraid that we'd have to write a lot of comparison code for different data types over there.

core/src/main/scala/spark/ui/env/EnvironmentUI.scala
```diff
@@ -58,8 +58,8 @@ private[spark] class EnvironmentUI(sc: SparkContext) {
 
     val propertyHeaders = Seq("Name", "Value")
     def propertyRow(kv: (String, String)) = <tr><td>{kv._1}</td><td>{kv._2}</td></tr>
-    val sparkPropertyTable = UIUtils.listingTable(propertyHeaders, propertyRow, sparkProperties)
-    val otherPropertyTable = UIUtils.listingTable(propertyHeaders, propertyRow, otherProperties)
+    val sparkPropertyTable = UIUtils.listingTable(propertyHeaders, propertyRow, sparkProperties.sorted)
```
pwendell: Hey Karen - would you mind just sorting these things when they are defined rather than here? Could make it nicer if we want to re-use them.

karenfeng:
screen shot 2013-08-06 at 2 18 23 pm
screen shot 2013-08-06 at 2 10 04 pm

core/src/main/scala/spark/ui/jobs/IndexPage.scala
```diff
@@ -69,11 +69,19 @@ private[spark] class IndexPage(parent: JobProgressUI) {
               {Utils.memoryBytesToString(listener.totalShuffleWrite)}
             </li>
          }
-          <li><strong>Active Stages Number:</strong> {activeStages.size} </li>
-          <li><strong>Completed Stages Number:</strong> {completedStages.size} </li>
-          <li><strong>Failed Stages Number:</strong> {failedStages.size} </li>
-          <li><strong>Scheduling Mode:</strong> {parent.sc.getSchedulingMode}</li>
-
+         <li><a href="#active">
+             <strong>Active Stages Number:</strong>
```
pwendell: Could we change these to just say "Active Stages", "Completed Stages", etc with no "Number"?

Also, can we try it with the link just around e.g. "Active Stages", not the colon, space or count?

pwendell:
@mateiz @karenfeng It's really too bad they don't make it easy to sort it from Javascript. The problem with sorting it on our own is that it doesn't set the little "down arrow" in the column header, so it's not at all obvious that the table is sortable. I've seen a few people click around the UI and none of them even realized that you could sort the tables :(

core/src/main/scala/spark/ui/jobs/IndexPage.scala
```diff
@@ -83,11 +91,11 @@ private[spark] class IndexPage(parent: JobProgressUI) {
...
         Seq()
       }} ++
-      <h3>Active Stages : {activeStages.size}</h3> ++
+      <h3 id="active">Active Stages : {activeStages.size}</h3> ++
       activeStagesTable.toNodeSeq++
-      <h3>Completed Stages : {completedStages.size}</h3> ++
+      <h3 id="completed">Completed Stages : {completedStages.size}</h3> ++
       completedStagesTable.toNodeSeq++
-      <h3>Failed Stages : {failedStages.size}</h3> ++
+      <h3 id ="failed">Failed Stages : {failedStages.size}</h3> ++
```
mateiz, line 98: One small comment: remove the space in front of the colon in these headings

mateiz:
@pwendell yeah, that's a good point about the arrow. Maybe it would be possible to add some JQuery code that finds all sortable tables and calls the stuff above on them (e.g. `$("table.sortable").each(function(table) { ... })`. We can do it in a later PR though so as to not complicate this one too much.

karenfeng:
@pwendell, is this what you meant?
screen shot 2013-08-06 at 3 14 05 pm

karenfeng:
screen shot 2013-08-06 at 3 44 50 pm

karenfeng:
screen shot 2013-08-06 at 4 32 34 pm

karenfeng:
screen shot 2013-08-06 at 4 52 08 pm

mateiz:
Merged this by hand due to a conflict. Thanks Karen!

PR #783, opened 2013-08-06
expose HDFS stats as executor metrics 

anfeng:
This PR exposes HDFS stats via Executor metrics.

Currently, we support 2 HDFS schemes:
- hdfs ... Hadoop distributed file system
- file ... Hadoop local file system

For each scheme, we have the following gauges:
- Number of bytes written
- Number of read operations
- Number of large read operations
- Number of write operations


core/src/main/scala/spark/executor/ExecutorSource.scala
```diff
@@ -2,9 +2,25 @@ package spark.executor
...
 
+import org.apache.hadoop.fs.FileSystem
+import org.apache.hadoop.hdfs.DistributedFileSystem
+import org.apache.hadoop.fs.LocalFileSystem
+
 import spark.metrics.source.Source
 
+
 class ExecutorSource(val executor: Executor) extends Source {
+  private def file_stats(scheme: String) : FileSystem.Statistics = {
```
pwendell: In scala it is idomatic to use `Option[FileSystem.Statistics]` as the return type instead of returning possibly null.

It is also common to write loops like this in a functional manner:

```
 import scala.collection.JavaConversions._
 FileSystem.getAllStatistics().filter(s => s.getScheme.equals(scheme)).headOption
```

core/src/main/scala/spark/executor/ExecutorSource.scala
```diff
@@ -27,4 +43,84 @@ class ExecutorSource(val executor: Executor) extends Source {
...
+
+  //Gauge got the # of bytes read in HDFS distributed file system
+  metricRegistry.register(MetricRegistry.name("filesystem", "hdfs", "bytesRead"), new Gauge[Long] {
+    override def getValue: Long = {
+      val stats = file_stats("hdfs")
+      return if (stats==null) 0 else stats.getBytesRead()
+    }
+  })
+
+  //Gauge got the # of bytes written in HDFS distributed file system
```
pwendell: We usually put a space after the comment string `//`

ryanlecompte: Also, no snake case please. :-)

pwendell:
Hey @anfeng - this is nifty! I made a few comments about style... we usually try to use Scala conventions for things like list iteration and optional values. Overall, looks good.

core/src/main/scala/spark/executor/ExecutorSource.scala
```diff
@@ -27,4 +43,84 @@ class ExecutorSource(val executor: Executor) extends Source {
...
+  metricRegistry.register(MetricRegistry.name("filesystem", "hdfs", "bytesWritten"), new Gauge[Long] {
+    override def getValue: Long = {
+      val stats = file_stats("hdfs")
+      return if (stats==null) 0 else stats.getBytesWritten()
```
pwendell: Also, these can be written like this:

```
 override def getValue = file_stats("hdfs").map(s => s.getBytesWritten()).getOrElse(0L)
```

core/src/main/scala/spark/executor/ExecutorSource.scala
pwendell: oh ya, this should be `fileStats` /cc @ryanlecompte

core/src/main/scala/spark/executor/ExecutorSource.scala
```diff
@@ -27,4 +37,54 @@ class ExecutorSource(val executor: Executor) extends Source {
   metricRegistry.register(MetricRegistry.name("threadpool", "maxPool", "size"), new Gauge[Int] {
     override def getValue: Int = executor.threadPool.getMaximumPoolSize()
   })
+
```
stephenh, line 45: Could add a utility method:

``` scala
private def registerFileSystemStat[T](name: String, scheme: String, f: FileSystem.Statistics => T, defaultValue: T) = {
  metricRegistry.register(MetricRegistry.name("filesystem", scheme, name), new Guage[T] {
    override def getValue: T = fileStats(scheme).map(f).getOrElse(defaultValue)
  })
}

registerFileSystemStat("hdfs", "bytesRead", _.getBytesRead(), 0L)
registerFileSystemStat("hdfs", "bytesWritten", _.getBytesWrriten(), 0L)
...
```

(Sorry, just reading pull requests and got distracted. Typing off the top of my head, but it should work.)

markhamstra:
Could you please cleanup the git history on this, or re-submit based on the current state of the master branch?  There are really only three useful commits here, and the remainder of the 12 are just noise.

anfeng:
I have created a new PR #785 with clean commits. Let's close this old PR.

PR #784, opened 2013-08-06
Add MetricsServlet for Spark metrics system

jerryshao:
This patch adds  metrics servlet for spark's jetty server, users can get metrics snapshot using http request to master, work and client driver, a json format of metrics data will be returned.

### DIFF
--- a/conf/metrics.properties.template
+++ b/conf/metrics.properties.template
@@ -41,16 +41,11 @@
 #    customize metrics system. You can also put the file in ${SPARK_HOME}/conf
 #    and it will be loaded automatically.
 #    5. MetricsServlet is added by default as a sink in master, worker and client
-#    driver, you can send http request "/metrics" to get a snapshot of all the
-#    registered metrics in json format. For master, requests "/metrics/master" and
-#    "/metrics/applications" can be sent seperately to get metrics snapshot of
-#    instance master and applications.
+#    driver, you can send http request "/metrics/json" to get a snapshot of all the
+#    registered metrics in json format. For master, requests "/metrics/master/json" and
+#    "/metrics/applications/json" can be sent seperately to get metrics snapshot of
+#    instance master and applications. MetricsServlet may not be configured by self.
 #
-
-# Change MetricsServlet's property
-#*.sink.servlet.uri=/metrics
-#
-#*.sink.servlet.sample=false
 
 # Enable JmxSink for all instances by class name
 #*.sink.jmx.class=spark.metrics.sink.JmxSink
--- a/conf/metrics.properties.template
+++ b/conf/metrics.properties.template
@@ -3,8 +3,8 @@
 #  This file configures Spark's internal metrics system. The metrics system is
 #  divided into instances which correspond to internal components.
 #  Each instance can be configured to report its metrics to one or more sinks.
-#  Accepted values for [instance] are "master", "worker", "executor", "driver", 
-#  and "applications". A wild card "*" can be used as an instance name, in 
+#  Accepted values for [instance] are "master", "worker", "executor", "driver",
+#  and "applications". A wild card "*" can be used as an instance name, in
 #  which case all instances will inherit the supplied property.
 #
 #  Within an instance, a "source" specifies a particular set of grouped metrics.
@@ -19,7 +19,7 @@
 #  A "sink" specifies where metrics are delivered to. Each instance can be
 #  assigned one or more sinks.
 #
-#  The sink|source field specifies whether the property relates to a sink or 
+#  The sink|source field specifies whether the property relates to a sink or
 #  source.
 #
 #  The [name] field specifies the name of source or sink.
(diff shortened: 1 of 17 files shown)


core/src/main/scala/spark/deploy/master/ui/MasterWebUI.scala
```diff
@@ -57,7 +57,7 @@ class MasterWebUI(val master: ActorRef, requestedPort: Int) extends Logging {
     }
   }
 
-  val handlers = Array[(String, Handler)](
+  var handlers = Array[(String, Handler)](
```
pwendell: Rather than making this a `var` and mutating to it from outside, it would be better to leave it as a `val`. When you initialize it, check if the master has a metrics servlet and add it if it does. To do this, you might need to pass a reference to the master in the constructor to MasterWebUI.

core/src/main/scala/spark/deploy/master/Master.scala
```diff
@@ -65,6 +65,15 @@ private[spark] class Master(host: String, port: Int, webUiPort: Int) extends Act
   val applicationMetricsSystem = MetricsSystem.createMetricsSystem("applications")
   val masterSource = new MasterSource(this)
 
+  // Add default MetricsServlet handler to web ui
```
pwendell: This logic should be moved into `MasterWebUI.scala` when the handler list is first created.

core/src/main/scala/spark/deploy/worker/Worker.scala
```diff
@@ -101,6 +101,12 @@ private[spark] class Worker(
     logInfo("Spark home: " + sparkHome)
     createWorkDir()
     webUi = new WorkerWebUI(this, workDir, Some(webUiPort))
+
+    // Add default MetricsServlet handlers to webUi
+    metricsSystem.metricsServlet foreach { m =>
```
pwendell: It would be good to do the same thing here, have the UI check when its creating the handler list whether to add the metrics servlet.

core/src/main/scala/spark/metrics/MetricsSystem.scala
```diff
@@ -126,7 +129,11 @@ private[spark] class MetricsSystem private (val instance: String) extends Loggin
         val sink = Class.forName(classPath)
           .getConstructor(classOf[Properties], classOf[MetricRegistry])
           .newInstance(kv._2, registry)
-        sinks += sink.asInstanceOf[Sink]
+        if (kv._1 =="servlet") {
```
pwendell: add a space after the `==` operator.

core/src/main/scala/spark/ui/SparkUI.scala
```diff
@@ -43,8 +43,15 @@ private[spark] class SparkUI(sc: SparkContext) extends Logging {
   val jobs = new JobProgressUI(sc)
   val env = new EnvironmentUI(sc)
   val exec = new ExecutorsUI(sc)
+
+  // Add MetricsServlet handlers by default
+  val metricsServletHandlers = SparkEnv.get.metricsSystem.metricsServlet match {
```
pwendell: You can write this more conisely as:
`val metricsServletHandlers = SparkEnv.get.metricsSystem.metricsServlet.map(s => s.getHandlers).toSeq`

conf/metrics.properties.template
```diff
@@ -28,18 +28,29 @@
...
+#    3. Wild card properties can be overridden by more specific properties.
+#    For example, master.sink.console.period takes precedence over
 #    *.sink.console.period.
 #    4. A metrics specific configuration
 #    "spark.metrics.conf=${SPARK_HOME}/conf/metrics.properties" should be
 #    added to Java properties using -Dspark.metrics.conf=xxx if you want to
 #    customize metrics system. You can also put the file in ${SPARK_HOME}/conf
 #    and it will be loaded automatically.
+#    5. MetricsServlet is added by default as a sink in master, worker and client
+#    driver, you can send http request "/metrics" to get a snapshot of all the
```
pwendell: Could we make thee paths:

`/metrics/json`
`/metrics/master/json`
`/metrics/applications/json`

To make it consistent with the existing UI JSON endpoints? Also that way we can add an HTML version of this in the future.

core/src/main/scala/spark/metrics/MetricsConfig.scala
```diff
@@ -36,7 +36,9 @@ private[spark] class MetricsConfig(val configFile: Option[String]) extends Loggi
   var propertyCategories: mutable.HashMap[String, Properties] = null
 
   private def setDefaultProperties(prop: Properties) {
-    // empty function, any default property can be set here
+    prop.setProperty("*.sink.servlet.class", "spark.metrics.sink.MetricsServlet")
+    prop.setProperty("master.sink.servlet.uri", "/metrics/master")
```
pwendell: Let's actually not make this configurable, we can just hard code it.

pwendell:
Hey @jerryshao thanks for adding this! It looks really good. It would be great to get this into 0.8. I've made some minor comments - the biggest thing is to just have the UI's pull the metrics handlers from their parent components rather than adding it from outside. Would you be able to take a look? Thanks again - Patrick

jerryshao:
Hi @pwendell , thanks for your review. I've updated the code as you pointed out. 

For configuration of MetricsServlet, I wrote configuration here in `setDefaultProperties` is to ensure the consistency and convenience for `MetricsSystem` to load sinks, including MetricsServlet. But treat MetricsServlet as a unique sink with hard code configuration may break the consistency of code, so that's my reason to add here. 

On the one hand, now I add notes in `metrics.properties.template` to tell people that there's no need to configure MetricsServlet, also expose no configuration.  On the other hand I think it would be better to leave a space for people who understand the internals to be able to configure this.

pwendell:
Hey @jerryshao - okay, sounds good. Jenkins, test this please.

PR #785, opened 2013-08-06
expose HDFS file system stats via Executor metrics

anfeng:
This PR exposes HDFS stats via Executor metrics.

We support 2 HDFS schemes:
- hdfs ... Hadoop distributed file system
- file ... Hadoop local file system

For each scheme, we have the following gauges:
- Number of bytes written
- Number of read operations
- Number of large read operations
- Number of write operations

### DIFF
--- a/core/src/main/scala/spark/executor/ExecutorSource.scala
+++ b/core/src/main/scala/spark/executor/ExecutorSource.scala
@@ -2,9 +2,24 @@ package spark.executor
 
 import com.codahale.metrics.{Gauge, MetricRegistry}
 
+import org.apache.hadoop.fs.FileSystem
+import org.apache.hadoop.hdfs.DistributedFileSystem
+import org.apache.hadoop.fs.LocalFileSystem
+
+import scala.collection.JavaConversions._
+
 import spark.metrics.source.Source
 
 class ExecutorSource(val executor: Executor) extends Source {
+  private def fileStats(scheme: String) : Option[FileSystem.Statistics] =
+    FileSystem.getAllStatistics().filter(s => s.getScheme.equals(scheme)).headOption
+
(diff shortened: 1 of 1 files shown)


rxin:
Jenkins, test this please.

rxin:
Jenkins, test this please.

Jenkins, add to whitelist.

core/src/main/scala/spark/executor/ExecutorSource.scala
```diff
@@ -2,9 +2,24 @@ package spark.executor
...
+
+import scala.collection.JavaConversions._
+
 import spark.metrics.source.Source
 
 class ExecutorSource(val executor: Executor) extends Source {
+  private def fileStats(scheme: String) : Option[FileSystem.Statistics] =
+    FileSystem.getAllStatistics().filter(s => s.getScheme.equals(scheme)).headOption
+
+  private def registerFileSystemStat[T](scheme: String, name: String, f: FileSystem.Statistics => T, defaultValue: T) = {
```
pwendell: The line limit is 100 character, so this should be wrapped after the `(`

rxin:
Jenkins, test this again .. test this please ...

pwendell:
Hey @anfeng - the code looks good here. However, I just tried to run this both locally and on a cluster and in both cases it didn't work. It looks like FileSystem.getAllStatistics() is just returning an empty list.

Have you tested this at all? Either locally or on a cluster? It doesn't seem to work for me. Is this supported by all HDFS versions?

pwendell:
Ah actually I just realized the tests I was running only read/write shuffle
data... Not Hdfs! Let me run this again tomorrow and if it looks okay I can
merge it.

---

sent from my phone

PR #786, opened 2013-08-06
Java fixes, tests and examples for ALS, KMeans

shivaram:
This change addresses some of the issues with using MLLib from Java programs. Specifically it fixes the `train` interface to not use Ints or Doubles and also renames methods to make them easier to call. 

As a way to check if the Java interface works, we have two additions: a. A set of Java examples in the `examples` project that read a data file and run the algorithms. b. A set of unit tests which check if the basic methods can be called and work as expected.

### DIFF
--- a/examples/src/main/java/spark/mllib/JavaALS.java
+++ b/examples/src/main/java/spark/mllib/JavaALS.java
@@ -39,9 +39,9 @@ public class  JavaALS {
   static class ParseRating extends Function<String, Rating> {
     public Rating call(String line) {
       StringTokenizer tok = new StringTokenizer(line, ",");
-      Integer x = Integer.parseInt(tok.nextToken());
-      Integer y = Integer.parseInt(tok.nextToken());
-      Double rating = Double.parseDouble(tok.nextToken());
+      int x = Integer.parseInt(tok.nextToken());
+      int y = Integer.parseInt(tok.nextToken());
+      double rating = Double.parseDouble(tok.nextToken());
       return new Rating(x, y, rating);
     }
   }
--- a/examples/src/main/java/spark/mllib/JavaALS.java
+++ b/examples/src/main/java/spark/mllib/JavaALS.java
@@ -1,0 +1,87 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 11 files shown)


examples/src/main/java/spark/mllib/JavaALS.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+public class  JavaALS {
+
+  static class ParseRating extends Function<String, Rating> {
+    public Rating call(String line) {
+      StringTokenizer tok = new StringTokenizer(line, ",");
+      Integer x = Integer.parseInt(tok.nextToken());
+      Integer y = Integer.parseInt(tok.nextToken());
+      Double rating = Double.parseDouble(tok.nextToken());
+      return new Rating(x, y, rating);
```
mateiz, line 45: I'm curious here, Shivaram, couldn't you pass in primitive types? If that's a problem with Rating being a case class, we should make it not be one, so that it can also be initialized in Java.

mllib/src/test/scala/spark/mllib/recommendation/JavaALSSuite.java
```diff
@@ -0,0 +1,110 @@
...
+import spark.api.java.JavaSparkContext;
+
+import org.jblas.DoubleMatrix;
+
+public class JavaALSSuite implements Serializable {
+  private transient JavaSparkContext sc;
+
+  @Before
+  public void setUp() {
+    sc = new JavaSparkContext("local", "JavaLogisticRegressionSuite");
```
mateiz: This should say ALS

mateiz:
This looks good, but I just made a couple of comments.

examples/src/main/java/spark/mllib/JavaALS.java
shivaram, line 45: No there aren't any problems with case classes. I was under the wrong assumption that `Integer.parseInt` returned an `Integer` rather than `int` . I have removed the extra constructor now and changed the types here as well.

shivaram:
Thanks for the review. Using native types in the case class constructor seems to work fine, so I just deleted the java-specific code.

mateiz:
Great, thanks! Going to merge this in.

apivovarov:
I think these new 2 classes have wrong path or package

public class  JavaALS
path: examples/src/main/java/spark/mllib/JavaALS.java
package spark.mllib.examples;

public class JavaKMeans
path: examples/src/main/java/spark/mllib/JavaKMeans.java
package spark.mllib.examples;

apivovarov:
I also noticed that spark-mllib dependency was not added to hadoop2-yarn profile
so, maven build is brocken

apivovarov:
I fixed both issues. PR: https://github.com/mesos/spark/pull/804

PR #788, opened 2013-08-07
For standalone mode, add worker local env setting of SPARK_JAVA_OPTS as ...

shane-huang:
...default and let application env override default options if applicable

I think SPARK_JAVA_OPTS should be obtained not only from application env, but also from worker's local settings. Sometimes different worker machines have different hardware configurations and we have to enable specific JVM options for certain workers instead of using the same settings for all. I suggest we use worker's local setting as default and let application settings override the worker's setting if applicable. 

Signed-off-by: shane-huang shengsheng.huang@intel.com

### DIFF
--- a/conf/spark-env.sh.template
+++ b/conf/spark-env.sh.template
@@ -16,4 +16,9 @@
 # - SPARK_WORKER_PORT / SPARK_WORKER_WEBUI_PORT
 # - SPARK_WORKER_INSTANCES, to set the number of worker instances/processes
 #   to be spawned on every slave machine
+# - SPARK_JAVA_OPTS, to set the jvm options for executor backend. Note: This is
+#   only for node-specific options, whereas app-specific options should be set
+#   in the application.
+#   Examples of node-speicic options : -Dspark.local.dir, GC related options.
+#   Examples of app-specific options : -Dspark.serializer
 
--- a/core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
+++ b/core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
@@ -111,6 +111,7 @@ private[spark] class ExecutorRunner(
     val libraryOpts = getAppEnv("SPARK_LIBRARY_PATH")
       .map(p => List("-Djava.library.path=" + p))
       .getOrElse(Nil)
+    val workerLocalOpts = Option(getenv("SPARK_JAVA_OPTS")).map(Utils.splitCommandString).getOrElse(Nil)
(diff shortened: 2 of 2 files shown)


mateiz:
Shane, I'm curious, what's a use case for this? I think that at the very least you should only add in workerLocalOpts if userOpts is not set, because otherwise they might not be compatible. I'm afraid that a user will accidentally configure one thing on their cluster (e.g. Netty shuffle) and another thing on their job, and get processes that can't talk to each other.

mateiz:
Jenkins, this is ok to test

shane-huang:
Since JVM allows command line options overriding and userOpts comes after workerLocalOpts, the options set in userOpts will override those set in workerLocalOpts (the “defaults”). I can barely think of any incompatible issues. Do you have an example when this way leads to incompatible issues?

An example to justify this pull request is that we may need different settings for spark.local.dir on different workers if some workers have less spindles than others. Also, sometimes we may need to install certain profiling tool on only one of the workers and add specific JVM options to collect traces.

To make it more general, I think some options are meant to be set by application users, while some options are admin-level options that are not supposed to be set by application users – actually application users may not even know how many spindles and what the folder permissions are on each cluster node. We’d better draw a line between options for users and those for administrators

Bringing the topic to the next level, I think we should make a clean separation between the functionalities for application developer/user and platform administrator. Our customers often complain that it’s not clear which scripts/options are used by application users and which are used by administrators. The current “run” script in spark home directory by default uses SPARK_MEM as heap size and it’s not suitable to run user applications if not changed or configured, so we write our own “run” scripts to run applications. You must know that Hadoop already separates the scripts used by admins and users in different folders (sbin and bin). I think this problem is practical. I’m now working to find a decent way to do the separation in spark and will later post a proposal for discussion on mailing list.

mateiz:
Okay, that does makes sense. I was worried that with some options, like -Dspark.serializer=spark.KryoSerializer, this would cause nodes to be unable to talk to each other, if one node has the option and one doesn't. But because of the local directory for example, it may make sense to do this. Can you add one thing though -- edit spark-env.sh.template and write there that SPARK_JAVA_OPTS is only for node-specific options there, whereas app-specific ones should be set in the application?

This question of user vs admin settings will also need to be taken into account when we add a "real" configuration system to Spark instead of Java system properties (https://spark-project.atlassian.net/browse/SPARK-544). Maybe you can put together your proposal based on that. Also, I do like the idea of separating bin and sbin, and in fact I'd be happy to drop "run" also to make that less confusing.

mateiz:
Thanks for changing the spark-env.sh too. Going to merge this.

PR #789, opened 2013-08-07
Adding Scala version of PageRank example

MLnick:
See https://spark-project.atlassian.net/browse/SPARK-760

### DIFF
--- a/examples/src/main/scala/spark/examples/SparkPageRank.scala
+++ b/examples/src/main/scala/spark/examples/SparkPageRank.scala
@@ -1,0 +1,51 @@
+package spark.examples
+
+import spark.SparkContext._
+import spark.SparkContext
+
+
+/**
+ * Computes the PageRank of URLs from an input file. Input file should
+ * be in format of:
+ * URL         neighbor URL
+ * URL         neighbor URL
+ * URL         neighbor URL
+ * ...
+ * where URL and their neighbors are separated by space(s).
+ */
+object SparkPageRank {
+
+  def main(args: Array[String]) {
+    if (args.length < 3) {
+      System.err.println("Usage: PageRank <master> <file> <number_of_iterations>")
+      System.exit(1)
+    }
+    var iters = args(2).toInt
+    val ctx = new SparkContext(args(0), "PageRank", System.getenv("SPARK_HOME"), Seq(System.getenv("SPARK_EXAMPLES_JAR")))
+    val lines = ctx.textFile(args(1), 1)
+    val links = lines.map(s => {
+        val parts = s.split("\\s+")
+        (parts(0), parts(1))
+      }).distinct().groupByKey().cache()
+    var ranks = links.mapValues(v => 1.0)
+
+    for (i <- 1 to iters) {
+      val contribs = links.join(ranks).values.flatMap{ case (urls, rank) => {
+        val size = urls.size
+        urls.map(url => (url, rank / size))
+      }}
+
+      ranks = contribs.groupByKey().mapValues(ranks => {
+        val sumRanks = ranks.foldLeft(0.0)(_ + _)
+        0.15 + sumRanks * 0.85
+      })
+    }
+
+    val output = ranks.collect()
+    output.foreach(tup => println(tup._1 + " has rank: " + tup._2 + "."))
+
+    ctx.stop()
+    System.exit(0)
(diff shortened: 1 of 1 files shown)


mateiz:
Jenkins, add to whitelist

examples/src/main/scala/spark/examples/SparkPageRank.scala
```diff
@@ -0,0 +1,51 @@
...
+        val parts = s.split("\\s+")
+        (parts(0), parts(1))
+      }).distinct().groupByKey().cache()
+    var ranks = links.mapValues(v => 1.0)
+
+    for (i <- 1 to iters) {
+      val contribs = links.join(ranks).values.flatMap{ case (urls, rank) => {
+        val size = urls.size
+        urls.map(url => (url, rank / size))
+      }}
```
mateiz: You could have just one set of curly braces here -- you don't need the ones after the `=>`. So for example this is valid Scala:

```
List((1,2), (3,4)).flatMap { case (a, b) =>
  println(a)
  1 to b
}
```

examples/src/main/scala/spark/examples/SparkPageRank.scala
```diff
@@ -0,0 +1,51 @@
...
+
+      ranks = contribs.groupByKey().mapValues(ranks => {
+        val sumRanks = ranks.foldLeft(0.0)(_ + _)
+        0.15 + sumRanks * 0.85
+      })
```
mateiz: Ditto on the inner braces here -- just use braces instead of parens after mapValues

examples/src/main/scala/spark/examples/SparkPageRank.scala
```diff
@@ -0,0 +1,51 @@
...
+      System.err.println("Usage: PageRank <master> <file> <number_of_iterations>")
+      System.exit(1)
+    }
+    var iters = args(2).toInt
+    val ctx = new SparkContext(args(0), "PageRank", System.getenv("SPARK_HOME"), Seq(System.getenv("SPARK_EXAMPLES_JAR")))
+    val lines = ctx.textFile(args(1), 1)
+    val links = lines.map(s => {
+        val parts = s.split("\\s+")
+        (parts(0), parts(1))
+      }).distinct().groupByKey().cache()
```
mateiz: Ditto on the parens/braces here; also unindent the body to look like:

```
    val links = lines.map{ s =>
      val parts = s.split("\\s+")
      (parts(0), parts(1))
    }.distinct().groupByKey().cache()
```

examples/src/main/scala/spark/examples/SparkPageRank.scala
```diff
@@ -0,0 +1,51 @@
...
+ * Computes the PageRank of URLs from an input file. Input file should
+ * be in format of:
+ * URL         neighbor URL
+ * URL         neighbor URL
+ * URL         neighbor URL
+ * ...
+ * where URL and their neighbors are separated by space(s).
+ */
+object SparkPageRank {
+
```
mateiz: No need for this blank line

mateiz:
@MLnick this looks good to me, modulo some small style comments!

examples/src/main/scala/spark/examples/SparkPageRank.scala
```diff
@@ -0,0 +1,50 @@
...
+    }.distinct().groupByKey().cache()
+    var ranks = links.mapValues(v => 1.0)
+
+    for (i <- 1 to iters) {
+      val contribs = links.join(ranks).values.flatMap{ case (urls, rank) =>
+        val size = urls.size
+        urls.map(url => (url, rank / size))
+      }
+
+      ranks = contribs.groupByKey().mapValues{ranks =>
```
mateiz, line 37: Oh, sorry, one more thing, which is fairly important: change this to `contribs.reduceByKey(_ + _).mapValues(0.15 + 0.85 * _)`. That's going to be more efficient because it will do a local reduction on each node before communicating on the network. If the Java PageRank doesn't do that, I'll fix it there as well.

mateiz:
Actually, I'll merge this and make the fix myself since it's small. Thanks Nick!

MLnick:
Thanks for that - was away without laptop access this weekend so couldn't get to it.
—

PR #791, opened 2013-08-07
Fix race condition in ConnectionManager.

pwendell:
This fixes a race condition in the ConnectionManager.

The specific condition involves write channels. There is a race where a new message was added to a channel while an existing message was being written. Depending on the order of thread scheduling, it was possible for the ongoing write handler to disable the write interest on the FD and it have it never get properly re-enabled.

This code simplifies the write path a bit in a way that also removes the race. I've put it here for testing, please don't merge yet.


rxin:
@harsha2010

alig:
Looks good to me, I think you squashed some bugs. 

I thought there was a race left, but I think it's _fine_. Please double check!
(1) A write is being processed. It finishes. Registers uninterest. Outbox synchronized block finishes inside write, returns to run() inside the calling thread.
(2) Context switch happens, new data is written, interest is registered.
(3) The single threaded select loop calls triggerWrite() which notices that the key is already being worked on, so it just returns.
(4) Context switch happens back to run() inside the thread, which now removes the key from the hash table.

All 4 above steps might happen, but I think it's OK because interest will continue to be set so the function will be triggered again from the select loop, in which case, this time, it will properly launch a new thread and process the request.

mridulm:
This does not look like a good change.
There is indeed a race condition I had not originally considered : namely, 
thr1: no chunks returned for connA, pre-empted.
thr2: write, register
thr3: selecter thread woken up, registered, register(), wakeup, connA writable, thr1 still not returned, so we do not attempt to enable write.

This was indeed an oversight : and should be handled by checking if registerInterest() was invoked on the Connection explicitly subsequent to entering the loop (a fetch/reset flag should do).

The change as introduced can cause busy wait until thr1 actually returns, and spreading the state-management of selector-channel all over the place (that was one of the reason to consolidate it into that code).

mridulm:
Also, I would move registerInterest() in send outside the outbox sync block - had not noticed that earlier.

pwendell:
Hey @mridulm  - Could you submit a patch with your proposed fix and we can look at it as well?

What busy-waiting are you specifically referring to? Just the fact that this particular FD will continue to be select()'ed over-and-over? It might be better to have a solution that avoids that, but given that this is such a rare codepath I'm not sure it justifies adding a ton of additional complexity.

pwendell:
@mridulm actually I guess it could occur fairly commonly to have a new write added when one is outstanding... much more common than this particular race. Happy to look at a patch which deals with this nicely.

mridulm:
@pwendell Done - #795 - have CC'ed you on it.
Unfortunately, I have not tested it due to time constraints.
If there are any issues with it, please let me know.

pwendell:
Closing - let's migrate the discussion to #795 .

haoyuan:
The test ran well using this patch. No hung issue has been detected. Just for the record.

mridulm:
Hi @haoyuan, will it be possible to test #795  ?
It is to address the same issue coded up slightly differently.

Thanks,
Mridul

PR #792, opened 2013-08-08
Make SparkListener interface asynchronous, and fix UI bugs

kayousterhout:
In addition to making the SparkListener interface asynchronous, this request fixes 4 bugs in the UI:

One bug caused the UI to crash if you try to look at a job's status
before any of the tasks have finished.

The second bug was a concurrency issue where two different threads
(the scheduling thread and a UI thread) could be reading/updating
the data structures in JobProgressListener concurrently (SPARK-810).

The third bug mis-used an Option, also causing the UI to crash
under certain conditions.

The 4th bug allowed web UI users to add entries in the JobProgressListener's hash maps that would never be deleted if they went to a page for a stage that didn't exist (thus causing a memory leak, albeit at small scale).

### DIFF
--- a/core/src/main/scala/spark/scheduler/DAGScheduler.scala
+++ b/core/src/main/scala/spark/scheduler/DAGScheduler.scala
@@ -102,7 +102,7 @@ class DAGScheduler(
 
   private[spark] val stageToInfos = new TimeStampedHashMap[Stage, StageInfo]
 
-  private[spark] val sparkListeners = ArrayBuffer[SparkListener]()
+  private val listenerEventProcessor = new SparkListenerEventProcessor()
 
   var cacheLocs = new HashMap[Int, Array[List[String]]]
 
@@ -135,6 +135,10 @@ class DAGScheduler(
         DAGScheduler.this.run()
       }
     }.start()
+  }
+
+  def addSparkListener(listener: SparkListener) {
+    listenerEventProcessor.addListener(listener)
   }
 
   private def getCacheLocs(rdd: RDD[_]): Array[List[String]] = {
@@ -334,7 +338,7 @@ class DAGScheduler(
           // Compute very short actions like first() or take() with no parent stages locally.
           runLocally(job)
         } else {
-          sparkListeners.foreach(_.onJobStart(SparkListenerJobStart(job, properties)))
+          listenerEventProcessor.addEvent(SparkListenerJobStart(job, properties))
           idToActiveJob(runId) = job
           activeJobs += job
           resultStageToJob(finalStage) = job
@@ -348,11 +352,11 @@ class DAGScheduler(
         handleExecutorLost(execId)
 
       case begin: BeginEvent =>
-        sparkListeners.foreach(_.onTaskStart(SparkListenerTaskStart(begin.task, begin.taskInfo)))
+        listenerEventProcessor.addEvent(SparkListenerTaskStart(begin.task, begin.taskInfo))
 
       case completion: CompletionEvent =>
-        sparkListeners.foreach(_.onTaskEnd(SparkListenerTaskEnd(completion.task,
-                               completion.reason, completion.taskInfo, completion.taskMetrics)))
+        listenerEventProcessor.addEvent(SparkListenerTaskEnd(
+          completion.task, completion.reason, completion.taskInfo, completion.taskMetrics))
         handleTaskCompletion(completion)
 
       case TaskSetFailed(taskSet, reason) =>
@@ -363,7 +367,7 @@ class DAGScheduler(
         for (job <- activeJobs) {
           val error = new SparkException("Job cancelled because SparkContext was shut down")
           job.listener.jobFailed(error)
-          sparkListeners.foreach(_.onJobEnd(SparkListenerJobEnd(job, JobFailed(error, None))))
+          listenerEventProcessor.addEvent(SparkListenerJobEnd(job, JobFailed(error, None)))
         }
         return true
     }
@@ -513,8 +517,8 @@ class DAGScheduler(
     // must be run listener before possible NotSerializableException
     // should be "StageSubmitted" first and then "JobEnded"
     val properties = idToActiveJob(stage.priority).properties
-    sparkListeners.foreach(_.onStageSubmitted(
-      SparkListenerStageSubmitted(stage, tasks.size, properties)))
+    listenerEventProcessor.addEvent( 
+      SparkListenerStageSubmitted(stage, tasks.size, properties))
     
     if (tasks.size > 0) {
       // Preemptively serialize a task to make sure it can be serialized. We are catching this
@@ -560,8 +564,7 @@ class DAGScheduler(
       }
       logInfo("%s (%s) finished in %s s".format(stage, stage.name, serviceTime))
       stage.completionTime = Some(System.currentTimeMillis)
-      val stageComp = StageCompleted(stageToInfos(stage))
-      sparkListeners.foreach{_.onStageCompleted(stageComp)}
+      listenerEventProcessor.addEvent(StageCompleted(stageToInfos(stage)))
       running -= stage
     }
     event.reason match {
@@ -585,7 +588,7 @@ class DAGScheduler(
                     activeJobs -= job
                     resultStageToJob -= stage
                     markStageAsFinished(stage)
-                    sparkListeners.foreach(_.onJobEnd(SparkListenerJobEnd(job, JobSucceeded)))
+                    listenerEventProcessor.addEvent(SparkListenerJobEnd(job, JobSucceeded))
                   }
                   job.listener.taskSucceeded(rt.outputId, event.result)
                 }
@@ -732,7 +735,7 @@ class DAGScheduler(
       val job = resultStageToJob(resultStage)
       val error = new SparkException("Job failed: " + reason)
       job.listener.jobFailed(error)
-      sparkListeners.foreach(_.onJobEnd(SparkListenerJobEnd(job, JobFailed(error, Some(failedStage)))))
+      listenerEventProcessor.addEvent(SparkListenerJobEnd(job, JobFailed(error, Some(failedStage))))
       idToActiveJob -= resultStage.priority
       activeJobs -= job
       resultStageToJob -= resultStage
--- a/core/src/main/scala/spark/scheduler/DAGScheduler.scala
+++ b/core/src/main/scala/spark/scheduler/DAGScheduler.scala
@@ -102,7 +102,7 @@ class DAGScheduler(
 
   private[spark] val stageToInfos = new TimeStampedHashMap[Stage, StageInfo]
 
-  private val listenerEventProcessor = new SparkListenerEventProcessor()
+  private val listenerBus = new SparkListenerBus()
 
   var cacheLocs = new HashMap[Int, Array[List[String]]]
 
@@ -138,7 +138,7 @@ class DAGScheduler(
   }
 
   def addSparkListener(listener: SparkListener) {
-    listenerEventProcessor.addListener(listener)
+    listenerBus.addListener(listener)
   }
 
   private def getCacheLocs(rdd: RDD[_]): Array[List[String]] = {
@@ -338,7 +338,7 @@ class DAGScheduler(
           // Compute very short actions like first() or take() with no parent stages locally.
           runLocally(job)
         } else {
-          listenerEventProcessor.addEvent(SparkListenerJobStart(job, properties))
+          listenerBus.post(SparkListenerJobStart(job, properties))
           idToActiveJob(runId) = job
           activeJobs += job
           resultStageToJob(finalStage) = job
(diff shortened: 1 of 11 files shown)


rxin:
Any idea whether these changes affect the throughput of the scheduler? Can we reduce the number of locks with concurrent data structures?

pwendell:
This lock is only contended when someone is actively refreshing the ui.
Therefore I think its sufficient to just make sure nothing inside of the
synchronized statements is expensive. Would be good to check that over.

---

sent from my phone

core/src/main/scala/spark/ui/jobs/StagePage.scala
```diff
@@ -36,100 +36,100 @@ private[spark] class StagePage(parent: JobProgressUI) {
...
-          <h2>Summary Metrics</h2> No tasks have started yet
-          <h2>Tasks</h2> No tasks have started yet
-        </div>
-      return headerSparkPage(content, parent.sc, "Stage Details: %s".format(stageId), Jobs)
-    }
+      val tasks = listener.stageToTaskInfos(stageId).toSeq.sortBy(_._1.launchTime)
 
-    val tasks = listener.stageToTaskInfos(stageId).toSeq
+      val shuffleRead = listener.stageToShuffleRead.getOrElse(stageId, 0L) > 0
+      val shuffleWrite = listener.stageToShuffleWrite.getOrElse(stageId, 0L) > 0
```
markhamstra: val shuffleRead = listener.stageToShuffleRead.exists {case (k, v) => k == stageId && v > 0)
val shuffleWrite = listener.stageToShuffleWrite.exists {case (k, v) => k == stageId && v > 0)

markhamstra: ...probably a little better performance:

val shuffleRead = listener.stageToShuffleRead.get(stageId).exists(_ > 0)
val shuffleWrite = listener.stageToShuffleWrite.get(stageId).exists(_ > 0)

rxin:
Patrick, isn't onStageCompleted/onStageSubmitted/onTaskEnd/onJobEnd are all methods that would be invoked during a normal job execution, even without UI?

kayousterhout:
Yeah I think the point Patrick was making is that there's only contention
for the lock when someone's looking at the UI.

pwendell:
I said the lock would not be contended, not that the lock wouldn't be
acquired. The big issue is if the lock is contended.

pwendell:
However, this may be an issue if the page takes a long time to render, so actually we should try this with concurrent data structures.

markhamstra:
@pwendell I'm presuming that you are presuming and haven't actually tested the relative performance of the non-UI code with and without the locking?  While I'm not inclined to disagree with the presumption that acquiring mostly-uncontended locks won't notably slow down onStageCompleted et al., it would be nice to have actual data.

Reynold's suggestion of using concurrent data structures (such as scala.collection.mutable.ConcurrentMap --scala.collection.concurrent.Map for 2.10+) would clean up the need for explicit synchronization, but shouldn't entail much of a performance difference.  Obviously, things can't be left broken just to gain performance, so if the current UI architecture with locking does turn out to be slowing the non-UI code significantly, a re-design may be in order.

kayousterhout:
Patrick is correct that the non-UI code doesn't slow down with the locking
(I've verified this).  But, with a big job (many thousands of tasks), the
time to generate the UI code becomes significant, so if someone's
refreshing the UI a bunch, scheduling throughput is significantly affected.
 Using a concurrent hash set to store the list of tasks should help
this...will change the code to do that and test the throughput again!

mateiz:
Jenkins, add to whitelist

mateiz:
FYI Kay, it looks like you'll also have to rebase this on the latest master branch, or merge in master, because it can no longer merge automatically. I do like the idea of switching to data structures that avoid blocking the scheduler though.

kayousterhout:
Ah thanks for the head's up!  I'm actually about to close this pull request
and open a different one that uses a different approach, so will make sure
to rebase on master before doing that (will send that out in a few hours).
 With the current architecture of the UI, using lock-free data structures
and/or fine-grained locking in the UI is exceedingly complex, so I'm going
to implement the UI in a slightly different way where it processes the
updates from the scheduler asynchronously (so we don't have to worry about
the scheduler thread blocking on the UI, and then can use the simpler
coarse-grained locking).

mateiz:
Cool, that makes sense. Maybe in the long term we should make the whole SparkListener API be asynchronous this way, to avoid having this problem with other consumers.

markhamstra:
@mateiz Precisely what I was thinking when I did a quick think yesterday about what it would take to make the WebUI updates async.  Sounds like a good 0.9 task.

kayousterhout:
Realized I didn't need to do a new pull request for this!  I have verified that this change does not affect the throughput of the scheduler, either in normal operation, or when someone is aggressively loading UI pages.

The throughput of the scheduler _does_ decrease by about 2x if someone is aggressively loading UI pages (based on YourKit, it looks like this is due to additional GC), but this isn't a regression from this change.

core/src/main/scala/spark/scheduler/SparkListenerEventProcessor.scala
```diff
@@ -0,0 +1,62 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package spark.scheduler
+
+import scala.collection.mutable.ArrayBuffer
+
+import java.util.concurrent.LinkedBlockingQueue
```
pwendell: This import should come before the scala imports. The style is:

import java.*

import scala.*

import other

import other

core/src/main/scala/spark/scheduler/SparkListenerEventProcessor.scala
```diff
@@ -0,0 +1,62 @@
...
+
+/** Asynchronously passes SparkListenerEvents to registered SparkListeners. */
+class SparkListenerEventProcessor() {
+  /* sparkListeners is not thread safe, so this assumes that listeners are all added before any
+   * SparkListenerEvents occur. */
```
mateiz: This is probably not a great assumption to leave in.. can you just add a synchronized block around access to this, and around the per-event code in the thread below?

core/src/main/scala/spark/scheduler/SparkListenerEventProcessor.scala
```diff
@@ -0,0 +1,62 @@
...
+  private val sparkListeners = ArrayBuffer[SparkListener]()
+  private val eventQueue = new LinkedBlockingQueue[SparkListenerEvents]
```
pwendell: We might want to handle the case where this queue gets too big, e.g. the listeners actually can't keep up. Maybe we could check if the queue is really long and throw an exception. The alternative is that we get an OOM in the JVM, which will be really hard to understand. Maybe this is just paranoia...

core/src/main/scala/spark/ui/jobs/JobProgressListener.scala
```diff
@@ -9,6 +9,12 @@ import spark.scheduler.cluster.TaskInfo
 import spark.executor.TaskMetrics
 import collection.mutable
 
+/** Tracks task-level information to be displayed in the UI.
```
mateiz: Small note on style: leave the line with `/**` empty and start the text below, like this:

```
/**
 * Tracks task-level information ...
```

core/src/main/scala/spark/scheduler/SparkListenerEventProcessor.scala
pwendell: It seems like initializing this with a capacity is what we want.

core/src/main/scala/spark/ui/jobs/JobProgressListener.scala
```diff
@@ -39,129 +45,145 @@ private[spark] class JobProgressListener(val sc: SparkContext) extends SparkList
   override def onJobStart(jobStart: SparkListenerJobStart) {}
 
   override def onStageCompleted(stageCompleted: StageCompleted) = {
-    val stage = stageCompleted.stageInfo.stage
-    poolToActiveStages(stageToPool(stage)) -= stage
-    activeStages -= stage
-    completedStages += stage
-    trimIfNecessary(completedStages)
+    this.synchronized {
```
mateiz: Another note on Scala style: instead of putting `this.synchronized` inside a method, write it this way:

```
override def onStageCompleted(stageCompleted: StageCompleted): Unit = synchronized {
  // body
}
```

I guess this is slightly weirder but it's common throughout our code for methods that are fully synchronized, so we should stick with it.

markhamstra:
Do we want to consider making SparkListenerEventProcessor an Actor so that it can be more easily relocated?

core/src/main/scala/spark/scheduler/SparkListenerEventProcessor.scala
kayousterhout: I just made this a SynchronizedBuffer -- but let me know if there's some reason that using a synchronized block around it is better.

kayousterhout:
Thanks everyone for the comments -- all addressed!  @markhamstra, I considered making it an actor but was concerned that was too heavyweight.  In the past when I've tried adding actors that need to be called in the scheduler loop, it significantly decreases scheduling throughput.

kayousterhout:
Also there seems to be some Jenkins issue where it can't properly fetch the code from Git, so the tests are "failing".  I've run them locally but not sure how to fix the Jenkins problem?

core/src/main/scala/spark/scheduler/SparkListenerEventProcessor.scala
```diff
@@ -0,0 +1,72 @@
...
+  }.start()
+
+  def addListener(listener: SparkListener) {
+    sparkListeners += listener
+  }
+
+  def addEvent(event: SparkListenerEvents) {
+    val eventAdded = eventQueue.offer(event)
+    if (!eventAdded) {
+      logError("Dropping SparkListenerEvent because no remaining room in event queue. " +
```
pwendell: It might be better to just exit the JVM in this situation rather than write out an endless stream of error logs. Maybe it could log one error message and then shutdown.

pwendell: Alternatively, you could just print this at most once. That way the person can see the error message in their log but it won't, e.g. fill up the disk on their machine.

markhamstra:
This is a definite improvement, and probably about as far as SparkListener changes can be taken in 0.8, but there is still more that may need to be done with an API-breaking change in 0.9.  Making SparkListener itself be at least asynchronous and maybe also relocatable is probably the correct approach to making the model robust against the possibility of users registering many and/or slow and/or resource hungry listeners.  In other words, the processing of the eventQueue in SparkListenerEventProcessor would go from making synchronous method calls on a sequence of in-process listeners to asynchronously forwarding the queued events to a sequence of listeners, each of which may be out-of-process or even off-node.

mateiz:
I actually wouldn't overcomplicate this now, because 1) it's not clear yet how people will use it, 2) slow listeners are always going to be a problem (we can't just buffer messages for them), and 3) the SparkListener API is semi-internal anyway because it directly exposes scheduling concepts, such as stages. Let's see how it gets used -- people can always set up their own second-level thing to forward messages elsewhere.

core/src/main/scala/spark/scheduler/SparkListenerEventProcessor.scala
```diff
@@ -0,0 +1,74 @@
...
+package spark.scheduler
+
+import java.util.concurrent.LinkedBlockingQueue
+
+import scala.collection.mutable.{ArrayBuffer, SynchronizedBuffer}
+
+import spark.Logging
+
+/** Asynchronously passes SparkListenerEvents to registered SparkListeners. */
+class SparkListenerEventProcessor() extends Logging {
```
mateiz: I would rename this to just `SparkListenerBus`. Also, make it `private[spark]` since users don't make it directly.

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -102,7 +102,7 @@ class DAGScheduler(
 
   private[spark] val stageToInfos = new TimeStampedHashMap[Stage, StageInfo]
 
-  private[spark] val sparkListeners = ArrayBuffer[SparkListener]()
+  private val listenerEventProcessor = new SparkListenerEventProcessor()
```
mateiz: Same here with calling it "listenerBus"

core/src/main/scala/spark/scheduler/SparkListenerEventProcessor.scala
```diff
@@ -0,0 +1,74 @@
...
+        }
+      }
+    }
+  }.start()
+
+  def addListener(listener: SparkListener) {
+    sparkListeners += listener
+  }
+
+  def addEvent(event: SparkListenerEvents) {
```
mateiz: This can just be called `post` (I think it will be a bit clearer)

mateiz:
Thanks for the changes, Kay. I made a couple of comments on naming to make the names shorter, but otherwise it looks good.

kayousterhout:
Name changing done -- thanks Matei!!

mateiz:
Thanks a lot, Kay! I've merged this manually to deal with some small conflicts.

PR #793, opened 2013-08-08
Create a custom Kryo serializer for WrappedArray to work around Chill's inability to serialize WrappedArray's of specialized tuples.

rxin:
Once chill fixes the problem, we can revert this change (but we should keep the unit tests I added).

### DIFF
--- a/core/src/main/scala/spark/KryoSerializer.scala
+++ b/core/src/main/scala/spark/KryoSerializer.scala
@@ -27,6 +27,12 @@ import serializer.{SerializerInstance, DeserializationStream, SerializationStrea
 import spark.broadcast._
 import spark.storage._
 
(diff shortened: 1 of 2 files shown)


mateiz:
Can you add some tests with arrays of primitive types (e.g. Array[Int])? Those might be tricky depending on how WrappedArray.toArray works for them (since you don't have a class manifest for T).

ryanlecompte:
I just submitted a pull request for chill here: https://github.com/twitter/chill/pull/115

It fixes this issue. Hopefully it will get merged soon and we can get a new chill hotfix released.

rxin:
Actually my fix no longer works. It looks like it has trouble overriding the serializer for WrappedArray, and the Kryo one is still used. I will submit another temporary fix ...

PR #794, opened 2013-08-08
A quick bandaid fix for sc.parallelize for specialized tuples

rxin:
Again, this can be reverted once chill can handle WrappedArray of specialized tuples.


mateiz:
@rxin do you still want to merge this, or wait until Chill makes a bugfix release? I think they wanted to post one quickly.

ryanlecompte:
I think we can just upgrade now. Oscar just notified me that chill 0.3.1 is
on sonatype with the fix!

rxin:
Let's close this one and pull in the new chill release.

PR #795, opened 2013-08-08
Fix bug reported in PR 791 : a race condition in ConnectionManager and Connection

mridulm:
CC @pwendell
rewrite of PR #791

I have not tested this unfortunately, due to time constraints - please take a look and review.

### DIFF
--- a/core/src/main/scala/spark/network/Connection.scala
+++ b/core/src/main/scala/spark/network/Connection.scala
@@ -251,12 +251,15 @@ class SendingConnection(val address: InetSocketAddress, selector_ : Selector,
     }
   }
 
+  // outbox is used as a lock - ensure that it is always used as a leaf (since methods which 
+  // lock it are invoked in context of other locks)
   private val outbox = new Outbox(1)
   /*
-    This is orthogonal to whether we have pending bytes to write or not - and satisfies a slightly different purpose.
-    This flag is to see if we need to force reregister for write even when we do not have any pending bytes to write to socket.
-    This can happen due to a race between adding pending buffers, and checking for existing of data as detailed in
-    https://github.com/mesos/spark/pull/791#issuecomment-22294729
+    This is orthogonal to whether we have pending bytes to write or not - and satisfies a slightly 
+    different purpose. This flag is to see if we need to force reregister for write even when we 
+    do not have any pending bytes to write to socket.
+    This can happen due to a race between adding pending buffers, and checking for existing of 
+    data as detailed in https://github.com/mesos/spark/pull/791
    */
   private var needForceReregister = false
   val currentBuffers = new ArrayBuffer[ByteBuffer]()
@@ -346,7 +349,8 @@ class SendingConnection(val address: InetSocketAddress, selector_ : Selector,
             outbox.getChunk() match {
               case Some(chunk) => {
                 val buffers = chunk.buffers
-                // If we have 'seen' pending messages, then reset flag - since we handle that as normal registering of event (below)
+                // If we have 'seen' pending messages, then reset flag - since we handle that as normal 
+                // registering of event (below)
                 if (needForceReregister && buffers.exists(_.remaining() > 0)) resetForceReregister()
                 currentBuffers ++= buffers
               }
--- a/core/src/main/scala/spark/network/Connection.scala
+++ b/core/src/main/scala/spark/network/Connection.scala
@@ -45,11 +45,14 @@ abstract class Connection(val channel: SocketChannel, val selector: Selector,
   channel.socket.setKeepAlive(true)
   /*channel.socket.setReceiveBufferSize(32768) */
 
+  @volatile private var closed = false
   var onCloseCallback: Connection => Unit = null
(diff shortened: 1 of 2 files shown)


pwendell:
Hey @mridulm, thanks for contributing this.

One question, what happens if there is a context switch between line 126 and 127 of the Connection Manager, and in that time a new message is added to the outbox via `send()`? For example:

```
 val needReregister = register || conn.resetForceReregister()  [say this returns false]

 --> other thread calls send() and conn.registerInterest()  

 if (needReregister && conn.changeInterestForWrite()) {
      conn.registerInterest()
 }
```

mridulm:
If I understood the scenario correctly, you are referring to :

assume forceReset = false
thr1: after resetForceReregister. <pre-empted>
thr2: send() - set forceReset = true. <pre-empted>
thr1 = does not register interest

Correct ?
In send - we always register interest, so thr1 not registering interest
will not have an impact.

Or did I get the scenario/flow wrong ?
I want to ensure there are no bugs remaining in this, since everything else
depends on this !

Thanks,
Mridul

mridulm:
To add to above, in case the scenario was - will the write event register get dropped since there is a writerunnable already for the key : 
note that on selector finding socket writable, we would not ignore it - since thr1 above is within the lock already.
So the triggerWrite on socket writable event will only enter the sync block after thr1 finishes : allowing the trigger to actually trigger a write() call.

mateiz:
Jenkins, this is ok to test

pwendell:
Hey @mridul - yep, that actually looks fine - thanks for explaining. I'm
testing this on a cluster right now to try and make sure things are working
OK.

mridulm:
Hi Matei,

  Even though the change is pretty small (in comparison to most PR's :) ),
since this is a relatively critical codepath, I would like this to be
pretty thoroughly reviewed.
The issue Patrick unearthed (thx !) could have had interesting side effects
... so more eyes on the codepaths (ConnectionManager and Connection) would help !

Regards,
Mridul

pwendell:
Ya, @mateiz do you mind taking a look and see if it looks correct to you? This is a very important codepath. While this solution is a bit complicated, it does avoid busy-waiting and it also fixes the race condition, so I'm +1 for merging it.

After testing this on a cluster, no further races were witnessed. Of course, these aren't deterministic failures :)

mateiz:
Patrick, has Ali looked at it, since he's been looking at this code?

pwendell:
I think he has, but I'm going to ask him to do a final look through and give a +1. @alig - could you take a look?

alig:
Looks good to me @mridulm @pwendell @mateiz . Just one thing I don't quite get: `resetForceReregister()`, do we need it? 

Say we change the order of lines 110, 111-113 in ConnectionManager:

``` Scala
if (writeRunnableStarted.contains(key)) return
if (conn.changeInterestForWrite()) conn.unregisterInterest()
```

Now, first time the select loop returns you get to the second `if` expression, executing `unregisterInterest()`. Subsequent selection calls for that key will immediately return because the key is in `writeRunnableStarted`.

Now, `send(message)` automatically triggers interest for its key. Do we need the logic around `resetForceReregister()` then? We'd still `registerInterest()` when register is true.

Am I missing something?

mridulm:
If we do not unregister from selector, next selection operation (when we
return due to writeRunnableStarted containing the key) will return
immediately with the same key in writable set - resulting in busy loop of
the selector thread.

Regards,
Mridul

alig:
But whenever the boolean flag is true (writeRunnableStarted contains the key),
it must have already been unregistered, right? So how could it get called
again?

mridulm:
Consider something like this :

No writable launched for key1
selectorThr:   triggerWrite, unregister, no writable present, launch thread, back to sleep.
writeRunnableThr:   invoke write, process/preempt
thr1: invokes send - registerInterest - wakes up selectorThr.

Here, we have :
a) writeRunnableThr is either within write() or has returned but not entered the writeRunnableStarted sync block.
b) writeRunnableStarted has the key1.

For selectorThr, if key is not unregistered from selector, we will have something like :
selector.select(), selected key1 for write, triggerWrite returns immediately since (b), goes back to selector.select(), wakes up immediately : and this will continue indefinitely ... until thr1 removes key1 from writeRunnableStarted

Ideally, this last bit should happen asap (due to non blocking writes), but then that is left to vagaries of nio impl, OS scheduling, etc - and a busy loop might actually compound the problem.

I am not sure if I am clarifying the actual query you have, or I am misunderstanding the intent !

Regards,
Mridul

alig:
That explains it. There is some chance of busy-looping due to a concurrent write. That justifies `resetForceReregister`. Thanks for clarifying @mridulm.

+1 on this. Maybe @haoyuan can test it to make sure it works too. 

--Ali

haoyuan:
Sure. The test takes hours. Will get back tonight or tomorrow.

haoyuan:
The test ran well using this patch. No hung issue has been detected.

mateiz:
Alright, going to merge it then. Thanks @mridulm, Ali and Patrick.

PR #796, opened 2013-08-08
Bootstrap re-design

pwendell:
This patch uses a new bootstrap theme for the UI. It also changes some of the header sizes and improves the buttons in the log page. This is merged from an earlier patch that Karen submitted.


pwendell:
Jenkins, test this please.

core/src/main/resources/spark/ui/static/webui.css
```diff
@@ -5,6 +5,10 @@
   padding: 0;
 }
 
+body {
+  font-size: 15px !important;
```
mateiz, line 9: Patrick, why change this at all instead of leaving it to Bootstrap's default? The default is 14px I believe.

core/src/main/resources/spark/ui/static/webui.css
```diff
@@ -46,6 +50,7 @@
   padding: 0;
   padding-top: 7px;
   padding-left: 4px;
+  line-height: 15px !important;
```
mateiz, line 53: Same comment here

PR #797, opened 2013-08-08
Update to Chill 0.3.1

mateiz:


ryanlecompte:
Thanks for putting this together, @mateiz. I wanted to do it but got caught up with work. Looks great.

mateiz:
Merging it since it also passed the tests.

PR #798, opened 2013-08-09
Add option to perform synchronous stopping of the cluster

pwendell:
This is necessary when writing automated scripts that stop and start the cluster.

My bash skills leave something to be desired - if there is a more elegant way to implement any of this, please let me know.


mateiz:
Seems pretty useful, but have you tested it out on both Mac and Linux? Sometimes these behaviors only work on one.

bin/spark-daemon.sh
```diff
@@ -29,7 +29,7 @@
 #   SPARK_NICENESS The scheduling priority for daemons. Defaults to 0.
 ##
 
-usage="Usage: spark-daemon.sh [--config <conf-dir>] [--hosts hostlistfile] (start|stop) <spark-command> <spark-instance-number> <args...>"
+usage="Usage: spark-daemon.sh [--config <conf-dir>] [--hosts hostlistfile] (start|stop) <spark-command> <spark-instance-number> <wait-flag> <args...>"
 
 # if no args specified, show usage
 if [ $# -le 1 ]; then
```
mateiz, line 35: Technically this should be -le 3 now.

pwendell:
I'm actually going to close this one for now. A more elegant way to do this would be to have a version of `sc.stop()` that blocks until all of the executors are cleaned up. That would solve a super-set of the cases that we have here... this only covers the case where you are also terminating the standalone cluster. There is also the case where you stop a given job and want to wait until it has relinquished all resources.

PR #800, opened 2013-08-09
Pull HBASE_VERSION in the head of sbt build

dlyubimov:
Small nitpicking on SBT build as has been discussed on the list. Now that Spark pulls HBase jar and drops it into lib_managed, i think it should be far more obvious it does that, in the build.


mateiz:
Thanks for the PR, this looks useful.

mateiz:
Jenkins, this is ok to test

mateiz:
Thanks for the fix.

PR #803, opened 2013-08-11
Hadoop integration #1

jey:
Initial patch to allow one spark binary to target multiple hadoop versions. Has a few bugs that I'll submit fixes for shortly:
- `sbt clean` doesn't work
- unncessarily recompiles some source files when `$SPARK_HADOOP_VERSION` is changed
- doesn't work with hadoop-0.23.x


core/src/main/scala/org/apache/hadoop/mapred/SparkHadoopMapRedUtil.scala
```diff
@@ -0,0 +1,45 @@
...
+    val klass = firstAvailableClass("org.apache.hadoop.mapred.TaskAttemptContextImpl", "org.apache.hadoop.mapred.TaskAttemptContext")
+    val ctor = klass.getDeclaredConstructor(classOf[JobConf], classOf[TaskAttemptID])
+    ctor.newInstance(conf, attemptId).asInstanceOf[TaskAttemptContext]
+  }
+
+  def newTaskAttemptID(jtIdentifier: String, jobId: Int, isMap: Boolean, taskId: Int, attemptId: Int) = {
+    new TaskAttemptID(jtIdentifier, jobId, isMap, taskId, attemptId)
+  }
+
+  private def firstAvailableClass(first: String, second: String): Class[_] = {
```
mateiz, line 37: Small code organization suggestion -- factor this into spark.Utils because you have another copy of this method in HadoopMapReduceUtil

project/SparkBuild.scala
```diff
@@ -59,10 +43,23 @@ object SparkBuild extends Build {
...
+  lazy val activeProjects = Seq[ClasspathDependency](core, repl, examples, bagel, streaming, mllib, tools) ++ maybeYarn
+  lazy val everything = Seq[ProjectReference](core, repl, examples, bagel, streaming, mllib, tools, runpath) ++ maybeYarnRef
+  lazy val maybeYarn = if(isYarnMode) Seq[ClasspathDependency](yarn) else Seq[ClasspathDependency]()
+  lazy val maybeYarnRef = if(isYarnMode) Seq[ProjectReference](yarn) else Seq[ProjectReference]()
+
   // A configuration to set an alternative publishLocalConfiguration
   lazy val MavenCompile = config("m2r") extend(Compile)
   lazy val publishLocalBoth = TaskKey[Unit]("publish-local", "publish local for m2 and ivy")
 
+  lazy val hadoopVersion = scala.util.Properties.envOrElse("SPARK_HADOOP_VERSION", "1.0.4")
```
mateiz: Let's make this 1.2.1 by default since that's the current stable Hadoop version and I believe it's protocol-compatible with 1.0.4. Change this in update-classpath.sh as well.

core/src/main/scala/org/apache/hadoop/mapred/SparkHadoopMapRedUtil.scala
jey, line 37: Just gave that a shot, but turns out that `spark.Utils` is private to the package. Do we want to increase its visibility enough to be available from `org.apache.hadoop`?

jey:
Another known bug: some unit tests won't pass when running against YARN

project/SparkBuild.scala
jey: Will do

mridulm:
Any reason to move yarn from inside core to outside ? I dont think it makes sense outside of core ...

CC'ing @tgravescs

jey:
We are working to unify the Spark binaries so it won't have to be rebuilt for each Hadoop version. This involved moving YARN support out of core because the YARN APIs are not available when building against MapReduce v1 versions of Hadoop. The only effect that has on YARN users is that the `SPARK_JAR` environment variable has to be set to point to the `spark-yarn` assembly JAR instead of the `spark-core` assembly used previously.

mateiz:
Hey Jey, for the default Hadoop version, I actually suggested 1.2.1, not 1.1.2.

jey:
Oops. Fixed.

jey:
Here's a pretty strange error from Jenkins, looks like possibily a bug in Scala's JPropertiesWrapper's toMap method? Or maybe some other thread is caling System.setProperty at around the same time?

```
[info] DriverSuite:
Exception in thread "main" java.util.ConcurrentModificationException
    at java.util.Hashtable$Enumerator.next(Hashtable.java:1200)
    at scala.collection.JavaConversions$JPropertiesWrapper$$anon$3.next(JavaConversions.scala:933)
    at scala.collection.JavaConversions$JPropertiesWrapper$$anon$3.next(JavaConversions.scala:930)
    at scala.collection.Iterator$class.foreach(Iterator.scala:772)
    at scala.collection.JavaConversions$JPropertiesWrapper$$anon$3.foreach(JavaConversions.scala:930)
    at scala.collection.IterableLike$class.foreach(IterableLike.scala:73)
    at scala.collection.JavaConversions$JPropertiesWrapper.foreach(JavaConversions.scala:903)
    at scala.collection.TraversableOnce$class.toMap(TraversableOnce.scala:256)
    at scala.collection.JavaConversions$JPropertiesWrapper.toMap(JavaConversions.scala:903)
    at spark.SparkContext.<init>(SparkContext.scala:254)
    at spark.DriverWithoutCleanup$.main(DriverSuite.scala:53)
    at spark.DriverWithoutCleanup.main(DriverSuite.scala)
[info] - driver should exit after finishing *** FAILED ***
[info]   SparkException was thrown during property evaluation. (DriverSuite.scala:37)
[info]   Message: Process List(./run, spark.DriverWithoutCleanup, local-cluster[2,1,512]) exited with code 1
[info]   Occurred at table row 1 (zero based, not counting headings), which had values (
[info]     master = local-cluster[2,1,512]
[info]   )
```

Jenkins, retest this please.

jey:
BTW, unit tests pass under YARN too after 1a0607a fixed a silly mistake in `bin/update-classpath.sh`

jey:
Jenkins, retest this please.

mridulm:
Moving yarn code out of core does not look like a good decision - it does not make sense as a top level module.
It is better to simply have different profile which adds the yarn specific targets and subdirectories for building the code - the default building for non-yarn case which does not include the yarn sub-dir.

jey:
It's simpler to have a different module, especially due to Maven insanity. The only way this materially affects YARN users is that they need to provide the path to the spark-yarn assembly instead of the spark-core assembly when launching a YARN job.

jey:
I'm happy to consider the profile-based approach, but please list the benefits, as they are not obvious to me.

mateiz:
Yeah, the core problem is that the YARN code won't work unless you are linking against a YARN-enabled Hadoop, and we don't want to be publishing separate spark-core artifacts for every Hadoop version. YARN users will just have to add spark-yarn as an additional dependency until either Hadoop improves its packaging (e.g. adds some kind of yarn-client package that still works with old Hadoop versions) or YARN is widely enough deployed that we don't want to run on non-YARN clusters (unlikely in the near future).

jey:
I'm closing this PR and will submit a new one targeted against mesos/master instead of mesos/branch-0.8

mridulm:
I am not commenting about including yarn artifacts within the same spark core jar - but where the code is hosted within spark source tree. Hosting it as a top level directory does not make much sense when it has no standalone value since it is very closely tied to spark core (and specializing it in a implementation dependent way).
Compare with other top level directories - bagel, ml, shark, doc, etc : it makes sense for them to exist there right now, not so for yarn directory.

Having two jars make sense - since expected functionality is different : and we can generate different artifacts when driven by different profiles. Btw, we would not need to include mesos jars while building spark yarn jar - if it is getting separated out.

mateiz:
The folder structure in SBT/Maven just reflects modules, not anything about standalone value. I don't see any reason to complicate the build file to do this. We've actually fought a lot with the profiles in Maven and SBT to try todo conditional builds, and it's really painful. This stuff will still get compiled, tested, etc together.

PR #804, opened 2013-08-11
Fixed path to JavaALS.java and JavaKMeans.java, fixed hadoop2-yarn profi...

apivovarov:
...le


shivaram:
Jenkins, this is ok to test

shivaram:
Thanks for catching this. Fixes look good to me -- Will merge them once Jenkins passes.

apivovarov:
My pleasure. 
What you think if I change hadoop dependency from 2.0.2 to 2.0.5 in hadoop2-yarn profile in parent pom.xml?

shivaram:
I wouldn't recommend changing the dependency in this PR. You could open another PR and I guess @mridulm or @mateiz will take a look. Also any reasons why this should be changed ?

apivovarov:
I made a separate PR for yarn.version change https://github.com/mesos/spark/pull/806
latest hadoop-2.0.x version is 2.0.5

rxin:
Alex, do you mind rebasing the branch, so the extra commits (the original one and the reverting one) about 2.0.2->2.0.5 don't show up on this one?

PR #806, opened 2013-08-11
Changed yarn.version to 2.0.5 in pom.xml

apivovarov:


apivovarov:
one of the hadoop2 distro (probably the only one prod ready) Wandisco WDD v3.6 is based on hadoop-2.0.5
we use spark / shark with this distro. This is why we need this fix

mateiz:
Looks good, thanks.

PR #807, opened 2013-08-11
Change scala.Option to Guava Optional in Java APIs

JoshRosen:
This pull request changes the public Java APIs to use Guava's `Optional` in place of `scala.Option`, since `Optional` will be friendlier to Java users ([discussion from dev list](https://mail-archives.apache.org/mod_mbox/incubator-spark-dev/201308.mbox/%3CCAOEPXP5iWWTSEhJGq5eafpPY1Dxp8ZTAfuhX9w-2%2BaQ3hApbfA%40mail.gmail.com%3E)).

I added a test case for leftOuterJoin in the Java API (I didn't implement 100% test coverage here because there's symmetry between the different *OuterJoin methods).

I also updated the Java API completeness checker's rule for handling `Option`, rewriting it to `Optional` when it appears in return types and dropping it (e.g. accepting nulls) when it appears in function arguments.

### DIFF
--- a/core/src/main/scala/spark/api/java/JavaPairRDD.scala
+++ b/core/src/main/scala/spark/api/java/JavaPairRDD.scala
@@ -41,6 +41,8 @@ import spark.Partitioner._
 import spark.RDD
 import spark.SparkContext.rddToPairRDDFunctions
 
+import com.google.common.base.Optional
+
 class JavaPairRDD[K, V](val rdd: RDD[(K, V)])(implicit val kManifest: ClassManifest[K],
   implicit val vManifest: ClassManifest[V]) extends JavaRDDLike[(K, V), JavaPairRDD[K, V]] {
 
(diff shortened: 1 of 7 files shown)


core/src/main/scala/spark/api/java/JavaPairRDD.scala
```diff
@@ -41,6 +41,8 @@ import spark.Partitioner._
 import spark.RDD
 import spark.SparkContext.rddToPairRDDFunctions
 
+import com.google.common.base.Optional
```
pwendell: Hey Josh - we're trying to alphabetize imports now. Would you mind bumping this up in the import list?

JoshRosen: @pwendell 

I think we were already alphabetizing within each "block" of imports, so are we extending this to a total alphabetical ordering of all imports?

Should `com.google.common.base.Optional` appear at the top of the imports list?  Or should the `com.google.*` imports come before the `spark.*` imports in a separate section?

Can we document the new import style in https://github.com/mesos/spark/blob/master/docs/contributing-to-spark.md?

mateiz: I think Patrick is saying that com should go in the block that currently has org.\* above. Basically the blocks are the following:
- Java standard library
- Scala standard library
- Third-party packages
- Spark packages

Within each block, stuff should be alphabetized. But you added com below them as a fifth block.

JoshRosen: I've fixed this now.

mateiz:
Okay, thanks. Merging this.

PR #811, opened 2013-08-13
implementing SPARK-865: Add the equivalent of ADD_JARS to PySpark

AndreSchumacher:
Implementing

https://spark-project.atlassian.net/browse/SPARK-865

Now the ADD_FILES environment variable is parsed as a list of ':' separated files that are handed over to SparkContext to be distributed and made available to the nodes.


python/pyspark/shell.py
```diff
@@ -7,10 +7,15 @@
 import pyspark
 from pyspark.context import SparkContext
 
+# this is the equivalent of ADD_JARS
+add_files =  os.environ.get("ADD_FILES").split(':') if os.environ.get("ADD_FILES") != None else None
```
mateiz, line 11: It would be better to use commas here to be consistent with ADD_JARS and to make it work well on Windows too, where : can be in drive names. Also, looks like you have two spaces after the = sign.

AndreSchumacher, line 11: Thanks for pointing this out. I sent a revised pull request (with the
second commit squashed).

mateiz, line 11: Cool, thanks! I'll close this one then and merge that one.

PR #812, opened 2013-08-13
Create SparkContext in beforeAll for MLLib tests

shivaram:
This overcomes test failures that occur using Maven.


mateiz:
Looks good, thanks!

PR #813, opened 2013-08-13
Implementing SPARK-865: Add the equivalent of ADD_JARS to PySpark

AndreSchumacher:
Now ADD_FILES uses a comma as file name separator.


mateiz:
Thanks Andre!

PR #816, opened 2013-08-13
should call reviveOffer in fine-grained mesos mode after tasks successfully finished 

xiajunluan:
[Spark-872]I think we should call reviveOffer in statusUpdate function to request resource, In scheduler.statusUpdate function, it calls reviveOffer only for TASK_LOST and TASK_FAILED, so it need deal with TASK_FINISHED scenario, this improvement will enhance performance significantly.

### DIFF
--- a/core/src/main/scala/spark/scheduler/mesos/MesosSchedulerBackend.scala
+++ b/core/src/main/scala/spark/scheduler/mesos/MesosSchedulerBackend.scala
@@ -286,6 +286,9 @@ private[spark] class MesosSchedulerBackend(
         }
       }
       scheduler.statusUpdate(tid, state, status.getData.asReadOnlyByteBuffer)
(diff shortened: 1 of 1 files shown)


mateiz:
In this mode, Mesos itself should send back offers when our refusals time out after 1 second. Also, we call reviveOffers when a TaskSetManager is submitted. What kind of performance difference have you seen with this, and with what Mesos version?

Calling reviveOffers on every task finish can also be quite expensive because Mesos will list all the free nodes in the cluster for us.

mateiz:
OK, I'd like to ask the Mesos guys about this, because I think calling reviveOffers on every task-finished will be very expensive on a larger cluster. In older versions of Mesos, the system made a new resource offer as soon as a task finished. Has this changed recently? CC @benh

PR #818, opened 2013-08-13
Properly account for killed tasks.

kayousterhout:
The TaskState class's isFinished() method didn't return true for
KILLED tasks, which means some resources are never reclaimed
for tasks that are killed. This also made it inconsistent with the
isFinished() method used by CoarseMesosSchedulerBackend.

I don't think this is an issue now because the KILLED state isn't used -- 
but it will be a problem in the future.

### DIFF
--- a/core/src/main/scala/spark/TaskState.scala
+++ b/core/src/main/scala/spark/TaskState.scala
@@ -24,9 +24,11 @@ private[spark] object TaskState
 
   val LAUNCHING, RUNNING, FINISHED, FAILED, KILLED, LOST = Value
 
+  val FINISHED_STATES = Set(FINISHED, FAILED, KILLED, LOST)
+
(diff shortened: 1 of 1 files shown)


core/src/main/scala/spark/TaskState.scala
```diff
@@ -26,7 +26,7 @@ private[spark] object TaskState
 
   type TaskState = Value
 
-  def isFinished(state: TaskState) = Seq(FINISHED, FAILED, LOST).contains(state)
+  def isFinished(state: TaskState) = Seq(FINISHED, FAILED, KILLED, LOST).contains(state)
```
ryanlecompte: Should we use a `Set` instead of a `Seq` here?

mateiz:
Looks good, thanks.

PR #819, opened 2013-08-13
Change SVM to use {0,1} labels

shivaram:
With this change, all classification models will use {0,1} labels in ML Lib. This change also introduces a data-validation pass that can check if the user's input matches what we expect
@atalwalkar - As we discussed, this should be extensible to check for other properties as well.

Thanks to @etrain for catching this and @pxinghao for help with debugging.

### DIFF
--- a/mllib/src/main/scala/spark/mllib/classification/LogisticRegression.scala
+++ b/mllib/src/main/scala/spark/mllib/classification/LogisticRegression.scala
@@ -17,12 +17,13 @@
 
 package spark.mllib.classification
 
+import scala.math.round
+
 import spark.{Logging, RDD, SparkContext}
 import spark.mllib.optimization._
 import spark.mllib.regression._
 import spark.mllib.util.MLUtils
-
-import scala.math.round
+import spark.mllib.util.DataValidators
 
 import org.jblas.DoubleMatrix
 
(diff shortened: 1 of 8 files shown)


jey:
Why do we use `{0, 1}` as the labels? AFAIK,  `{-1, 1}` is more conventional and intuitive in terms of the underlying algorithms. Maybe it depends on the audience; {0,1} makes sense if it's a boolean that means "is-member-of-class?".

pxinghao:
I think the reason we decided on {0,1} is due to the more natural extension to multi-class classification.

jey:
Good point, makes sense

atalwalkar:
I agree that {-1,1} is used more by the SVM folks, but {0,1} is also
commonly used, and is the standard in statistics, which is why we went with
it.  The main goal is to be consistent across all algorithms, and this PR
satisfies this goal.

mateiz:
Guys, how ready is this to merge? It would be nice to get a release candidate out for 0.8 soon.

shivaram:
Hey Matei -- Sorry I got caught up and didn't get to finishing this. I will post an update to this and #830 by tonight/tomorrow.

shivaram:
I've added the option to turn off data validation and added a test case for that as well. Also I did some minor refactoring to make all the variables look similar in GLM - They have a default value but can be overriden

etrain:
Thanks for adding the option to disable validation - should save us some overhead in the case where the data is already sane.

Looks good to me - thanks Shivaram!

PR #822, opened 2013-08-13
Adding GC Stats to TaskMetrics (and three small fixes)

pwendell:
The main change here is that we add time the JVM spent in GC for reach task to `TaskMetrics` and the web UI.

There are also three minor fixes related to terminology/rendering of the UI pages.

### DIFF
--- a/core/src/main/scala/spark/executor/Executor.scala
+++ b/core/src/main/scala/spark/executor/Executor.scala
@@ -158,7 +158,7 @@ private[spark] class Executor(executorId: String, slaveHostname: String, propert
         case t: Throwable => {
           val serviceTime = (System.currentTimeMillis() - taskStart).toInt
           val metrics = attemptedTask.flatMap(t => t.metrics)
-          metrics.foreach{m =>
+          metrics.foreach {m =>
             m.executorRunTime = serviceTime
             m.jvmGCTime = getTotalGCTime - startGCTime
           }
--- a/core/src/main/scala/spark/executor/Executor.scala
+++ b/core/src/main/scala/spark/executor/Executor.scala
@@ -17,18 +17,17 @@
 
 package spark.executor
 
-import java.io.{File, FileOutputStream}
-import java.net.{URI, URL, URLClassLoader}
+import java.io.{File}
+import java.lang.management.ManagementFactory
+import java.nio.ByteBuffer
 import java.util.concurrent._
 
-import org.apache.hadoop.fs.FileUtil
+import scala.collection.JavaConversions._
+import scala.collection.mutable.HashMap
 
-import scala.collection.mutable.{ArrayBuffer, Map, HashMap}
-
-import spark.broadcast._
 import spark.scheduler._
 import spark._
-import java.nio.ByteBuffer
+
 
 /**
  * The Mesos executor for Spark.
@@ -116,6 +115,9 @@ private[spark] class Executor(executorId: String, slaveHostname: String, propert
       context.statusUpdate(taskId, TaskState.RUNNING, EMPTY_BYTE_BUFFER)
       var attemptedTask: Option[Task[Any]] = None
       var taskStart: Long = 0
+      def getTotalGCTime = ManagementFactory.getGarbageCollectorMXBeans.map(g => g.getCollectionTime).sum
+      val startGCTime = getTotalGCTime
+
       try {
         SparkEnv.set(env)
         Accumulators.clear()
@@ -132,6 +134,7 @@ private[spark] class Executor(executorId: String, slaveHostname: String, propert
           m.hostname = Utils.localHostName
           m.executorDeserializeTime = (taskStart - startTime).toInt
           m.executorRunTime = (taskFinish - taskStart).toInt
+          m.jvmGCTime = getTotalGCTime - startGCTime
         }
         //TODO I'd also like to track the time it takes to serialize the task results, but that is huge headache, b/c
         // we need to serialize the task metrics first.  If TaskMetrics had a custom serialized format, we could
@@ -155,7 +158,10 @@ private[spark] class Executor(executorId: String, slaveHostname: String, propert
         case t: Throwable => {
           val serviceTime = (System.currentTimeMillis() - taskStart).toInt
           val metrics = attemptedTask.flatMap(t => t.metrics)
-          metrics.foreach{m => m.executorRunTime = serviceTime}
+          metrics.foreach{m =>
+            m.executorRunTime = serviceTime
+            m.jvmGCTime = getTotalGCTime - startGCTime
+          }
           val reason = ExceptionFailure(t.getClass.getName, t.toString, t.getStackTrace, metrics)
           context.statusUpdate(taskId, TaskState.FAILED, ser.serialize(reason))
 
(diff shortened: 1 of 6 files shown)


core/src/main/scala/spark/scheduler/local/LocalScheduler.scala
```diff
@@ -210,7 +218,10 @@ private[spark] class LocalScheduler(threads: Int, val maxFailures: Int, val sc:
       case t: Throwable => {
         val serviceTime = System.currentTimeMillis() - taskStart
         val metrics = attemptedTask.flatMap(t => t.metrics)
-        metrics.foreach{m => m.executorRunTime = serviceTime.toInt}
+        metrics.foreach{ m =>
```
mateiz: Add a space before the {

core/src/main/scala/spark/executor/Executor.scala
```diff
@@ -155,7 +158,10 @@ private[spark] class Executor(executorId: String, slaveHostname: String, propert
         case t: Throwable => {
           val serviceTime = (System.currentTimeMillis() - taskStart).toInt
           val metrics = attemptedTask.flatMap(t => t.metrics)
-          metrics.foreach{m => m.executorRunTime = serviceTime}
+          metrics.foreach{m =>
```
mateiz: Add a space before the {

core/src/main/scala/spark/scheduler/local/LocalScheduler.scala
```diff
@@ -202,6 +209,7 @@ private[spark] class LocalScheduler(threads: Int, val maxFailures: Int, val sc:
       val serviceTime = System.currentTimeMillis() - taskStart
       logInfo("Finished " + taskId)
       deserializedTask.metrics.get.executorRunTime = serviceTime.toInt
+      deserializedTask.metrics.get.jvmGCTime = getTotalGCTime - startGCTime
```
rxin, line 211: Does this lead to double counting if two tasks are running at the same time and a GC happens?

rxin, line 211: It is actually accurate at a per-task level - we just can't report the aggregated time. Do we ever aggregate them?

pwendell, line 211: GC stops the entire JVM, so it does "double count" but these are IMO correct. The main point here is to make it easy to see if a slow task is caused by a GC pause.

We can add total time spent in GC as well in a subsequent PR. It's a bit confusing how you should count the total though... for instance do you "double count" or not, if multiple tasks are running during a GC event. Another useful thing to add would be at the executor level add a metric for what proportion of overall time is spent in GC.

rxin, line 211: For young gen concurrent gc's, are the time reported here too?

I think this is ok as long as we don't report aggregated gc time based on these numbers, for reasons pointed out by both of us earlier.

pwendell, line 211: It's up the the individual collector to decide what to report. I'm not sure how the young collector in marksweep (called `PS Scavenge`) returns collection time.

Here is an example of code that treats collectors differently, but they just do young vs old (whereas we probably want concurrent vs not).
http://svn.cometd.org/trunk/cometd-java/cometd-java-client/src/test/java/org/cometd/client/StatisticsHelper.java

pwendell, line 211: One thing is we could exclude `PS Scavenge` if it's present... do you think that would be better? Then we'd only include stop-the-world events:

Also this is a good link as well:
http://docs.neo4j.org/chunked/stable/configuration-jvm.html

pwendell, line 211: Actually @rxin - maybe we should stick with the simplest thing at first, just use the JVM's own statistic. We can do finer grained reporting a bit later, such as breaking down between old/new GC and concurrent/stop the world.

rxin, line 211: i agree

pwendell:
Okay I addressed @mateiz's comments and looks like @rxin is okay with this as-is. Let me know if there are any other comments.

core/src/main/scala/spark/ui/jobs/StagePage.scala
```diff
@@ -154,10 +157,17 @@ private[spark] class StagePage(parent: JobProgressUI) {
...
+      {if (shuffleRead) {
+        <td>{metrics.flatMap{m => m.shuffleReadMetrics}.map{s =>
+          Utils.memoryBytesToString(s.remoteBytesRead)}.getOrElse("")}</td>
+      }}
+      {if (shuffleWrite) {
+        <td>{metrics.flatMap{m => m.shuffleWriteMetrics}.map{s =>
+          Utils.memoryBytesToString(s.shuffleBytesWritten)}.getOrElse("")}</td>
+      }}
+      <td sorttable_customkey={gcTime.toString}>
+        {if (gcTime > 0) {parent.formatDuration(gcTime)} else ""}
```
mateiz: Does this need braces around parent.formatDuration?

core/src/main/scala/spark/scheduler/local/LocalScheduler.scala
```diff
@@ -31,6 +34,7 @@ import spark.scheduler._
 import spark.scheduler.cluster._
 import spark.scheduler.cluster.SchedulingMode.SchedulingMode
 import akka.actor._
+import management.ManagementFactory
```
mateiz: Use the full package name for this (java.lang.management.ManagementFactory)

core/src/main/scala/spark/executor/Executor.scala
```diff
@@ -155,7 +158,10 @@ private[spark] class Executor(executorId: String, slaveHostname: String, propert
         case t: Throwable => {
           val serviceTime = (System.currentTimeMillis() - taskStart).toInt
           val metrics = attemptedTask.flatMap(t => t.metrics)
-          metrics.foreach{m => m.executorRunTime = serviceTime}
+          metrics.foreach {m =>
```
mateiz: For these, I actually meant the right style is `metrics.foreach { m =>` when they're multiline, but on second thought, maybe just do `for (m <- metrics)`

core/src/main/scala/spark/scheduler/local/LocalScheduler.scala
```diff
@@ -210,7 +218,10 @@ private[spark] class LocalScheduler(threads: Int, val maxFailures: Int, val sc:
       case t: Throwable => {
         val serviceTime = System.currentTimeMillis() - taskStart
         val metrics = attemptedTask.flatMap(t => t.metrics)
-        metrics.foreach{m => m.executorRunTime = serviceTime.toInt}
+        metrics.foreach {m =>
```
mateiz: Ditto here

core/src/main/scala/spark/scheduler/local/LocalScheduler.scala
mateiz: Actually this looks like a duplicate import anyway

mateiz:
Just saw some other minor issues, but otherwise it seems good.

core/src/main/scala/spark/executor/Executor.scala
pwendell: Where in the style guide does it say to do that? I could only find a single-line reference for foreach and it didn't put a space after the open paren:
http://docs.scala-lang.org/style/method-invocation.html

mateiz: I'm just saying based on other code we have that uses braces. For parens it doesn't make sense.

pwendell:
@mateiz thanks for the style catches. These are fixed now.

mateiz:
Cool, thanks.

PR #823, opened 2013-08-14
Added a fast and low-memory append-only map for shuffle operations

mateiz:
This is an attempt to reduce the CPU cost and memory usage of shuffles by taking advantage of the properties their hashmaps need. In particular, the hashmaps there are append-only, and a common operation is updating a key's value based on the old value. The included AppendOnlyMap class uses open hashing to use less space than Java's (by not having a linked list per bucket), does not support deletes, and has a changeValue operation to update a key in place without following the hash chain twice.

This is just an experiment now because it remains to test it in real Spark apps, but in micro-benchmarks against java.util.HashMap, scala.collection.mutable.HashMap, this is 20-30% smaller and 10-40% faster depending on the number and type of keys. It's also noticeably faster than fastutil's Object2ObjectOpenHashMap.

### DIFF
--- a/core/src/main/scala/spark/util/AppendOnlyMap.scala
+++ b/core/src/main/scala/spark/util/AppendOnlyMap.scala
@@ -1,0 +1,241 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package spark.util
+
+/**
+ * A simple open hash table optimized for the append-only use case, where keys
+ * are never removed, but the value for each key may be changed.
+ *
+ * This implementation uses quadratic probing with a power-of-2 hash table
+ * size, which is guaranteed to explore all spaces for each key (see
+ * http://en.wikipedia.org/wiki/Quadratic_probing).
+ *
+ * TODO: Cache the hash values of each key? java.util.HashMap does that.
+ */
+private[spark]
+class AppendOnlyMap[K, V](initialCapacity: Int = 64) extends Iterable[(K, V)] with Serializable {
+  if (!isPowerOf2(initialCapacity)) {
+    throw new IllegalArgumentException("Initial capacity must be power of 2")
+  }
+  if (initialCapacity >= (1 << 30)) {
+    throw new IllegalArgumentException("Can't make capacity bigger than 2^29 elements")
+  }
+
+  private var capacity = initialCapacity
+  private var curSize = 0
+
+  // Holds keys and values in the same array for memory locality; specifically, the order of
+  // elements is key0, value0, key1, value1, key2, value2, etc.
+  private var data = new Array[AnyRef](2 * capacity)
+
+  // Treat the null key differently so we can use nulls in "data" to represent empty items.
+  private var haveNullValue = false
+  private var nullValue: V = null.asInstanceOf[V]
+
+  private val LOAD_FACTOR = 0.7
+
+  /** Get the value for a given key */
+  def apply(key: K): V = {
+    val k = key.asInstanceOf[AnyRef]
+    if (k.eq(null)) {
+      return nullValue
+    }
+    val mask = capacity - 1
+    var pos = rehash(k.hashCode) & mask
+    var i = 1
+    while (true) {
+      val curKey = data(2 * pos)
+      if (curKey.eq(k) || curKey.eq(null) || curKey == k) {
+        return data(2 * pos + 1).asInstanceOf[V]
+      } else {
+        val delta = i
+        pos = (pos + delta) & mask
+        i += 1
+      }
+    }
+    return null.asInstanceOf[V]
+  }
+
+  /** Set the value for a key */
+  def update(key: K, value: V) {
+    val k = key.asInstanceOf[AnyRef]
+    if (k.eq(null)) {
+      if (!haveNullValue) {
+        incrementSize()
+      }
+      nullValue = value
+      haveNullValue = true
+      return
+    }
+    val isNewEntry = putInto(data, k, value.asInstanceOf[AnyRef])
+    if (isNewEntry) {
+      incrementSize()
+    }
+  }
+
+  /**
+   * Set the value for key to updateFunc(hadValue, oldValue), where oldValue will be the old value
+   * for key, if any, or null otherwise. Returns the newly updated value.
+   */
+  def changeValue(key: K, updateFunc: (Boolean, V) => V): V = {
+    val k = key.asInstanceOf[AnyRef]
+    if (k.eq(null)) {
+      if (!haveNullValue) {
+        incrementSize()
+      }
+      nullValue = updateFunc(haveNullValue, nullValue)
+      haveNullValue = true
+      return nullValue
+    }
+    val mask = capacity - 1
+    var pos = rehash(k.hashCode) & mask
+    var i = 1
+    while (true) {
+      val curKey = data(2 * pos)
+      if (curKey.eq(null)) {
+        val newValue = updateFunc(false, null.asInstanceOf[V])
+        data(2 * pos) = k
+        data(2 * pos + 1) = newValue.asInstanceOf[AnyRef]
+        incrementSize()
+        return newValue
+      } else if (curKey.eq(k) || curKey == k) {
+        val newValue = updateFunc(true, data(2*pos + 1).asInstanceOf[V])
+        data(2*pos + 1) = newValue.asInstanceOf[AnyRef]
+        return newValue
+      } else {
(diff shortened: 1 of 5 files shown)


core/src/main/scala/spark/util/AppendOnlyMap.scala
```diff
@@ -0,0 +1,241 @@
...
+ *
+ * This implementation uses quadratic probing with a power-of-2 hash table
+ * size, which is guaranteed to explore all spaces for each key (see
+ * http://en.wikipedia.org/wiki/Quadratic_probing).
+ *
+ * TODO: Cache the hash values of each key? java.util.HashMap does that.
+ */
+private[spark]
+class AppendOnlyMap[K, V](initialCapacity: Int = 64) extends Iterable[(K, V)] with Serializable {
+  if (!isPowerOf2(initialCapacity)) {
```
rxin: Maybe change this to 

```
require(isPowerOf2(initialCapacity), "Initial capacity must be power of 2")
```

and ditto for initialCapacity >= 1 << 30

mateiz:
FYI, for anyone interested, my benchmark vs fastutil and Java/Scala maps is at http://www.cs.berkeley.edu/~matei/maptest.tgz.

mateiz:
You can run it with, for example, `sbt/sbt "run 10000 10000000"`. Vary the first number to change the number of keys -- things like cache locality matter more or less with different numbers of them.

core/src/main/scala/spark/util/AppendOnlyMap.scala
rxin: Actually, it'd be better to get the user to pass in an arbitrary initial size, and then we just round it up to the next power of 2.

core/src/main/scala/spark/util/AppendOnlyMap.scala
```diff
@@ -0,0 +1,241 @@
...
+  /**
+   * Put an entry into a table represented by data, returning true if
+   * this increases the size of the table or false otherwise. Assumes
+   * that "data" has at least one empty slot.
+   */
+  private def putInto(data: Array[AnyRef], key: AnyRef, value: AnyRef): Boolean = {
+    val mask = (data.length / 2) - 1
+    var pos = rehash(key.hashCode) & mask
+    var i = 1
+    while (true) {
```
ryanlecompte, line 186: Are we sure we'll always break out of this `while (true)`?

ryanlecompte:
Exciting stuff! Looking forward to any shuffle benchmarks/comparisons in a real cluster with this change.

core/src/main/scala/spark/util/AppendOnlyMap.scala
mateiz, line 186: Yes, this is what the quadratic probing method and power-of-2 table size guarantees (see http://en.wikipedia.org/wiki/Quadratic_probing)

core/src/main/scala/spark/util/AppendOnlyMap.scala
```diff
@@ -0,0 +1,241 @@
...
+ * are never removed, but the value for each key may be changed.
+ *
+ * This implementation uses quadratic probing with a power-of-2 hash table
+ * size, which is guaranteed to explore all spaces for each key (see
+ * http://en.wikipedia.org/wiki/Quadratic_probing).
+ *
+ * TODO: Cache the hash values of each key? java.util.HashMap does that.
+ */
+private[spark]
+class AppendOnlyMap[K, V](initialCapacity: Int = 64) extends Iterable[(K, V)] with Serializable {
```
wannabeast, line 31: Might as well `[K <: AnyRef, V <: AnyRef]` and avoid all of the downcasting to `AnyRef`.

mateiz, line 31: Unfortunately that won't make it work for primitive types, which we allow in some clients that use this class (e.g. Aggregator and CoGroupedRDD). The casting is ugly but it seems necessary until we build some specialized classes for those types, which would be a later project.

core/src/main/scala/spark/util/AppendOnlyMap.scala
```diff
@@ -0,0 +1,241 @@
...
+    while (pos < capacity) {
+      if (!data(2 * pos).eq(null)) {
+        putInto(newData, data(2 * pos), data(2 * pos + 1))
+      }
+      pos += 1
+    }
+    data = newData
+    capacity = newCapacity
+  }
+
```
mateiz, line 223: Just dealt with this by rounding capacity up to the next power of 2, as Reynold suggested

core/src/main/scala/spark/util/AppendOnlyMap.scala
wannabeast, line 186: My concern was that `incrementSize()`, and consequently `growTable()`, is only called _after_ this attempts to add an entry, meaning that there must be at least one free slot on method entry, else the loop never terminates. `initialCapacity >= 1` is `require`d, so a free slot exists on the first call; none would exist on subsequent calls only if we could increment `curSize` to `capacity` without growing the table. This could happen (modulo multiplication rounding errors) only if `LOAD_FACTOR >= 1`.

One could iterate only `capacity` times, then fall through to an error message, like in CLRS, but this would add a check per iteration.

I was thinking that this search logic could be extracted and shared with `apply()` and `changeValue()`. Don't know whether the Scala or JIT compiler would inline it for the same performance. But then I realized that the most likely case, which should be tested first, differs between `putInto()` and `apply()`/`changeValue()`; the former expects to find an empty slot, the latter expect to find the given key.

Last week you were talking of using double hashing...? I notice that [fastutil seemed to have a performance issue with that and replaced it with linear probing](http://fastutil.di.unimi.it/docs/overview-summary.html).

core/src/main/scala/spark/util/AppendOnlyMap.scala
```diff
@@ -0,0 +1,228 @@
...
+    val k = key.asInstanceOf[AnyRef]
+    if (k.eq(null)) {
+      return nullValue
+    }
+    val mask = capacity - 1
+    var pos = rehash(k.hashCode) & mask
+    var i = 1
+    while (true) {
+      val curKey = data(2 * pos)
+      if (curKey.eq(k) || curKey.eq(null) || curKey == k) {
```
wannabeast, line 59: When `curKey.eq(null)`, the key isn't present, and `data(2 * pos + 1)` is returned, which should be `null`. Clearer to say

``` scala
else if (curKey.eq(null)) { return null.asInstanceOf[V] }
```

in parallel with the similar logic in `changeValue()`.

core/src/main/scala/spark/util/AppendOnlyMap.scala
```diff
@@ -0,0 +1,228 @@
...
+
+  // Holds keys and values in the same array for memory locality; specifically, the order of
+  // elements is key0, value0, key1, value1, key2, value2, etc.
+  private var data = new Array[AnyRef](2 * capacity)
+
+  // Treat the null key differently so we can use nulls in "data" to represent empty items.
+  private var haveNullValue = false
+  private var nullValue: V = null.asInstanceOf[V]
+
+  private val LOAD_FACTOR = 0.7
```
wannabeast, line 46: I'd add `assert(LOAD_FACTOR > 0 && LOAD_FACTOR < 1)` to ensure that `incrementSize()` doesn't misbehave.

core/src/main/scala/spark/util/AppendOnlyMap.scala
```diff
@@ -0,0 +1,228 @@
...
+    var i = 1
+    while (true) {
+      val curKey = data(2 * pos)
+      if (curKey.eq(null)) {
+        val newValue = updateFunc(false, null.asInstanceOf[V])
+        data(2 * pos) = k
+        data(2 * pos + 1) = newValue.asInstanceOf[AnyRef]
+        incrementSize()
+        return newValue
+      } else if (curKey.eq(k) || curKey == k) {
```
wannabeast, line 112: I'd place this conditional block first, since one presumably expects to find the key immediately.

wannabeast:
Are these the only two use cases for `AppendOnlyMap`, or should I expect others?

core/src/main/scala/spark/util/AppendOnlyMap.scala
```diff
@@ -0,0 +1,228 @@
...
+
+  private val LOAD_FACTOR = 0.7
+
+  /** Get the value for a given key */
+  def apply(key: K): V = {
+    val k = key.asInstanceOf[AnyRef]
+    if (k.eq(null)) {
+      return nullValue
+    }
+    val mask = capacity - 1
```
wannabeast, line 54: Shouldn't `mask` be an instance variable that you update in conjunction with `capacity`?

core/src/main/scala/spark/util/AppendOnlyMap.scala
```diff
@@ -0,0 +1,228 @@
...
+        val delta = i
+        pos = (pos + delta) & mask
+        i += 1
+      }
+    }
+    return null.asInstanceOf[V]
+  }
+
+  /** Set the value for a key */
+  def update(key: K, value: V) {
```
wannabeast, line 71: Return type `: Unit`.

wannabeast:
It should be noted that the Scala library also has an `OpenHashMap`, but it's probably slower due to an extra level of indirection to the `OpenEntry[Key, Value]` elements of its internal `Array`.

wannabeast:
So far I've found no discernible difference in the performance of k-means with this change.

sparkkmeans2 performance

mateiz:
Ah, interesting, thanks for looking at it. K-means is actually fairly CPU-heavy due to code in the application itself, so it might be better to try another thing (e.g. the group-by test in spark-perf). But at least the good news is that it doesn't hurt performance either.

Anyway, I'm not going to merge this in 0.8.0 until we have more tests.

mateiz:
Closing this since it's now here: https://github.com/apache/incubator-spark/pull/44

PR #826, opened 2013-08-14
Fixed 2 bugs in executor UI (incl. SPARK-877)

kayousterhout:
1) UI crashed if the executor UI was loaded before any tasks started. (SPARK-877)
2) The total tasks was incorrectly reported due to using string (rather
than int) arithmetic.


rxin:
Jenkins, test this please.

rxin:
Thanks, Kay!

PR #827, opened 2013-08-14
Fixed empty security group name

lucarosellini:
Using the region option the scripts seems to not be able to resolve security groups names using the reservation object. Resolving security group names using the instance object seems to work.


pwendell:
Hey @lucarosellini,

Sorry for the delay in getting to this, we just merged in some fairly large outstanding changes to the ec2 scripts.

I haven't been able to re-produce this bug. Could you show me what command you were using to launch the cluster. What I tried is launching a cluster for a specific region:

```
./spark-ec2 launch testbug -i XXX -k XXX -t m1.xlarge -s 3 -r us-west-2
```

pwendell:
Hey @lucarosellini - any chance you could let me know what command you used. If not, I might close this until we have a way to re-produce it.

PR #828, opened 2013-08-14
Scheduler fixes and improvements

mateiz:
This is a work in progress to simplify and improve a few things in the TaskSet scheduler and the way we handle locality. Here's a summary of changes (see also the commit messages):
- Eliminated most hostPort based data structures and just used executor IDs instead, since the hostPort vs host stuff was confusing (and not checkable with static typing, leading to ugly debug code), and hostPorts are not provided by Mesos. The hostPort stuff was there from before when we could easily pass executorIDs through.
- Removed non-local fallback stuff in ClusterScheduler that tried to launch less-local tasks on a node once the local ones were all assigned. This change didn't work because many cluster schedulers send offers for just one node at a time (even the standalone and YARN ones do so as nodes join the cluster one by one). Thus, lots of non-local tasks would be assigned even though a node with locality for them would be able to receive tasks just a short time later.
- Added multi-level delay scheduling in ClusterTaskSetManager so you can have different waits for each locality level. This covers part of the use case for the fallback stuff above (for users who wanted to skip node-local waits and just go for rack-local).
- Updated the way ClusterTaskSetManager handles racks: instead of enqueueing a task to a separate queue for all the hosts in the rack, which would create lots of large queues, have one queue per rack name.
- Added periodic revival of offers in StandaloneSchedulerBackend (this was an option in the YARN patch but seems good to have by default).
- Renamed map output tracker "generations" to "epochs".
- Simplified and cleaned up a bunch of the code.

The only thing missing is new unit tests for this stuff, which I'm working on, but the code appears to run fine so I wanted to put it up for others to try.

### DIFF
--- a/core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
+++ b/core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
@@ -40,12 +40,10 @@ private[spark] class ExecutorRunner(
     val memory: Int,
     val worker: ActorRef,
     val workerId: String,
-    val hostPort: String,
+    val host: String,
     val sparkHome: File,
     val workDir: File)
   extends Logging {
-
-  Utils.checkHostPort(hostPort, "Expected hostport")
 
   val fullId = appId + "/" + execId
   var workerThread: Thread = null
@@ -92,7 +90,7 @@ private[spark] class ExecutorRunner(
   /** Replace variables such as {{EXECUTOR_ID}} and {{CORES}} in a command argument passed to us */
   def substituteVariables(argument: String): String = argument match {
     case "{{EXECUTOR_ID}}" => execId.toString
-    case "{{HOSTNAME}}" => Utils.parseHostPort(hostPort)._1
+    case "{{HOSTNAME}}" => host
     case "{{CORES}}" => cores.toString
     case other => other
   }
--- a/core/src/main/scala/spark/rdd/ZippedPartitionsRDD.scala
+++ b/core/src/main/scala/spark/rdd/ZippedPartitionsRDD.scala
@@ -55,29 +55,15 @@ abstract class ZippedPartitionsBaseRDD[V: ClassManifest](
   }
 
   override def getPreferredLocations(s: Partition): Seq[String] = {
-    // TODO(matei): Fix this for hostPort
-
-    // Note that as number of rdd's increase and/or number of slaves in cluster increase, the computed preferredLocations below
-    // become diminishingly small : so we might need to look at alternate strategies to alleviate this.
-    // If there are no (or very small number of preferred locations), we will end up transferred the blocks to 'any' node in the
-    // cluster - paying with n/w and cache cost.
-    // Maybe pick a node which figures max amount of time ?
-    // Choose node which is hosting 'larger' of some subset of blocks ?
-    // Look at rack locality to ensure chosen host is atleast rack local to both hosting node ?, etc (would be good to defer this if possible)
-    val splits = s.asInstanceOf[ZippedPartitionsPartition].partitions
-    val rddSplitZip = rdds.zip(splits)
-
-    // exact match.
-    val exactMatchPreferredLocations = rddSplitZip.map(x => x._1.preferredLocations(x._2))
-    val exactMatchLocations = exactMatchPreferredLocations.reduce((x, y) => x.intersect(y))
-
-    // Remove exact match and then do host local match.
-    val exactMatchHosts = exactMatchLocations.map(Utils.parseHostPort(_)._1)
-    val matchPreferredHosts = exactMatchPreferredLocations.map(locs => locs.map(Utils.parseHostPort(_)._1))
-      .reduce((x, y) => x.intersect(y))
-    val otherNodeLocalLocations = matchPreferredHosts.filter { s => !exactMatchHosts.contains(s) }
-
-    otherNodeLocalLocations ++ exactMatchLocations
+    val parts = s.asInstanceOf[ZippedPartitionsPartition].partitions
+    val prefs = rdds.zip(parts).map { case (rdd, p) => rdd.preferredLocations(p) }
+    // Check whether there are any hosts that match all RDDs; otherwise return the union
+    val exactMatchLocations = prefs.reduce((x, y) => x.intersect(y))
+    if (!exactMatchLocations.isEmpty) {
+      exactMatchLocations
+    } else {
+      prefs.flatten.distinct
+    }
   }
 
   override def clearDependencies() {
--- a/core/src/main/scala/spark/rdd/ZippedPartitionsRDD.scala
+++ b/core/src/main/scala/spark/rdd/ZippedPartitionsRDD.scala
@@ -55,6 +55,8 @@ abstract class ZippedPartitionsBaseRDD[V: ClassManifest](
   }
 
   override def getPreferredLocations(s: Partition): Seq[String] = {
+    // TODO(matei): Fix this for hostPort
+
     // Note that as number of rdd's increase and/or number of slaves in cluster increase, the computed preferredLocations below
     // become diminishingly small : so we might need to look at alternate strategies to alleviate this.
     // If there are no (or very small number of preferred locations), we will end up transferred the blocks to 'any' node in the
--- a/core/src/main/scala/spark/scheduler/DAGSchedulerEvent.scala
+++ b/core/src/main/scala/spark/scheduler/DAGSchedulerEvent.scala
@@ -54,9 +54,7 @@ private[spark] case class CompletionEvent(
     taskMetrics: TaskMetrics)
   extends DAGSchedulerEvent
 
-private[spark] case class ExecutorGained(execId: String, hostPort: String) extends DAGSchedulerEvent {
-  Utils.checkHostPort(hostPort, "Required hostport")
-}
+private[spark] case class ExecutorGained(execId: String, host: String) extends DAGSchedulerEvent
 
 private[spark] case class ExecutorLost(execId: String) extends DAGSchedulerEvent
 
--- a/core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
+++ b/core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
@@ -37,18 +37,22 @@ import java.util.{TimerTask, Timer}
  */
 private[spark] class ClusterScheduler(val sc: SparkContext)
   extends TaskScheduler
-  with Logging {
-
+  with Logging
+{
   // How often to check for speculative tasks
   val SPECULATION_INTERVAL = System.getProperty("spark.speculation.interval", "100").toLong
+
   // Threshold above which we warn user initial TaskSet may be starved
   val STARVATION_TIMEOUT = System.getProperty("spark.starvation.timeout", "15000").toLong
+
   // How often to revive offers in case there are pending tasks - that is how often to try to get
   // tasks scheduled in case there are nodes available : default 0 is to disable it - to preserve existing behavior
-  // Note that this is required due to delayed scheduling due to data locality waits, etc.
-  // TODO: rename property ?
-  val TASK_REVIVAL_INTERVAL = System.getProperty("spark.tasks.revive.interval", "0").toLong
+  // Note that this is required due to delay scheduling due to data locality waits, etc.
+  // TODO(matei): move to StandaloneSchedulerBackend?
+  val TASK_REVIVAL_INTERVAL = System.getProperty("spark.scheduler.revival.interval", "1000").toLong
 
+  // TODO(matei): replace this with something that only affects levels past PROCESS_LOCAL;
+  // basically it can be a "cliff" for locality
   /*
    This property controls how aggressive we should be to modulate waiting for node local task scheduling.
    To elaborate, currently there is a time limit (3 sec def) to ensure that spark attempts to wait for node locality of tasks before
@@ -71,7 +75,8 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
    If cluster is rack aware, then setting it to RACK_LOCAL gives best tradeoff and a 3x - 4x performance improvement while minimizing IO impact.
    Also, it brings down the variance in running time drastically.
     */
-  val TASK_SCHEDULING_AGGRESSION = TaskLocality.parse(System.getProperty("spark.tasks.schedule.aggression", "NODE_LOCAL"))
+  val TASK_SCHEDULING_AGGRESSION = TaskLocality.withName(
+    System.getProperty("spark.tasks.schedule.aggression", "NODE_LOCAL"))
 
   val activeTaskSets = new HashMap[String, TaskSetManager]
 
@@ -89,16 +94,11 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
   // Which executor IDs we have executors on
   val activeExecutorIds = new HashSet[String]
 
-  // TODO: We might want to remove this and merge it with execId datastructures - but later.
-  // Which hosts in the cluster are alive (contains hostPort's) - used for process local and node local task locality.
-  private val hostPortsAlive = new HashSet[String]
-  private val hostToAliveHostPorts = new HashMap[String, HashSet[String]]
-
   // The set of executors we have on each host; this is used to compute hostsAlive, which
   // in turn is used to decide when we can attain data locality on a given host
-  private val executorsByHostPort = new HashMap[String, HashSet[String]]
+  private val executorsByHost = new HashMap[String, HashSet[String]]
 
-  private val executorIdToHostPort = new HashMap[String, String]
+  private val executorIdToHost = new HashMap[String, String]
 
   // JAR server, if any JARs were added by the user to the SparkContext
   var jarServer: HttpServer = null
@@ -138,7 +138,7 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
     schedulableBuilder.buildPools()
     // resolve executorId to hostPort mapping.
     def executorToHostPort(executorId: String, defaultHostPort: String): String = {
-      executorIdToHostPort.getOrElse(executorId, defaultHostPort)
+      executorIdToHost.getOrElse(executorId, defaultHostPort)
     }
 
     // Unfortunately, this means that SparkEnv is indirectly referencing ClusterScheduler
@@ -146,13 +146,12 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
     SparkEnv.get.executorIdToHostPort = Some(executorToHostPort)
   }
 
-
   def newTaskId(): Long = nextTaskId.getAndIncrement()
 
   override def start() {
     backend.start()
 
-    if (JBoolean.getBoolean("spark.speculation")) {
+    if (System.getProperty("spark.speculation", "false").toBoolean) {
       new Thread("ClusterScheduler speculation check") {
         setDaemon(true)
 
@@ -172,6 +171,7 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
 
 
     // Change to always run with some default if TASK_REVIVAL_INTERVAL <= 0 ?
+    // TODO(matei): remove this thread
     if (TASK_REVIVAL_INTERVAL > 0) {
       new Thread("ClusterScheduler task offer revival check") {
         setDaemon(true)
@@ -201,7 +201,7 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
       schedulableBuilder.addTaskSetManager(manager, manager.taskSet.properties)
       taskSetTaskIds(taskSet.id) = new HashSet[Long]()
 
-      if (hasReceivedTask == false) {
+      if (!hasReceivedTask) {
         starvationTimer.scheduleAtFixedRate(new TimerTask() {
           override def run() {
             if (!hasLaunchedTask) {
@@ -214,7 +214,7 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
           }
         }, STARVATION_TIMEOUT, STARVATION_TIMEOUT)
       }
-      hasReceivedTask = true;
+      hasReceivedTask = true
     }
     backend.reviveOffers()
   }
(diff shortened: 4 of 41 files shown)


core/src/main/scala/spark/scheduler/cluster/ClusterTaskSetManager.scala
```diff
@@ -801,4 +655,39 @@ private[spark] class ClusterTaskSetManager(sched: ClusterScheduler, val taskSet:
   override def hasPendingTasks(): Boolean = {
     numTasks > 0 && tasksFinished < numTasks
   }
+
+  private def getLocalityWait(level: TaskLocality.TaskLocality): Long = {
+    val defaultWait = System.getProperty("spark.locality.wait", "3000")
```
pwendell, line 679: Matei - what are the semantics of these different waits? Say that you want a task to bump from PROCESS_LOCAL to NODE_LOCAL after 10 seconds. Would you set the wait to 10 seconds for PROCESS_LOCAL or NODE_LOCAL? It would be good to make that clear in the docs.

It might make sense to have the default wait be different for different levels so that the "turn key" performance is pretty good. For instance, it's probably fine to make the PROCESS_LOCAL -> NODE_LOCAL switch after only a few seconds. But going from PROCESS_LOCAL to RACK_LOCAL should be a lot longer.

core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
```diff
@@ -40,13 +40,11 @@ private[spark] class ExecutorRunner(
     val memory: Int,
     val worker: ActorRef,
     val workerId: String,
-    val hostPort: String,
+    val host: String,
     val sparkHome: File,
```
mridulm, line 44: I am not very sure about this - iirc this needs to be hostport - since we can have multiple executors in the same node.

core/src/main/scala/spark/rdd/ZippedPartitionsRDD.scala
```diff
@@ -55,27 +55,15 @@ abstract class ZippedPartitionsBaseRDD[V: ClassManifest](
...
+    val parts = s.asInstanceOf[ZippedPartitionsPartition].partitions
+    val prefs = rdds.zip(parts).map { case (rdd, p) => rdd.preferredLocations(p) }
+    // Check whether there are any hosts that match all RDDs; otherwise return the union
+    val exactMatchLocations = prefs.reduce((x, y) => x.intersect(y))
+    if (!exactMatchLocations.isEmpty) {
+      exactMatchLocations
+    } else {
+      prefs.flatten.distinct
+    }
   }
```
mridulm, line 67: If there are both "host:port" and "host" preferred locations, this will result in loosing the "host" preferred locations - right ?
Will that happen - I am not sure, but we should be defending against future changes imo.

Same applies to other similar changes ...

core/src/main/scala/spark/scheduler/DAGSchedulerEvent.scala
```diff
@@ -54,9 +54,7 @@ private[spark] case class CompletionEvent(
     taskMetrics: TaskMetrics)
   extends DAGSchedulerEvent
 
-private[spark] case class ExecutorGained(execId: String, hostPort: String) extends DAGSchedulerEvent {
-  Utils.checkHostPort(hostPort, "Required hostport")
-}
+private[spark] case class ExecutorGained(execId: String, host: String) extends DAGSchedulerEvent
```
mridulm, line 58: How will this handle the case of multiple containers in a single node ? (same host, different ports)

core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
```diff
@@ -169,27 +127,6 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
...
-            } catch {
-              case e: InterruptedException => {}
-            }
-
-            if (hasPendingTasks()) backend.reviveOffers()
-          }
-        }
-      }.start()
-    }
   }
```
mridulm, line 140: The rationale behind this thread was to revive offers periodically : and not necessarily have to wait for a task event (finished/failed/etc) to reissue revive requests for all 'free' workers.

core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
```diff
@@ -235,172 +172,53 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
...
+  def resourceOffers(offers: Seq[WorkerOffer]): Seq[Seq[TaskDescription]] = synchronized {
+    SparkEnv.set(sc.env)
+
+    // Mark each slave as alive and remember its hostname
+    for (o <- offers) {
+      executorIdToHost(o.executorId) = o.host
+      if (!executorsByHost.contains(o.host)) {
+        executorsByHost(o.host) = new HashSet[String]()
+        executorGained(o.executorId, o.host)
       }
```
mridulm, line 194: resourceOffers can be invoked from different threads (akka and others too ?), so we might want to have this around sync block - not just for locking, but also to ensure changes are reflected across threads.

core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
```diff
@@ -536,80 +353,56 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
...
+   * For example, given <h1, [o1, o2, o3]>, <h2, [o4]>, <h1, [o5, o6]>, returns
+   * [o1, o5, o4, 02, o6, o3]
+   */
   def prioritizeContainers[K, T] (map: HashMap[K, ArrayBuffer[T]]): List[T] = {
     val _keyList = new ArrayBuffer[K](map.size)
     _keyList ++= map.keys
 
     // order keyList based on population of value in map
     val keyList = _keyList.sortWith(
+      // TODO(matei): not sure why we're using getOrElse if keyList = map.keys... see if it matters
```
mridulm: getOrElse is not really required - was probably an oversight on my part.
I think it was one of those templatized copy/paste from java to scala code conversions.

core/src/main/scala/spark/scheduler/cluster/ClusterTaskSetManager.scala
```diff
@@ -171,166 +127,86 @@ private[spark] class ClusterTaskSetManager(sched: ClusterScheduler, val taskSet:
...
-          val rackOpt = scheduler.getRackForHost(h)
-          if (rackOpt.isDefined) {
-            val hostsOpt = scheduler.getCachedHostsForRack(rackOpt.get)
-            if (hostsOpt.isDefined) {
-              hosts ++= hostsOpt.get
-            }
-          }
+  // Figure out which locality levels we have in our TaskSet, so we can do delay scheduling
+  val myLocalityLevels = computeValidLocalityLevels()
+  val localityWaits = myLocalityLevels.map(getLocalityWait) // Time to wait at each level
```
mridulm, line 144: Unless I am misreading the diffs - computeValidLocalityLevels, etc is a val in the class right ?
Wont that not need to be recomputed each time tasks are added/finished ?

core/src/main/scala/spark/scheduler/cluster/ClusterTaskSetManager.scala
```diff
@@ -171,166 +127,86 @@ private[spark] class ClusterTaskSetManager(sched: ClusterScheduler, val taskSet:
...
-        val host = Utils.parseHostPort(hostPort)._1
-        val hostList = pendingTasksForHost.getOrElseUpdate(host, ArrayBuffer())
-        hostList += index
+      if (sched.hasExecutorsAliveOnHost(loc.host)) {
+        addTo(pendingTasksForHost.getOrElseUpdate(loc.host, new ArrayBuffer))
+        for (rack <- sched.getRackForHost(loc.host)) {
+          addTo(pendingTasksForRack.getOrElseUpdate(rack, new ArrayBuffer))
+        }
+        hadAliveLocations = true
       }
```
mridulm, line 178: Will this handle the case of new workers coming up on the same host or rack (for this block and the block above) ?
I was not able to come to a conclusion either way ...

core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
```diff
@@ -235,172 +172,53 @@ private[spark] class ClusterScheduler(val sc: SparkContext)
...
-          Utils.checkHostPort(hostPort)
-
-          val numProcessLocalTasks =  math.max(0, math.min(manager.numPendingTasksForHostPort(hostPort), availableCpus(i)))
-          if (numProcessLocalTasks > 0){
-            val list = processLocalOffers.getOrElseUpdate(hostPort, new ArrayBuffer[Int])
-            for (j <- 0 until numProcessLocalTasks) list += i
+          val execId = offers(i).executorId
+          val host = offers(i).host
+          for (task <- manager.resourceOffer(execId, host, availableCpus(i))) {
+            tasks(i) += task
```
mridulm, line 216: This actually leads to potentially suboptimal scheduling.
Consider the case of 4 workers ready for scheduling, with the level set to rack local (anything worse than instance local).
The best scheduling will be in the order instance local, node local, rack local, others.

But that need not necessarily be gauranteed here - for example, if task1 has offers(0) rack local, offers(1) node local and offers(2) instance local.
Ideally, we will want task1 to be scheduled on offers(2), not even on offers(1) ... which is what the earlier messy blocks were attempting to do - to minimize intra node (added as an after thought thanks to @rxin's suggestion !), followed by inter node, and then intra-rack data movement.

mridulm:
Hi Matei,

  I have added a few comments, unfortunately I am unable to go through this at the level of detail I would like to - apologies for that.

In general, a few things though :

a) There were a few sync blocks removed - we might need to re-introduce those back.
(Not just for MT-safety, which might not be a big issue, but also to ensure visibility of updates since same thread need not necessarily be used by akka/etc).

b) I noticed change from hostPort to host, executorId - would it not be better to remove host also ?
The earlier code was more of a hack - since I was trying to reconcile pre-executorId codebase with significant drift in yarn branch which had host -> hostPort change.

c) I added a few comments about scheduling in general in the code, and why we hacked in reviveOffers thread : coupled with my earlier mail on this subject, hopefully I have not messed up explaining my rationale for those changes !

Thanks,
Mridul

core/src/main/scala/spark/rdd/ZippedPartitionsRDD.scala
mateiz, line 67: No, I've decided not to allow host:port in preferred locations. Instead, Tasks that have a preferred executor can pass an executorID as part of their TaskLocation object.

core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
mateiz, line 44: It's only used to set the host below in this file (no other references), so it should be fine.

core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
mateiz, line 140: I've moved this into StandaloneSchedulerBackend. Specifically, that has a new line to schedule a ReviveOffers message to itself every 1000 ms (or whatever the timeout is)

core/src/main/scala/spark/scheduler/DAGSchedulerEvent.scala
mateiz, line 58: Using the executorID. That's how it gets used below.

core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
mateiz, line 194: Yeah, sorry it's not super clear, but I just changed the code style to add it at the top:

```
def resourceOffers(offers: Seq[WorkerOffer]): Seq[Seq[TaskDescription]] = synchronized { ...
```

This is a more common style for making a whole method synchronized in Scala

core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
mateiz, line 216: This is a good point, though it will only happen if some TaskSets have already waited enough to go to a non-local level and a local node happens to become available in the same set of resource offers. It can't really happen when the cluster is busy and only one node is becoming free at a time  (the most common case), or when the cluster was idle and a new job is starting (because then it won't have waited). That said though, I'll see if I can change this to loop through the locality levels as before. My main concern was that I didn't want to launch _non-local_ tasks as soon as we exhaust the local spots in this resource offer if those TaskSets haven't waited enough.

mridulm, line 216: When we have individual offers being made, you are right - this will not be an issue : since we have only a single executor being offered anyway.
But at start of a job, when we have a bunch of offers, this will result in suboptimal scheduling (primarily between instance local and node local) - but can also happen when multiple nodes are passed on to resource offers - when we have node failures and re-allocations, etc.

core/src/main/scala/spark/scheduler/cluster/ClusterTaskSetManager.scala
mateiz, line 144: Yeah, for simplicity I decided not to recompute it. Technically we could optimize things a bit by jumping straight to the next locality level when we exhaust stuff at one level (for example, if we had some process-local tasks and some that could only be node-local, we should jump to the latter as soon as we're done scheduling the process-local ones). I don't think the number of levels ever needs to increase, since even if nodes fail, tasks will only be added to a subset of the previous levels or the noPrefs list (e.g. a task that had both process- and node-local locations might now only have node). I'll look into this.

core/src/main/scala/spark/scheduler/cluster/ClusterTaskSetManager.scala
mateiz, line 178: Because we're creating queues based on the host name and rack name now, this should work, as long as sched.getRackForHost returned the right rack name.

core/src/main/scala/spark/deploy/worker/ExecutorRunner.scala
mridulm, line 44: can we remove references to host also ?
iirc we can infer the host from the executorId right ? This will remove all references to host when we can avoid it ...

core/src/main/scala/spark/rdd/ZippedPartitionsRDD.scala
mridulm, line 67: ah, that is an excellent change ! had not noticed that

mateiz:
Thanks for the feedback, Mridul, this is super helpful! I'll address some of the issues you brought up.

core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
mridulm, line 140: cool ! i missed out on this, my apologies.

core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
mridulm, line 194: There are changes in ClusterTaskSetManager also which removed synch ... I think I added the comment at the wrong location.
Ideally ClusterTaskSetManager should only get invoked from here, but they were added to be defensive in case of future changes for consistency of update to shared state.

... and I had not noticed the synchronized at the top of method definition - thanks for pointing that out btw !

core/src/main/scala/spark/scheduler/cluster/ClusterTaskSetManager.scala
mateiz, line 679: Good idea; I'll add a doc page on this. Basically there is a separate jump between each pair of levels: for example process -> (wait 3 sec) -> node -> (wait 3 more sec) -> rack. You can configure each wait separately.

My instinct was that it's not obvious which jumps will cost more, so I'd rather have a simple default setting. For example, going from process to node could be really bad for data that's expensive to serialize, while going from node to rack might be very cheap with a fast network.

pwendell, line 679: I see - so I guess my main question then was whether this defines the level we are jumping _to_ or jumping _from_. I guess since you can't set this for `ANY` it's the level you are jumping _from_.

core/src/main/scala/spark/scheduler/cluster/ClusterTaskSetManager.scala
```diff
@@ -171,166 +127,86 @@ private[spark] class ClusterTaskSetManager(sched: ClusterScheduler, val taskSet:
...
+  // We then move down if we manage to launch a "more local" task.
+  var currentLocalityIndex = 0    // Index of our current locality level in validLocalityLevels
+  var lastLaunchTime = System.currentTimeMillis()  // Time we last launched a task at this level
 
-        hosts
+  /**
+   * Add a task to all the pending-task lists that it should be on. If readding is set, we are
+   * re-adding the task so only include it in each list if it's not already there.
+   */
+  private def addPendingTask(index: Int, readding: Boolean = false) {
```
ryanlecompte, line 156: minor typo: readding

ryanlecompte, line 156: oh, nevermind! re-adding, not readding :)

core/src/main/scala/spark/scheduler/cluster/ClusterTaskSetManager.scala
mridulm, line 178: Looks good, thanks for clarifying !

core/src/main/scala/spark/scheduler/cluster/ClusterScheduler.scala
mridulm, line 216: To add to above (late night comments are not always very optimal :) )
There were two intentions behind the earlier piece of code :

a) Given a seq of WorkerOffer, ensure that the order in which they are passed on to TaskSetManager is such that relatively optimal scheduling of workers occurs.
Namely, prioritize instance local, node local, rack local and any in that order.

b) Once all optimal schedules have happened, expand the scope so that non optimal schedules are also done such that cluster utilization is mazimized (within bounds controlled via aggression and time spent waiting).
These are also done in order of our prioritity - so even if aggression was set to ANY, we will try to do rack local schedules first, before ANY.

This tries to ensure that we balance optimal scheduling with maximizing utilization of cluster resources.
Invocation of resourceOffers with single workoffer happens when a task finishes/fails; which is unfortunately not very optimal - I never got around to resolving this and relying only on reviveOffers() (bulk submission of all available workers)

Regards,
Mridul

mateiz, line 216: Yes, I'll implement point a), but for point b), as I said above, it's not a good idea. Single worker offers are the common case while a job is running (e.g. as it finishes the first series of tasks and launches another), and it fundamentally wouldn't make sense to delay them to wait for multiple workers to finish because then you'd be wasting cluster resources. It's not just an issue when a job begins -- this aggression logic degrades locality throughout the job runtime. For example, if you have a job with cached data (process-local) that is running multiple waves of tasks, some tasks will be scheduled non-locally in the current code, leading to a very expensive transfer of a block across nodes when you could've just waited a few seconds to run on the right node. With my new code, it's still very easy to minimize the initial wait if that's a concern, and you get the additional bonus of making the process-local wait time different from the node/rack one.

mridulm, line 216: Hmm, on the contrary, the aggression logic is supposed to make it user
controllable on whether the tradeoff of waiting for optimal scheduling vs
maximising cluster utilization while maximising locality on the set of one
or more offers.
Particularly with locality wait and larger number of workers, the
reviveOffers thread will be invoking the method with nontrivial number of
offers - optimizing the single offer case was left for later (see below).

In our tests (ml apps with disk or mem ser on non trivial input datasets)
introduction of this one change speedup of jobs by 2.5x - 4x (iirc) with no
other change in system : with setting it to the default having no perf
change w.r.t earlier perf of jobs in our measurements.
As I mentioned before, idling large number of workers is extremely
suboptimal for our usecases.
I am sure for a large number of other applications of spark this is not a
concern : if entire working set fits in memory and is hosted in non-ser
mode, etc - which is why it would be good to allow its customization via
flags - while keeping it at defaults to handle common case.

Re single worker case : we can simply disable aggression code for this
(earlier, the last iteration expanded the locality constraint) when in this
last iteration and num offers = 1 : this was not done since reviveOffers
thread was optional unlike now.
This should disallow non host or instance local scheduling until timeout
for single offer case.

Regards
Mridul

mateiz, line 216: As I said, you can get the same behavior by controlling the different timeouts. As far as I understood it, the biggest change for you was actually from skipping node locality to go to rack; this can be done easily by setting the node locality wait to 0. If you also want to skip process locality, you can try that too, though I don't think this is usually a good idea. The new code is also simpler than the old aggression code, more efficient (doesn't allocate a whole bunch of hashmaps on every resource offer), and avoids some real performance regressions I saw with that change. If you guys try it out and see some performance problems that can't be solved with these settings, I'm happy to revisit it.

BTW, I've also pushed an update for this that now goes through the locality levels in order for the whole set of available offers.

mridulm, line 216: Let me clarify the performance issue we face, in the current code, if we
set node locality wait to 0 - the scaling is extremely bad due to heavy
data movement at job start (it is worse than non zero node locality value
actually - so it is pointless to set that atleast in our expts).
Particularly when dataset is large. Effectively, almost all schedules end
up being rack local.

Setting it to anything non zero (however small - in our expts even 1ms -
since effectively it is a much higher wait), the cluster utilization drops
drastically.
Before introducing this change, we had experimented with reducing locality
wait to 0ms and 1ms in earlier codebase - and since that did not help
matters, we modified the scheduling ... initially aggression was
effectively any - which improved situation much better, and adding rack
support improved it even further while minimizing io overhead but at
slightly decreased cluster utilization (so it was a tradeoff between both).

Scheduling is something we need to heavily experiment with - considering
the non trivial gains we saw with relatively small changes.

mateiz, line 216: Mridul, as I said, please try it with my change and spark.locality.wait.node = 0. If you want the exact same behavior as your old code, you can also set spark.locality.wait.process=0. The behavior will be _exactly_ the same:
- When multiple offers come in, Spark will look at _all_ of them together and try to assign only process-local tasks first, then node-local, then rack-local, etc. Note that I added the logic to go through these levels in order across the set of offers.
- If you set both node and process waits to 0, it will immediately jump up to rack to fill in all the remaining spots.

I really don't see how it's different, except that I'm doing this with significantly simpler code.

I've been working on similar stuff in Hadoop for a long time (wrote the Hadoop Fair Scheduler in 2008, which uses the same algorithm for locality and has been running fine on 1000+ node clusters since then at Facebook, many Cloudera users, etc), so it's not like this is an untested change. There was definitely a problem before for large clusters where you don't have executors on most of the data-containing nodes, but the combination of rack-awareness and the ability to configure these delays and to optimize across a set of offers fixes that. I'm not trying to remove code because I don't understand it -- I just think that it's important for this logic (and all of Spark's scheduling code) to be dead-simple in order to be easy to optimize and understandable to new developers.

> Let me clarify the performance issue we face, in the current code, if we set node locality wait to 0 - the scaling is extremely bad due to heavy data movement at job start (it is worse than non zero node locality value actually - so it is pointless to set that atleast in our expts). Particularly when dataset is large. Effectively, almost all schedules end up being rack local. Setting it to anything non zero (however small - in our expts even 1ms - since effectively it is a much higher wait), the cluster utilization drops drastically. Before introducing this change, we had experimented with reducing locality wait to 0ms and 1ms in earlier codebase - and since that did not help matters, we modified the scheduling ... initially aggression was effectively any - which improved situation much better, and adding rack support improved it even further while minimizing io overhead but at slightly decreased cluster utilization (so it was a tradeoff between both). Scheduling is something we need to heavily experiment with - considering the non trivial gains we saw with relatively small changes.

mridulm, line 216: Ah, I see my disconnect now.
The local PR I had was not updated with commit 1d6da02 - hence the bad perf
numbers I observed (it takes a couple of days to finish) : and commented
based on analysis of the local code - my apologies !
Preliminary results seem to indicate reasonably close enough numbers -
thanks for clarifying.

mateiz:
Jenkins, please retest this

mateiz:
Alright, I think this is pretty much done; I also added some unit tests and found a bug with them (ClusterTaskSetManager was setting currentLocalityIndex to allowedLocality instead of taskLocality).

One final improvement might be to add support for rack locality in non-YARN modes, but we can do that as a separate PR.

mridulm:
I ran some tests with this PR (as pulled on saturday). The results were mixed - 

The setup:
a) baseline - spark.tasks.revive_periodic.interval=200, spark.tasks.schedule.aggression=RACK
b) pr applied, enabled - spark.locality.wait.process=0 spark.locality.wait.node=0 spark.locality.wait.rack=3000 (to mirror (a) as closely as possible).
c) pr applied, but not enabled - so run with default config.

The relative performance numbers below are for 4 runs of the expts, and both min and avg of the time taken via 'time' command (so end to end time).

For lower number of nodes ( <= 10) , (b) performed better : from 2% to 10% for the experiments. For (c), the results were break even to worse by 4% or so.

But as the number of nodes increase, the relative performance worsens, and somewhere between 20 to 25, (b) becomes worse than (a).
At 100 nodes, (b) is about 18 % worse and (c) is about 41% worse.

I did not try with large numbers though (partly due to diminishing returns for the dataset, and partly because the graphs were diverging anyway, so intutively it should worsen).

This is much better than earlier though (without 2a4ed10).
Also note that the relative difference between (b) and (c) is much lower than earlier (with and without aggression) - which can be taken as a positive change : default out of box performance is better.

I am planning to rerun the expt set once again with the latest PR, just to be sure.

mateiz:
I actually did fix a small bug in it yesterday, where it wasn't properly remembering the locality of each launched task.

Instead of c), can you also try setting process spark.locality.wait.process=3000, spark.locality.wait.node=0, and spark.locality.wait.rack=3000? This is a new configuration possible with this PR that lets you launch tasks on HDFS in a rack-local manner while still getting good locality for cached data. (Is your test actually using any caching by the way?)

mateiz:
Also, if you want to match a), you need to set spark.scheduler.revive.interval=200 in the new code too.

mridulm:
I was trying to replicate exactly similar test env for both baseline and the pr change (hence spark.locality.wait.process=0).

I did not notice that revive thread timeout was configurable - will change the test to use spark.scheduler.revive.interval = 200, thanks !

mridulm:
Looks great now Matei !
With the latest commits the preliminary numbers with PR enabled is within margin error with baseline.

mateiz:
Thanks for testing it, Mridul. It also does fine on Patrick's regression tests (https://github.com/amplab/spark-perf), so I'm going to merge it in.

PR #830, opened 2013-08-14
Adding Linear Regression, and refactoring Ridge Regression.

etrain:
Adding vanilla Linear Regression (no regularization) and refactoring Ridge Regression (Linear Regression + L2 Regularization) to use SGD -  MLlib. Adding test cases (scala+java) and augmenting the data generators for test support.

See https://github.com/mesos/spark/pull/765 for prior discussion of a similar pull request. Now refactored under @shivaram's updated GLM framework.


mateiz:

mllib/src/main/scala/spark/mllib/regression/LinearRegression.scala
```diff
@@ -0,0 +1,168 @@
...
+import org.jblas.DoubleMatrix
+
+/**
+ * Regression model trained using LinearRegression.
+ *
+ * @param weights Weights computed for every feature.
+ * @param intercept Intercept computed for this model.
+ */
+class LinearRegressionModel(
+                  override val weights: Array[Double],
```
shivaram, line 33: Style nit: This should be 4 spaces here. A bunch of other constructors and function definitions have the same spacing issue.

mllib/src/main/scala/spark/mllib/regression/RidgeRegression.scala
```diff
@@ -18,200 +18,159 @@
...
+/**
+ * Train a regression model with L2-regularization using Stochastic Gradient Descent.
+ */
+class RidgeRegressionWithSGD private (
+    var stepSize: Double,
+    var numIterations: Int,
+    var regParam: Double,
+    var miniBatchFraction: Double,
+    var addIntercept: Boolean)
+    extends GeneralizedLinearAlgorithm[RidgeRegressionModel]
```
shivaram, line 53: One thing you could consider doing is to have RidgeRegression extend LinearRegression and just set the updater in the constructor. I guess you'll need to make the LinearRegression constructor `protected` for that.

mllib/src/main/scala/spark/mllib/util/LinearRegressionDataGenerator.scala
```diff
@@ -0,0 +1,98 @@
...
+import spark.{RDD, SparkContext}
+import spark.mllib.regression.LabeledPoint
+
+/**
+ * Generate sample data used for LinearRegression. This class generates
+ * uniformly random values for every feature and adds Gaussian noise with mean `eps` to the
+ * response variable `Y`.
+ *
+ */
+object LinearRegressionDataGenerator {
```
shivaram: Is there any difference between this and the `RidgeRegressionDataGenerator` ? If not I'd suggest removing the ridge one. Also is Lasso covered by this data generator as well ? I think overall it will be better if we can re-use data generators or may be add command line flags if they need to be a little different

mllib/src/test/java/spark/mllib/regression/JavaLinearRegressionSuite.java
```diff
@@ -0,0 +1,96 @@
...
+        int nPoints = 10000;
+        double A = 2.0;
+        double[] weights = {-1.5, 1.0e-2};
+
+        JavaRDD<LabeledPoint> testRDD = sc.parallelize(LinearRegressionSuite.generateLinearRegressionInputAsList(A,
+                weights, nPoints, 42), 2).cache();
+        List<LabeledPoint> validationData =
+                LinearRegressionSuite.generateLinearRegressionInputAsList(A, weights, nPoints, 17);
+
+        LinearRegressionWithSGD svmSGDImpl = new LinearRegressionWithSGD();
```
shivaram: this should be named linearRegImpl

mllib/src/test/java/spark/mllib/regression/JavaRidgeRegressionSuite.java
```diff
@@ -0,0 +1,96 @@
...
+        int nPoints = 10000;
+        double A = 2.0;
+        double[] weights = {-1.5, 1.0e-2};
+
+        JavaRDD<LabeledPoint> testRDD = sc.parallelize(RidgeRegressionSuite.generateRidgeRegressionInputAsList(A,
+                weights, nPoints, 42), 2).cache();
+        List<LabeledPoint> validationData =
+                RidgeRegressionSuite.generateRidgeRegressionInputAsList(A, weights, nPoints, 17);
+
+        RidgeRegressionWithSGD svmSGDImpl = new RidgeRegressionWithSGD();
```
shivaram: svmSGD -> ridgeSGD

mllib/src/test/scala/spark/mllib/regression/RidgeRegressionSuite.scala
```diff
@@ -24,6 +25,37 @@ import org.scalatest.FunSuite
...
+  def generateRidgeRegressionInputAsList(
+    intercept: Double,
+    weights: Array[Double],
+    nPoints: Int,
+    seed: Int): java.util.List[LabeledPoint] = {
+    seqAsJavaList(generateRidgeRegressionInput(intercept, weights, nPoints, seed))
+  }
+
+
+  // Generate noisy input of the form Y = x.dot(weights) + intercept + noise
```
shivaram: On a similar note to the data generators, we can reuse this function from `LinearRegressionSuite` ?

mllib/src/test/scala/spark/mllib/regression/RidgeRegressionSuite.scala
```diff
@@ -38,31 +70,31 @@ class RidgeRegressionSuite extends FunSuite with BeforeAndAfterAll {
...
+  // Test if we can correctly learn Y = 3 + 10*X1 + 10*X2 when
   // X1 and X2 are collinear.
   test("multi-collinear variables") {
-    val rnd = new Random(43)
-    val x1 = Array.fill[Double](20)(rnd.nextGaussian())
-
-    // Pick a mean close to mean of x1
-    val rnd1 = new Random(42) //new NormalDistribution(0.1, 0.01)
-    val x2 = Array.fill[Double](20)(0.1 + rnd1.nextGaussian() * 0.01)
+    val testRDD = RidgeRegressionDataGenerator.generateRidgeRDD(sc, 100, 2, 0.0, intercept=3.0).cache()
```
shivaram: This is even better ! -- I think we can get rid of the generateRDD method and just use the one from the generator in both Java and Scala ?

mllib/src/main/scala/spark/mllib/util/RidgeRegressionDataGenerator.scala
```diff
@@ -48,7 +48,8 @@ object RidgeRegressionDataGenerator {
     nexamples: Int,
     nfeatures: Int,
     eps: Double,
-    nparts: Int = 2) : RDD[LabeledPoint] = {
+    nparts: Int = 2,
+    intercept: Double = 0.0) : RDD[LabeledPoint] = {
     org.jblas.util.Random.seed(42)
     // Random values distributed uniformly in [-0.5, 0.5]
     val w = DoubleMatrix.rand(nfeatures, 1).subi(0.5)
```
shivaram, line 103: This line is misleading -- Can you delete it ? We seem to set the first two features to 10 always if I am reading things right.. Why don't we have two versions of this function ? One where we take in the feature values as Array[Double] for the test and the other for the generator where we pick randomly ?

shivaram:
Hey Evan, Thanks for adding this. This mostly looks good to me -- I have a bunch of inline comments mostly relating to code-reuse for data generators etc. 

Also do you want to add the model broadcast for SGD in the PR ?

mllib/src/main/scala/spark/mllib/regression/RidgeRegression.scala
etrain, line 53: In that case we'd probably want to do the same for Lasso - and we'd need to decide how we want to handle the existence of the regularization parameter.

mllib/src/main/scala/spark/mllib/util/LinearRegressionDataGenerator.scala
etrain: I had the same thought as I was copying/pasting code - let's just call it "LinearModelDataGenerator" and be done with it?

mllib/src/test/java/spark/mllib/regression/JavaRidgeRegressionSuite.java
etrain: ah, good catch on this one and the last, thx.

mllib/src/test/scala/spark/mllib/regression/RidgeRegressionSuite.scala
etrain: I started to do this, and ran into some typing issues calling the DataGenerators from java - will see if I can work around those - I agree it makes sense to have this all in one place.

mllib/src/main/scala/spark/mllib/util/RidgeRegressionDataGenerator.scala
etrain, line 103: Hmm - since we're centralizing anyway, I'm going to take one pass over the thing and be done with it. A data generator will take some list of weights, an optional intercept, an n, a d, and an error term, as well as a "nparts" parameter.

etrain:
See my comments above, and let me know if you have objections - will address!

shivaram:
Agree with your idea -- One central data generator will be very good.

etrain:
Addressed these issues, and centralized data generation facilities for linear models - let's try to keep from writing that code again ;). Let's save weight broadcasting for a separate PR since doing it correctly may be expensive memory-wise and I want to make sure we manage that correctly.

mateiz:
Guys, how ready is this to merge? It would be nice to get a release candidate out for 0.8 soon.

etrain:
Sorry for the delayed response. The Jenkins build failures are caused by the upstream merge (286d7ed) - @shivaram mentioned he was going to cherry pick the other commits into a new PR, but I guess that hasn't happened yet? I will be able to take a closer look at this on Tuesday/Wednesday.

shivaram:
Sorry adding a new unit test for ridge regression took way longer than I expected -- I've rebased things and opened #863

PR #831, opened 2013-08-15
A few small scheduler / job description changes.

rxin:
1. Renamed SparkContext.addLocalProperty to setLocalProperty. And allow this function to unset a property.
2. Renamed SparkContext.setDescription to setCurrentJobDescription.
3. Throw an exception if the fair scheduler allocation file is invalid.

### DIFF
--- a/core/src/main/scala/spark/SparkContext.scala
+++ b/core/src/main/scala/spark/SparkContext.scala
@@ -267,16 +267,20 @@ class SparkContext(
       localProperties.value = new Properties()
   }
 
-  def addLocalProperty(key: String, value: String) {
-    if(localProperties.value == null) {
+  def setLocalProperty(key: String, value: String) {
+    if (localProperties.value == null) {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/spark/SparkContext.scala
```diff
@@ -267,16 +267,20 @@ class SparkContext(
...
+      localProperties.value.remove(key)
+    } else {
+      localProperties.value.setProperty(key, value)
+    }
   }
 
   /** Set a human readable description of the current job. */
-  def setDescription(value: String) {
-    addLocalProperty(SparkContext.SPARK_JOB_DESCRIPTION, value)
+  def setCurrentJobDescription(value: String) {
```
mateiz: Maybe rename this to setJobDescription

rxin:
Changed. Let's wait for Jenkins to verify the tests.

PR #832, opened 2013-08-15
Coalesced RDD with locality

alig:
Coalesced RDD which tries to pack the parents so that there exists one preferred location for the coalesced RDD that most of the coalesced partitions have in common.

### DIFF
--- a/core/src/main/scala/spark/SparkContext.scala
+++ b/core/src/main/scala/spark/SparkContext.scala
@@ -615,6 +615,15 @@ class SparkContext(
   }
 
   /**
+   * Gets the locality information associated with the partition in a particular rdd
+   * @param rdd of interest
+   * @param partition to be looked up for locality
+   * @return list of preferred locations for the partition
+   */
+  def getPreferredLocs(rdd: RDD[_], partition: Int): List[String] =
+    dagScheduler.getPreferredLocs(rdd, partition)
+
+  /**
    * Adds a JAR dependency for all tasks to be executed on this SparkContext in the future.
    * The `path` passed can be either a local file, a file in HDFS (or other Hadoop-supported
    * filesystems), or an HTTP, HTTPS or FTP URI.
--- a/core/src/main/scala/spark/SparkContext.scala
+++ b/core/src/main/scala/spark/SparkContext.scala
@@ -56,8 +56,7 @@ import spark.deploy.LocalSparkCluster
 import spark.partial.{ApproximateEvaluator, PartialResult}
 import spark.rdd.{CheckpointRDD, HadoopRDD, NewHadoopRDD, UnionRDD, ParallelCollectionRDD,
   OrderedRDDFunctions}
-import spark.scheduler.{DAGScheduler, DAGSchedulerSource, ResultTask, ShuffleMapTask, SparkListener,
-  SplitInfo, Stage, StageInfo, TaskScheduler}
+import spark.scheduler._
 import spark.scheduler.cluster.{StandaloneSchedulerBackend, SparkDeploySchedulerBackend,
   ClusterScheduler, Schedulable, SchedulingMode}
 import spark.scheduler.local.LocalScheduler
@@ -65,6 +64,10 @@ import spark.scheduler.mesos.{CoarseMesosSchedulerBackend, MesosSchedulerBackend
 import spark.storage.{StorageStatus, StorageUtils, RDDInfo, BlockManagerSource}
 import spark.ui.SparkUI
 import spark.util.{MetadataCleaner, TimeStampedHashMap}
+import scala.Some
+import spark.scheduler.StageInfo
+import spark.storage.RDDInfo
+import spark.storage.StorageStatus
 
 /**
  * Main entry point for Spark functionality. A SparkContext represents the connection to a Spark
@@ -620,7 +623,7 @@ class SparkContext(
    * @param partition to be looked up for locality
    * @return list of preferred locations for the partition
    */
-  private [spark] def getPreferredLocs(rdd: RDD[_], partition: Int): List[String] = {
+  private [spark] def getPreferredLocs(rdd: RDD[_], partition: Int): Seq[TaskLocation] = {
     dagScheduler.getPreferredLocs(rdd, partition)
   }
 
--- a/core/src/main/scala/spark/SparkContext.scala
+++ b/core/src/main/scala/spark/SparkContext.scala
@@ -620,7 +620,7 @@ class SparkContext(
    * @param partition to be looked up for locality
    * @return list of preferred locations for the partition
    */
-  def getPreferredLocs(rdd: RDD[_], partition: Int): List[String] =
+  private [spark] def getPreferredLocs(rdd: RDD[_], partition: Int): List[String] =
     dagScheduler.getPreferredLocs(rdd, partition)
 
   /**
--- a/core/src/main/scala/spark/SparkContext.scala
+++ b/core/src/main/scala/spark/SparkContext.scala
@@ -620,8 +620,9 @@ class SparkContext(
    * @param partition to be looked up for locality
    * @return list of preferred locations for the partition
    */
-  private [spark] def getPreferredLocs(rdd: RDD[_], partition: Int): List[String] =
+  private [spark] def getPreferredLocs(rdd: RDD[_], partition: Int): List[String] = {
     dagScheduler.getPreferredLocs(rdd, partition)
+  }
 
   /**
    * Adds a JAR dependency for all tasks to be executed on this SparkContext in the future.
--- a/core/src/main/scala/spark/rdd/CoalescedRDD.scala
+++ b/core/src/main/scala/spark/rdd/CoalescedRDD.scala
@@ -60,6 +60,27 @@ private[spark] case class CoalescedRDDPartition(
  *
  * This transformation is useful when an RDD with many partitions gets filtered into a smaller one,
  * or to avoid having a large number of small tasks when processing a directory with many files.
+ *
+ * If there is no locality information (no preferredLocations) in the parent RDD, then the coalescing
+ * is very simple: chunk parents that are close in the Array in chunks.
+ * If there is locality information, it proceeds to pack them with the following three goals in mind:
+ *
+ * (1) Balance the groups so they roughly have the same number of parent partitions
+ * (2) Achieve locality per partition, i.e. there exists one machine which most parent partitions prefer
+ * (3) Be efficient, i.e. O(n) algorithm for n parent partitions (underlying problem is likely NP-hard)
+ * (4) Balance preferred machines, i.e. avoid as much as possible picking the same preferred machine
+ *
+ * Furthermore, it is assumed that the parent RDD may have many partitions, e.g. 100 000.
+ * We assume the final number of desired partitions is small, e.g. less than 1000.
+ *
+ * The algorithm tries to assign unique preferred machines to each partition. If the number of desired
+ * partitions is greater than the number of preferred machines (can happen), it needs to start picking
+ * duplicate preferred machines. This is determined using coupon collector estimation (2n log(n)).
+ * The load balancing is done using power-of-two randomized bins-balls with one twist: it tries to
+ * also achieve locality. This is done by allowing a slack (balanceSlack) between two bins.
+ * If two bins are within the slack in terms of balance, the algorithm will assign partitions
+ * according to locality. (contact alig for questions)
+ *
  */
 class CoalescedRDD[T: ClassManifest](
     @transient var prev: RDD[T],
--- a/core/src/main/scala/spark/rdd/CoalescedRDD.scala
+++ b/core/src/main/scala/spark/rdd/CoalescedRDD.scala
@@ -19,11 +19,13 @@ package spark.rdd
 
 import spark.{Dependency, OneToOneDependency, NarrowDependency, RDD, Partition, TaskContext}
 import java.io.{ObjectOutputStream, IOException}
+import scala.collection.mutable
 
 private[spark] case class CoalescedRDDPartition(
     index: Int,
     @transient rdd: RDD[_],
-    parentsIndices: Array[Int]
+    parentsIndices: Array[Int],
+    prefLoc: String = ""
   ) extends Partition {
   var parents: Seq[Partition] = parentsIndices.map(rdd.partitions(_))
 
@@ -33,6 +35,8 @@ private[spark] case class CoalescedRDDPartition(
     parents = parentsIndices.map(rdd.partitions(_))
     oos.defaultWriteObject()
   }
+
+  def getPreferredLocation = prefLoc
 }
 
 /**
@@ -45,20 +49,24 @@ private[spark] case class CoalescedRDDPartition(
  */
 class CoalescedRDD[T: ClassManifest](
     @transient var prev: RDD[T],
-    maxPartitions: Int)
+    maxPartitions: Int,
+    balanceSlack: Double = 0.20 )
   extends RDD[T](prev.context, Nil) {  // Nil since we implement getDependencies
 
   override def getPartitions: Array[Partition] = {
-    val prevSplits = prev.partitions
-    if (prevSplits.length < maxPartitions) {
-      prevSplits.map(_.index).map{idx => new CoalescedRDDPartition(idx, prev, Array(idx)) }
-    } else {
-      (0 until maxPartitions).map { i =>
-        val rangeStart = ((i.toLong * prevSplits.length) / maxPartitions).toInt
-        val rangeEnd = (((i.toLong + 1) * prevSplits.length) / maxPartitions).toInt
-        new CoalescedRDDPartition(i, prev, (rangeStart until rangeEnd).toArray)
-      }.toArray
+    val res = mutable.ArrayBuffer[CoalescedRDDPartition]()
+    val targetLen = math.min(prev.partitions.length, maxPartitions)
+    val packer = new PartitionCoalescer(maxPartitions, prev, balanceSlack)
+
+    packer.setupGroups(targetLen)   // setup the groups (bins) and preferred locations
+    packer.throwBalls()             // assign partitions (balls) to each group (bins)
+
+    for ((pg, i) <- packer.groupArr.zipWithIndex) {
+      val ids = pg.list.map(_.index).toArray
+      res += new CoalescedRDDPartition(i, prev, ids, pg.prefLoc)
     }
+
+    res.toArray
   }
 
   override def compute(split: Partition, context: TaskContext): Iterator[T] = {
@@ -78,4 +86,181 @@ class CoalescedRDD[T: ClassManifest](
     super.clearDependencies()
     prev = null
   }
+}
+
+
+class PartitionCoalescer(maxPartitions: Int, prev: RDD[_], balanceSlack: Double) {
+
+  // this class just keeps iterating and rotating infinitely over the partitions of the RDD
+  // next() returns the next preferred machine that a partition is replicated on
+  // the rotator first goes through the first replica copy of each partition, then second, then third
+  private class RotateLocations(prev: RDD[_]) extends Iterator[String] {
+
+    private var it: Iterator[String] = resetIterator()
+    override val isEmpty = !it.hasNext
+
+    // initializes/resets to start iterating from the beginning
+    private def resetIterator() = {
+      val i1 =  prev.partitions.view.map( (p: Partition) =>
+      { if (prev.preferredLocations(p).length > 0) Some(prev.preferredLocations(p)(0)) else None } )
+      val i2 =  prev.partitions.view.map( (p: Partition) =>
+      { if (prev.preferredLocations(p).length > 1) Some(prev.preferredLocations(p)(1)) else None } )
+      val i3 =  prev.partitions.view.map( (p: Partition) =>
+      { if (prev.preferredLocations(p).length > 2) Some(prev.preferredLocations(p)(2)) else None } )
+      val res = List(i1,i2,i3)
+      res.view.flatMap(x => x).flatten.iterator // fuses the 3 iterators (1st replica, 2nd, 3rd) into one iterator
+    }
+
+    // hasNext() is false iff there are no preferredLocations for any of the partitions of the RDD
+    def hasNext(): Boolean = !isEmpty
+
+    // return the next preferredLocation of some partition of the RDD
+    def next(): String = {
+      if (it.hasNext)
+        it.next()
+      else {
+        it = resetIterator() // ran out of preferred locations, reset and rotate to the beginning
+        it.next()
+      }
+    }
+  }
+
+  case class PartitionGroup(prefLoc: String = "") {
+    var list = mutable.MutableList[Partition]()
+
+    def size = list.size
+
+    // returns number of partitions that got locality in this group
+    def local = {
+      var loc: Int = 0
+      list.foreach(p => if (prev.preferredLocations(p).contains(prefLoc)) loc += 1)
+      loc
+    }
+
+    override def toString(): String = {
+      val localityStr = if (size == 0) "0" else (local*100. / size).toInt.toString
+      "PartitionGroup(\"" + prefLoc + "\") size: " + size + " locality: " + localityStr +"% \n"
+      //    list.map("\t\t" + _.toString).mkString("\n") + "\n"
+    }
+  }
+
+  def compare(o1: PartitionGroup, o2: PartitionGroup): Boolean = o1.size < o2.size
+  def compare(o1: Option[PartitionGroup], o2: Option[PartitionGroup]): Boolean =
+    if (o1 == None) false else if (o2 == None) true else compare(o1.get, o2.get)
+
+  val rnd = new scala.util.Random(7919) // keep this class deterministic
+
+  // each element of groupArr represents one coalesced partition
+  val groupArr = mutable.ArrayBuffer[PartitionGroup]()
+
+  // hash used to check whether some machine is already in groupArr
+  val groupHash = mutable.Map[String, mutable.MutableList[PartitionGroup]]()
+
+  // determines the tradeoff between load-balancing the partitions sizes and their locality
+  // e.g. balanceSlack=0.10 means that it allows up to 10% imbalance in favor of locality
+  val slack = (balanceSlack * maxPartitions).toInt
+
+  private var noLocality = true  // if true if no preferredLocations exists for parent RDD
+
+  /**
+   * Sorts and gets the least element of the list associated with key in groupHash
+   * The returned PartitionGroup is the least loaded of all groups that represent the machine "key"
+   * @param key string representing a partitioned group on preferred machine key
+   * @return Option of PartitionGroup that has least elements for key
+   */
+  private def getLeastGroupHash(key: String): Option[PartitionGroup] = {
+    groupHash.get(key).map(_.sortWith(compare).head)
+  }
+
+  /**
+   * Initializes targetLen partition groups and assigns a preferredLocation
+   * This uses coupon collector to estimate how many preferredLocations it must rotate through until it has seen
+   * most of the preferred locations (2 * n log(n))
+   * @param targetLen
+   */
+  def setupGroups(targetLen: Int) {
+    val rotIt = new RotateLocations(prev)
+
+    // deal with empty rotator case, just create targetLen partition groups with no preferred location
+    if (!rotIt.hasNext()) {
+      (1 to targetLen).foreach(x => groupArr += PartitionGroup())
+      return
+    }
+
+    noLocality = false
+
+    // number of iterations needed to be certain that we've seen most preferred locations
+    val expectedCoupons2 = 2 * (math.log(targetLen)*targetLen + targetLen + 0.5).toInt
+    var numCreated = 0
+    var tries = 0
+
+    // rotate through until either targetLen unique/distinct preferred locations have been created OR
+    // we've rotated expectedCoupons2, in which case we have likely seen all preferred locations, i.e.
+    // likely targetLen >> number of preferred locations (more buckets than there are machines)
+    while (numCreated < targetLen && tries < expectedCoupons2) {
+      tries += 1
+      val nxt = rotIt.next()
+      if (!groupHash.contains(nxt)) {
+        val pgroup = PartitionGroup(nxt)
+        groupArr += pgroup
+        groupHash += (nxt -> (mutable.MutableList(pgroup))) // use list in case we have multiple groups for same machine
+        numCreated += 1
+      }
+    }
+
+    while (numCreated < targetLen) {  // if we don't have enough partition groups, just create duplicates
+    val nxt = rotIt.next()
+      val pgroup = PartitionGroup(nxt)
+      groupArr += pgroup
+      groupHash.get(nxt).get += pgroup
+      numCreated += 1
+    }
+  }
+
+  /**
+   * Takes a parent RDD partition and decides which of the partition groups to put it in
+   * Takes locality into account, but also uses power of 2 choices to load balance
+   * It strikes a balance between the two use the balanceSlack variable
+   * @param p partition (ball to be thrown)
+   * @return partition group (bin to be put in)
+   */
+  def pickBin(p: Partition): PartitionGroup = {
+    val pref = prev.preferredLocations(p).map(getLeastGroupHash(_)).sortWith(compare) // least loaded bin of replicas
+    val prefPart = if (pref == Nil) None else pref.head
+
+    val r1 = rnd.nextInt(groupArr.size)
+    val r2 = rnd.nextInt(groupArr.size)
+    val minPowerOfTwo = if (groupArr(r1).size < groupArr(r2).size) groupArr(r1) else groupArr(r2) // power of 2
+    if (prefPart== None) // if no preferred locations, just use basic power of two
+      return minPowerOfTwo
+
+    val prefPartActual = prefPart.get
+
+    if (minPowerOfTwo.size + slack <= prefPartActual.size)  // more imbalance than the slack allows
+      return minPowerOfTwo  // prefer balance over locality
+    else {
+      return prefPartActual // prefer locality over balance
+    }
+  }
+
+  def throwBalls() {
+    if (noLocality) {  // no preferredLocations in parent RDD, no randomization needed
+      if (maxPartitions > groupArr.size) { // just return prev.partitions
+        for ((p,i) <- prev.partitions.zipWithIndex) {
+          groupArr(i).list += p
+        }
+      } else { // old code, just splits them grouping partitions that are next to each other in the array
+        (0 until maxPartitions).foreach { i =>
+          val rangeStart = ((i.toLong * prev.partitions.length) / maxPartitions).toInt
+          val rangeEnd = (((i.toLong + 1) * prev.partitions.length) / maxPartitions).toInt
+          (rangeStart until rangeEnd).foreach{ j => groupArr(i).list += prev.partitions(j) }
+        }
+      }
+    } else {
+      for (p <- prev.partitions) { // throw every partition (ball) into a partition group (bin)
+        pickBin(p).list += p
+      }
+    }
+  }
+
 }
--- a/core/src/main/scala/spark/rdd/CoalescedRDD.scala
+++ b/core/src/main/scala/spark/rdd/CoalescedRDD.scala
@@ -36,7 +36,21 @@ private[spark] case class CoalescedRDDPartition(
     oos.defaultWriteObject()
   }
 
+  /**
+   * Gets the preferred location for this coalesced RDD partition. Most parent indices should prefer this machine.
+   * @return preferred location
+   */
   def getPreferredLocation = prefLoc
+
+  /**
+   * Computes how many of the parents partitions have getPreferredLocation as one of their preferredLocations
+   * @return locality of this coalesced partition between 0 and 1
+   */
+  def localFraction :Double = {
+    var loc: Int = 0
+    parents.foreach(p => if (rdd.preferredLocations(p).contains(getPreferredLocation)) loc += 1)
+    if (parents.size == 0) 0.0 else (loc.toDouble / parents.size.toDouble)
+  }
 }
 
 /**
@@ -55,13 +69,9 @@ class CoalescedRDD[T: ClassManifest](
 
   override def getPartitions: Array[Partition] = {
     val res = mutable.ArrayBuffer[CoalescedRDDPartition]()
-    val targetLen = math.min(prev.partitions.length, maxPartitions)
     val packer = new PartitionCoalescer(maxPartitions, prev, balanceSlack)
 
-    packer.setupGroups(targetLen)   // setup the groups (bins) and preferred locations
-    packer.throwBalls()             // assign partitions (balls) to each group (bins)
-
-    for ((pg, i) <- packer.groupArr.zipWithIndex) {
+    for ((pg, i) <- packer.getPartitions.zipWithIndex) {
       val ids = pg.list.map(_.index).toArray
       res += new CoalescedRDDPartition(i, prev, ids, pg.prefLoc)
     }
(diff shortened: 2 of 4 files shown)


core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -19,11 +19,13 @@ package spark.rdd
...
 import spark.{Dependency, OneToOneDependency, NarrowDependency, RDD, Partition, TaskContext}
 import java.io.{ObjectOutputStream, IOException}
+import scala.collection.mutable
 
 private[spark] case class CoalescedRDDPartition(
     index: Int,
     @transient rdd: RDD[_],
-    parentsIndices: Array[Int]
+    parentsIndices: Array[Int],
+    prefLoc: String = ""
```
rxin: Make this a transient val, and rename it preferredLoc. Then you can remove the getPreferredLocation method.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -33,6 +35,24 @@ private[spark] case class CoalescedRDDPartition(
     parents = parentsIndices.map(rdd.partitions(_))
```
rxin, line 44: This is not your change. But I am wondering if this parents = ... is needed because in the constructor, we are already setting the parents.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -33,6 +35,24 @@ private[spark] case class CoalescedRDDPartition(
...
+  def getPreferredLocation = prefLoc
+
+  /**
+   * Computes how many of the parents partitions have getPreferredLocation
+   * as one of their preferredLocations
+   * @return locality of this coalesced partition between 0 and 1
+   */
+  def localFraction :Double = {
+    var loc: Int = 0
+    parents.foreach(p => if (rdd.preferredLocations(p).contains(getPreferredLocation)) loc += 1)
```
rxin: Nice handling of error cases. The first two lines we can simplify it a little bit to

```
val loc = parents.count(p => rdd.preferredLocations(p).contains(getPreferredLocation))
```

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -42,23 +62,44 @@ private[spark] case class CoalescedRDDPartition(
  *
  * This transformation is useful when an RDD with many partitions gets filtered into a smaller one,
  * or to avoid having a large number of small tasks when processing a directory with many files.
+ *
+ * If there is no locality information (no preferredLocations) in the parent, then the coalescing
+ * is very simple: chunk parents that are close in the Array in chunks.
+ * If there is locality information, it proceeds to pack them with the following three goals:
+ *
+ * (1) Balance the groups so they roughly have the same number of parent partitions
+ * (2) Achieve locality per partition, i.e. there exists one machine which most parent splits prefer
```
rxin: Let's change all references to "split" to "partition". "split" was a term we used previously. Now we are using the term "partition" only. In source code, you still see the use of "split" because they are old code, but we definitely should avoid using "split" in documentation.

mateiz:

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -42,23 +62,44 @@ private[spark] case class CoalescedRDDPartition(
...
+ * it tries to also achieve locality. This is done by allowing a slack (balanceSlack) between two
+ * bins. If two bins are within the slack in terms of balance, the algorithm will assign partitions
+ * according to locality. (contact alig for questions)
+ *
  */
 class CoalescedRDD[T: ClassManifest](
     @transient var prev: RDD[T],
-    maxPartitions: Int)
+    maxPartitions: Int,
+    balanceSlack: Double = 0.10 )
```
rxin: Remove the space before )

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -42,23 +62,44 @@ private[spark] case class CoalescedRDDPartition(
...
-    } else {
-      (0 until maxPartitions).map { i =>
-        val rangeStart = ((i.toLong * prevSplits.length) / maxPartitions).toInt
-        val rangeEnd = (((i.toLong + 1) * prevSplits.length) / maxPartitions).toInt
-        new CoalescedRDDPartition(i, prev, (rangeStart until rangeEnd).toArray)
-      }.toArray
+    val res = mutable.ArrayBuffer[CoalescedRDDPartition]()
+    val packer = new PartitionCoalescer(maxPartitions, prev, balanceSlack)
+
+    for ((pg, i) <- packer.getPartitions.zipWithIndex) {
```
rxin: Instead of a for loop, if you use a map, you can get rid of the ArrayBuffer

```
packer.getPartitions.zipWithIndex.map { case(pg, i) =>
  ...
}.toArray
```

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +119,209 @@ class CoalescedRDD[T: ClassManifest](
...
+  /**
+   * Returns the preferred machine for the split. If split is of type CoalescedRDDPartition,
+   * then the preferred machine will be one which most parent splits prefer too.
+   * @param split
+   * @return the machine most preferred by split
+   */
+  override def getPreferredLocations(split: Partition): Seq[String] = {
+    if (split.isInstanceOf[CoalescedRDDPartition])
+      List(split.asInstanceOf[CoalescedRDDPartition].getPreferredLocation)
+    else
```
rxin: Why calling super? We should just return Nil if the partition is not of proper type. Also, for if/else statements, always wrap them with curly braces.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +119,209 @@ class CoalescedRDD[T: ClassManifest](
...
+    }
+  }
+
+  private def throwBalls() {
+    if (noLocality) {  // no preferredLocations in parent RDD, no randomization needed
+      if (maxPartitions > groupArr.size) { // just return prev.partitions
+        for ((p,i) <- prev.partitions.zipWithIndex) {
+          groupArr(i).list += p
+        }
+      } else { // old code, just splits them grouping partitions that are next to each other
```
rxin: remove "old old" since most people won't have a clue what old cold means here.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +119,209 @@ class CoalescedRDD[T: ClassManifest](
...
+  // the rotator first goes through the first replica copy of each partition, then second, third
+  // the iterators return type is a tuple: (replicaString, partition)
+  private class RotateLocations(prev: RDD[_]) extends Iterator[(String, Partition)] {
+
+    private var it: Iterator[(String, Partition)] = resetIterator()
+    override val isEmpty = !it.hasNext
+
+    // initializes/resets to start iterating from the beginning
+    private def resetIterator() = {
+      val i1 =  prev.partitions.view.map( (p: Partition) =>
```
rxin: Write map's like this

```
val i1 =  prev.partitions.view.map { p: Partition =>
  ....
}
```

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +119,209 @@ class CoalescedRDD[T: ClassManifest](
...
+  private def getLeastGroupHash(key: String): Option[PartitionGroup] = {
+    groupHash.get(key).map(_.sortWith(compare).head)
+  }
+
+  def addPartToPGroup(part : Partition, pgroup : PartitionGroup) : Boolean = {
+    if (!initialHash.contains(part)) {
+      pgroup.list += part           // already assign this element
+      initialHash += (part -> true) // needed to avoid assigning partitions to multiple buckets
+      true
+    } else false
```
rxin: wrap the else with { }. ditto for other if/else statements too

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +119,209 @@ class CoalescedRDD[T: ClassManifest](
...
+    if (split.isInstanceOf[CoalescedRDDPartition])
+      List(split.asInstanceOf[CoalescedRDDPartition].getPreferredLocation)
+    else
+      super.getPreferredLocations(split)
+  }
+
+}
+
+
+class PartitionCoalescer(maxPartitions: Int, prev: RDD[_], balanceSlack: Double) {
```
rxin: In general, this looks like a single funciton. Maybe we should restructure it to turn this into an Object, and define a coalesce function the CoalescedRDD can call?

rxin:
Hi Ali - this looks pretty good. Thanks for doing it. I left some comments on style and design. I didn't look closely into the algorithm itself though.

alig:
Thanks for taking time and commenting on this. I'll fix them ASAP.

--Ali

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig, line 44: Agree. I'm leaving it for now because I don't want to break anything. But since it's an immutable, setting it once in the constructor should be sufficient.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig: Done

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig: Done

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig: Done

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig: done

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig: Good catch, done!

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig: Thanks, wrapped the if/else's in `{ }`.

We are overriding the super class. The type checker will not catch giving it just an actual Partition, so I'm just deferring that to the base class which can implement it however it wants. Say it wants to throw an exception or something in the future.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig: Done

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig: Thanks, will do (and done). That also allows me to use uniform syntax for partial functions, indeed useful!

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig: Done here and elsewhere, except in one liners, say:

``` scala
val ratio = if (denominator==0) 0 else numerator/denominator
```

Can change those too if you prefer, but it is concise and unlikely that someone by mistakes makes it multi-line without braces.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig: Good idea, created a functional object.

mridulm:
It would be better to abstract out the functionality in this PR and reuse it elsewhere too.
For example, shuffle can leverage rack locality of map output locations to optimize on minimal cross rack local traffic for example.

rxin:
That's a good idea. Let's do it in a separate patch later.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -19,20 +19,32 @@ package spark.rdd
...
     parents = parentsIndices.map(rdd.partitions(_))
     oos.defaultWriteObject()
   }
+
+  /**
+   * Computes how many of the parents partitions have getPreferredLocation
+   * as one of their preferredLocations
+   * @return locality of this coalesced partition between 0 and 1
+   */
+  def localFraction :Double = {
```
mateiz: Formatting is weird here.. should be `localFraction: Double`

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +109,215 @@ class CoalescedRDD[T: ClassManifest](
...
+
+}
+
+case class PartitionGroup(prefLoc: String = "") {
+  var list = mutable.ListBuffer[Partition]()
+
+  def size = list.size
+}
+
+object coalescePartitions {
```
mateiz: This is not a good name format for an object or class -- it should start with a capital letter, and be a noun. Call it PartitionCoalescer maybe. Also, I don't think we need an `object` for it.. users could call it with just `new PartitionCoalescer().run(stuff)`.

mateiz: Also, mark this and the class above as `private[spark]`.

mateiz:
Hey, one comment on this -- it's currently using the parent RDD's `preferredLocations` to determine where to run, which will only work if you run `coalesce` directly on a `HadoopRDD`, without any maps or other operations in-between. This is because preferred locations are not inherited down the chain. To make this work in more general cases, where I think it will still be quite useful, it should use `DAGScheduler.getPreferredLocs`. @rxin, is that okay? What is the use case in Shark (what is the lineage graph you call coalesce on)?

rxin:
The current use case is to operate directly on a hadoop rdd. I agree in general it'd be better to call getPreferredLocs from DAGScheduler (actually we need to create a recursive getPreferredLocs similar to the one in DAGScheduerl, but shouldn't take cache locality into account). 

This one as is is already a big improvement. We can perhaps change to getPreferredLocs in a later PR.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig: Yeah weird, fixed! Thanks.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig: Done. Changed name back to the original PartitionCoalescer and added the functionality in run() and marked classes as private[spark]

alig:
Ok, implemented functionality to augment `SparkContext` and `DAGScheduler` with methods to get the _current_ preferred locations of derived RDDs. I made the extensions to `DAGScheduler` thread safe. Also added tests that now test whether derived RDDs are scheduled for locality too. Please have a look @mateiz and @rxin.

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -774,7 +777,13 @@ class DAGScheduler(
...
   }
 
-  private def getPreferredLocs(rdd: RDD[_], partition: Int): List[String] = {
+  /**
+   * Synchronized method that might be called from other threads.
+   * @param rdd whose partitions are to be looked at
+   * @param partition to lookup locality information for
+   * @return list of machines that are preferred by the partition
+   */
+  private[spark] def getPreferredLocs(rdd: RDD[_], partition: Int): List[String] = synchronized {
```
rxin: I think it is best to just implement a recursive version of getPreferredLocs in the RDD class itself, without taking memory locality into account there. There are two reasons:
1. This method is called pretty often in the scheduler, and can become a pretty serious bottleneck if we synchronized it.
2. I don't think we should use memory locality in the coalesced rdd's preferred locations, because the life span of a coalesced rdd's preferred locations is the same as the rdd itself, but the life span of the cache locations vary.

mateiz: I actually disagree with this -- although cache locations might change, we should make sure we have sensible behavior when people do want to coalesce a cached RDD. For example, a common use case I can imagine is doing a filter or local reduce on a big cached RDD, then coalescing the result of that to be more efficient to deal with. In most cases (e.g. HDFS files) we'll be coalescing non-cached RDDs anyway and only caching after the coalesce.

The synchronized here shouldn't be a problem. The DAGScheduler event loop is single-threaded so it will have no trouble acquiring the lock, not to mention it only acquires the lock once per processEvent() call, not separately on each getPreferredLocs() call.

core/src/main/scala/spark/SparkContext.scala
```diff
@@ -616,6 +616,15 @@ class SparkContext(
   }
 
   /**
+   * Gets the locality information associated with the partition in a particular rdd
+   * @param rdd of interest
+   * @param partition to be looked up for locality
+   * @return list of preferred locations for the partition
+   */
+  def getPreferredLocs(rdd: RDD[_], partition: Int): List[String] =
```
mateiz: This should be `private[spark]`

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -17,22 +17,37 @@
...
   }
+
+  /**
+   * Computes how many of the parents partitions have getPreferredLocation
+   * as one of their preferredLocations
+   * @return locality of this coalesced partition between 0 and 1
+   */
+  def localFraction: Double = {
+    val loc = parents.count(p => rdd.context.getPreferredLocs(rdd, p.index)
+      .contains(preferredLocation))
```
mateiz: Would look better as

```
val loc = parents.count(p =>
  rdd.context.getPreferredLocs(rdd, p.index).contains(preferredLocation))
```

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +112,221 @@ class CoalescedRDD[T: ClassManifest](
...
+   * @param partition
+   * @return the machine most preferred by split
+   */
+  override def getPreferredLocations(partition: Partition): Seq[String] = {
+    if (partition.isInstanceOf[CoalescedRDDPartition]) {
+      List(partition.asInstanceOf[CoalescedRDDPartition].preferredLocation)
+    } else {
+      super.getPreferredLocations(partition)
+    }
+  }
```
mateiz, line 113: Why would the partition passed not be a CoalescedRDDPartition? Have you seen this happen?

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -42,28 +57,47 @@ private[spark] case class CoalescedRDDPartition(
...
+ * Furthermore, it is assumed that the parent RDD may have many partitions, e.g. 100 000.
+ * We assume the final number of desired partitions is small, e.g. less than 1000.
+ *
+ * The algorithm tries to assign unique preferred machines to each partition. If the number of
+ * desired partitions is greater than the number of preferred machines (can happen), it needs to
+ * start picking duplicate preferred machines. This is determined using coupon collector estimation
+ * (2n log(n)). The load balancing is done using power-of-two randomized bins-balls with one twist:
+ * it tries to also achieve locality. This is done by allowing a slack (balanceSlack) between two
+ * bins. If two bins are within the slack in terms of balance, the algorithm will assign partitions
+ * according to locality. (contact alig for questions)
```
mateiz: This is a great doc comment on the algorithm but I'd move it to the `PartitionCoalescer` class to put it closer to the code. In `CoalescedRDD` you can just say it tries to take into account both locality and load-balancing.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +112,221 @@ class CoalescedRDD[T: ClassManifest](
...
+      List(partition.asInstanceOf[CoalescedRDDPartition].preferredLocation)
+    } else {
+      super.getPreferredLocations(partition)
+    }
+  }
+
+}
+
+private[spark] class PartitionCoalescer(maxPartitions: Int, prev: RDD[_],
+                                        balanceSlack: Double) {
```
mateiz: This can fit on a 100-character line. If it doesn't, move the `private[spark]` to a line by itself and leave the `class` stuff below.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +112,221 @@ class CoalescedRDD[T: ClassManifest](
...
+  protected def compare(o1: PartitionGroup, o2: PartitionGroup): Boolean = o1.size < o2.size
```
mateiz: The members of this class probably don't need to be protected, because we won't have people extending it. You can just remove those protected annotations to reduce clutter.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +112,221 @@ class CoalescedRDD[T: ClassManifest](
...
+  protected def compare(o1: Option[PartitionGroup], o2: Option[PartitionGroup]): Boolean =
+    if (o1 == None) false else if (o2 == None) true else compare(o1.get, o2.get)
+
+  protected val rnd = new scala.util.Random(7919) // keep this class deterministic
+
+  // each element of groupArr represents one coalesced partition
+  protected val groupArr = mutable.ArrayBuffer[PartitionGroup]()
```
mateiz: Just import ArrayBuffer and ListBuffer instead of prefixing them with mutable. IntelliJ recommends using the prefix for Map because there's also an immutable.Map, so that's fine, but these guys have no immutable counterparts.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +112,221 @@ class CoalescedRDD[T: ClassManifest](
...
+
+  // gets the *current* preferred locations from the DAGScheduler (as opposed to the static ones)
+  protected def currentPreferredLocations(rdd: RDD[_], part: Partition) : Seq[String] =
+    rdd.context.getPreferredLocs(rdd, part.index)
+
+  // this class just keeps iterating and rotating infinitely over the partitions of the RDD
+  // next() returns the next preferred machine that a partition is replicated on
+  // the rotator first goes through the first replica copy of each partition, then second, third
+  // the iterators return type is a tuple: (replicaString, partition)
+  protected class RotateLocations(prev: RDD[_]) extends Iterator[(String, Partition)] {
```
mateiz: Name this LocationIterator to make it clearer that it's an interator

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +112,221 @@ class CoalescedRDD[T: ClassManifest](
...
+      { if (currentPreferredLocations(prev, p).length > 0)
+        Some((currentPreferredLocations(prev, p)(0),p)) else None } }
+      val i2 =  prev.partitions.view.map{ p: Partition =>
+      { if (currentPreferredLocations(prev, p).length > 1)
+        Some((currentPreferredLocations(prev, p)(1),p)) else None } }
+      val i3 =  prev.partitions.view.map{ p: Partition =>
+      { if (currentPreferredLocations(prev, p).length > 2)
+        Some((currentPreferredLocations(prev, p)(2),p)) else None } }
+      val res = List(i1,i2,i3)
+      res.view.flatMap(x => x).flatten.iterator // fuses the 3 iterators (1st replica, 2nd, 3rd)
```
mateiz: If you make i1, i2 and such iterators, you can just do `i1 ++ i2 ++ i3`

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +112,221 @@ class CoalescedRDD[T: ClassManifest](
...
+  protected class RotateLocations(prev: RDD[_]) extends Iterator[(String, Partition)] {
+
+    protected var it: Iterator[(String, Partition)] = resetIterator()
+    override val isEmpty = !it.hasNext
+
+    // initializes/resets to start iterating from the beginning
+    protected def resetIterator() = {
+      val i1 =  prev.partitions.view.map{ p: Partition =>
+      { if (currentPreferredLocations(prev, p).length > 0)
+        Some((currentPreferredLocations(prev, p)(0),p)) else None } }
```
mateiz: Just make these iterators instead of views, e.g.:

```
val i1 = prev.partitions.iterator.map(p => currentPreferredLocations(prev, p)).filter(_.length > 0).map(s => s(0))
```

Also since `currentPreferredLocations` is only called on `prev` as the RDD, I'd make it just use that by default instead of having a parameter.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +112,221 @@ class CoalescedRDD[T: ClassManifest](
...
+      val res = List(i1,i2,i3)
+      res.view.flatMap(x => x).flatten.iterator // fuses the 3 iterators (1st replica, 2nd, 3rd)
+    }
+
+    // hasNext() is false iff there are no preferredLocations for any of the partitions of the RDD
+    def hasNext(): Boolean = !isEmpty
+
+    // return the next preferredLocation of some partition of the RDD
+    def next(): (String, Partition) = {
+      if (it.hasNext)
```
mateiz, line 199: Put braces around the if body even if it's one line

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +112,221 @@ class CoalescedRDD[T: ClassManifest](
...
+  protected val rnd = new scala.util.Random(7919) // keep this class deterministic
+
+  // each element of groupArr represents one coalesced partition
+  protected val groupArr = mutable.ArrayBuffer[PartitionGroup]()
+
+  // hash used to check whether some machine is already in groupArr
+  protected val groupHash = mutable.Map[String, mutable.ListBuffer[PartitionGroup]]()
+
+  // hash used for the first maxPartitions (to avoid duplicates)
+  protected val initialHash = mutable.Map[Partition, Boolean]()
```
mateiz: If you just use it to test membership, this can be a Set instead of Map

core/src/main/scala/spark/rdd/CoalescedRDD.scala
mateiz: Also, BTW, there's probably no reason to use ListBuffer over ArrayBuffer in this code

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +112,221 @@ class CoalescedRDD[T: ClassManifest](
...
+        val pgroup = PartitionGroup(nxt_replica)
+        groupArr += pgroup
+        addPartToPGroup(nxt_part, pgroup)
+        groupHash += (nxt_replica -> (mutable.ListBuffer(pgroup))) // list in case we have multiple
+        numCreated += 1
+      }
+    }
+
+    while (numCreated < targetLen) {  // if we don't have enough partition groups, create duplicates
+    var (nxt_replica, nxt_part) = rotIt.next()
```
mateiz: Indent seems wrong

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +112,221 @@ class CoalescedRDD[T: ClassManifest](
...
+  }
+
+  protected def throwBalls() {
+    if (noLocality) {  // no preferredLocations in parent RDD, no randomization needed
+      if (maxPartitions > groupArr.size) { // just return prev.partitions
+        for ((p,i) <- prev.partitions.zipWithIndex) {
+          groupArr(i).list += p
+        }
+      } else { // no locality available, then simply split partitions based on positions in array
+        (0 until maxPartitions).foreach { i =>
```
mateiz: Clearer to write this as `for (i <- 0 until maxPartitions)`; same with the foreach over a range below

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +112,221 @@ class CoalescedRDD[T: ClassManifest](
...
+   */
+  def run() : Array[PartitionGroup] = {
+    setupGroups(math.min(prev.partitions.length, maxPartitions))   // setup the groups (bins)
+    throwBalls()             // assign partitions (balls) to each group (bins)
+    getPartitions
+  }
+}
+
+private[spark] case class PartitionGroup(prefLoc: String = "") {
+  var list = mutable.ListBuffer[Partition]()
```
mateiz: Same here, probably no need for ListBuffer instead of ArrayBuffer.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
```diff
@@ -78,4 +112,221 @@ class CoalescedRDD[T: ClassManifest](
...
+        }
+      }
+    } else {
+      for (p <- prev.partitions if (!initialHash.contains(p))) { // throw every partition into group
+        pickBin(p).list += p
+      }
+    }
+  }
+
+  protected def getPartitions : Array[PartitionGroup] = groupArr.filter( pg => pg.size > 0).toArray
```
mateiz: Don't put a space before colons

core/src/test/scala/spark/RDDSuite.scala
```diff
@@ -173,6 +174,61 @@ class RDDSuite extends FunSuite with SharedSparkContext {
...
       null)
   }
+  test("cogrouped RDDs with locality") {
+    // RDD with locality preferences spread (non-randomly) over 6 machines, m0 through m5
+    val data = sc.makeRDD((1 to 9).map( i => (i, (i to (i+2)).map{ j => "m" + (j%6)} )))
+    val coalesced1 = data.coalesce(3)
+    assert(coalesced1.collect().toList.sorted === (1 to 9).toList, "Data got *lost* in coalescing")
+
+    val splits = coalesced1.glom().collect().map(_.toList).toList
+    assert(splits.length == 3, "Supposed to coalesce to 3 but got " + splits.length)
```
mateiz: Use `===` instead of `==`

core/src/test/scala/spark/RDDSuite.scala
```diff
@@ -173,6 +174,61 @@ class RDDSuite extends FunSuite with SharedSparkContext {
...
+
+    assert(splits.foldLeft(true)
+      ( (x,y) => if (!x) false else y.length >= 1) === true, "Some partitions were empty")
```
mateiz: Simpler to write this as `splits.forall(_.length >= 1)`

core/src/test/scala/spark/RDDSuite.scala
```diff
@@ -173,6 +174,61 @@ class RDDSuite extends FunSuite with SharedSparkContext {
...
+
+    // If we try to coalesce into more partitions than the original RDD, it should just
+    // keep the original number of partitions.
+    val coalesced4 = data.coalesce(20)
+    assert(coalesced4.glom().collect().map(_.toList).toList.sortWith(
+      (x, y) => if (x.isEmpty) false else if (y.isEmpty) true else x(0) < y(0)) === (1 to 9).
+      map(x => List(x)).toList, "Tried coalescing 9 partitions to 20 but didn't get 9 back")
```
mateiz: This is a super-long expression.. split it into smaller ones.

core/src/test/scala/spark/RDDSuite.scala
```diff
@@ -173,6 +174,61 @@ class RDDSuite extends FunSuite with SharedSparkContext {
...
+    // large scale experiment
+    import collection.mutable
+    val rnd = scala.util.Random
+    val partitions = 10000
+    val numMachines = 50
+    val machines = mutable.ListBuffer[String]()
+    (1 to numMachines).foreach(machines += "m"+_)
+
+    val blocks = (1 to partitions).map( i => (i, (i to (i+2))
+      .map{ j => machines(rnd.nextInt(machines.size)) } ))
```
mateiz: The line splitting here is not great; it makes it hard to see that the .map is on the range i to i+2, and all this stuff is in a tuple. Feel free to either use temporary variables or use blocks, e.g.

```
(1 to partitions).map { i =>
  val locs = (i to (i+2)).map(_ => machines(rnd.nextInt(machines.size))
  (i, locs)
}
```

mateiz: Also, if this is just using 3 random hosts, there's no point using (i to (i+2)).. you can use (1 to 3) or just `Array.fill(3)(machines(rnd.nextInt(machines.size))`

core/src/test/scala/spark/RDDSuite.scala
```diff
@@ -173,6 +174,61 @@ class RDDSuite extends FunSuite with SharedSparkContext {
...
+
+    val data2 = sc.makeRDD(blocks) // .map( i => i*2 )
+    val coalesced2 = data2.coalesce(numMachines*2)
+
+    // test that you get over 90% locality in each group
+    val minLocality = coalesced2.partitions
+      .map( part => part.asInstanceOf[CoalescedRDDPartition].localFraction )
```
mateiz: Don't leave spaces after open parens or before close parens

core/src/test/scala/spark/RDDSuite.scala
```diff
@@ -173,6 +174,61 @@ class RDDSuite extends FunSuite with SharedSparkContext {
...
+
+    // If we try to coalesce into more partitions than the original RDD, it should just
+    // keep the original number of partitions.
+    val coalesced4 = data.coalesce(20)
+    assert(coalesced4.glom().collect().map(_.toList).toList.sortWith(
+      (x, y) => if (x.isEmpty) false else if (y.isEmpty) true else x(0) < y(0)) === (1 to 9).
+      map(x => List(x)).toList, "Tried coalescing 9 partitions to 20 but didn't get 9 back")
+
+
+    // large scale experiment
```
mateiz, line 203: Instead of putting this at the bottom of a big "cogrouped RDD with locality" test, make it a separate test case. It will make it easier to see what's failing.

core/src/test/scala/spark/RDDSuite.scala
```diff
@@ -173,6 +174,61 @@ class RDDSuite extends FunSuite with SharedSparkContext {
...
+      .foldLeft(1.)( (perc, loc) => math.min(perc,loc) )
+    assert(minLocality >= 0.90, "Expected 90% locality but got " + (minLocality*100.).toInt + "%")
+
+    // test that the groups are load balanced with 100 +/- 20 elements in each
+    val maxImbalance = coalesced2.partitions
+      .map( part => part.asInstanceOf[CoalescedRDDPartition].parents.size )
+      .foldLeft(0)((dev, curr) => math.max(math.abs(100-curr),dev))
+    assert(maxImbalance <= 20, "Expected 100 +/- 20 per partition, but got " + maxImbalance)
+
+    // TDD: Test for later when we have implemented functionality to get locality from DAGScheduler
```
mateiz: This is done now, so no need to keep this comment, right?

mateiz:
Hey Ali, I've looked through this now too and made some detailed comments, almost entirely on Scala style. The actual algorithm and tests look good. Thanks for putting this together!

core/src/main/scala/spark/scheduler/DAGScheduler.scala
mateiz: BTW, one other question on this -- have you tested it with 100K partitions again? If it's slow, then one solution would be to add a "bulk" version of getPreferredLocs that returns the locations for all partitions of an RDD at once as an array of lists, and do this just once in PartitionCoalescer, then look into that array instead of calling this.

core/src/main/scala/spark/SparkContext.scala
alig: Done

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig: Done

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig, line 113: Nope, just defensive programming since the compiler wouldn't be able to catch that. Let me know if I should just assume it is.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig: Done

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig: Done

core/src/main/scala/spark/rdd/CoalescedRDD.scala
mateiz, line 113: Yeah you should, because I think passing something than the RDD's own Partition objects here would be a logic failure somewhere else, and we should catch it.

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig: Done!

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig: Done

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig: Done

core/src/main/scala/spark/rdd/CoalescedRDD.scala
alig, line 113: Ok, so should I just assume it's the right type or should I check for it and throw an exception? Note that the parent anyway just returns Nil, so I could just return Nil too. Just let me know.

mateiz, line 113: Yes, just cast it as if it were the right type.

core/src/main/scala/spark/scheduler/DAGScheduler.scala
alig: Just for the record, 100k partitions coalesced into 2k takes 850ms on my MBA. Doing it in bulk didn't buy much, perhaps 10ms on average (repeating the experiment 20 times to get the std.dev down to negligible numbers). So for now, we're making one call to DAGScheduler for each partition of the RDD.

alig:
Done with all of Matei's comments.

mateiz:
Jenkins, please retest this

mateiz:
Jenkins, retest this please

mateiz:
@alig this looks good to me, the only thing I'd like to do is run the scheduling throughput perf-test to make sure it doesn't get affected. After that I'll merge it.

alig:
Sure @mateiz, I'll do that report it here.

alig:
Jenkins, retest this please

alig:
Ok, so ran the performance tests for the master branch and this pull-request:

Results for the current master branch:

``` bash
scheduling-tput:      5.603,  6.545,  4.659,  4.659, 21.521
scala-agg-by-key:    29.658, 17.575, 21.255, 21.255, 69.928
scala-sort-by-key:    9.269,  0.765,  9.214,  9.214,  11.19
scala-count:         17.582, 17.169,  6.689,  6.689, 55.479
scala-count-w-fltr:  15.128, 17.194,  7.087,  7.087, 55.917
```

Results for this pull-request:

``` bash
scheduling-tput:      4.703,  4.719,  4.522,  4.522, 16.616
scala-agg-by-key:    30.365, 20.058,  21.46,  21.46, 75.604
scala-sort-by-key:    9.184,  0.719,  8.932,  8.932, 10.904
scala-count:         17.852, 20.271,  6.349,  6.349, 54.761
scala-count-w-fltr:  14.886, 16.503, 13.971, 13.971, 56.561
```

The numbers are in order: median, std.dev, min, first, last. 
As a side note, I did actually take time to check carefully that the two different performance tests were using their respective branches of the source code and not just using the same source twice.

mateiz:
OK, that looks like it didn't break anything. Going to merge it then.

PR #833, opened 2013-08-15
Various UI improvements.

rxin:
Made various changes to the UI, including making naming, size more consistent.

Also changed some column names to make them more self-evident and added counts (e.g. how many active tasks are running) to table titles.

### DIFF
--- a/core/src/main/scala/spark/deploy/master/ui/IndexPage.scala
+++ b/core/src/main/scala/spark/deploy/master/ui/IndexPage.scala
@@ -53,7 +53,7 @@ private[spark] class IndexPage(parent: MasterWebUI) {
     val workers = state.workers.sortBy(_.id)
     val workerTable = UIUtils.listingTable(workerHeaders, workerRow, workers)
 
-    val appHeaders = Seq("ID", "Name", "Cores", "Memory per Node", "Launch Time", "User",
+    val appHeaders = Seq("ID", "Name", "Cores", "Memory per Node", "Submitted Time", "User",
       "State", "Duration")
     val activeApps = state.activeApps.sortBy(_.startTime).reverse
     val activeAppsTable = UIUtils.listingTable(appHeaders, appRow, activeApps)
--- a/core/src/main/scala/spark/deploy/master/ui/IndexPage.scala
+++ b/core/src/main/scala/spark/deploy/master/ui/IndexPage.scala
@@ -53,7 +53,7 @@ private[spark] class IndexPage(parent: MasterWebUI) {
     val workers = state.workers.sortBy(_.id)
     val workerTable = UIUtils.listingTable(workerHeaders, workerRow, workers)
 
-    val appHeaders = Seq("ID", "Description", "Cores", "Memory per Node", "Submit Time", "User",
+    val appHeaders = Seq("ID", "Name", "Cores", "Memory per Node", "Launch Time", "User",
       "State", "Duration")
     val activeApps = state.activeApps.sortBy(_.startTime).reverse
     val activeAppsTable = UIUtils.listingTable(appHeaders, appRow, activeApps)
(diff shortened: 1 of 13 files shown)


core/src/main/scala/spark/deploy/master/ui/IndexPage.scala
```diff
@@ -53,7 +53,7 @@ private[spark] class IndexPage(parent: MasterWebUI) {
     val workers = state.workers.sortBy(_.id)
     val workerTable = UIUtils.listingTable(workerHeaders, workerRow, workers)
 
-    val appHeaders = Seq("ID", "Description", "Cores", "Memory per Node", "Submit Time", "User",
+    val appHeaders = Seq("ID", "Name", "Cores", "Memory per Node", "Launch Time", "User",
```
pwendell: "Launch Time" might imply to the user that the application has actually launched executors. It's possible an application is submitted but is waiting for resources, this could be interpreted to seem like it already launched and be confused why it isn't making progress. Isn't "Submit time" strictly more accurate?

For "Description" - that is to be consistent with the internal code, but I do like "Name" better so let's definitely change that.

core/src/main/scala/spark/ui/jobs/StagePage.scala
```diff
@@ -122,16 +123,19 @@ private[spark] class StagePage(parent: JobProgressUI) {
             if (hasShuffleRead) shuffleReadQuantiles else Nil,
             if (hasShuffleWrite) shuffleWriteQuantiles else Nil)
 
-          val quantileHeaders = Seq("Metric", "Min", "25%", "50%", "75%", "Max")
+          val quantileHeaders = Seq("Metric", "Min (0th percentitle)", "25th percentile",
```
pwendell: It might be okay to just say "Min" and "Max". Or if you do want to use parens, you could try this:

0th percentile (Min)   |   25th Percentile    |    75th Percentile    |    100th percentile (Max)

That way it's more consistent.

core/src/main/scala/spark/ui/jobs/StagePage.scala
```diff
@@ -122,16 +123,19 @@ private[spark] class StagePage(parent: JobProgressUI) {
...
           def quantileRow(data: Seq[String]): Seq[Node] = <tr> {data.map(d => <td>{d}</td>)} </tr>
           Some(listingTable(quantileHeaders, quantileRow, listings))
         }
 
       val content =
-        summary ++ <h2>Summary Metrics</h2> ++ summaryTable.getOrElse(Nil) ++
-          <h2>Tasks</h2> ++ taskTable;
+        summary ++
+        <h4>Summary Metrics for {numCompleted} Completed Tasks</h4> ++
+        <div>{summaryTable.getOrElse("No tasks have reported their execution metrics yet.")}</div> ++
```
pwendell: nit: but it might better to say "No tasks have reported metrics yet." rather than "execution metrics", we don't use that term anywhere else and it sounds a little bit like "executor metrics" which is a different thing.

pwendell:
Hey @rxin - a few minor suggestions but this looks really good. The only sticking point is the "Launch Time" label. Once you've addressed that I'm +1 to merge this.

core/src/main/scala/spark/ui/jobs/StagePage.scala
mateiz: I agree there's no point saying "0th percentile" and "100th percentile" in parentheses. One other change might be to replace "50th percentile" with "median".

rxin:
Updated. I also added some css properties to avoid overflowing tables.

rxin:
Thats actually not me. Jenkins, test this please.

pwendell:
Looks good to me!

PR #834, opened 2013-08-15
Updated json output to allow for display of worker state

Daemoen:
Ops teams need to ensure that the cluster is functional and performant.  Having to scrape the html source for worker state won't work reliably, and will be slow.  By exposing the state in the json output, ops teams are able to ensure a fully functional environment by querying for the json output and parsing for dead nodes.


rxin:
Jenkins, test this please.

rxin:
Actually I accidentally merged this one. This pull request as is won't build, because WorkerInfo doesn't have the getState method. I am working on something related to this file. I will fix it based on that.

rxin:
See #841

PR #835, opened 2013-08-15
Send Task results through the block manager when larger than Akka frame size (fixes SPARK-669).

kayousterhout:
This change requires adding an extra failure mode: tasks can complete
successfully, but the result gets lost or flushed from the block manager
before it's been fetched.

Thanks Josh for writing the original version of this patch!

### DIFF
--- a/core/src/main/scala/org/apache/spark/SparkContext.scala
+++ b/core/src/main/scala/org/apache/spark/SparkContext.scala
@@ -146,7 +146,7 @@ class SparkContext(
   }
 
   // Create and start the scheduler
-  private var taskScheduler: TaskScheduler = {
+  private[spark] var taskScheduler: TaskScheduler = {
     // Regular expression used for local[N] master format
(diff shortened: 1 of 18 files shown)


pwendell:
I had to kill this test because it had been running for more than 1 hour. Checkout the output here:

https://amplab.cs.berkeley.edu/jenkins/job/SparkPullRequestBuilder/604/console

mateiz:
Thanks for implementing this, Kay! I haven't had a chance to read through it yet but I'll take a close look early this week.

kayousterhout:
Cool sounds good. Also, "TaskResultGetter" is a bad name -- Patrick
suggested "taskResultFetcher" or "taskResultResolver" but will wait to
change it until you look at it in case you have a better name idea.

kayousterhout:
I updated this to master and also implemented Patrick's naming suggestion

rxin:
Closing this one since it has been resubmitted to the asf repo

PR #837, opened 2013-08-15
SPARK-878: Adding of zip and egg files to PYTHONPATH

AndreSchumacher:
Previously it seems that only the local working directory was added to sys,path. See the original discussion:

https://groups.google.com/forum/#!searchin/spark-users/pyfiles/spark-users/jG8VC17vTe4/z_DhBoRWuAMJ

This pull request lets PySpark's SparkContext record the *.zip and *.egg files. These are then recorded in Shark's PythonRDD, which serializes their names into the worker's input stream, similarly to the broadcast variables. Then the worker adds them to its sys.path prior to the function execution. Also a test case for the local mode is included.


core/src/main/scala/spark/api/python/PythonRDD.scala
```diff
@@ -79,6 +81,10 @@ private[spark] class PythonRDD[T: ClassManifest](
             dataOut.writeInt(broadcast.value.length)
             dataOut.write(broadcast.value)
           }
+	  // Python includes (*.zip and *.egg files)
```
JoshRosen: The indentation here is wrong; maybe your editor is mixing tabs and spaces?

python/pyspark/context.py
```diff
@@ -103,11 +104,14 @@ def __init__(self, master, jobName, sparkHome=None, pyFiles=None,
...
+        root_dir = SparkFiles.getRootDirectory()
+        sys.path.append(root_dir)
+
         # Deploy any code dependencies specified in the constructor
+        self._python_includes = list()
         for path in (pyFiles or []):
-            self.addPyFile(path)
-        SparkFiles._sc = self
-        sys.path.append(SparkFiles.getRootDirectory())
+                self.addPyFile(path, root_dir)
```
JoshRosen: Similarly, this appears to be over-indented.

python/pyspark/context.py
```diff
@@ -257,7 +261,16 @@ def addPyFile(self, path):
         HTTP, HTTPS or FTP URI.
         """
         self.addFile(path)
-        filename = path.split("/")[-1]
+        (dirname, filename) = os.path.split(path) # dirname may be directory or HDFS/S3 prefix
+
+        if filename.endswith('.zip') or filename.endswith('.ZIP') or filename.endswith('.egg'):
+            self._python_includes.append(filename)
+            sys.path.append(os.path.join(SparkFiles.getRootDirectory(), filename)) # for tests in local mode
+            #if sys.platform == 'win32' or sys.platform == 'cygwin': # we may be running on windows
```
JoshRosen: Please remove old code instead of commenting it out.

JoshRosen: Also, you could also use [os.pathsep](http://docs.python.org/2/library/os.html#os.pathsep) to get the platform-specific path separator character.

python/pyspark/rdd.py
```diff
@@ -759,7 +759,7 @@ def batched_func(split, iterator):
         env = MapConverter().convert(self.ctx.environment,
                                      self.ctx._gateway._gateway_client)
         python_rdd = self.ctx._jvm.PythonRDD(self._prev_jrdd.rdd(),
-            pipe_command, env, self.preservesPartitioning, self.ctx.pythonExec,
+            pipe_command, env, ListConverter().convert(self.ctx._python_includes, self.ctx._gateway._gateway_client), self.preservesPartitioning, self.ctx.pythonExec,
```
JoshRosen: Because this line is a little long, I'd extract the list conversion and place it on its own line (like how the environment map `env` is converted above).

core/src/main/scala/spark/api/python/PythonRDD.scala
AndreSchumacher: Good point. Either the editor or its user :) Do you btw have some automated checks for these formatting issues? I would be interested in learning to use these.

JoshRosen: You could probably use [scalastyle](http://www.scalastyle.org/) to check for issues like this, although I haven't tried it.

python/pyspark/context.py
JoshRosen: `addPyFile` only takes one argument, so I don't think this will work.  I guess we don't have any tests for adding pyFiles through SparkContext's constructor?

AndreSchumacher:
Thanks a lot for the comments, Josh! (and finding that one bug.. indeed adding files via the constructor doesn't seem to be covered by the tests currently). Also thanks to Jey for fixing my broken git clone :)

Sent a new pull request

PR #839, opened 2013-08-16
Currying RDD.zipPartitions 

jegonzal:
The previous argument ordering to zipPartitions requires additional type annotation and is slightly more difficult to use.  By currying the method we are able to automatically resolve the types of the arguments to the final map function.  Previously we would have written:

``` scala
rdd1.zipPartitions( (iter1: Iterator[T1], iter2: Iterator[T2], iter3: Iterator[T2]) =>  /* some code */, rdd2, rdd3)
```

now we can write:

``` scala
rdd1.zipPartitions(rdd2, rdd3)( (iter1, iter2, iter3) => /*some code */ )
```

I also updated the Java interface to reflect the reordering (without currying).

### DIFF
--- a/core/src/main/scala/spark/RDD.scala
+++ b/core/src/main/scala/spark/RDD.scala
@@ -515,22 +515,19 @@ abstract class RDD[T: ClassManifest](
    * *same number of partitions*, but does *not* require them to have the same number
    * of elements in each partition.
    */
-  def zipPartitions[B: ClassManifest, V: ClassManifest](
-      f: (Iterator[T], Iterator[B]) => Iterator[V],
-      rdd2: RDD[B]): RDD[V] =
+  def zipPartitions[B: ClassManifest, V: ClassManifest]
+      (rdd2: RDD[B])
+      (f: (Iterator[T], Iterator[B]) => Iterator[V]): RDD[V] =
     new ZippedPartitionsRDD2(sc, sc.clean(f), this, rdd2)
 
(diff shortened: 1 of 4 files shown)


mridulm:
Since I do not have enough context, when was zipPartitions introduced
as an api ?
If there is user code which already uses this api, this would be an
incompatible api change ...

jegonzal:
My understanding is that this is a relatively new feature (introduced by @shivaram).  However, since users may already be calling the current versions of the function in the master branch I have added them back as deprecated.  It might make sense to remove them completely as of the final 0.8 release?

shivaram:
zipPartitions is not present in 0.7 (or earlier). My take is that we can change the API before 0.8 is out as the older API has never been a part of any release. @mateiz any thoughts on this ?

mateiz:
Yes, let's change it for 0.8.

rxin:
I think we can just remove the API since it is introduced in an unreleased 0.8 version.

mateiz:
Yeah agreed on removing the old API. Does MLI use it? We can change it before release.

shivaram:
(cc @etrain for MLI updates) 
Joey, the change looks good to me. One request: Could you rebase and squash all the commits into one ?

jegonzal:
The commits have been squashed.

shivaram:
Jenkins, test this please.

etrain:
MLI isn't using this yet, so it's good with me!

mateiz:
Merged this in -- thanks Joey!

PR #840, opened 2013-08-16
Implementing SPARK-878 for PySpark: adding zip and egg files to context ...

AndreSchumacher:
...and passing it down to workers which add these to their sys.path


core/src/main/scala/spark/api/python/PythonRDD.scala
```diff
@@ -79,6 +81,10 @@ private[spark] class PythonRDD[T: ClassManifest](
             dataOut.writeInt(broadcast.value.length)
             dataOut.write(broadcast.value)
           }
+          // Python includes (*.zip and *.egg files)
+          dataOut.writeInt(pythonIncludes.length)
+          for (f <- pythonIncludes)
+            PythonRDD.writeAsPickle(f, dataOut)
```
mateiz, line 87: Small style fix: put braces around the for loop body here even if it's one line, in case we add lines later:

```
      for (f <- pythonIncludes) {
        PythonRDD.writeAsPickle(f, dataOut)
      }
```

AndreSchumacher:
Added brackets as suggested by Matei.

python/pyspark/tests.py
```diff
@@ -125,6 +125,17 @@ def func():
         from userlibrary import UserClass
         self.assertEqual("Hello World!", UserClass().hello())
 
+    def test_add_egg_file_locally(self):
+        # To ensure that we're actually testing addPyFile's effects, check that
+        # this fails due to `userlibrary` not being on the Python path:
+        def func():
+            from userlib import UserClass
+        self.assertRaises(ImportError, func)
```
mateiz, line 133: Indentation looks wrong here -- there are probably some tabs

python/pyspark/worker.py
```diff
@@ -49,15 +49,29 @@ def main(infile, outfile):
     split_index = read_int(infile)
     if split_index == -1:  # for unit tests
         return
+
+    # fetch name of workdir
     spark_files_dir = load_pickle(read_with_length(infile))
     SparkFiles._root_directory = spark_files_dir
     SparkFiles._is_running_on_worker = True
-    sys.path.append(spark_files_dir)
+    #print "read workdir"
```
mateiz: No need to keep these commented-out lines with print in the code (same for the two below)

mateiz:
Sorry, saw a couple of other small issues -- otherwise it looks good!

python/pyspark/tests.py
JoshRosen, line 133: The indentation here looks fine to me (`pep8` didn't complain about a mixture of tabs and spaces).  This line, the assertRaises, isn't meant to be part of func().

python/pyspark/rdd.py
```diff
@@ -758,8 +758,10 @@ def batched_func(split, iterator):
         class_manifest = self._prev_jrdd.classManifest()
         env = MapConverter().convert(self.ctx.environment,
                                      self.ctx._gateway._gateway_client)
+        java_list = ListConverter().convert(self.ctx._python_includes,
```
JoshRosen: Minor nitpick, but `includes` would be a more descriptive name than `java_list`.

python/pyspark/tests.py
mateiz, line 133: Oh, got it, that's true.

AndreSchumacher, line 133: I need to figure out how to make emacs replace tabs by spaces (as vim does). Sorry about that. But yeah as Josh said it seems to be OK. At least the test ran through.

AndreSchumacher:
A general question: do you prefer people to amend the commits contained in a pull request (and force-push to their cloned repository) or should these fixes that happen during the code review be committed as separate commits?

AndreSchumacher:
So I think all the suggested changes should be there now. Thanks for the feedback! I guess that one is ready then?

mateiz:
Cool, thanks! I've merged this in.

To answer your question about commits, it's fine to either append commits or do git rebase to squash small commits into one and then force-push a new branch. Ideally we only want separate commits for changes that are significant, but we've done a mix.

PR #841, opened 2013-08-16
Use the JSON formatter from Scala library and removed dependency on lift-json.

rxin:
It made the JSON creation slightly more complicated, but reduces one external dependency. The scala library also properly escape "/" (which lift-json doesn't).

This subsumes #834.

### DIFF
--- a/core/src/main/scala/spark/deploy/JsonProtocol.scala
+++ b/core/src/main/scala/spark/deploy/JsonProtocol.scala
@@ -17,7 +17,7 @@
 
 package spark.deploy
 
-import net.liftweb.json.JsonDSL._
+import scala.util.parsing.json.{JSONArray, JSONObject, JSONType}
 
 import spark.deploy.DeployMessages.{MasterStateResponse, WorkerStateResponse}
 import spark.deploy.master.{ApplicationInfo, WorkerInfo}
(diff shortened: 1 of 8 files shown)


pwendell:
Okay this looks good - did you test it in a browser and just make sure everything renders OK?

core/src/main/scala/spark/deploy/JsonProtocol.scala
```diff
@@ -25,61 +25,63 @@ import spark.deploy.worker.ExecutorRunner
...
+  def writeWorkerInfo(obj: WorkerInfo): JSONType = JSONObject(Map(
+    "id" -> obj.id,
+    "host" -> obj.host,
+    "port" -> obj.port,
+    "webuiaddress" -> obj.webUiAddress,
+    "cores" -> obj.cores,
+    "coresused" -> obj.coresUsed,
+    "memory" -> obj.memory,
+    "memoryused" -> obj.memoryUsed,
+    "state" -> obj.state.toString
```
pwendell, line 38: This and a few other things look like they were added. Did you mean to do this?

pwendell, line 38: Ah nvm, I see the other PR.

pwendell:
Ya looks good to me.

PR #842, opened 2013-08-16
Added stageId <--> jobId mapping in DAGScheduler

markhamstra:
I'm not completely certain that this is ready to be merged yet, but I think it is ready to gather some comments.

Outside of DAGScheduler.scala, the changes are almost entirely just renaming -- I resolved the confusion of the notions of priority and job identity by consistently relabeling old, FIFO-only-based usages of "priority" to now be "jobId".  The old usage of "priority" was misleading since we now have the fair scheduler infrastructure that doesn't use the old "priority" to order priority except to break what would otherwise be priority ties.  Now we can speak consistently about jobs having jobIds that are unique and always increasing.  Stages contain the jobId rof the job which was responsible for creating them.  Again, all of that is really just renaming or thinking about things slightly differently without changing anything functionally.

There are also a few easy changes within DAGScheduler.scala that just suppress a bunch of code analyzer warnings in IntelliJ -- I got tired of looking at them.

The bigger changes in the DAGScheduler have to do with the new jobIdToStageIds and stageIdToJobIds maps and the associated methods to build them up and tear them down.  (I also renamed idToStage to stageIdToStage to be more consistent and less ambiguous.)  The stageIdToJobIds map gets built up as new stages are created, the jobIdToStageIds map gets sync'ed-with/built-up from stageIdToJobIds when jobs are submitted, and both maps are modified in concert when jobs are completed or aborted.

There's a fair amount of walking through entire data structures to create and maintain these maps, so I was concerned about the performance hit.  I haven't rigorously tested just the relative performance of the DAGScheduler in isolation with and without these changes, but I have closely watched the total time reported by a complete run of Spark's unit tests, and there appears to be no significant difference -- a complete run takes almost 13.5 minutes when my machine is fully warmed up, and the difference with the new stageId <--> jobId mapping is only a couple of seconds.  Some of that is probably because, while a lot more is done when a new stage is created, that actually results in the short-circuiting of later calls to, e.g., getShuffleMapStage.  

The last bits outside the DAGScheduler have to do with SparkListenerJobStart, which now sends along an array containing all of the stageIds that the job depends on.

With all of this in place, it should now be simple for the WebUI to aggregate the appropriate stages to report progress and other information for an entire job, and it should also be easy for Ram's killJob stuff to kill all of the stages associated with a job while not killing any that are also associated with another job.

From looking at some instrumented output, I'm fairly certain that I've got things at least close to right.  I don't think this is quite ready to merge yet, since the JobLogger still probably needs to be tweaked, and I need to come to an understanding of why stages seem to be left alive by the ThreadingSuite.  Outside of that, this new code seems to do a nice job preventing the sizes of various DAGScheduler data structures from growing without bound, and does so while not breaking any of the unit tests.

### DIFF
--- a/core/src/main/scala/org/apache/spark/MapOutputTracker.scala
+++ b/core/src/main/scala/org/apache/spark/MapOutputTracker.scala
@@ -235,6 +235,10 @@ private[spark] class MapOutputTracker extends Logging {
     }
   }
 
+  def has(shuffleId: Int): Boolean = {
+    cachedSerializedStatuses.get(shuffleId).isDefined || mapStatuses.contains(shuffleId)
+  }
+
   def getSerializedLocations(shuffleId: Int): Array[Byte] = {
     var statuses: Array[MapStatus] = null
     var epochGotten: Long = -1
@@ -247,12 +251,12 @@ private[spark] class MapOutputTracker extends Logging {
         case Some(bytes) =>
           return bytes
         case None =>
-          statuses = mapStatuses(shuffleId)
+          statuses = mapStatuses.getOrElse(shuffleId, Array[MapStatus]())
           epochGotten = epoch
       }
     }
     // If we got here, we failed to find the serialized locations in the cache, so we pulled
-    // out a snapshot of the locations as "locs"; let's serialize and return that
+    // out a snapshot of the locations as "statuses"; let's serialize and return that
     val bytes = serializeStatuses(statuses)
     logInfo("Size of output statuses for shuffle %d is %d bytes".format(shuffleId, bytes.length))
     // Add them into the table only if the epoch hasn't changed while we were working
@@ -261,7 +265,7 @@ private[spark] class MapOutputTracker extends Logging {
         cachedSerializedStatuses(shuffleId) = bytes
       }
     }
-    return bytes
+    bytes
   }
 
   // Serialize an array of map output locations into an efficient byte format so that we can send
--- a/core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
@@ -289,46 +289,58 @@ class DAGScheduler(
   }
 
   private def jobIdToStageIdsRemove(jobId: Int) {
+    def removeStage(stageId: Int) {
+      // data structures based on Stage
+      stageIdToStage.get(stageId).foreach { s => {
+        stageToInfos -= s
+        if (pendingTasks.contains(s)) {
+          logError("Tasks still pending for stage %d even though there are no more jobs registered for that stage."
+                   .format(stageId))
+          pendingTasks -= s
+        }
+        if (waiting.contains(s)) {
+          logError("Still waiting on stage %d even though there are no more jobs registered for that stage."
+                   .format(stageId))
+          waiting -= s
+        }
+        if (running.contains(s)) {
+          logError("Stage %d still running even though there are no more jobs registered for that stage."
+                   .format(stageId))
+          running -= s
+        }
+        if (failed.contains(s)) {
+          logError("Stage %d still registered as failed even though there are no more jobs registered for that stage."
+                   .format(stageId))
+          failed -= s
+        }
+      }}
+      // data structures based on StageId
+      stageIdToStage -= stageId
+      stageIdToJobIds -= stageId
+
+      logDebug("After removal of stage %d, remaining stages = %d".format(stageId, stageIdToStage.size))
+    }
+
     if (!jobIdToStageIds.contains(jobId)) {
       logError("Trying to remove unregistered job " + jobId)
     } else {
       val registeredStages = jobIdToStageIds(jobId)
       if (registeredStages.isEmpty) {
         logError("No stages registered for job " + jobId)
-      } else { stageIdToJobIds.filterKeys(stageId => registeredStages.contains(stageId)).foreach {
-        case (stageId, jobSet) => {
-          if (!jobSet.contains(jobId)) {
-            logError("Job %d not registered for stage %d even though that stage was registered for the job"
-              .format(jobId, stageId))
-          } else {
-            jobSet -= jobId
-          }
-          if (jobSet.isEmpty) {
-            stageIdToStage.get(stageId).foreach { s =>
-              stageToInfos -= s
-              if (pendingTasks.contains(s)) {
-                logError("Tasks still pending for stage " + stageId + " even though there are no more jobs registered for that stage.")
-                pendingTasks -= s
-              }
-              if (waiting.contains(s)) {
-                logError("Still waiting on stage " + stageId + " even though there are no more jobs registered for that stage.")
-                waiting -= s
-              }
-              if (running.contains(s)) {
-                logError("Stage " + stageId + " still running even though there are no more jobs registered for that stage.")
-                running -= s
-              }
-              if (failed.contains(s)) {
-                logError("Stage " + stageId + " still registered as failed even though there are no more jobs registered for that stage.")
-                failed -= s
-              }
+      } else {
+        stageIdToJobIds.filterKeys(stageId => registeredStages.contains(stageId)).foreach {
+          case (stageId, jobSet) => {
+            if (!jobSet.contains(jobId)) {
+              logError("Job %d not registered for stage %d even though that stage was registered for the job"
+                       .format(jobId, stageId))
+            } else {
+              jobSet -= jobId
             }
-            stageIdToStage -= stageId
-            stageIdToJobIds -= stageId
-            logDebug("After removal of stage %d, remaining stages = %d".format(stageId, stageIdToStage.size))
+            if (jobSet.isEmpty) { // nobody needs this stage any more
+              removeStage(stageId)
+            }
           }
-        }
-      }}
+        }}
       jobIdToStageIds -= jobId
     }
   }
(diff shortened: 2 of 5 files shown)


rxin:
Thanks for doing this, Mark.

I'd actually recommend you to separate this as two pull requests since they are fairly independent. The renaming pull request should be much easier to merge, while the other one would require a lot of testing. Note that unit test runtime by no means test the scheduler's throughput, since the runtime is dominated by failure recovery, shuffle, and other things.

I believe in the perf benchmark suite Patrick has been working on (going to open source soon or have already open sourced) includes a scheduler throughput test that just launches a large number of empty tasks. We should use that one to test the scheduler throughput here.

markhamstra:
No disagreement that just timing the unit tests isn't an adequate measure of the change in scheduler performance.  I was just using what I had as a rough check that things hadn't gone horribly wrong.  If Patrick's suite can be used to do some better testing, that's great.

markhamstra:
UPDATE: I broke out renaming and other minor changes into a separate pull request; so after https://github.com/mesos/spark/pull/844 is merged, this one reduces to just the stageId <--> jobId mapping stuff.

markhamstra:
So, a few things worth talking about:
1. This PR essentially boils down to moving at least part of the logic from the JobLogger directly into the DAGScheduler.
2. At the moment, that work is duplicated in both DAGScheduler and JobLogger, but JobLogger should be trimmed down once we decide exactly what gets done in DAGScheduler.
3. JobLogger originally ran in its own thread; now runs via the new async SparkListener handling, so again in a separate thread.  This PR brings some of that work directly into the scheduler, with obvious performance concerns.  However, if testing bears out that the performance hit is not too severe, then I think the change is worth it to not just track the jobs <--> stages relationship, but also to control the growth of DAGScheduler data structures, and doing so while keeping the concurrency issues manageable.
4. Enough of the jobs <--> stages information needs to be sent out from the DAGScheduler to SparkListeners so that they don't need to duplicate much of the work in order to get all the information they need.  Right now, I'm just sending out an array of associated StageIds within a SparkListenerJobStart, but that may not be enough information exported or the best way of making the information available to SparkListeners.
5. Right now, this PR is maximally aggressive about cleaning up when a job completes or aborts and a stage is not part of another job.  That's probably not the right thing to do if another job will soon launch and want to make use of one of the earlier job's stages.  Putting the no-longer-in-a-job stages onto a queue for a while (how long?) before cleaning them out of the DAGScheduler data structures may make sense as a way to give other jobs a chance to re-use old stages.

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -108,6 +108,10 @@ class DAGScheduler(
 
   val nextStageId = new AtomicInteger(0)
 
+  val jobIdToStageIds = new TimeStampedHashMap[Int, HashSet[Int]]
+
+  val stageIdToJobIds = new TimeStampedHashMap[Int, HashSet[Int]]
```
mateiz: You should call clearOldValues() on these two maps in DAGScheduler.cleanup()

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -260,6 +265,58 @@ class DAGScheduler(
...
+        val parentsWithoutThisJobId = parents.filter(p => !stageIdToJobIds.get(p.id).exists(_.contains(jobId)))
+        registerJobIdWithStageList(parentsWithoutThisJobId ++ stages.tail)
+      }
+    }
+    registerJobIdWithStageList(List(stage))
+  }
+
+  private def jobIdToStageIdsAdd(jobId: Int) {
+    val stageSet = jobIdToStageIds.getOrElseUpdate(jobId, new HashSet[Int]())
+    stageIdToJobIds.foreach{case (stageId, jobSet) =>
```
mateiz: Small style thing -- put spaces around the brace:

```
stageIdToJobIds.foreach { case (stageId, jobSet) =>
```

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -260,6 +265,58 @@ class DAGScheduler(
...
+      }
+    }
+  }
+
+  private def jobIdToStageIdsRemove(jobId: Int) {
+    if (!jobIdToStageIds.contains(jobId))
+      logError("Trying to remove unregistered job " + jobId)
+    else {
+      val registeredStages = jobIdToStageIds(jobId)
+      if (registeredStages.isEmpty)
```
mateiz: Put braces around the if and else bodies, even if they're one-line, to match the style of the rest of the code

core/src/main/scala/spark/scheduler/DAGScheduler.scala
```diff
@@ -260,6 +265,58 @@ class DAGScheduler(
...
+              .format(jobId, stageId))
+          else
+            jobSet -= jobId
+          if (jobSet.isEmpty) {
+            stageIdToStage.get(stageId).foreach{s =>
+              pendingTasks -= s
+              waiting -= s
+              running -= s
+              failed -= s
+            }
```
mateiz: I'm curious, why can this happen -- basically it would mean we decided to remove a job before all its stages were finished?

mateiz:
Hey Mark, I made a few comments on the PR. Is this still a work in progress on your side as you said in the original message, or is it mostly done?

core/src/main/scala/spark/scheduler/DAGScheduler.scala
markhamstra: Not sure what your "this" refers to.  There shouldn't be be stages left in the pendingTasks, waiting, running or failed queues at this point, so if there are, then I should at least log the unexpected occurrence.  If the unexpected occurs, do you want to fail, adopt this "clean up for a mistake someplace else" approach, or try to do something like TimeStampedHashMap to periodically clean up the leaks?

Within jobIdToStageIdsRemove, I'm also now looking at cleaning up stageIdToJobIds, stageToInfos and shuffleToMapStage so that they don't continue to grow.  The first two are easy, but I've got some work to do on the third.

markhamstra:
Still work in progress.  I'm adding "data structures are now empty" assertions at the end of the tests in DAGSchedulerSuite and need to do some work still on shuffleToMapStage in general and the handling of clean up in the zero partition and local job cases.

After that, it should be down to addressing the five points in my "things worth talking about" list.

core/src/main/scala/spark/scheduler/DAGScheduler.scala
mateiz: Yeah, I meant the cleanup code here. You should at least log an error if the jobSet is empty but the stage is still in pendingTasks, waiting, etc. Otherwise, this will make it harder to catch such bugs.

markhamstra:
Updated.  Still WIP.

core/src/main/scala/spark/scheduler/DAGScheduler.scala
markhamstra: Done

markhamstra:
UPDATED: The DAGSchedulerSuite now includes checks in all tests except the "zero split job", asserting that all DAGScheduler data structures are empty at the end of each test job.  These were mostly fairly easy, but cleaning up shuffleToMapStage is considerably more involved, since it is not just internal to the DAGScheduler, but also involves external state in the MapOutputTracker.

In order to reuse map outputs that are still available for a stage that has previously been cleaned out of the DAGScheduler (and avoid re-submitting such stages), I've split DAGScheduler.newStage into two pieces: the first to initialize new stages, the second to re-use or re-constitute stages using output locations available in the mapOutputTracker.

Along the way, I also had to fix SPARK-864.  That bug resulted from trying to submit stages and missing tasks for a job that is no longer active (thus failing in the idToActiveJob(stage.jobId) lookup when submitMissingTasks tries to get appropriate properties for the job.)  With this PR, the properties for the earliest job associated with the stage are used, or the stage is aborted if there are no longer any active jobs associated with the stage.

With the exception of handling zero-partition RDDs and passing sufficient information to SparkListeners, I now consider this PR to be feature complete, and I'll move on to performance regression testing.

markhamstra:
I've got spark-perf numbers: 

scheduling-tput
  --num-tasks=10000 --num-trials=10 --inter-trial-wait=30, 

  baseline: 4.955, 0.156, 4.539, 4.955, 4.766
  PR842:    5.016, 0.106, 4.786, 5.071, 4.892

scala-agg-by-key
  --num-trials=10 --inter-trial-wait=30 --num-partitions=1 --reduce-tasks=1 --random-seed=5 --persistent-type=memory  --num-records=200 --unique-keys=1 --key-length=10 --unique-values=1 --value-length=10, 

  baseline: 0.047, 0.002, 0.045, 0.049, 0.047
  PR842:    0.049, 0.003, 0.045, 0.050, 0.049

scala-sort-by-key
  --num-trials=10 --inter-trial-wait=30 --num-partitions=1 --reduce-tasks=1 --random-seed=5 --persistent-type=memory  --num-records=20 --unique-keys=1 --key-length=10 --unique-values=1 --value-length=10, 

  baseline: 0.045, 0.004, 0.041, 0.046, 0.047
  PR842:    0.046, 0.003, 0.041, 0.048, 0.052

scala-count
  --num-trials=10 --inter-trial-wait=30 --num-partitions=1 --reduce-tasks=1 --random-seed=5 --persistent-type=memory  --num-records=200 --unique-keys=1 --key-length=10 --unique-values=1 --value-length=10, 

  baseline: 0.018, 0.002, 0.017, 0.022, 0.017
  PR842:    0.018, 0.003, 0.017, 0.025, 0.018

scala-count-w-fltr
  --num-trials=10 --inter-trial-wait=30 --num-partitions=1 --reduce-tasks=1 --random-seed=5 --persistent-type=memory  --num-records=200 --unique-keys=1 --key-length=10 --unique-values=1 --value-length=10, 

  baseline: 0.022, 0.002, 0.02, 0.025, 0.020
  PR842:    0.024, 0.003, 0.018, 0.025, 0.020

mateiz:
Cool, that looks good! I'd like to wait until 0.8.0 is out to merge this but it's definitely appreciated.

markhamstra:
It's definitely getting there.  I'm curious why DriverSuite is failing at least some of the time and why the Python tests are frequently complaining about [tasks still pending for a stage even though there are no more jobs needing that stage](https://github.com/mesos/spark/pull/842/files#L1R320).  Not sure yet whether either of these are real bugs or just test artifacts.  I'll look through some worker logs looking for more such logError messages, but I'll go out on a limb and claim that any such messages are as likely the result of weaknesses in existing code that the new checks are revealing as they are the result of mistakes in this PR.

Anyway, I still haven't resolved the handling of zero-partition RDDs by the DAGScheduler, so those weirdos will continue to cause data structure bloat -- but no worse than they were before, so still a large net reduction in bloat for all other circumstances.  The other question is just what the JobLogger and WebUI need in order to avoid duplicating reference tracking work.  Probably should be sending both jobIdToStageIds and stageIdToJobIds out to SparkListeners on job start.  Each of these remaining issues can easily and without harm be resolved in follow-on pull requests, I think.

markhamstra:
Rebased and fixed some typos and grammatical niggles.

markhamstra:
Addressed the pendingTasks bloat -- it wasn't just for Python tests, but for every job.  I think I got the life-cycle issues correct (i.e. will always be a set of task associated with `stage` when handling a successful task completion, and it's possible that the stage -> task set may already have been removed when a task is resubmitted), but take a close look @mateiz . 

UPDATE: Changed my mind.  Better and much simpler to leave handleTaskCompletion alone, to expect pendingTasks(stage).isEmpty when a stage completes, and to handle that expectation and final cleanup in removeStage.

mateiz:
Jenkins, retest this please

PR #845, opened 2013-08-17
Add randomSplit and kFoldSplit cross validations to mllib

aht:


mllib/src/test/scala/spark/mllib/crossvalidation/CrossValidationSuite.scala
```diff
@@ -0,0 +1,55 @@
+package spark.mllib.crossvalidation
```
mateiz, line 18: Please add an Apache license header to this file too

mateiz:
Jenkins, this is ok to test

mllib/src/test/scala/spark/mllib/crossvalidation/CrossValidationSuite.scala
aht, line 18: Done.

shivaram:
Hi @aht,
Thanks for sending this out. Splitting an RDD into disjoint subsets is a pretty useful feature that has been requested before. I have a few style comments that I'll add to the files, but here are some high-level comments on the design:
- I think the RDD can be moved into `core/src/main/scala/spark/rdd`. I think random/disjoint split is a pretty general operation and not specific to just ML Lib.
- On the same note, right now the `RandomSplitRDD` only creates two splits. I think it would be better to generalize this to do `k` disjoint splits. @atalwalkar can provide more details, but afaik there are other algorithms where a k-way split would be useful.
- Going down this path one way to do this would to be just create disjoint RDDs by a top level operation in `RDD` that returns a list of RDDs each of which may have a subset of the splits. In some sense this would be the opposite of `coalesce` where we go from many partitions to few partitions. Optionally we could also shuffle the data before doing the split etc. This would be a slightly bigger change though @mateiz / @rxin - any thoughts on this ?

mllib/src/main/scala/spark/mllib/crossvalidation/CrossValidation.scala
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ */
+
+package spark.mllib.crossvalidation
+
+import spark.RDD
+import java.util.Random
+
+object CrossValidation {
+  /** Return an Iterator of size k of (train, test) RDD Tuple,
```
shivaram: The comment style we use in spark is something like 

```
/**
 * Return an iterator
 */
```

mllib/src/main/scala/spark/mllib/crossvalidation/CrossValidation.scala
```diff
@@ -0,0 +1,48 @@
...
+    * whose respective size ratio are approximately (trainingSize, 1-trainingSize).
+    * The train & test data across k split are sampled using different seed at each iteration.
+    */
+  def randomSplit[T](rdd: RDD[T], k: Int, trainingSize: Double, seed: Long)(implicit _cm: ClassManifest[T]): Iterator[(RDD[T], RDD[T])] = {
```
shivaram: This line looks like it is > 100 characters ? Also you can just use `cm: ClassManifest`, we don't typically use variables with under-score prefix.

mllib/src/main/scala/spark/mllib/crossvalidation/CrossValidation.scala
```diff
@@ -0,0 +1,48 @@
...
+    require(0 < trainingSize && trainingSize < 1)
+    val rg = new Random(seed)
+    (1 to k).map(_ => rg.nextInt).map(z =>
```
shivaram: multi-line map calls and for loops are usually indented with curly braces. For example

```
(1 to k).map { z =>
}
```

mllib/src/main/scala/spark/mllib/crossvalidation/RandomSplitRDD.scala
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  override def getPreferredLocations(split: Partition): Seq[String] =
+    firstParent[T].preferredLocations(split.asInstanceOf[SampledRDDPartition].prev)
+
+  override def compute(splitIn: Partition, context: TaskContext): Iterator[T] = {
+    val split = splitIn.asInstanceOf[SampledRDDPartition]
+    val rand = new Random(split.seed)
+    if (isTraining) {
+      firstParent[T].iterator(split.prev, context).filter(x => {val z = rand.nextDouble; z < lower || z >= upper})
```
shivaram: Similarly to the comment above, this could be 

```
iterator(split.prev, context).filter { x =>
  val z = rand.nextDouble
  z < lower || z >= upper
}
```

mllib/src/test/scala/spark/mllib/crossvalidation/CrossValidationSuite.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  test("randomSplit") {
+    val data = sc.parallelize(1 to 1000, 2)
+    val trainSize = 0.85
+    for (seed <- 1 to 5) {
+      for (split <- CrossValidation.randomSplit(data, 5, trainSize, seed)) {
+        val (train, test) = split
+        val trainArray = train.collect
+        val testSet = test.collect.toSet
+        assertEquals(trainSize, trainArray.size/1000.0, 0.025)
```
shivaram, line 46: We encourage using scalatest assert statements instead of junit asserts. For example this would be `assert(trainSize === 0.025)`

aht, line 46: actually this one is a assert trainArray.size/1000.0 == 0.85 with epsilon 0.025

aht:
> On the same note, right now the RandomSplitRDD only creates two splits. I think it would be better to generalize this to do k disjoint splits. @atalwalkar can provide more details, but afaik there are other algorithms where a k-way split would be useful.

@shivaram this patch have the kFoldSplit function which does the usual k-way split often used in practice. Note that it split the rdd into k (train, test) pairs. The implementation does a k-way split but only in essence, arranging it so that ~1/k of the data goes to the test set and the rest go to the train set.

shivaram:
Right, I saw the k-way split in this patch -- I just think it would be better to have the RDD operation be more general and support k-disjoint splits from a given RDD. Then we can do k-fold cross validation with a wrapper function that computes on a `union` of k-1 RDDs for training. 

Also while I agree that k (test, train) pairs is useful, I was referring to clustering algorithms like Dirichlet Process means (written by @pxinghao) which uses `k` disjoint splits, one for every iteration.

aht:
I can add easily a function to return k disjoint splits of a RDD, by returning k RandomSplitRDD with different sliding window of (upper, lower). That's what the kFoldSplit function does anyway, but in a different return format.

Proposed signature:  disjointRandomSplit(rdd, k) => Array(rdd1, rdd2, ..., rddk)

Are we waiting for any input from @rxin or @mateiz or @atalwalkar regarding the following?
- whether to put RandomSplitRDD into the spark.rdd package
- whether to surface rdd.disjointRandomSplit(k) as a top level RDD operation

shivaram:
Right -- the proposed signature would be adding an operator `split(k: Int): Seq[RDD[T]]` to RDD.scala -- Also the other issue is that right now each of the `k` splits actually touches the entire dataset since we use a filter-based approach. I think that an approach where we split the input partitions of the base RDD to get disjoint splits would be better as then each of the disjoint RDDs won't need to process the entire dataset. (Optionally we could shuffle before the split)

And yes, I'd say lets wait for @mateiz thoughts on this as this is a change to the RDD interface.

aht:
@shivaram Note that rdd.sample() currently also use the filter-based approach. 

If the k-way splits are partition-based then we have to rehash/shuffle into (multiple of) k even-sized partitions in order to have a guarantee about the split sizes training so shuffling is not optional, and it is quite expensive.

PR #846, opened 2013-08-18
Two minor RDD refactoring

rxin:
1. Removed mapSideCombine option in CoGroupedRDD and partitionBy function.
2. Moved the shuffle serializer setting from constructor of respected RDDs to a separate method named setSerializer.


mateiz:
Looks good, thanks.

PR #847, opened 2013-08-19
Allow subclasses of Product2 in all key-value related classes

rxin:
(ShuffleDependency, PairRDDFunctions, etc)

This change would allow more performance optimizations for Spark applications, e.g.:
1. Across elements, reuse a mutable implementation of Tuple2 to reduce temporary object creation (unfortunately the current iterator model makes it harder for escape analysis to do on stack allocation for these temporary tuples).
2. Making the key element of the tuple transient. This can result in huge shuffle savings when the application is using Spark in BSP fashion (the key is often used to route the message to a specific partition; once the partition code is calculated, the key is no longer needed).

### DIFF
--- a/core/src/main/scala/spark/PairRDDFunctions.scala
+++ b/core/src/main/scala/spark/PairRDDFunctions.scala
@@ -165,7 +165,7 @@ class PairRDDFunctions[K: ClassManifest, V: ClassManifest](self: RDD[(K, V)])
 
     def reducePartition(iter: Iterator[(K, V)]): Iterator[JHashMap[K, V]] = {
       val map = new JHashMap[K, V]
-      for ((k, v) <- iter) {
+      iter.foreach { case (k, v) =>
         val old = map.get(k)
         map.put(k, if (old == null) v else func(old, v))
       }
@@ -173,7 +173,7 @@ class PairRDDFunctions[K: ClassManifest, V: ClassManifest](self: RDD[(K, V)])
     }
 
     def mergeMaps(m1: JHashMap[K, V], m2: JHashMap[K, V]): JHashMap[K, V] = {
-      for ((k, v) <- m2) {
+      m2.foreach { case (k, v) =>
         val old = m1.get(k)
         m1.put(k, if (old == null) v else func(old, v))
       }
--- a/core/src/main/scala/spark/PairRDDFunctions.scala
+++ b/core/src/main/scala/spark/PairRDDFunctions.scala
@@ -21,9 +21,8 @@ import java.nio.ByteBuffer
 import java.util.{Date, HashMap => JHashMap}
 import java.text.SimpleDateFormat
 
-import scala.collection.Map
+import scala.collection.{mutable, Map}
 import scala.collection.mutable.ArrayBuffer
-import scala.collection.mutable.HashMap
 import scala.collection.JavaConversions._
 
 import org.apache.hadoop.conf.Configuration
@@ -38,7 +37,6 @@ import org.apache.hadoop.mapred.OutputFormat
 
 import org.apache.hadoop.mapreduce.lib.output.{FileOutputFormat => NewFileOutputFormat}
 import org.apache.hadoop.mapreduce.{OutputFormat => NewOutputFormat, RecordWriter => NewRecordWriter, Job => NewAPIHadoopJob, HadoopMapReduceUtil}
-import org.apache.hadoop.security.UserGroupInformation
 
 import spark.partial.BoundedDouble
 import spark.partial.PartialResult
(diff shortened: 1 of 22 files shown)


core/src/main/scala/spark/PairRDDFunctions.scala
```diff
@@ -378,7 +374,13 @@ class PairRDDFunctions[K: ClassManifest, V: ClassManifest](
...
    * Return the key-value pairs in this RDD to the master as a Map.
    */
-  def collectAsMap(): Map[K, V] = HashMap(self.collect(): _*)
+  def collectAsMap(): Map[K, V] = {
+    val data = self.toArray()
+    val map = new mutable.HashMap[K, V]
+    map.sizeHint(data.length)
+    data.foreach { case(k, v) => map.put(k, v) }
+    map
+  }
```
mateiz, line 381: Is this really faster than the HashMap constructor?

ryanlecompte, line 381: It would appear so:

``` scala
scala> val th = new ichi.bench.Thyme

scala> val items = Array.tabulate(10000) { i => (i, i) }

scala> def a(items: Seq[(Int, Int)]): Map[Int, Int] = Map(items: _*)

scala> def b(items: Seq[(Int, Int)]): Map[Int, Int] = { val map = new mutable.HashMap[Int, Int]; map.sizeHint(items.size); items.foreach { case (k, v) => map.put(k, v) }; map }

scala> th.pbench { a(items) }
Benchmark (2540 calls in 7.554 s)
  Time:    2.352 ms   95% CI 2.260 ms - 2.443 ms   (n=20)
  Garbage: 560.2 us   (n=344 sweeps measured)

scala> th.pbench { b(items) }
Benchmark (5100 calls in 5.399 s)
  Time:    961.5 us   95% CI 919.1 us - 1.004 ms   (n=20)
  Garbage: 83.59 us   (n=101 sweeps measured)
```

Thyme is great for benchmarking this kind of stuff (might be useful to use Thyme to compare the AppendOnlyHashMap vs. scala.collection.mutable.OpenHashMap too).

https://github.com/Ichoran/thyme

rxin, line 381: While it is faster, it exposes a problem that a lot of internal Spark code uses case(k, v) to pattern match, and this would be problematic in the case of non-tuple2 product2.... I will update the PR later.

rxin:
I ended up making only the implementations of RDDs and shuffle related files taking _ <: Product2, but kept PairRDDFunctions as in (still requiring Tuple2). I think this strikes a good balance between keeping the APIs consistent (for PairRDDFunctions) and allowing performance optimizations (if one was to exploit performance at this level, they should be able to construct jobs out of concrete implementations of RDDs directly).

Scala gurus, please take a look, especially on all the type system stuff. @ryanlecompte

ryanlecompte:
I took a brief look and it looks fine overall. Just a couple things:
- I haven't looked in detail, but are we sure that introducing `MutablePair` won't introduce any race conditions anywhere? For example, just ensuring that multiple tasks within a single JVM aren't trying to concurrently do things like `mp._1 = ...`.
- Would be nice if we didn't have to perform `asInstanceOf[XYZ]` all over the place. Are we sure that it's safe to do this in the new places that we're doing it in this pull request? Just want to make sure we are appeasing the compiler vs. making any new assumptions about types of RDDs. The places that I saw it done looked fine, though.

Nice work!

rxin:
For the first one, the user has to know what he is doing.

For the 2nd one, I don't think I actually introduced more casts; if anything, I removed some.

rxin:
Added a new test for sorting.

ryanlecompte:
Perfect, thanks for the clarification Reynold.

core/src/main/scala/spark/util/MutablePair.scala
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package spark.util
+
+
+/** A tuple of 2 elements.
+  *  @param  _1   Element 1 of this MutablePair
+  *  @param  _2   Element 2 of this MutablePair
```
mateiz: You should explain the purpose of this a bit better here (say it's for situations where we want to reuse one object and avoid allocation). Also, for the comment style, use this:

```
/**
 * A tuple of 2 elements
 * ...
 */
```

It's slightly different from the "official" Scala style but it's what we've used throughout.

core/src/main/scala/spark/util/MutablePair.scala
```diff
@@ -0,0 +1,34 @@
...
+  */
+case class MutablePair[@specialized(Int, Long, Double, Char, Boolean/*, AnyRef*/) T1,
+                      @specialized(Int, Long, Double, Char, Boolean/*, AnyRef*/) T2]
+  (var _1: T1,var _2: T2)
```
mateiz: Space after the comma

core/src/main/scala/spark/util/MutablePair.scala
```diff
@@ -0,0 +1,34 @@
...
+  extends Product2[T1, T2]
+{
+
```
mateiz: No need for a blank line here

rxin:
Ok I added more tests and incorporated the review feedback.

rxin:
Jenkins, test this please.

mateiz:
Looks good, thanks.

PR #848, opened 2013-08-19
IndexedRDD for accelerated joins

jegonzal:
In many applications (especially graph computation and machine learning) we are iteratively joining model parameters (vertices) with data (edges).  In these cases it can be beneficial to pre-organize the records within each partition to share a similar structure.  

Logically the `IndexedRDD[K,V]` extends the `RDD[(K,V)]` and provides the same functionality.  An IndexedRDD is constructed by using the `PairRDDFunctions.indexed` method:

``` scala
val tbl: IndexedRDD[String, Int] = sc.parallelize(Array( ("1",1), ("1", -1), ("3",3), ("5",5) )).indexed() 
```

The `tbl.index` can be then applied to other RDDs that share the same key set:

``` scala
val tbl: IndexedRDD[String, Int] = sc.parallelize(Array( ("1",1), ("1", -1), ("3",3), ("5",5) )).indexed()
val tbl2: IndexedRDD[String, Float] = sc.parallelize(Array( ("1",1.0), ("3",3.0), ("5",5.0) )).indexed(tbl.index)  
```

By sharing the same index across multiple RDDs, we guarantee that the values within each RDD is organized identically enabling fast join operations.  Furthermore, because the index is prematerialized (and cached) we can use it to accelerate join operations with `RDD[(K,V)]` that have not been indexed.


jegonzal:
This is still a prototype but I wanted to open the discussion on the design so I can incorporate feedback early.

core/src/main/scala/spark/rdd/IndexedRDD.scala
```diff
@@ -0,0 +1,550 @@
...
+
+  override def getPreferredLocations(s: Partition): Seq[String] = 
+    tuples.getPreferredLocations(s)
+
+
+  override def cache: IndexedRDD[K,V] = {
+    index.cache
+    valuesRDD.cache
+    return this
+  }
```
mateiz: You should be overriding persist() instead of cache() here to work with all storage levels

core/src/main/scala/spark/rdd/IndexedRDD.scala
```diff
@@ -0,0 +1,550 @@
...
+   * Construct a new IndexedRDD that is indexed by only the keys in the RDD
+   */
+   def reindex(): IndexedRDD[K,V] = IndexedRDD(this)
+
+
+  /**
+   * Pass each value in the key-value pair RDD through a map function without changing the keys;
+   * this also retains the original RDD's partitioning.
+   */
+  def mapValues[U: ClassManifest](f: V => U): IndexedRDD[K, U] = {
```
mateiz: I'm somewhat worried that all of these PairRDDFunctions functions have to be reimplemented here. It would be better for PairRDDFunctions to test whether the RDD it's built on has an index and execute differently in that case.

jegonzal: Agreed!  Actually, I looked at several designs that would have allowed me to override just cogroup, combinebykey, and mapValues and and then rely on the rest of PairRDDFunctions.  In the end I decided to implement a separate version of the functions since many of them are not needed (they take partitioner information which is already constrained by the index) and the return type is an RDD[(K,V)] which would make it difficult to access the index.

mateiz: I see, that is a good point. Basically I think the solution would be to have _all_ RDDs have an optional Index field, and use that in PairRDDFunctions the same way we use the optional Partitioner. But let's discuss in person to see whether that makes sense.

mridulm:
I have not looked into this in great detail, but shouldn't cogroup, etc
(anything using multiple rdd's) validate the partitioners are equal before
comparing partition index, etc ?
Or is the github mobile site messing with me !

Regards
Mridul

jegonzal:
I have made some organizational changes based on suggestions from @rxin and @mateiz.  

To RDD I added: 

``` scala
  def pairRDDFunctions[K, V](
      implicit t: T <:< (K, V), k: ClassManifest[K], v: ClassManifest[V]): 
      PairRDDFunctions[K, V] = {
    new PairRDDFunctions(this.asInstanceOf[RDD[(K,V)]])
  }
```

This interesting piece of code returns a PairRDDFunctions wrapper for this RDD.  By delegating the construction of the pairRDDFunctions object to the RDD class, specializations of RDD can return different implementations of the PairRDDFunctions.  For example, the new IndexedRDD class overrides this function to return: 

``` scala
override def pairRDDFunctions[K1, V1](
      implicit t: (K, V) <:< (K1,V1), k: ClassManifest[K1], v: ClassManifest[V1]): 
    PairRDDFunctions[K1, V1] = {
    new IndexedRDDFunctions[K1,V1](this.asInstanceOf[IndexedRDD[K1,V1]])
  }
```

I updated the implicit construction of PairRDDFunctions to call `rdd.pairRDDFunctions`  

I modified PairRDDFunctions to provide a ClassManifest for intermediate types.  This is needed for some PairRDD implementations that are array backed (i.e., IndexedRDD).

jegonzal:
In response to @mridulm, the IndexedRDD exploits the partitioning of the index.  If both RDDs are indexed (IndexedRDDs) and have the same index then both must be partitioned identically and even arranged identically within each partition enabling cogroup to be achieve using zip.  If both RDDs are indexed with a different index then the standard shuffle logic is applied though the index is used to help pre-construct hash maps.

mridulm:
To give an example of what I was referring to - take a look at cogroup in IndexedRDDFunctions, 

self and other IndexedRDDFunctions need not be partitioned by the same partitioner - and so direct index (value) comparisons in the case match can result in incorrectness ?

Unfortunately, I do not have time to go over this PR - so will derfer to someone else to comment on this further.

AndreSchumacher:
Hey Joey, this looks nice. Do you happen to have some microbenchmarks for the performance gains of indexed versus non-indexed cogroup? Also you you mentioned the applications to graph algorithms. What other operations on RDD's do you think could benefit from having an index from this context?

jegonzal:
Unfortunately, I don't have any good benchmarks to evaluate joins so I am in the process of creating some.  Does anyone have suggestions?

PR #849, opened 2013-08-20
Small fixes to web UI

mateiz:
- Use SPARK_PUBLIC_DNS environment variable if set (for EC2); this fixes SPARK-833 (https://spark-project.atlassian.net/browse/SPARK-833)
- Use a non-ephemeral port (3030 instead of 33000) by default
- Updated test to use non-ephemeral port too

### DIFF
--- a/core/src/test/scala/spark/ui/UISuite.scala
+++ b/core/src/test/scala/spark/ui/UISuite.scala
@@ -30,8 +30,9 @@ class UISuite extends FunSuite {
     val (jettyServer1, boundPort1) = JettyUtils.startJettyServer("localhost", startPort, Seq())
     val (jettyServer2, boundPort2) = JettyUtils.startJettyServer("localhost", startPort, Seq())
 
-    assert(boundPort1 === startPort + 1)
-    assert(boundPort2 === startPort + 2)
+    // Allow some wiggle room in case ports on the machine are under contention
+    assert(boundPort1 > startPort && boundPort1 < startPort + 10)
+    assert(boundPort2 > boundPort1 && boundPort2 < boundPort1 + 10)
   }
 
   test("jetty binds to port 0 correctly") {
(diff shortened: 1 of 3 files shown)


core/src/test/scala/spark/ui/UISuite.scala
```diff
@@ -24,7 +24,7 @@ import org.eclipse.jetty.server.Server
 
 class UISuite extends FunSuite {
   test("jetty port increases under contention") {
-    val startPort = 33333
+    val startPort = 3030
     val server = new Server(startPort)
     server.start()
     val (jettyServer1, boundPort1) = JettyUtils.startJettyServer("localhost", startPort, Seq())
```
rxin, line 30: While you are at this one, maybe we can change the test to just check startPort + delta > boundPort1 > startPort, and boundPort1 + delta > boundPort2 > boundPort1? 

Setting delta to 10 would get rid of the occasional jenkins failures.

mateiz, line 30: Yeah good idea

mateiz:
Jenkins, please retest this

ec2/spark_ec2.py
```diff
@@ -178,7 +178,7 @@ def launch_cluster(conn, opts, cluster_name):
     master_group.authorize('tcp', 50030, 50030, '0.0.0.0/0')
     master_group.authorize('tcp', 50070, 50070, '0.0.0.0/0')
     master_group.authorize('tcp', 60070, 60070, '0.0.0.0/0')
-    master_group.authorize('tcp', 33000, 33010, '0.0.0.0/0')
+    master_group.authorize('tcp', 3030, 3030, '0.0.0.0/0')
```
shivaram: On a similar note, allow some wiggle room here in case you want to launch two `spark-shell` at the same time on EC2 ?

mateiz: OK, done.

mateiz:
Merging this since it passed the tests; thanks for the feedback.

pwendell:
@mateiz  - Could you also update `docs/configuration.md` which currently lists the default as 33000?

mateiz:
Good catch; done.

PR #850, opened 2013-08-20
Remove -optimize from compiler flags

mateiz:
According to several reports (e.g. http://stackoverflow.com/questions/6704233/what-does-the-optimise-scala-compiler-flag-do), this flag doesn't do a lot in practice, and indeed on our own performance benchmarks (http://github.com/amplab/spark-perf) it makes no difference. Removing it makes the build about 50% faster.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -387,7 +387,7 @@
             <useZincServer>true</useZincServer>
             <args>
               <arg>-unchecked</arg>
-              <arg>-optimise</arg>
+              <arg>-deprecation</arg>
             </args>
             <jvmArgs>
               <jvmArg>-Xms64m</jvmArg>
--- a/project/SparkBuild.scala
+++ b/project/SparkBuild.scala
@@ -67,7 +67,7 @@ object SparkBuild extends Build {
     organization := "org.spark-project",
     version := "0.8.0-SNAPSHOT",
     scalaVersion := "2.9.3",
-    scalacOptions := Seq("-unchecked", "-optimize", "-deprecation"),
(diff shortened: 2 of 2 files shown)


markhamstra:
The optimizer in 2.11.x will be quite a bit different, though, so we should remember to re-evaluate this issue once Spark is using those Scala versions.

mateiz:
Yeah, we'll definitely revisit it for 2.11.

pwendell:
I saw zero difference in the regression tests so I'm +1 to remove this. Looks like it needs to be brought up to date though.

wannabeast:
I have seen the flag optimize the Java byte code, in Spark, but I can't recall where. The JIT compiler may be performing the same optimization, keeping you from seeing any performance difference. It's probably visible as a reduction of the JAR size.

mateiz:
I see, IMO let's wait until 0.8.1 then. No point doing this right now. But I do like the idea of reducing the JAR file size.

wannabeast:
It could increase the JAR size, too, like if it does loop unrolling. You'd just have to check it.

I remember now where I saw its effect: I was looking to improve performance of your `AppendOnlyMap` by extracting common subexpressions like `2 * pos`, but I found that `-optimise` did that already, putting it into a (JVM) local variable.

aarondav:
(We're now on the 0.8.1 branch so I vote to merge this as soon as convenient, because build times are pretty annoying right now.)

rxin:
Submitted a new pull request to the asf git repo: https://github.com/apache/incubator-spark/pull/17

PR #851, opened 2013-08-21
Removed meaningless types

markhamstra:
This PR doesn't change the functionality of the code, but it does remove the misleading implication that a `MutablePair[T1,T2]` can only be equal to a `MutablePair[T3,T4]` if `T1` is the same as `T3` and `T2` is the same as `T4`.  Because of type erasure, there is no such constraint on the types.

If we actually want to disallow equality for MutablePairs of different types, then this is one way to accomplish that:

``` scala
override def canEqual(that: Any): Boolean = that match {
  case m: MutablePair[_,_] => this.getClass.isAssignableFrom(m.getClass)
  case _ => false
}
```

Note that this results in `MutablePair(28, 42) != MutablePair(28L, '*')` and other such comparisons that would currently be `true`, and that we may like to keep that way.


rxin:
Thanks.

PR #852, opened 2013-08-21
Remove quotes from CLASSPATH in the run script

tobert:
The quotes introduced in b4905c383bf24143966d8b2d0ae69ef9ee7c3b0e
break spark-shell and probably other scripts that rely on the
'run' script. This patch removes them. Tested on a couple
Linux machines.

### DIFF
--- a/run
+++ b/run
@@ -165,7 +165,7 @@ else
   EXTRA_ARGS="$JAVA_OPTS"
 fi
 
-command="$RUNNER -cp \"$CLASSPATH\" $EXTRA_ARGS $@"
+command="$RUNNER -cp $CLASSPATH $EXTRA_ARGS $@"
 if [ "$SPARK_PRINT_LAUNCH_COMMAND" == "1" ]; then
   echo "Spark Command: $command"
   echo "========================================"


velvia:
To be specific, this breaks distributions created with "make-distribution.sh" or anything else that relies on having "*" in the CLASSPATH.

pwendell:
Are these equivalent?

```
command="$RUNNER -cp $CLASSPATH $EXTRA_ARGS $@"
exec $command
```

and 

```
exec "$RUNNER" -cp "$CLASSPATH" $EXTRA_ARGS "$@"
```

I was trying to make sure things were consistent with how it launched before b4905c3. But I'm not sure if the proposed fix is equivalent either... is it? Whatever we introduce needs to be equivalent to the second one here.

tobert:
"$CLASSPATH" and $CLASSPATH are in effect the same as long as there is no whitespace in the classpath.  Keeping the quotes might be a good idea just to prevent problems. Shall I send a new PR?

pwendell:
Hey Al - ya I'd like to keep the quotes because I think it's designed to
just be defensive in case people put weird stuff in there. Note that
earlier we actually put quotes around RUNNER and $@ also - I actually
should have ported those as well in b4905c3 but it looks like i forgot.

mateiz:
We actually want to support spaces in arguments and in the path to runner and in the arguments ("$@"), so it seems that exec $COMMAND broke that. It doesn't behave the same as exec "$RUNNER" "$@". For example:

```
matei@mbp:~$ text="b   c"
matei@mbp:~$ runner="echo"
matei@mbp:~$ "$runner" "$text"
b   c
matei@mbp:~$ command="$runner $text"
matei@mbp:~$ $command
b c
```

ScrapCodes:
" note that variables which may intentionally be empty must not
 be quoted: otherwise an empty string will appear as a command line
 argument, and java will think that is the program to run. "

Source: https://github.com/scala/scala/blob/master/test/partest#L133

So ideally, anything passed as arguments to java should not be quoted if it is expected to be empty/unset.

mateiz:
I think I've fixed this issue as part of this pull request: https://github.com/mesos/spark/pull/857. Although it's still under review, do you mind testing it? With this PR you have to do sbt/sbt assembly instead of package to build Spark.

mateiz:
I'm closing this because #857 deals with it. Thanks for bringing up the issue though!

PR #853, opened 2013-08-21
Implementing SPARK-838: Add DoubleRDDFunctions methods to PySpark

AndreSchumacher:


python/pyspark/statcounter.py
```diff
@@ -0,0 +1,115 @@
...
+        self.m2 = 0.0  # Running variance numerator (sum of (x - mean)^2)
+
+        for v in values:
+            self.merge(v)
+            
+    # Add a value into this StatCounter, updating the internal statistics.
+    def merge(self, value):
+        # check if value is numeric
+        # (based on http://stackoverflow.com/questions/4187185/how-can-i-check-if-my-python-object-is-a-number)
+        if isinstance(value, numbers.Number):
```
mateiz: Hey Andre, this check looks expensive to do on every value; I'd rather remove it and rely on a Python exception here

python/pyspark/rdd.py
```diff
@@ -357,6 +358,63 @@ def count(self):
...
+        2.0
+        """
+        return self.stats().mean()
+
+    def variance(self):
+        """
+        Compute the variance of this RDD's elements.
+
+        >>> sc.parallelize([1, 2, 3]).variance()
+        0.6666666666666666
```
mateiz: It might be better to pick values that give an integer variance so this doesn't get affected by implementation-dependent rounding

python/pyspark/rdd.py
```diff
@@ -357,6 +358,63 @@ def count(self):
...
+        """
+        return self.stats().variance()
+
+    def stdev(self):
+        """
+        Compute the standard deviation of this RDD's elements.
+
+        >>> sc.parallelize([1, 2, 3]).stdev()
+        0.816496580927726
```
mateiz: Same here on the rounding / printing

mateiz:
Actually as you see, on https://amplab.cs.berkeley.edu/jenkins/job/SparkPullRequestBuilder/683/console, the test failures are because of those floating-point numbers. Would be good to pick some values where the results are integers.

python/pyspark/statcounter.py
AndreSchumacher: OK, good idea.

python/pyspark/rdd.py
AndreSchumacher: I was expecting the rounding to be consistent among implementations but it seems that was a too strong assumption. Will choose different examples.

AndreSchumacher:
Thanks, Matei! Will change the examples.

python/pyspark/rdd.py
JoshRosen: I think that Python 2.7 changed the float rounding behavior.  From http://docs.python.org/dev/whatsnew/2.7.html:

> Conversions from long integers and regular integers to floating point now round differently, returning the floating-point number closest to the number. This doesn’t matter for small integers that can be converted exactly, but for large numbers that will unavoidably lose precision, Python 2.7 now approximates more closely.

JoshRosen:
You could maybe use [doctest.ELLIPSIS](http://docs.python.org/2/library/doctest.html#doctest.ELLIPSIS) to only match the first few decimal places.

AndreSchumacher:
Thanks Josh, doctest.ELLIPSIS seems to work nicely at least on my local machine. Pushed the changes, let's see what Jenkins says.

mateiz:
Cool, it passes! I'm going to merge this then. Feel free to resolve the issue on JIRA.

PR #854, opened 2013-08-21
Synced sbt and maven builds to use the same dependencies, etc.

markhamstra:


mateiz:
Looks good; thanks for catching these discrepancies.

PR #855, opened 2013-08-21
Update build docs

jey:
Documented the new build and made it a bit user-friendlier by allowing configuration to happen in `conf/spark-env.sh`


README.md
```diff
@@ -43,10 +43,48 @@ locally with one thread, or "local[N]" to run locally with N threads.
...
+For convenience, these variables may also be set through the `conf/spark-env.sh` file
+described below.
+
+When developing a Spark application, specify the Hadoop version by adding the
+"hadoop-client" artifact to your project's dependencies. For example, if you're
+using Hadoop 0.20.203 and build your application using SBT, add this to
+`libraryDependencies`:
+
+    // "force()" is required because "0.20.203" is less than Spark's default of "1.0.4"
+    "org.apache.hadoop" % "hadoop-client" % "0.20.203" force()
```
rxin: hadoop-client doesn't have a 0.20 version, does it?

README.md
```diff
@@ -43,10 +43,48 @@ locally with one thread, or "local[N]" to run locally with N threads.
...
 Spark uses the Hadoop core library to talk to HDFS and other Hadoop-supported
-storage systems. Because the HDFS API has changed in different versions of
+storage systems. Because the protocols have changed in different versions of
 Hadoop, you must build Spark against the same version that your cluster runs.
-You can change the version by setting the `HADOOP_VERSION` variable at the top
-of `project/SparkBuild.scala`, then rebuilding Spark.
+You can change the version by setting the `SPARK_HADOOP_VERSION` environment
+when building Spark.
+
+For Apache Hadoop versions 1.x, 0.20.x, Cloudera CDH MRv1, and other Hadoop
```
mateiz: Here's another mention of 0.20.x

jey: fixed

README.md
```diff
@@ -16,7 +16,7 @@ Spark requires Scala 2.9.3 (Scala 2.10 is not yet supported). The project is
 built using Simple Build Tool (SBT), which is packaged with it. To build
 Spark and its example programs, run:
 
-    sbt/sbt package
+    sbt/sbt package assembly
```
mateiz, line 19: Sorry, noticed one other thing -- it's enough to say sbt/sbt assembly, no need to do package assembly. You can fix it in the instructions below as well.

README.md
```diff
@@ -43,10 +43,48 @@ locally with one thread, or "local[N]" to run locally with N threads.
...
+For convenience, these variables may also be set through the `conf/spark-env.sh` file
+described below.
+
+When developing a Spark application, specify the Hadoop version by adding the
+"hadoop-client" artifact to your project's dependencies. For example, if you're
+using Hadoop 1.0.1 and build your application using SBT, add this to
+`libraryDependencies`:
+
+    // "force()" is required because "1.0.1" is less than Spark's default of "1.0.4"
+    "org.apache.hadoop" % "hadoop-client" % "1.0.1" force()
```
mateiz: IMO let's not mention `force()` because 1.0.4 is backward-compatible with other 1.0.x versions, and that's the earliest hadoop-client version.

mateiz: Maybe update this line to say 1.2.1 for example to make it more interesting

README.md
```diff
@@ -43,10 +43,48 @@ locally with one thread, or "local[N]" to run locally with N threads.
...
+
+If your project is built with Maven, add this to your POM file's `<dependencies>` section:
+
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-client</artifactId>
+      <!-- the brackets are needed to tell Maven that this is a hard dependency on version "1.0.1" exactly -->
+      <version>[1.0.1]</version>
```
mateiz: Ditto on this one

conf/spark-env.sh.template
```diff
@@ -19,6 +19,14 @@
...
 #   Examples of app-specific options : -Dspark.serializer
 
+# Hadoop version to build against
+# export SPARK_HADOOP_VERSION=1.0.4
+
+# Uncomment this when using a Hadoop version with YARN
+# export SPARK_WITH_YARN=true
+
+# Extra arguments to pass to `java` when building with SBT
+# export EXTRA_BUILD_ARGS="$EXTRA_BUILD_ARGS -Xmx8g"
```
mateiz: Let's not put these in spark-env.sh; I'd prefer if people specify them only on the command line, even though it technically works in spark-env.sh.

sbt/sbt
```diff
@@ -17,12 +17,15 @@
...
-  EXTRA_ARGS="-Djava.library.path=$MESOS_HOME/lib/java"
-fi
-
 export SPARK_HOME=$(cd "$(dirname $0)/.." 2>&1 >/dev/null ; pwd)
 export SPARK_TESTING=1  # To put test classes on classpath
 
-java -Xmx1200m -XX:MaxPermSize=350m -XX:ReservedCodeCacheSize=128m $EXTRA_ARGS -jar $SPARK_HOME/sbt/sbt-launch-*.jar "$@"
+if [ -f "$SPARK_HOME/conf/spark-env.sh" ]; then
+  source "$SPARK_HOME/conf/spark-env.sh"
+fi
```
mateiz: As mentioned above, I wouldn't read this here -- let's minimize the number of ways to configure the build.

mateiz:
Made a few other small comments; other than that, it looks good.

sbt/sbt
rxin: Also while you are at this, can you bump up the reserved code cache size to 256MB?

README.md
jey, line 19: Unfortunately, `assembly` does not actually invoke `package` at present:

```
jey@kallisti:~/work/spark$ find . -name spark-\*.jar | xargs rm
jey@kallisti:~/work/spark$ find . -name spark-\*.jar
jey@kallisti:~/work/spark$ sbt/sbt assembly 2>&1 | tail -n1
[success] Total time: 27 s, completed Aug 21, 2013 5:56:19 PM
jey@kallisti:~/work/spark$ find . -name spark-\*.jar
./repl/target/spark-repl-assembly-0.8.0-SNAPSHOT.jar
./yarn/target/spark-yarn-assembly-0.8.0-SNAPSHOT.jar
./streaming/target/spark-streaming-assembly-0.8.0-SNAPSHOT.jar
./core/target/spark-core-assembly-0.8.0-SNAPSHOT.jar
```

mateiz, line 19: Oh, got it. Let's leave it like this for now then.. I'll see if I can make the examples use the assembly instead of the package later. The only reason we needed package is for SPARK_EXAMPLES_JAR.

sbt/sbt
jey: How about if we source a `conf/spark-build.sh` instead? I personally would prefer to not always have to type out all the environment variables every time I run sbt.

mateiz: I'd rather not add this for now; let's see how people use it. The reason is that most people will never rebuild Spark after the first time.

jey: Agreed; this would be for the benefit of Spark developers.

jey:
I've made the requested changes. I removed the ability to configure the build in `conf/spark-env.sh` for an expedient merge, but I also think this should be revisited soon. Being able to put the configuration in a flie would be good for users and developers: the configuration can just be written down once, much like `configure`-script based build systems. This is good for cluster admins so that builds are repeatable and self-documenting, and good for Spark devs for the same reasons in addition to saving the work of typing the same configuration parameters repeatedly.

mateiz:
Okay, thanks. I'll think more about the build issue -- basically I'm worried about having a really confusing array of config files. People can also write their own script to build. But I do understand the benefit of more self-documenting builds.

One other thing I may do is change the assembly name to include the Hadoop version -- that will make it easier to see when you're running with the wrong one.

PR #856, opened 2013-08-22
Re-add removed dependency to fix build under Hadoop 0.23.9

jey:
This kludge, removed in ff6f1b0500a9a75b61c225f3dc80d4026e98bcd5, is actually needed to convince SBT into successfully building with Hadoop 0.23.9.

The root cause is some differences in error handling between Ivy2 and Maven when parsing internally-inconsistent POM files such as the one for commons-daemon.

### DIFF
--- a/project/SparkBuild.scala
+++ b/project/SparkBuild.scala
@@ -172,6 +172,7 @@ object SparkBuild extends Build {
       "log4j" % "log4j" % "1.2.17",
       "org.slf4j" % "slf4j-api" % slf4jVersion,
       "org.slf4j" % "slf4j-log4j12" % slf4jVersion,
+      "commons-daemon" % "commons-daemon" % "1.0.10",  // workaround for bug HADOOP-9407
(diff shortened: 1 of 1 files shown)


markhamstra:
Sorry, Jey.  Which hadoop versions actually needs this?

And the meta question: Do we want to continue to support those versions?

jey:
0.23.9

jey:
Also affects 2.0.4 as documented in https://issues.apache.org/jira/browse/HADOOP-9407

mateiz:
Yeah, unfortunately I think we have to keep this in here. As far as I can tell, 0.23.9 is still the latest stable release on the 0.23 branch.

markhamstra:
How about adding a comment and link to the JIRA issue so that we are not tempted to remove it again?

jey:
@markhamstra: Good idea, done.

mateiz:
Cool, merging this.

PR #857, opened 2013-08-23
Change build and run instructions to use assemblies

mateiz:
This PR makes Spark invocation saner by using an assembly JAR to find all of Spark's dependencies instead of adding all the JARs in lib_managed. It also packages the examples into an assembly and uses that as SPARK_EXAMPLES_JAR. Finally, it replaces the old "run" script with two better-named scripts: "run-examples" for examples, and "spark-class" for Spark internal classes (e.g. REPL, master, etc). This is also designed to minimize the confusion people have in trying to use "run" to run their own classes; it's not meant to do that, but now at least if they look at it, they can modify run-examples to do a decent job for them.

As part of this, Bagel's examples are also now properly moved to the examples package instead of bagel.

This is still a work in progress, but I'm putting it up for comments. It's not yet fully tested with Maven, and there's still a TODO to make Maven's examples build as an assembly. In addition, I'd like to make the examples assemblies not include spark-core and all its dependencies, by listing spark-core as "provided". This seems easy to do in Maven but tougher in SBT.


mateiz:
Forgot to add -- this currently has a bug with PySpark too because py4j.jar is in python/lib instead of core/lib. I'll fix that soon.

mateiz:
Jenkins, retest this please

mateiz:
Jenkins, retest this please

mateiz:
Jenkins, retest this please

mateiz:
Jenkins, retest this please

mateiz:
Jenkins, retest this please

pwendell:
Matei, there are several examples in streaming that still reference the `run` script in the javadoc. It would be good to update those to refer to the new scripts. To find these, just do this in `$SPARK_HOME`: `grep "\./run" * -R`.

mateiz:
Jenkins, retest this please

mateiz:
Jenkins, retest this please

mateiz:
:scream_cat: finally! Now to break it with a few more commits...

mateiz:
Alright, I've now updated this to work with Maven too. @c0s, @markhamstra, can you look at this for the Maven changes? Basically the changes are the following:
- Maven will always build an assembly JAR from the "assembly" project and another one from "examples", to match SBT
- These are the only things added to the classpath (with the examples added only if we're in `run-example`)
- The BigTop assembly that @c0s added is now built with `-Pbigtop-dist`
- The `repl-bin` assembly that ClearStory added can still be built with `-Prepl-bin`; I think we can deprecate this and add another way to make Debian packages in the future, but I wanted to keep it as is for now since I don't want to expand the scope of packaging changes in 0.8.

I've tested that the unit tests, pyspark, spark-shell etc all work with Maven builds.

ghost:
Will give it a fresh look in the morning - thanks a lot for crossing all t's!

mateiz:
Yeah, just to be clear, the examples are _not_ going to be on the classpath for user jobs, but I wanted to make them and their dependencies into a big JAR so they can add themselves there properly when you submit them to a cluster (as ordinary Spark apps). Before this patch, the examples _were_ on the classpath all the time, which is bad, because they brought in dependencies on finicky stuff like HBase.

mateiz:
I'm going to merge this because a bunch of other things depend on it. If you try it on Maven and see any problems, let me know.

PR #858, opened 2013-08-23
Ensure that scripts correctly follow symbolic links to determine the framework directory

ianbuss:
Often it may be desirable for packaged versions to create symbolic links to the executor scripts. As they currently stand, the scripts will mis-identify the framework directory using `dirname` to be the directory in which the symlink resides.

The pull request uses a python one-liner as `readlink`'s behaviour can be platform dependent.

### DIFF
--- a/pyspark
+++ b/pyspark
@@ -18,7 +18,7 @@
 #
 
 # Figure out where the Scala framework is installed
-FWDIR="$(cd `dirname $0`; pwd)"
+FWDIR=$(python -c "import os,sys; print os.path.dirname(os.path.realpath(\"$0\"))")
 
 # Export this as SPARK_HOME
 export SPARK_HOME="$FWDIR"
--- a/run
+++ b/run
@@ -20,7 +20,7 @@
 SCALA_VERSION=2.9.3
 
 # Figure out where the Scala framework is installed
-FWDIR="$(cd `dirname $0`; pwd)"
+FWDIR=$(python -c "import os,sys; print os.path.dirname(os.path.realpath(\"$0\"))")
 
 # Export this as SPARK_HOME
 export SPARK_HOME="$FWDIR"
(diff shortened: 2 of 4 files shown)


jey:
Jenkins, please test this.

pwendell:
Jenkins, test this please

pwendell:
Unfortunately we can't assume the driver has python installed. Could you give an example of a case where the current approach fails?

ianbuss:
An example use case: packaging conventions often result in symbolic links to executables being placed in `/usr/bin` or similar, sometimes via `/etc/alternatives`. A straight symlink in these cases breaks spark's executor scripts. I suppose it might be argued it should be up to the package author for the particular system to provide a wrapper around the executor scripts to call the script correctly. In fact, I've just discovered this is the approach taken by those adding Spark to Bigtop: https://github.com/apache/bigtop (see lines 148-164 of https://github.com/apache/bigtop/blob/master/bigtop-packages/src/common/spark/install_spark.sh in particular).

I also agree that it shouldn't be assumed that python is installed and that, if any, a "pure bash" solution is desirable.

In view of these points it might be worth dropping this PR and folding it into any work done on SPARK-665.

mateiz:
I actually don't think we should be running spark-executor through the system PATH. Jobs will be configured with a path to Spark. If you do want to run them through the system PATH, maybe just install a script there that cds into the right directory and calls our existing spark-executor? That way we don't need Python and such.

mateiz:
(basically the script in /usr/bin would just say `<SPARK_HOME>/spark-executor "$@"`)

ianbuss:
Agreed. Looking into it further, packages built around other frameworks (e.g. Hadoop) also provide wrapper scripts. Am closing the PR. Thanks for the comments.

PR #860, opened 2013-08-23
Fix IDE project generation under SBT

jey:
This fixes the problem with generation of IDEA and Eclipse projects described in https://github.com/mesos/spark/pull/750#issuecomment-23078354


mateiz:
Looks good, thanks.

PR #861, opened 2013-08-23
Pyspark sampling function

AndreSchumacher:
This is more of a draft than an actual Pull Request. In the JIRA Josh was proposing to fall back to Python's native random in case NumPy is not installed, while preserving the results. I find that hard to implement given that the random generators used may be very different, although one could emulate drawing values from a Poisson distribution via random.expovariate() I guess. So I was wondering whether we could discuss this here.

Also note that takeSample() is still missing. I wanted to discuss sample() first before coming to that.
Update: added takeSample() based on RDD.scala

JIRA: https://spark-project.atlassian.net/browse/SPARK-651

### DIFF
--- a/python/pyspark/rdd.py
+++ b/python/pyspark/rdd.py
@@ -21,6 +21,7 @@ from collections import defaultdict
 from itertools import chain, ifilter, imap, product
 import operator
 import os
+import sys
 import shlex
 from subprocess import Popen, PIPE
 from tempfile import NamedTemporaryFile
@@ -32,6 +33,7 @@ from pyspark.serializers import batched, Batch, dump_pickle, load_pickle, \
 from pyspark.join import python_join, python_left_outer_join, \
     python_right_outer_join, python_cogroup
 from pyspark.statcounter import StatCounter
+from pyspark.rddsampler import RDDSampler
 
 from py4j.java_collections import ListConverter, MapConverter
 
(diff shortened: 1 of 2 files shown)


mateiz:
Hey Andre,

I agree with Josh that we should fall back on the default if numpy isn't available. I suggest to test whether numpy is available on the master first. If it is, always try to use it on the workers -- otherwise, always use the system default. If someone installs numpy on some workers but not others, or on the master but not workers, that's their fault.

AndreSchumacher:
OK, falling back on the default random.py when numpy is unavailable sounds good (as long as people don't expect the results to be the same).

Is there any more practical way to pass settings from master to workers than serializing it into its input stream, using SparkEnv maybe?

Thanks for the comments!

AndreSchumacher:
So I refactored things a bit. The whole drawing of random numbers now happens inside RDDSampler, which determines which of the corresponding functions to call. Based on a little experiment I did comparing the two Poisson generators they seem to come close.

If I understand correctly the RDDSampler should actually be created on the master when it's constructor is invoked. From them on the decision which of the libraries to use should be consistent among the workers. Would be great if someone who knows the code better could comment on that.

python/pyspark/rdd.py
```diff
@@ -165,14 +167,61 @@ def distinct(self):
...
+        fraction = 0.0
+        total = 0
+        multiplier = 3.0
+        # TODO: it seems expensive to start a job just to initialize this variable but it may
+        # not be possible to avoid it (Spark does the same)
+        initialCount = self.count()
+        maxSelected = 0
+
+        if (num < 0):
+            raise Exception("Negative number of elements requested")
```
JoshRosen: I'd raise [ValueError](http://docs.python.org/2/library/exceptions.html#exceptions.ValueError) here.

python/pyspark/rdd.py
```diff
@@ -165,14 +167,61 @@ def distinct(self):
...
+        else:
+            fraction = multiplier * (num + 1) / initialCount
+            total = num
+
+        samples = self.sample(withReplacement, fraction, seed).collect()
+    
+        # If the first sample didn't turn out large enough, keep trying to take samples;
+        # this shouldn't happen often because we use a big multiplier for their initial size
+        while len(samples) < total:
+            if seed > sys.maxint -2:
```
JoshRosen: There should be an extra space next to this minus sign.

AndreSchumacher:
Disabled the doctest tests for sample() and takeSample() for now since they are bound to fail in numpy-installed versus non-numpy cases.

AndreSchumacher:
Jenkins, retest this please

mateiz:
Hey Andre, I tried this out and it looks good. Thanks!

PR #863, opened 2013-08-26
Adding linear regression and refactoring Ridge regression to use SGD

shivaram:
Based on mesos/spark #830 contributed by @etrain. In addition to that pull request this includes a new unit test for ridge regression, centering variables before training in Lasso, Ridge and some minor style fixes.


aht:
I'm curious why you decided to completely remove RidgeRegression normal equations solver. It does not scale for large features set but it does work very well on tens or even hundreds of features. Plus we can compute analytically:
- generalized CV error
- standard error of coefficients
- p-values if we so desire

Is it possible to keep both version?

atalwalkar:
We decided to do this for simplicity / scalability reasons.  As you
mentioned, the 2nd-order method doesn't scale to larger feature sets, and
for the first release, we wanted to keep the structure of all of our
algorithms consistent.  After the first release we should strongly consider
(re) adding this normal equations solver back.

etrain:
Thanks for cleaning up the formatting issues Shivaram, as well as catching the centering/scaling thing. We should revisit this when we hit the sparse case, but for the dense case we're looking at now, I think this is the right thing to do.

PR #864, opened 2013-08-26
Revert json library change

rxin:
@shivaram

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -107,6 +107,10 @@
       <artifactId>scala-library</artifactId>
     </dependency>
     <dependency>
+      <groupId>net.liftweb</groupId>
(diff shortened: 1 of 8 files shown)


shivaram:
LGTM - One minor comment - Could you add a unit-test that actually checks if the JSON output is valid ?

pwendell:
A poor man's version of that test would be to just call `parse`() from LiftJson and see if it throws an exception. It would be good to check if this actually throws exceptions for invalid Json.

https://github.com/lift/lift/blob/master/framework/lift-base/lift-json/src/main/scala/net/liftweb/json/JsonParser.scala#L45

rxin:
I agree it'd be good to do that, but let's do it in a separate pr. This should go in asap since it is fixing something that is broken.

PR #865, opened 2013-08-26
Spark on Yarn should use yarn approved directories for spark.local.dir and tmp

tgravescs:
Spark on Yarn should use the directories that yarn allocates for the application rather then the user specified ones in spark.local.dir or /tmp.  This way Yarn can manage and cleanup the directories as needed and the user won't have to know how the yarn cluster is configured.

I also noticed the docs referred to the old and removed config --user so I removed that while I was updating.

### DIFF
--- a/core/src/main/scala/spark/executor/Executor.scala
+++ b/core/src/main/scala/spark/executor/Executor.scala
@@ -60,6 +60,13 @@ private[spark] class Executor(
     System.setProperty(key, value)
   }
 
+  // If we are in yarn mode, systems can have different disk layouts so we must set it
+  // to what Yarn on this system said was available. This will be used later when SparkEnv
+  // created.
+  if (java.lang.Boolean.valueOf(System.getProperty("SPARK_YARN_MODE", System.getenv("SPARK_YARN_MODE")))) {
+    System.setProperty("spark.local.dir", getYarnLocalDirs())
+  }
+
   // Create our ClassLoader and set it on this thread
   private val urlClassLoader = createClassLoader()
   private val replClassLoader = addReplClassLoaderIfNeeded(urlClassLoader)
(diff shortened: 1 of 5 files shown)


mateiz:
Jenkins, this is ok to test

mateiz:
Good catch, thanks Tom!

tgravescs:
Thanks Matei.  I just thought of another issue though.  I think the spark.local.dir and DiskStore rely on a homogenous disk setup on all the nodes. This isn't the case on yarn. We can have systems with 12 disk, some with 6 disk, etc.. all in the same cluster.   Its going to use the list from the node the application master starts on, which isn't necessarily the same as the ones the workers start on.  I'll have to look at that more.

mateiz:
Ah, true. I guess you could reset it in Executor? Or maybe there's a way to do it in the launch command on each node.

tgravescs:
Yeah I think that makes sense. I'll make sure it gets set when we launch the worker (for the node its on) and then have the Executor use the local version rather then the system properties obtained from the RegisteredExecutor call.

mateiz:
Yup, you could simply override it there. If a YARN_WORK_DIR or whatever variable is set, use that.

tgravescs:
I changed the Executor so it overrides spark.local.dirs if in yarn mode.  I'm not really happy about having to duplicate the getLocalDirs code or pollute Executor with YARN variables but I didn't see any other good spot to put that.  Let me know if you have better ideas.

core/src/main/scala/spark/executor/Executor.scala
```diff
@@ -107,6 +114,24 @@ private[spark] class Executor(
...
+    var localDirs = System.getenv("LOCAL_DIRS")
+    val yarnLocalSysDirs = Option(System.getenv("YARN_LOCAL_DIRS"))
+    yarnLocalSysDirs match {
+        case Some(s) => localDirs = s
+        case None => {
+          if ((localDirs == null) || (localDirs.isEmpty())) {
+            throw new Exception("Yarn Local dirs can't be empty")
+          }
+        }
+    }
```
mateiz, line 128: You can write this more idiomatically as

```
var localDirs = Option(System.getenv("YARN_LOCAL_DIRS"))
  .getOrElse(Option(System.getenv("LOCAL_DIRS"))
  .getOrElse("")
```

Then do your if test and return it. (Technically you can also do `.getOrElse(throw new Exception(...))` but that looks weirder.)

mateiz:
This fix sounds fine for now -- we may refactor it later if we come up with a common way to do it across cluster management systems (e.g. if we find that the Spark standalone mode or Mesos run into the same problem, which is likely to happen on heterogeneous hardware). I made one comment on simplifying it a bit.

tgravescs:
Thanks for the comments.  Updated the patch to refactor.

core/src/main/scala/spark/executor/Executor.scala
```diff
@@ -60,6 +60,13 @@ private[spark] class Executor(
     System.setProperty(key, value)
   }
 
+  // If we are in yarn mode, systems can have different disk layouts so we must set it
+  // to what Yarn on this system said was available. This will be used later when SparkEnv
+  // created.
+  if (java.lang.Boolean.valueOf(System.getProperty("SPARK_YARN_MODE", System.getenv("SPARK_YARN_MODE")))) {
```
mateiz: Sorry, one more thing -- why would this ever be set as a system property instead of an environment var? Right now you're checking for both.

tgravescs: I just copied that from the existing checks in SparkEnv to keep it consistent. It looks like the Yarn client sets both.  It sets the property before creating the client and then the env later to be passed to the app master and workers.  I would think the env would be enough.  I can look into it more, but perhaps that is a separate issue?

tgravescs: thinking about this more since this is in the Executor and on yarn that will only be set via the ENV variable I will change this to only look at the env variable.  I will also look more to see why it uses the property in some cases but I will do that under separate issue if that is ok with you?

mateiz:
Sure, sounds good. I've merged this in.

PR #866, opened 2013-08-27
Allow users to push s3 credentials to slaves

pwendell:
This allows users to push their AWS credentials into the Hadoop config files when creating a cluster. It is disabled by default.


pwendell:
This corresponds to this `spark-ec2` PR. It can be merged first though without any conflicts:
https://github.com/mesos/spark-ec2/pull/17/files

mateiz:
Patrick, will these also be visible to Spark? You may want to export them in spark-env.sh on the AMI. I don't think we add Hadoop's config files onto Spark's classpath by default, but if you set these guys in spark-env.sh Spark will use them.

pwendell:
I'm closing this and going to re-submit later with a different approach based on just setting this as env varaibles (which spark will now pickup in it's hadoop configuration).

PR #867, opened 2013-08-27
Spark on Yarn allow users to specify environment variables 

tgravescs:
This adds the ability for spark on yarn users to specify environment variables.  These are passed to the launched processes (master/executors).  One of the main use cases I'm looking at for this is for the user to be able to choose  which java version and architecture (32 bit or 64bit) to use by specifying the JAVA_HOME variable.


mateiz:

mateiz:
Merging this in -- thanks Tom.

PR #869, opened 2013-08-28
PySpark: implementing subtractByKey(), subtract() and keyBy()

AndreSchumacher:


AndreSchumacher:
Jenkins, test this please.

python/pyspark/rdd.py
```diff
@@ -754,6 +754,43 @@ def cogroup(self, other, numPartitions=None):
...
+        Return each (key, value) pair in C{self} that has no pair with matching key
+        in C{other}.
+
+        >>> x = sc.parallelize([("a", 1), ("b", 4), ("b", 5), ("a", 2)])
+        >>> y = sc.parallelize([("a", 3), ("c", None)])
+        >>> sorted(x.subtractByKey(y).collect())
+        [('b', 4), ('b', 5)]
+        """ 
+        filter_func = lambda tpl: len(tpl[1][0]) > 0 and len(tpl[1][1]) == 0
+        map_func = lambda tpl: [(tpl[0], val) for val in tpl[1][0]]
```
mateiz, line 768: FYI, when a lambda takes a tuple, you can actually "pattern-match" on the tuple like this:

```
flter_func = lambda (key, vals): len(vals[0]) > 0 and len(vals[1]) == 0
map_func = lambda (key, vals): [(a, val) for val in vals[0])
```

mateiz:
This looks great.. I made a small comment on a way you can make this kind of code cleaner but I've also merged it since it will be nice to have it in 0.8.

python/pyspark/rdd.py
AndreSchumacher, line 768: Thanks. Good point, that's a lot more readable.

PR #870, opened 2013-08-28
Don't send SIGINT / ctrl-c to Py4J gateway subprocess

JoshRosen:
This addresses [SPARK-885](https://spark-project.atlassian.net/browse/SPARK-885), a usability issue where PySpark's Java gateway process would be killed if the user hit ctrl-c, preventing new jobs from being run.  This behavior was annoying because I tend to reflexively hit ctrl-c if I want to start a blank input line.

This is based on an answer to http://stackoverflow.com/questions/5045771.

### DIFF
--- a/python/pyspark/java_gateway.py
+++ b/python/pyspark/java_gateway.py
@@ -17,6 +17,7 @@
 
 import os
 import sys
+import signal
 from subprocess import Popen, PIPE
 from threading import Thread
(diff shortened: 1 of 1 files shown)


JoshRosen:
Jenkins, retest this please.

mateiz:
Good catch, thanks.

PR #872, opened 2013-08-29
tweaks to the SparkLR example

wannabeast:
I've been using `spark.examples.SparkLR` for performance testing. Just passing along some enhancements to the class.

### DIFF
--- a/examples/src/main/scala/spark/examples/SparkLR.scala
+++ b/examples/src/main/scala/spark/examples/SparkLR.scala
@@ -80,8 +80,8 @@ object SparkLR {
       println("On iteration " + i)
       val zero = Vector.zeros(D)
       val gradient = points.map { p =>
-        (1 / (1 + exp(-p.y * (w dot p.x))) - 1) * p.y * p.x
-      }.fold(zero)(_ += _)
+        ((1 / (1 + exp(-p.y * (w dot p.x))) - 1) * p.y, p.x)
+      }.aggregate(zero)((sum,v) => sum saxpy (v._1,v._2), _ += _)
       w -= gradient
     }
 
--- a/examples/src/main/scala/spark/examples/SparkLR.scala
+++ b/examples/src/main/scala/spark/examples/SparkLR.scala
@@ -21,18 +21,32 @@ import java.util.Random
 import scala.math.exp
 import spark.util.Vector
 import spark._
+import com.esotericsoftware.kryo.Kryo
 
 /**
  * Logistic regression based classification.
  */
 object SparkLR {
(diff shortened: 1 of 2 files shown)


mateiz:
Jenkins, this is ok to test

examples/src/main/scala/spark/examples/SparkLR.scala
```diff
@@ -21,19 +21,33 @@ import java.util.Random
...
 object SparkLR {
-  val N = 10000  // Number of data points
+  case class DataPoint(x: Vector, y: Double)
+
+  class MyRegistrator extends KryoRegistrator {
+    override def registerClasses(kryo: Kryo) {
+      kryo.setRegistrationRequired(true)
+      
+      kryo.register(classOf[scala.collection.mutable.WrappedArray.ofRef[_]])
+      kryo.register(classOf[java.lang.Class[_]])
```
mateiz, line 37: Hey Mike, why do you need to register WrappedArray and Class? Doesn't seem like they'll occur in our data here.

wannabeast, line 37: I came up with that list by adding `setRegistrationRequired`, then fixing each exception that was thrown. So Kryo said they were used; I haven't considered why.

mateiz, line 37: I see. Did Kryo actually make a performance difference for you at all? We are not caching data in serialized form here, so it would only be used to send back task results, but that's just one Vector per partition. I think the WrappedArrays are because we somehow send that back within an array.

examples/src/main/scala/spark/examples/SparkLR.scala
```diff
@@ -59,9 +78,10 @@ object SparkLR {
 
     for (i <- 1 to ITERATIONS) {
       println("On iteration " + i)
+      val zero = Vector.zeros(D)
       val gradient = points.map { p =>
-        (1 / (1 + exp(-p.y * (w dot p.x))) - 1) * p.y * p.x
-      }.reduce(_ + _)
+        ((1 / (1 + exp(-p.y * (w dot p.x))) - 1) * p.y, p.x)
+      }.aggregate(zero)((sum,v) => sum saxpy (v._1,v._2), _ += _)
```
mateiz, line 84: Change the second argument to `sum.saxpy(v._1, v._2)`; we don't use infix notation for methods unless they're operators.

mateiz:
Sorry for the long delay, just made a couple more comments. I'm happy to merge this otherwise. You can either update it here or send a new PR against the apache/incubator-spark repo (the latter would be slightly easier).

PR #874, opened 2013-08-30
Fix removed block zero size log reporting

jerryshao:
`memSize` and `diskSize` of **removed** block will always report as 0.0 B, like this:

```
INFO BlockManagerMasterActor$BlockManagerInfo:Removed input-2-1377741792200 on xxx in memory (size: 0.0 B, free: 2.8 GB)
```

Cause input parameter `memSize` and `diskSize` calculated in BlockManager will always be zero. It would be better to display the actual size of this block when reporting this log.

Thanks
Jerry


rxin:
Jenkins, test this please.

rxin:
Thanks, Jerry!

PR #875, opened 2013-08-30
Fix broken build by removing addIntercept

shivaram:
While the two ML pull requests merged cleanly they changed things in the class hierarchy resulting a build break.
Sorry for the trouble !


shivaram:
Jenkins, test this please

PR #877, opened 2013-08-30
Doc improvements for 0.8

mateiz:
Several updates to URLs, organization of the docs, branding, etc, plus a new hardware provisioning guide.


mateiz:
Jenkins, retest this please

docs/_layouts/global.html
```diff
@@ -61,20 +61,25 @@
...
                                 <li><a href="quick-start.html">Quick Start</a></li>
-                                <li><a href="scala-programming-guide.html">Scala</a></li>
-                                <li><a href="java-programming-guide.html">Java</a></li>
-                                <li><a href="python-programming-guide.html">Python</a></li>
+                                <li><a href="scala-programming-guide.html">Spark in Scala</a></li>
+                                <li><a href="java-programming-guide.html">Spark in Java</a></li>
+                                <li><a href="python-programming-guide.html">Spark in Python</a></li>
+                                <li class="divider"></li>
                                 <li><a href="streaming-programming-guide.html">Spark Streaming</a></li>
+                                <li><a href="mllib-programming-guide.html">MLlib (Machine Learning)</a></li>
```
pwendell: This seems to reference a file that is not included in this PR.

docs/index.md
```diff
@@ -3,50 +3,51 @@ layout: global
...
 # A Note About Hadoop Versions
 
-Spark uses the Hadoop core library to talk to HDFS and other Hadoop-supported
+Spark uses the Hadoop-client library to talk to HDFS and other Hadoop-supported
 storage systems. Because the HDFS protocol has changed in different versions of
-Hadoop, you must build Spark against the same version that your cluster runs.
-You can change the version by setting the `HADOOP_VERSION` variable at the top
-of `project/SparkBuild.scala`, then rebuilding Spark (`sbt/sbt clean compile`).
+Hadoop, you must build Spark against the same version that your cluster uses.
+You can do this by setting the `SPARK_HADOOP_VERSION` variable when compiling:
```
pwendell: It might be good to mention the default Hadoop version here as well (it's in the quickstart, but many people read this page who don't come from the quickstart). In fact, maybe that should be a variable similar to site.SCALA_VERSION in the docs, so we make sure to update it when we bump hadoop defaults.

docs/index.md
```diff
@@ -3,50 +3,51 @@ layout: global
...
+
+    SPARK_HADOOP_VERSION=1.2.1 sbt/sbt assembly
+
+In addition, if you wish to run Spark on [YARN](running-on-yarn.md), you should also
+set `SPARK_YARN`:
```
pwendell: Maybe just say "set `SPARK_YARN` to 'true'". I realize it's evident from the example just below, but I got caught up here wondering what I was supposed to set it _to_ before reading the example. In other places we have boolean variables that need to be set to "1" so I'd err on the side of just making this very explicit.

docs/index.md
```diff
@@ -74,27 +80,27 @@ of `project/SparkBuild.scala`, then rebuilding Spark (`sbt/sbt clean compile`).
...
-* [AMP Camp](http://ampcamp.berkeley.edu/): a two-day training camp at UC Berkeley that featured talks and exercises
-  about Spark, Shark, Mesos, and more. [Videos](http://ampcamp.berkeley.edu/agenda-2012),
+* [Spark Homepage](http://spark.incubator.apache.org)
+* [Mailing Lists](http://spark.incubator.apache.org/mailing-lists.html): ask questions about Spark here
+* [AMP Camps](http://ampcamp.berkeley.edu/): a series of training camps at UC Berkeley that featured talks and
+  exercises about Spark, Shark, Mesos, and more. [Videos](http://ampcamp.berkeley.edu/agenda-2012),
   [slides](http://ampcamp.berkeley.edu/agenda-2012) and [exercises](http://ampcamp.berkeley.edu/exercises-2012) are
   available online for free.
-* [Code Examples](http://spark-project.org/examples.html): more are also available in the [examples subfolder](https://github.com/mesos/spark/tree/master/examples/src/main/scala/spark/examples) of Spark
+* [Code Examples](http://spark.incubator.apache.org/examples.html): more are also available in the [examples subfolder](https://github.com/mesos/spark/tree/master/examples/src/main/scala/spark/examples) of Spark
```
pwendell: It might be good to add the streaming examples here as well. They are at a separate path and it's not obvious how to find them (or that they exist), given the path displayed here.

docs/quick-start.md
```diff
@@ -142,7 +142,13 @@ resolvers ++= Seq(
   "Spray Repository" at "http://repo.spray.cc/")
```
pwendell: Unrelated - but we can delete all references to Spray now.

docs/quick-start.md
```diff
@@ -142,7 +142,13 @@ resolvers ++= Seq(
...
 {% endhighlight %}
 
-Of course, for sbt to work correctly, we'll need to layout `SimpleJob.scala` and `simple.sbt` according to the typical directory structure. Once that is in place, we can create a JAR package containing the job's code, then use `sbt run` to execute our example job.
+If you also wish to read data from Hadoop's HDFS, you will also need to add a dependency on `hadoop-client` for your version of HDFS:
```
pwendell, line 143: Hey, I'm just curious. Does this change mean that existing projects which build against Spark using Maven/SBT will break if they do not specify an HDFS version? It seems to me like that must be the case based on how this works.

If so, we should make sure to say this in the release notes.

mateiz, line 143: No, they actually link to hadoop 1.0.4 by default. This is only to override it with a newer one.

docs/hardware-provisioning.md
```diff
@@ -0,0 +1,70 @@
...
+* If at all possible, run Spark on the same nodes as HDFS. The simplest way is to set up a Spark
+[standalone mode cluster](spark-standalone.html) on the same nodes, and configure Spark and
+Hadoop's memory and CPU usage to avoid interference (for Hadoop, the relevant options are
+`mapred.child.java.opts` for the per-task memory and `mapred.tasktracker.map.tasks.maximum`
+and `mapred.tasktracker.reduce.tasks.maximum` for number of tasks). Alternatively, you can run
+Hadoop and Spark on a common cluster manager like [Mesos](running-on-mesos.html) or
+[Hadoop YARN](running-on-yarn.html).
+
+* If this is not possible, run Spark on different nodes in the same local-area network as HDFS.
+If your cluster spans multiple racks, include some Spark nodes on each rack.
```
pwendell: Depending on how people use spark, this might not be the most prudent advice. For instance, if they tend to load a dataset from HDFS once and then do a bunch of iterative and shuffle-heavy jobs on it, it would be better to just put all of the spark nodes co-resident on one rack. Many of the clusters I've seen so far actually do this - they just have a single rack with a bunch of beefy Spark nodes for this reason. It might be good to either remove this as the "default" recommendation, or give some sense of when it's better to put all spark nodes on the same rack.

docs/hardware-provisioning.md
```diff
@@ -0,0 +1,70 @@
...
+
+* For low-latency data stores like HBase, it may be preferrable to run computing jobs on different
+nodes than the storage system to avoid interference.
+
+# Local Disks
+
+While Spark can perform a lot of its computation in memory, it still uses local disks to store
+data that doesn't fit in RAM, as well as to preserve intermediate output between stages. We
+recommend having **4-8 disks** per node, configured _without_ RAID (just as separate mount points).
+In Linux, mount the disks with the [`noatime` option](http://www.centos.org/docs/5/html/Global_File_System/s2-manage-mountnoatime.html)
```
pwendell, line 33: We may want to recommend `nodiratime` also (or does `noatime` automatically set that?)

docs/configuration.md
```diff
@@ -347,6 +312,32 @@ Apart from these, the following properties are also available, and may be useful
 
 </table>
 
+# Environment Variables
+
+Certain Spark settings can also be configured through environment variables, which are read from the `conf/spark-env.sh`
```
pwendell, line 317: One thing that people (at least I) tend to do is use environment variables when running standalone jobs:

```
SPARK_MEM=X SPARK_CLASSPATH=Y scala ./my.spark.job
```

This line implies that these should only be used in the context of a spark-env.sh file. So are we just deprecating this approach?

This also reminds me it would be nice to have a doc that covers how to run a real production standalone job. Including things like how to deal with dependencies and the recommended way for setting these types of configuration options.

pwendell:
Hey @mateiz - this is really useful! A bunch of small comments...

docs/configuration.md
mateiz, line 317: Yes, that's exactly what I want to do. The environment variables are there from before we had system properties for memory use and SparkContext.addJar.

This doc here is supposed to be the doc on how to run a real production job -- let me know if something doesn't seem appropriate for production.

docs/hardware-provisioning.md
mateiz, line 33: It looks like it implies it: http://lwn.net/Articles/245002/

mateiz:
OK, updated with fixes.

pwendell:
Looks good to me @mateiz I think all of my comments are addressed now.

mateiz:
Alright, I've merged this in now. There may still be other things we want to tweak later though.

PR #878, opened 2013-08-30
Link the Spark UI up to the Yarn UI 

tgravescs:
This is to hook the Spark UI up nicely with the Yarn UI. Yarn requires the links to go through a proxy so I had to add in the base uri stuff. 

somewhat unrelated but found during testing I also added a configuration to allow the user to wait longer for the Spark context to be intialized.


core/src/main/scala/spark/ui/UIUtils.scala
```diff
@@ -25,33 +25,53 @@ import spark.SparkContext
...
+  private[spark] val storageStr = addBaseUri("/storage")
+  private[spark] val stagesStr = addBaseUri("/stages")
+  private[spark] val envStr = addBaseUri("/environment")
+  private[spark] val executorsStr = addBaseUri("/executors")
+  private[spark] val bootstrapMinCssStr = addBaseUri("/static/bootstrap.min.css")
+  private[spark] val webuiCssStr = addBaseUri("/static/webui.css")
+  private[spark] val bootstrapResponsiveCssStr = addBaseUri("/static/bootstrap-responsive.min.css")
+  private[spark] val sortTableStr = addBaseUri("/static/sorttable.js")
+  private[spark] val sparkLogoHdStr = addBaseUri("/static/spark-logo-77x50px-hd.png")
+  private[spark] val sparkLogoStr = addBaseUri("/static/spark_logo.png")
```
mateiz: I'd prefer to get rid of these variables and just use baseUri + "/stages", etc where it's required. It will be more obvious what's going on.

yarn/src/main/scala/spark/deploy/yarn/ApplicationMaster.scala
```diff
@@ -53,21 +55,21 @@ class ApplicationMaster(args: ApplicationMasterArguments, conf: Configuration) e
...
 
-      if (numCore > 0) {
+    //  if (numCore > 0) {
         // do not override - hits https://issues.apache.org/jira/browse/HADOOP-8406
         // TODO: Uncomment when hadoop is on a version which has this fixed.
         // args.workerCores = numCore
-      }
-    }
+    //  }
+    //}
```
mateiz: Why was this changed -- will it work in all YARN versions now (both 0.23.x and 2.0.x)?

mateiz: Oops, actually, I saw you commented it out rather than uncommenting it. Anyway it would be good to know which YARN versions it works in since this seems like an important thing to set.

core/src/main/scala/spark/ui/SparkUI.scala
```diff
@@ -78,7 +78,9 @@ private[spark] class SparkUI(sc: SparkContext) extends Logging {
     server.foreach(_.stop())
   }
 
-  private[spark] def appUIAddress = "http://" + host + ":" + boundPort.getOrElse("-1")
+  private[spark] def appHttpUIAddress = "http://" + appUIAddress
+  private[spark] def appUIAddress = host + ":" + boundPort.getOrElse("-1")
```
mateiz: Instead of having two variables, maybe we can always return the version with http:// and fix it in other places.

mateiz: (Or alternatively add http:// in the one place where you pass this to YARN)

yarn/src/main/scala/spark/deploy/yarn/ApplicationMaster.scala
tgravescs: yeah I actually just commented out the rest of the code that wasn't being used.  The jira references if fixed in hadoop 2.0.2-alpha but not in a 0.23 line.  I can look at pulling it into the 0.23 line if it applies.  Honestly I don't understand what the code is trying to do here. Do you know?  otherwise I'll talk to Mridul.  I don't know why it is linking the number of cores to a yarn config for the  minimum resource capability. This config is somewhat arbitrary as to what your cluster is configured just based on what the cluster admin wants the minimum memory increment to be.

For now I moved the comment up a bit and I'll look into it separately.

core/src/main/scala/spark/ui/SparkUI.scala
tgravescs: Actually yarn doesn't need the http:// so I would have to truncate the http:// off, which seems kind of messy to me.  There is only one place in spark that uses it right now.    I changed it so that this function doesn't have http:// and the one place in spark that uses it adds the http://.. Let me know what you prefer though.

core/src/main/scala/spark/ui/UIUtils.scala
tgravescs: done

tgravescs:
Thanks for the review Matei.
I also merged but it looks like a few of the package renames weren't applied to the spark on yarn so I'm updating those and will post fixes along with the fixes to the comments.

tgravescs:
I updated SPARK_CLASS_REGEX so it works for the examples.  I'm not sure if I got all the permutations though.

core/src/main/scala/org/apache/spark/ui/UIUtils.scala
```diff
@@ -25,38 +25,46 @@ import org.apache.spark.SparkContext
 private[spark] object UIUtils {
   import Page._
 
+  // Yarn has to go through a proxy so the base uri is provided and has to be on all links
+  private[spark] val uiRoot : String = Option(System.getenv("APPLICATION_WEB_PROXY_BASE")).
+                                         getOrElse("")
+
+  def addBaseUri(resource: String = ""): String = {
```
pwendell: For a very short def it's also OK to do:

```
  def addBaseUri(resource: String = "") = uiRoot + resource
```

core/src/main/scala/org/apache/spark/ui/UIUtils.scala
```diff
@@ -25,38 +25,46 @@ import org.apache.spark.SparkContext
...
+    return uiRoot + resource
+  }
+
   /** Returns a spark page with correctly formatted headers */
   def headerSparkPage(content: => Seq[Node], sc: SparkContext, title: String, page: Page.Value)
   : Seq[Node] = {
     val jobs = page match {
-      case Stages => <li class="active"><a href="/stages">Stages</a></li>
-      case _ => <li><a href="/stages">Stages</a></li>
+      case Stages => <li class="active"><a href={"%s/stages".format(UIUtils.addBaseUri())}>Stages</a></li>
```
pwendell: This and several other lines are now over 100 characters. Our style guide is max 100 characters, so please adjust them accordingly. For instance:

```
case Stages => 
  <li class="active"><a href={"%s/stages".format(UIUtils.addBaseUri())}>Stages</a></li>
```

pwendell: Also, for all these things, why not just do:

```
UIUtils.addBaseUri("/stages")
```

It seems a little cleaner than calling "%s/XX".format() every time.

core/src/main/scala/org/apache/spark/ui/UIUtils.scala
pwendell: To make this a bit more explicit, what about calling this `prependBaseUri`?

pwendell:
Hey @tgravescs, I added a few comments in the common UI code. Mostly they are small style things. Thanks!

core/src/main/scala/org/apache/spark/ui/UIUtils.scala
pwendell: Also, if you do the rename proposed above, this would be a bit more clear:

```
UIUtils.prependBaseUri("/stages)
```

core/src/main/scala/org/apache/spark/ui/UIUtils.scala
```diff
@@ -25,38 +25,46 @@ import org.apache.spark.SparkContext
 private[spark] object UIUtils {
   import Page._
 
+  // Yarn has to go through a proxy so the base uri is provided and has to be on all links
+  private[spark] val uiRoot : String = Option(System.getenv("APPLICATION_WEB_PROXY_BASE")).
+                                         getOrElse("")
```
hsaputra: Just small nit comment on style, I think continuation of code should be aligned with 4 spaces indent like:

```
// Yarn has to go through a proxy so the base uri is provided and has to be on all links
private[spark] val uiRoot : String = Option(System.getenv("APPLICATION_WEB_PROXY_BASE")).
    getOrElse("")
```

pwendell: Ah good point, yep we use "indent" rather than "hanging" style. However, we actually use two spaces, as described in the scala styleguide:

http://docs.scala-lang.org/style/indentation.html#line_wrapping

tgravescs:
Thanks for the comments. Updated the patch.

pwendell:
Jenkins, test this please.

pwendell:
Thanks @tgravescs looks good, I'm going to test this and then we can merge it.

pwendell:
Thanks @tgravescs  - I'm merging this!

tgravescs:
Thanks!

core/src/main/scala/org/apache/spark/ui/UIUtils.scala
hsaputra: Ah yes, my mistake it should have been 2 spaces =)

PR #880, opened 2013-08-31
Various UI tweaks

mateiz:
- Use "fluid" layout that can expand to wide browser windows (the old one was limited in width to 1200 px)
- Remove unnecessary `<hr>` lines
- Switch back to Bootstrap's default theme
- Use Bootstrap's standard nav style
- Make headers more consistent between deploy and app UIs
- Replace some inline CSS with stylesheets

Here's what it looks like:

screen shot 2013-08-31 at 4 08 05 pm


pwendell:
Hey Matei,

Two things here:

(1) The executors table doesn't seem to be fluid like the other ones, based on playing with the UI.
(2) Is there a way to uppoer-bound the width of the tables? It's a little ridiculous looking when you open full screen on a large monitor, it gets so wide that it hurts readability.

pwendell:
Hm - I think (1) might be caused by something weird with my build. Please ignore.

mateiz:
For 2, I actually would prefer not to upper-bound it, because most people with big monitors can make a narrower browser window. I'd rather make sure that people who want to see stuff in 1 row can do so. (Technically it would be possible to upper-bound it at a value larger than 1200, but it would require hacking BootStrap more extensively, which is painful).

pwendell:
OK

pwendell:
Looks good to me

mateiz:
Cool, thanks.

PR #882, opened 2013-09-01
Rename spark package to org.apache.spark

mateiz:
This is an initial commit whose purpose is to ~~kill GitHub~~ make tests pass. They do pass on my machine. Further changes will be needed to update the docs and README.

Edit: Those other updates are now complete too.


mateiz:
OK, I think this is ready now. FYI, I also added two super-small usability fixes: https://github.com/mateiz/spark/commit/c89977fe21410f7f9f7bbe37a675920d5636eb38 and https://github.com/mateiz/spark/commit/d800b543521a76189a51c4a8d808f0f5ee08cec1, since I found the previous behavior of these annoying when trying the examples in the docs.

mateiz:
Alright, I've merged this in.

jey:
Does Maven maybe have some way that our 0.8 and later POM files can specify that they are replacements for the corresponding packages from the `org.spark-project` group? So that people don't have to do all that crazy exclusion stuff by hand..

PR #883, opened 2013-09-01
Don't require the spark home environment variable to be set for standalone mode (change needed by SIMR)

alig:


mateiz:
Looks good, thanks.

PR #888, opened 2013-09-03
Decrease max class filename length for scalac

aarondav:
This causes smaller filenames to be generated. While 255 is the usual
filename limit, this can cause problems when running on systems using
some filesystem encryption schemes (namely, encrypted folders in Ubuntu).


mateiz:

aarondav:
Turns out this is insufficient to get it working. Closing this request.

PR #890, opened 2013-09-04
Fix hash bug

mridulm:
I did not see any other place where this idiom was used - but if I missed it, we got to add Utils.toHash there too.
Unfortunately, my response time might be spotty, apologies in advance.


mateiz:
Interesting, thanks for spotting this. Do you mind renaming this to Utils.nonNegativeHash though? Some use cases of hash functions are fine with negative values.

I guess the other option would be to use a Long in this calculation (`math.abs(block.hashCode.toLong)`), but that would be less obvious.

mateiz:

mateiz:
Merged this -- thanks Mridul.

PR #891, opened 2013-09-04
[SPARK-864]DAGScheduler Exception if we delete Worker and StandaloneExecutorBackend then add Worker

xiajunluan:
In fair scheduling mode, pool information of each stage(taskset) is stored in job(ActiveJob) that contain this stage. but in spark, each stage could be used more than one job, so we must not delete ActiveJob in idToActiveJob structure when job is ended, otherwise, once we need re-compute this stage , it will happens following exception

Exception in thread "DAGScheduler" java.util.NoSuchElementException: key not found: 2
    at spark.scheduler.DAGScheduler.spark$scheduler$DAGScheduler$$submitMissingTasks(DAGScheduler.scala:515)
    at spark.scheduler.DAGScheduler.spark$scheduler$DAGScheduler$$submitStage(DAGScheduler.scala:481)
    at spark.scheduler.DAGScheduler$$anonfun$resubmitFailedStages$3.apply(DAGScheduler.scala:383)
    at spark.scheduler.DAGScheduler$$anonfun$resubmitFailedStages$3.apply(DAGScheduler.scala:382)
    at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:34)
    at scala.collection.mutable.ArrayOps.foreach(ArrayOps.scala:38)
    at spark.scheduler.DAGScheduler.resubmitFailedStages(DAGScheduler.scala:382)
    at spark.scheduler.DAGScheduler.spark$scheduler$DAGScheduler$$run(DAGScheduler.scala:433)
    at spark.scheduler.DAGScheduler$$anon$1.run(DAGScheduler.scala:135)

### DIFF
--- a/core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
@@ -531,9 +531,16 @@ class DAGScheduler(
         tasks += new ResultTask(stage.id, stage.rdd, job.func, partition, locs, id)
       }
     }
+
+    val properties = if (idToActiveJob.contains(stage.jobId)) {
+      idToActiveJob(stage.jobId).properties
+    } else {
+      //this stage will be assigned to "default" pool
+      null
+    }
+
     // must be run listener before possible NotSerializableException
     // should be "StageSubmitted" first and then "JobEnded"
-    val properties = idToActiveJob(stage.jobId).properties
     listenerBus.post(SparkListenerStageSubmitted(stage, tasks.size, properties))
 
     if (tasks.size > 0) {


mateiz:
Andrew, I don't think this will work for long-term operation because that hashmap will grow indefinitely. Can you fix it in another way, e.g. assign the stageID to a new active job (the one that wants to recompute the stage)?

Matei

markhamstra:
Already fixed in https://github.com/mesos/spark/pull/842

mateiz:
I think we should still do a quick fix in 0.8.0 because #842 is too big a change to merge. Andrew, is there an easy way to implement what I said?

markhamstra:
I don't know that you can do it with a small change to the current code.  Right now, a stage doesn't know about any other jobId than that of the job that created it.  I can find another jobId in https://github.com/mesos/spark/issues/842 because of all the other work that I've done in that PR to track the linkages from the stage to other jobs that need it, but that information just isn't there in the current code.

xiajunluan:
@mateiz, I will fix it ASAP
@markhamstra, I will do a hot fix for 0.8.0 release, and I will also review your code to confirm if it is a better solution for all stage/job logic, thanks.

xiajunluan:
Make jobId of stage to be variable, and alway set it to be id of newest job that will use this stage.

markhamstra:
I'm pretty sure that breaks FIFO scheduling.

mateiz:
I think the important thing is to do it only if the job ID isn't in idToActiveJob. Am I missing something on that?

mateiz:
(Basically in the case where we were getting a "key not found" earlier, just use the latest job)

markhamstra:
What do you mean by "the latest job"?  Just the current state of nextJobId - 1?  Whatever you are using for "the latest job", somehow you have to know that the stage currently being submitted is actually part of that job so that the idToActiveJob("the latest job").properties makes sense for that stage and taskSet.

xiajunluan:
I think we could delete ActiveJob from idToActiveJob in cleanup function when cleaning shuffleToMapStage structure instead of end of the job, and idToActiveJob should remain ActiveJob structure that still is used by stages in shuffleToMapStage, is that the final solution?

markhamstra:
shuffleToMapStages isn't currently cleaned up at all except in the TimeStampedHashMap cleanup().  That only runs for Spark Streaming or if spark.cleaner.ttl is set.  In either case, spark.cleaner.ttl is typically a lengthy interval, which would allow for a lot of bloat of idToActiveJob.  And if you change the cleanup interval to be short, then you run the risk of eliminating still-valid, in-use data.

Sorry to constantly be coming back with negatives, but like I said before, I don't think that there is an easy solution that also controls data structure bloat.

mateiz:
Guys, the goal here is just to get a simple solution for 0.8.0. As long as we don't crash with an exception, it's okay. In my post above, when I said "latest job", I meant the job on whose behalf we are calling submitMissingTasks. We're submitting the missing tasks because some currently running stage (i.e. part of an active job) decided it needs results from a previous stage. Let's just see which stage _that_ running stage belonged to, and assign the old stage to this new job. Doesn't seem that hard to do -- we already have idToActiveJob for the running stage.

The idea of clearing idToActiveJob only when the TTL is set may also makes sense, but it seems riskier because we keep a lot of state in ActiveJob (e.g. the user's function and a job listener, which may reference some pretty large variables in the user's program).

A final option is to not assign these stages to _any_ job, and give them a special priority meaning "recompute this because it failed". That also seems fine for this fairly rare case.

mateiz:
Basically for the third option, all we'd do is 

```
val properties = if (idToActiveJob.contains(stage.jobId)) {
  idToActiveJob(stage.jobId).properties
} else {
  // a new Properties object with enough stuff for the scheduler to work; e.g. just add the job in the root pool
}
```

This code path worked before and only broke because we now look for an active job for the stage, which may not exist.

markhamstra:
Dumping into the default pool makes some sense to me.  There are actually several unresolved issues about how cross-job dependencies do and should work when the jobs are in different priority pools (and https://github.com/mesos/spark/issues/842 also doesn't address those), so your third option seems like a reasonable default behavior that mostly avoids the problem issues.

And you are right, Matei, that a lot of ActiveJob state can build up.  I was shown one of our logs today with gigabytes of memory tied up in ActiveJob state via idToActiveJob.

xiajunluan:
Ok, I will fix it based on mateiz's opinion.

mateiz:
Cool, thanks!

PR #893, opened 2013-09-04
SPARK-884: Add unit test to validate Spark JSON output

aarondav:
This unit test simply validates that the outputs of
the JsonProtocol methods are syntactically valid JSON.

### DIFF
--- a/core/src/test/scala/org/apache/spark/deploy/JsonProtocolSuite.scala
+++ b/core/src/test/scala/org/apache/spark/deploy/JsonProtocolSuite.scala
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.deploy
+
+import org.apache.spark.deploy.master.{WorkerInfo, ApplicationInfo}
+import org.scalatest.FunSuite
+import org.scalatest.matchers.ShouldMatchers
+import java.util.Date
+import net.liftweb.json.{JsonAST, JsonParser}
+import net.liftweb.json.JsonAST.JValue
+import org.apache.spark.deploy.worker.ExecutorRunner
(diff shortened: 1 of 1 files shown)


core/src/test/scala/org/apache/spark/deploy/JsonProtocolSuite.scala
```diff
@@ -0,0 +1,85 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.deploy
+
+import org.apache.spark.deploy.master.{WorkerInfo, ApplicationInfo}
```
pwendell: We use the following import style. 4 groups:
- java.X
- scala.X
- all other packages alphabetic order.
- org.apache.spark.X

Newline between each group, sorted within each group.

core/src/test/scala/org/apache/spark/deploy/JsonProtocolSuite.scala
```diff
@@ -0,0 +1,85 @@
...
+    val executors = List[ExecutorRunner]()
+    val finishedExecutors = List[ExecutorRunner](createExecutorRunner(), createExecutorRunner())
+    val stateResponse = new WorkerStateResponse("host", 8080, "workerId", executors, finishedExecutors,
+      "masterUrl", 4, 1234, 4, 1234, "masterWebUiUrl")
+    val output = JsonProtocol.writeWorkerState(stateResponse)
+    assertValidJson(output)
+  }
+
+  def createAppDesc() : ApplicationDescription = {
+    new ApplicationDescription("name", 4, 1234, null, "sparkHome", "appUiUrl")
```
pwendell: It would be nice to provide fake values for the things that are not actors, rather than just `null`. Those might be logged some day (if someone changes the protocol) and then we'll catch it if they produce bad JSON.

core/src/test/scala/org/apache/spark/deploy/JsonProtocolSuite.scala
```diff
@@ -0,0 +1,85 @@
...
+  }
+  def createWorkerInfo() : WorkerInfo = {
+    new WorkerInfo("id", "host", 8080, 4, 1234, null, 80, "publicAddress")
+  }
+  def createExecutorRunner() : ExecutorRunner = {
+    new ExecutorRunner("appId", 123, createAppDesc(), 4, 1234, null, "workerId", "host", null, null)
+  }
+
+  /* @throws ParseException if not valid json */
+  def assertValidJson(json: JValue) {
```
pwendell, line 85: We want to be able to distinguish from a test failure and an unexpected exception thrown during the tests. It would be better if this function called `parse`, then specifically caught a ParseException and then asserted a failure (I'm not sure exactly how to do this with scalatest, but at a minimum you could just `assert(false)` with a good error message).

pwendell:
Hey @ilikerps - thanks for this! Looks good overall, I added a few comments...

aarondav:
All comments addressed - thanks Patrick!

pwendell:
Thanks @ilikerps - this looks good to me!

PR #894, opened 2013-09-05
Updating assembly README to reflect recent changes in the build.

ghost:
Small fix.
Would be nice to have in branch-0.8 as well.


mateiz:
Looks good, thanks! Will also merge to 0.8.

PR #895, opened 2013-09-05
SPARK-821: Don't cache results when action run locally on driver

aarondav:
Caching the results of local actions (e.g., rdd.first()) causes the driver to
store entire partitions in its own memory, which may be highly constrained.
This patch simply makes the CacheManager avoid caching the result of all locally-run computations.

### DIFF
--- a/core/src/main/scala/org/apache/spark/CacheManager.scala
+++ b/core/src/main/scala/org/apache/spark/CacheManager.scala
@@ -69,8 +69,8 @@ private[spark] class CacheManager(blockManager: BlockManager) extends Logging {
           val elements = new ArrayBuffer[Any]
           logInfo("Computing partition " + split)
           elements ++= rdd.computeOrReadCheckpoint(split, context)
-          // Try to put this block in the blockManager
-          blockManager.put(key, elements, storageLevel, true)
+          // Persist the result, so long as the task is not running locally
+          if (!context.runningLocally) blockManager.put(key, elements, storageLevel, true)
           return elements.iterator.asInstanceOf[Iterator[T]]
         } finally {
           loading.synchronized {
--- a/core/src/main/scala/org/apache/spark/CacheManager.scala
+++ b/core/src/main/scala/org/apache/spark/CacheManager.scala
@@ -66,11 +66,13 @@ private[spark] class CacheManager(blockManager: BlockManager) extends Logging {
         }
         try {
           // If we got here, we have to load the split
-          val elements = new ArrayBuffer[Any]
           logInfo("Computing partition " + split)
-          elements ++= rdd.computeOrReadCheckpoint(split, context)
+          val computedValues = rdd.computeOrReadCheckpoint(split, context)
           // Persist the result, so long as the task is not running locally
-          if (!context.runningLocally) blockManager.put(key, elements, storageLevel, true)
+          if (context.runningLocally) return computedValues
+          val elements = new ArrayBuffer[Any]
+          elements ++= computedValues
+          blockManager.put(key, elements, storageLevel, true)
           return elements.iterator.asInstanceOf[Iterator[T]]
         } finally {
           loading.synchronized {
--- a/core/src/main/scala/org/apache/spark/CacheManager.scala
+++ b/core/src/main/scala/org/apache/spark/CacheManager.scala
@@ -69,7 +69,7 @@ private[spark] class CacheManager(blockManager: BlockManager) extends Logging {
           logInfo("Computing partition " + split)
           val computedValues = rdd.computeOrReadCheckpoint(split, context)
           // Persist the result, so long as the task is not running locally
-          if (context.runningLocally) return computedValues
+          if (context.runningLocally) { return computedValues }
           val elements = new ArrayBuffer[Any]
           elements ++= computedValues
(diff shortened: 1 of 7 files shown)


core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
```diff
@@ -478,7 +478,7 @@ class DAGScheduler(
       SparkEnv.set(env)
       val rdd = job.finalStage.rdd
       val split = rdd.partitions(job.partitions(0))
-      val taskContext = new TaskContext(job.finalStage.id, job.partitions(0), 0)
+      val taskContext = new TaskContext(job.finalStage.id, job.partitions(0), 0, true)
```
markhamstra: Prefer `new TaskContext(job.finalStage.id, job.partitions(0), 0, runningLocally = true)`

rxin:
Thanks for doing this, Aaron!
1. Can we remove the default value for runningLocally and explicitly set it to false in local tasks? This way, it is much more clear to whoever reads the code. Since this is internal code, we don't have to worry about maintaining api consistency. 
2. As Mark pointed out, use named parameters for boolean values so it is immediately obvious what "false" mean.
3. Can you add a unit test for this?

core/src/main/scala/org/apache/spark/CacheManager.scala
```diff
@@ -66,10 +66,12 @@ private[spark] class CacheManager(blockManager: BlockManager) extends Logging {
         }
         try {
           // If we got here, we have to load the split
-          val elements = new ArrayBuffer[Any]
           logInfo("Computing partition " + split)
-          elements ++= rdd.computeOrReadCheckpoint(split, context)
-          // Try to put this block in the blockManager
+          val computedValues = rdd.computeOrReadCheckpoint(split, context)
+          // Persist the result, so long as the task is not running locally
+          if (context.runningLocally) return computedValues
```
rxin: For one line ifs, wrap it with { }.

core/src/test/scala/org/apache/spark/CacheManagerSuite.scala
```diff
@@ -0,0 +1,89 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark
+
+import scala.Array
```
rxin: No need to import this one

core/src/test/scala/org/apache/spark/CacheManagerSuite.scala
```diff
@@ -0,0 +1,89 @@
...
+import scala.collection.mutable.ArrayBuffer
+
+import org.scalatest.{BeforeAndAfter, FunSuite}
+import org.scalatest.mock.EasyMockSugar
+
+import org.apache.spark.rdd.RDD
+import org.apache.spark.storage.{BlockManager, StorageLevel}
+import org.junit.Assert
```
rxin: Remove this import and change the junit assert to scala assert?

core/src/test/scala/org/apache/spark/CacheManagerSuite.scala
```diff
@@ -0,0 +1,89 @@
...
+
+// TODO: Test the CacheManager's thread-safety aspects
+class CacheManagerSuite extends FunSuite with BeforeAndAfter with EasyMockSugar {
+  val sc = new SparkContext("local", "test")
```
rxin: We need to clean this sc after the test finishes running. See other test suites for an example.

core/src/test/scala/org/apache/spark/CacheManagerSuite.scala
aarondav: Scala assert is actually less expressive than this -- assertEquals(), when it fails, returns the two values, rather than just saying they're not equal. Is there an equivalent for scalatest? I saw Assert.assertEquals being used elsewhere in our testing code.

rxin: You can use === in Scala assert to throw a more meaningful message. assertEquals is only used in the JavaAPISuite I believe.

aarondav: Thanks!

pwendell:
@ilikerps Looks great, I'm going to merge this.

hirakendu:
Perhaps explicitly specify type, e.g., `rdd = new RDD[Int](sc, Nil) {` .

rxin:
I pushed a fix: https://github.com/mesos/spark/commit/1e15feb5a314e7180328b9208054966e040eb2ad

Weird that Jenkins didn't catch this one in the pull request builder. Maybe Jenkins ran against an earlier version of Aaron's changes ...

shivaram:
So it looks like Jenkins had a build failure but the overall status was marked as success. Not sure why this happened though. 
(Search for Iterator[Nothing] in https://amplab.cs.berkeley.edu/jenkins/job/SparkPullRequestBuilder/842/consoleFull )

PR #898, opened 2013-09-06
SPARK-660: Add StorageLevel support in Python

aarondav:
It uses reflection... I am not proud of that fact, but it at least ensures
compatibility (sans refactoring of the StorageLevel stuff).

### DIFF
--- a/python/pyspark/context.py
+++ b/python/pyspark/context.py
@@ -281,16 +281,23 @@ class SparkContext(object):
 
 class StorageLevelReader:
     """
-    Mimics the Scala StorageLevel by directing all attribute requests
+    Mimics the Scala StorageLevel by delegating all attribute requests
     (e.g., StorageLevel.DISK_ONLY) to the JVM for reflection.
+    Memoizes results to reduce JVM call/memory overheads.
     """
 
     def __init__(self, sc):
         self.sc = sc
+        self.memoized = {}
 
     def __getattr__(self, name):
+        if name in self.memoized: 
+          return self.memoized[name]
+
         try:
-            return self.sc._jvm.PythonRDD.getStorageLevel(name)
+            storageLevel = self.sc._jvm.PythonRDD.getStorageLevelByName(name)
+            self.memoized[name] = storageLevel
+            return storageLevel
         except:
             print "Failed to find StorageLevel:", name
 
--- a/python/pyspark/context.py
+++ b/python/pyspark/context.py
@@ -27,6 +27,7 @@ from pyspark.broadcast import Broadcast
 from pyspark.files import SparkFiles
 from pyspark.java_gateway import launch_gateway
 from pyspark.serializers import dump_pickle, write_with_length, batched
+from pyspark.storagelevel import StorageLevel
 from pyspark.rdd import RDD
 
 from py4j.java_collections import ListConverter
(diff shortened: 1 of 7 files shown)


pwendell:
Is there a more pythonic way to do enumerations? This is a neat trick to match the syntax of Scala's `StorageLevel.MEMORY_ONLY`, etc but if there is a more standard way of doing enumerations in python, I think it would be better to do that, even if we just hard coded the storage level options in the Python code. @JoshRosen should take a look though, he maintains the python API.

pwendell:
Hm it actually looks like every enumeration technique in python boils down to this same syntax ClassName.enumValue, so maybe this is in line with that...

jey:
Perhaps we should memoize the values on the Python side since they're constants?

python/pyspark/context.py
```diff
@@ -279,6 +279,27 @@ def setCheckpointDir(self, dirName, useExisting=False):
...
+    def __getattr__(self, name):
+        if name in self.memoized: 
+          return self.memoized[name]
+
+        try:
+            storageLevel = self.sc._jvm.PythonRDD.getStorageLevelByName(name)
+            self.memoized[name] = storageLevel
+            return storageLevel
+        except:
+            print "Failed to find StorageLevel:", name
```
mateiz: Let's not print stuff to stdout; throw an exception in this case

python/pyspark/context.py
```diff
@@ -279,6 +279,27 @@ def setCheckpointDir(self, dirName, useExisting=False):
...
+    Mimics the Scala StorageLevel by delegating all attribute requests
+    (e.g., StorageLevel.DISK_ONLY) to the JVM for reflection.
+    Memoizes results to reduce JVM call/memory overheads.
+    """
+
+    def __init__(self, sc):
+        self.sc = sc
+        self.memoized = {}
+
+    def __getattr__(self, name):
```
mateiz: Instead of doing this and the reflection approach, I'd prefer to list the storage levels in Python. It's more repetitive, but it will be much easier to debug (e.g. no need to throw an exception below)

mateiz:
Thanks for adding this, Aaron. My comment though is that I'd prefer to list the storage levels in Python instead of using **getattr**.. that just creates a lot of complexity and error handling issues for a list of 6-7 constants. It's clever but not worth it for this use case.

python/pyspark/shell.py
```diff
@@ -29,6 +29,7 @@
 add_files = os.environ.get("ADD_FILES").split(',') if os.environ.get("ADD_FILES") != None else None
 
 sc = SparkContext(os.environ.get("MASTER", "local"), "PySparkShell", pyFiles=add_files)
+StorageLevel = sc.StorageLevel # alias StorageLevel to global scope
```
mateiz: Unfortunately I don't think this is a good way to make this visible -- it will only be available in the shell. How about the following approach:
- Create an enum called StorageLevel in Python (e.g. using these options: http://stackoverflow.com/questions/36932/how-can-i-represent-an-enum-in-python)
- In RDD.persist(), translate the Python StorageLevel object, which is just an identifier, to a Java one, using a hashmap that you initialize inside SparkContext (when you're connected to Java).

python/pyspark/context.py
```diff
@@ -47,6 +48,7 @@ class SparkContext(object):
     _active_spark_context = None
     _lock = Lock()
     _python_includes = None # zip and egg files that need to be added to PYTHONPATH
+    StorageLevel = None
```
mateiz: This seems like it was left in unnecessarily

mateiz:
Looks good, thanks for putting this together!

PR #899, opened 2013-09-06
to fix make-distribution.sh: cp conf/*.template did not work, changed to...

Du-Li:
the original command failed:
cp: cannot stat `/home/bigdatauser/Downloads/spark/conf/*.template': No such file or directory

it's necessary to copy the entire directory conf/ to dist/ because some local settings should be included for distribution.


make-distribution.sh
```diff
@@ -95,7 +95,8 @@ cp $FWDIR/assembly/target/scala*/*assembly*hadoop*.jar "$DISTDIR/jars/"
 
 # Copy other things
 mkdir "$DISTDIR"/conf
-cp -r "$FWDIR/conf/*.template" "$DISTDIR"
+#cp -r "$FWDIR/conf/*.template" "$DISTDIR"
+cp -r "$FWDIR/conf" "$DISTDIR"
```
mateiz, line 99: I think we shouldn't copy all of conf because we want to create a binary distribution that the user can configure later. It would be confusing for a "distribution" to include a specific configuration.

I think all you have to do is a `mkdir $DISTDIR/conf` and then `cp $FWDIR/conf/*.template $DISTDIR/conf`.

Also just as a note, never leave commented-out code in source files; we can see the changes in a diff.

mateiz:
I've actually fixed this myself in https://github.com/mesos/spark/pull/916 now. Thanks for reporting it.

PR #903, opened 2013-09-07
Fixed the bug that ResultTask was not properly deserializing outputId.

rxin:
Was reading through the scheduler code and found this potential bug.


mateiz:
That's definitely bad.. mind adding a test for it? I think it would happen if you use lookup() on a partitioned RDD.

rxin:
I have to admit that I don't fully understand what outputId does. 

I wasn't able to produce a problem with 

```
sc.parallelize(1 to 10, 2).map(x => (x,x)).partitionBy(new org.apache.spark.HashPartitioner(3)).lookup(1)
```

shivaram:
As far as I can see the outputId is only used on the driver (in DAGScheduler and the partial evaluators). So the serialization / deserialization doesn't really matter because of that ?

mateiz:
Ah, that is true. In that case I'll still merge the patch though in case people look at this code later and try to use it in tasks.

PR #907, opened 2013-09-08
Add better docs for coalesce.

stephenh:
Include the useful tip that if shuffle=true, coalesce can actually
increase the number of partitions.

This makes coalesce more like a generic `RDD.repartition` operation.

(Ideally this `RDD.repartition` could automatically choose either a coalesce or
a shuffle if numPartitions was either less than or greater than, respectively,
the current number of partitions.)

### DIFF
--- a/core/src/main/scala/org/apache/spark/rdd/RDD.scala
+++ b/core/src/main/scala/org/apache/spark/rdd/RDD.scala
@@ -282,8 +282,8 @@ abstract class RDD[T: ClassManifest](
    * Note: With shuffle = true, you can actually coalesce to a larger number
    * of partitions. This is useful if you have a small number of partitions,
    * say 100, potentially with a few partitions being abnormally large. Calling
-   * coalecse(1000, shuffle = true) will result in 1000 partitions with the
-   * data evenly distributed into each partition.
+   * coalesce(1000, shuffle = true) will result in 1000 partitions with the
+   * data distributed using a hash partitioner.
    */
   def coalesce(numPartitions: Int, shuffle: Boolean = false): RDD[T] = {
     if (shuffle) {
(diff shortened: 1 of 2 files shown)


mateiz:

core/src/test/scala/org/apache/spark/rdd/RDDSuite.scala
```diff
@@ -175,8 +175,14 @@ class RDDSuite extends FunSuite with SharedSparkContext {
     val coalesced5 = data.coalesce(1, shuffle = true)
     assert(coalesced5.dependencies.head.rdd.dependencies.head.rdd.asInstanceOf[ShuffledRDD[_, _, _]] !=
       null)
+
+    // when shuffling, we can increase the number of partitions
+    val coalesced6 = data.coalesce(20, shuffle = true)
+    assert(coalesced6.partitions.size === 20)
+    assert(coalesced6.collect().toList === (1 to 10).toList)
```
mateiz: This seems kind of brittle because it checks for the order of the elements, but the shuffle may have changed them. It just happens not to for integers because they hash to the right partitions. Mind changing the `toList` calls to `toSet`?

core/src/main/scala/org/apache/spark/rdd/RDD.scala
```diff
@@ -267,6 +267,23 @@ abstract class RDD[T: ClassManifest](
...
+   * you like (e.g. one node in the case of numPartitions = 1). To avoid this,
+   * you can pass shuffle = true. This will add a shuffle step, but means the
+   * current upstream partitions will be executed in parallel (per whatever
+   * the current partitioning is).
+   *
+   * Note: With shuffle = true, you can actually coalesce to a larger number
+   * of partitions. This is useful if you have a small number of partitions,
+   * say 100, potentially with a few partitions being abnormally large. Calling
+   * coalecse(1000, shuffle = true) will result in 1000 partitions with the
+   * data evenly distributed into each partition.
```
rxin: "evenly distributed" -> "distributed using a hash partitioner"

stephenh:
Good points, fixed both.

mateiz:
Looks good, thanks!

PR #908, opened 2013-09-08
Fix target JVM version in scala build

pwendell:
This patch makes the sbt build consistent with the maven build and sets the target JVM versions. This is important for cutting binary releases since we don't want to produce binaries with 1.7 code if someone compiles using JDK 1.7.


pwendell:
Jenkins, test this please.

mateiz:
Looks good, thanks.

PR #909, opened 2013-09-09
Fix an instance where full standalone mode executor IDs were passed to

mateiz:
StandaloneSchedulerBackend instead of the smaller IDs used within Spark
(that lack the application name).

This was reported by ClearStory in
https://github.com/clearstorydata/spark/pull/9.

Also fixed some messages that said slave instead of executor.

CC @markhamstra

### DIFF
--- a/core/src/main/scala/org/apache/spark/scheduler/cluster/SparkDeploySchedulerBackend.scala
+++ b/core/src/main/scala/org/apache/spark/scheduler/cluster/SparkDeploySchedulerBackend.scala
@@ -76,17 +76,17 @@ private[spark] class SparkDeploySchedulerBackend(
     }
   }
 
-  override def executorAdded(executorId: String, workerId: String, hostPort: String, cores: Int, memory: Int) {
+  override def executorAdded(fullId: String, workerId: String, hostPort: String, cores: Int, memory: Int) {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/scheduler/cluster/SparkDeploySchedulerBackend.scala
```diff
@@ -76,17 +76,17 @@ private[spark] class SparkDeploySchedulerBackend(
...
 
-  override def executorRemoved(executorId: String, message: String, exitStatus: Option[Int]) {
+  override def executorRemoved(fullId: String, message: String, exitStatus: Option[Int]) {
     val reason: ExecutorLossReason = exitStatus match {
       case Some(code) => ExecutorExited(code)
       case None => SlaveLost(message)
     }
-    logInfo("Executor %s removed: %s".format(executorId, message))
-    removeExecutor(executorId, reason.toString)
+    logInfo("Executor %s removed: %s".format(fullId, message))
```
rxin, line 89: Should we promote this to warning level?

mateiz, line 89: We already have an ERROR-level message in ClusterScheduler.executorLost so I think it's fine as is

PR #913, opened 2013-09-09
Missing metrics-ganglia dependency in core/pom.xml

markhamstra:
Previously defined dependency has to actually be used in core.


markhamstra:
Dupe

PR #914, opened 2013-09-09
Fine grain control of specifying ttl

mridulm:
Existing behavior is preserved (which is referenced in .
Using an enum for now - please suggest modification if inappropriate.

This is a hack to work around the problem I discussed with @matei : namely, enable cleaner to evict shuffle/etc blocks, while ensuring that there is no side effect in other parts of spark (like eviction of broadcast variables, etc).


mridulm:
Please do not commit yet - looks like broadcast blocks are sitll getting dropped, sigh ...

mridulm:
This PR does not address cases where a block is being used by a job (or its dependencies) - which is marked 'too old'. Actively used blocks should be exempted from cleanup - but unfortunately, that will need to be done in a later time.

For now, this sufficiently addresses our basic concerns - it is open for review.
@mateiz please let us know your comments, thanks.

mridulm:
Any updates on this ? Without this patch, for fairly large + fast jobs, master OOM's for us.

hsaputra:
Could you kindly move the pull request to the new Spark github repo under ASF mirror: https://github.com/apache/incubator-spark

Thx, Henry

core/src/main/scala/org/apache/spark/MapOutputTracker.scala
```diff
@@ -71,7 +71,7 @@ private[spark] class MapOutputTracker extends Logging {
   var cacheEpoch = epoch
   private val cachedSerializedStatuses = new TimeStampedHashMap[Int, Array[Byte]]
 
-  val metadataCleaner = new MetadataCleaner("MapOutputTracker", this.cleanup)
+  val metadataCleaner = new MetadataCleaner(MetadataCleanerType.MAP_OUTPUT_TRACKER, "MapOutputTracker", this.cleanup)
```
mateiz, line 74: I don't like that these take both an enum type and a string name as arguments -- let's have them take just the enum.

core/src/main/scala/org/apache/spark/storage/BlockManager.scala
```diff
@@ -548,9 +552,12 @@ private[spark] class BlockManager(
...
    * Put a new block of values to the block manager. Returns its (estimated) size in bytes.
    */
-  def put(blockId: String, values: ArrayBuffer[Any], level: StorageLevel,
-    tellMaster: Boolean = true) : Long = {
+  def put(blockId: String, values: ArrayBuffer[Any], level: StorageLevel, tellMaster: Boolean = true) : Long = {
+
+    put(blockId, values, level, tellMaster, false)
+  }
 
+  def put(blockId: String, values: ArrayBuffer[Any], level: StorageLevel, tellMaster: Boolean, broadcastBlock: Boolean) : Long = {
```
mateiz, line 560: So far we've been using the blockID instead of a flag to determine whether a block is a broadcast, an RDD, or something else. Please do it that way here instead of adding a new boolean flag just for broadcasts. In the cleanup part you can just check if the block ID starts with "broadcast_".

core/src/main/scala/org/apache/spark/util/MetadataCleaner.scala
```diff
@@ -53,9 +54,39 @@ class MetadataCleaner(name: String, cleanupFunc: (Long) => Unit) extends Logging
...
 }
 
+object MetadataCleanerType extends Enumeration("map_output_tracker", "spark_context", "http_broadcast", "dag_scheduler", "result_task",
+    "shuffle_map_task", "block_manager", "broadcast_vars") {
+
+  val MAP_OUTPUT_TRACKER, SPARK_CONTEXT, HTTP_BROADCAST, DAG_SCHEDULER, RESULT_TASK, SHUFFLE_MAP_TASK, BLOCK_MANAGER, BROADCAST_VARS = Value
+
+  type MetadataCleanerType = Value
+
+  def systemProperty(which: MetadataCleanerType.MetadataCleanerType) = "spark.cleaner.ttl." + which.toString
```
mateiz, line 64: Avoid underscores in the system properties -- use camelCase to match our other ones. I'd also be okay if these were full class names (MapOutputTracker, SparkContext, etc).

core/src/main/scala/org/apache/spark/util/MetadataCleaner.scala
```diff
@@ -53,9 +54,39 @@ class MetadataCleaner(name: String, cleanupFunc: (Long) => Unit) extends Logging
...
+}
+
 
 object MetadataCleaner {
+
+
+  // using only sys props for now : so that workers can also get to it while preserving earlier behavior.
   def getDelaySeconds = System.getProperty("spark.cleaner.ttl", "-1").toInt
```
mateiz, line 72: Let's call this getDefaultDelaySeconds or something like that; I'm not sure it gets called directly anymore after this change

mateiz:
Hey Mridul, sorry for the long delay on this, I was traveling after the 0.8 release. It looks good overall but I made a few comments to reduce the changes from existing code and make it more consistent. Please send it against the new repo too as Henry pointed out.

mridulm:
Thx Henry, done.
Matei, I have tried to address the review comments in the new PR, please do take a look - hopefully, I did not miss something.

hsaputra:
Cool! Thanks for your help Mridul

mateiz:
Closing this since it's now here: https://github.com/apache/incubator-spark/pull/32

PR #915, opened 2013-09-09
Get rid of / improve ugly NPE when Utils.deleteRecursively() fails

velvia:
listFiles() could return null if the I/O fails, and this currently results in an ugly NPE which is hard to diagnose.

### DIFF
--- a/core/src/main/scala/org/apache/spark/util/Utils.scala
+++ b/core/src/main/scala/org/apache/spark/util/Utils.scala
@@ -459,7 +459,9 @@ private[spark] object Utils extends Logging {
 
   private def listFilesSafely(file: File): Seq[File] = {
     val files = file.listFiles()
-    if (files == null) throw new IOException("Failed to list files for dir: " + file)
+    if (files == null) {
+      throw new IOException("Failed to list files for dir: " + file)
+    }
     files
   }
 
--- a/core/src/main/scala/org/apache/spark/util/Utils.scala
+++ b/core/src/main/scala/org/apache/spark/util/Utils.scala
@@ -457,12 +457,18 @@ private[spark] object Utils extends Logging {
   def newDaemonFixedThreadPool(nThreads: Int): ThreadPoolExecutor =
     Executors.newFixedThreadPool(nThreads, daemonThreadFactory).asInstanceOf[ThreadPoolExecutor]
 
+  private def listFilesSafely(file: File): Seq[File] = {
+    val files = file.listFiles()
+    if (files == null) throw new IOException("Failed to list files for dir: " + file)
+    files
+  }
+
   /**
    * Delete a file or directory and its contents recursively.
    */
   def deleteRecursively(file: File) {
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -457,12 +457,18 @@ private[spark] object Utils extends Logging {
   def newDaemonFixedThreadPool(nThreads: Int): ThreadPoolExecutor =
     Executors.newFixedThreadPool(nThreads, daemonThreadFactory).asInstanceOf[ThreadPoolExecutor]
 
+  private def listFilesSafely(file: File): Seq[File] = {
+    val files = file.listFiles()
+    if (files == null) throw new IOException("Failed to list files for dir: " + file)
```
hsaputra: Just small nit on style, do we prefer inline in the "if" check without else or something like:

if (files == null) {
  throw new IOException("Failed to list files for dir: " + file)
}

I remember Scala style guide prefer curly for if without else check.

jey: I believe the usual Spark style convention is to use the extra braces and put it on a separate line, because that is more future-proof (e.g. if additional code is added within the body of the `if`).

mateiz:
Jenkins, this is ok to test

core/src/main/scala/org/apache/spark/util/Utils.scala
velvia: Fixed with new patch.  For some reason these comments don't show up on the PR. 

-Evan
To be free is not merely to cast off one's chains, but to live in a way that respects & enhances the freedom of others. (#NelsonMandela)

hsaputra:
+1 looks good to me

mateiz:
This looks good to me, though it still doesn't quite solve the issue (if there's any way to solve it). We should look into an alternate way to delete these directories, e.g. Guava.

pwendell:
Hey @velvia - do you know what was causing the dir to not be listable? It would be good to detect or fix the root cause of this if we can.

velvia:
Patrick - at the time, the user running the Spark jobs was running into the
open file handles limit.   So, yeah, it's a pretty rare situation, not sure
if it is really solvable.

On Mon, Sep 9, 2013 at 9:08 PM, Patrick Wendell notifications@github.comwrote:

> Hey @velvia https://github.com/velvia - do you know what was causing
> the dir to not be listable? It would be good to detect or fix the root
> cause of this if we can.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/mesos/spark/pull/915#issuecomment-24132375
> .

## 

The fruit of silence is prayer;
the fruit of prayer is faith;
the fruit of faith is love;
the fruit of love is service;
the fruit of service is peace.  -- Mother Teresa

hsaputra:
I am not sure even using library like Guava will solve running out file handles for delete.

PR #916, opened 2013-09-09
Fix copy issue in https://github.com/mesos/spark/pull/899

mateiz:
This was reported in https://github.com/mesos/spark/pull/899. I'll fix it myself since it's a one-liner.


decster:
I don't think current code works:
  cp "$FWDIR/conf/_.template" "$DISTDIR"/conf
I compiled and got following error:
cp: /Users/decster/projects/spark/conf/_.template: No such file or directory
I think wildcard \* should not in quote
 -cp "$FWDIR/conf/_.template" "$DISTDIR"/conf
+cp "$FWDIR"/conf/_.template "$DISTDIR"/conf

PR #917, opened 2013-09-10
Document libgfortran dependency for MLBase

pwendell:


shivaram:
Hey @pwendell  

Sorry, I totally forgot about this. Thanks for fixing this !

pwendell:
Hey @mateiz @atalwalkar I'm going to merge this so I can cut a RC for 0.8.0. Please feel free to amend this with a new PR.

shivaram:
@mateiz This also affects the unit tests as our tests right now fail if gfortran is not installed. I think it'll be better to just catch an UnsatisfiedLinkError in a base class and then ignore tests if that happens ? Any other ideas ?

mateiz:
Nah, I'd rather have them fail right now. I don't think it's a big deal to expect developers to get this.

The one thing we may want to do later is catch the error and change it to something nicer within Spark (to say MLlib requires libfortran).

PR #919, opened 2013-09-10
Add explicit jets3t dependency, which is excluded in hadoop-client

mateiz:
Version 0.7.1 is required for Hadoop 1.0.4 -- newer versions won't work. For newer Hadoops, if Hadoop starts using a new jets3t, users can manually depend on it.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -37,6 +37,11 @@
       <artifactId>hadoop-client</artifactId>
     </dependency>
     <dependency>
+      <groupId>net.java.dev.jets3t</groupId>
+      <artifactId>jets3t</artifactId>
+      <version>0.7.1</version>
+    </dependency>
+    <dependency>
       <groupId>org.apache.avro</groupId>
       <artifactId>avro</artifactId>
     </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -374,6 +374,11 @@
         </exclusions>
       </dependency>
       <dependency>
+        <groupId>net.java.dev.jets3t</groupId>
+        <artifactId>jets3t</artifactId>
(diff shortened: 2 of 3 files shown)


core/pom.xml
```diff
@@ -37,6 +37,11 @@
       <artifactId>hadoop-client</artifactId>
     </dependency>
     <dependency>
+      <groupId>net.java.dev.jets3t</groupId>
+      <artifactId>jets3t</artifactId>
+      <version>0.7.1</version>
```
rxin, line 42: Do we need an explicit version here?

pwendell:
I tested this with Hadoop2 as well and it worked well.

core/pom.xml
markhamstra, line 42: No.  The version should be left unspecified so that the dependency specified in the spark/pom.xml will be used.

mateiz, line 42: Yes we do actually. Code that linked against jets3t 0.7.1 doesn't work with 0.8 and 0.9.

markhamstra, line 42: Yes, but you've already specified the version in spark/pom.xml.  Duplicating the version specification in spark/core/pom.xml is not necessary and is bad practice unless you have a specific (and commented) reason to override the previously specified version.

jey, line 42: @markhamstra's right, the subproject POMs inherit the version number from the [`<dependencyManagement>` section of the root POM](https://github.com/mateiz/spark/blob/f117dc6d0d1349c28b7e1c8cf7729d1aafde5d25/pom.xml#L379), and should not explicitly mention the version in the subproject's POM.

PR #920, opened 2013-09-10
Fixing SPARK-602: PythonPartitioner

AndreSchumacher:
Currently PythonPartitioner determines partition ID by hashing a byte-array
representation of PySpark's key. This PR lets PythonPartitioner use the
actual partition ID, which is required e.g. for sorting via PySpark.


AndreSchumacher:
One thing I am still worried about is the deserialization overhead on the scala side but on the other hand one saves one hash function call. If this turns out to degrade performance one could use the old version as default and limit usage of this new one for the sorting.

jey:
Hi Andre, could we also add a unit test that exercises this case?

AndreSchumacher:
Sorry, thought I had replied to this yesterday. After removing the conventional way of partition ID determination for a key (and replacing it by an Exception, as suggested by @jey ) and rerunning the tests I'm now fairly confident that nothing else relies on the older way of determining partition IDs. I understand it's maybe too late to be merged prior to the 0.8 release but since this is blocking PySpark-based sorting it would be nice to have it in there.

mateiz:
Andre, we can add this in 0.8.1. Probably too late to make a significant change for 0.8.0.

rxin:
@mateiz Do you want us to merge this?

core/src/main/scala/org/apache/spark/util/Utils.scala
```diff
@@ -70,6 +70,13 @@ private[spark] object Utils extends Logging {
     return ois.readObject.asInstanceOf[T]
   }
 
+  /** Deserialize a Long value (used for {@link org.apache.spark.api.python.PythonPartitioner}) */
+  def deserializeLongValue(bytes: Array[Byte]) = {
+    val bis = new ByteArrayInputStream(bytes)
+    val ois = new DataInputStream(bis)
+    ois.readLong()
+  }
```
mateiz, line 78: Creating a ByteArrayInputStream and DataInputStream just to read a Long seems wasteful; I would just read 8 bytes and shift them together here.

core/src/main/scala/org/apache/spark/api/python/PythonPartitioner.scala
```diff
@@ -30,6 +32,9 @@ import org.apache.spark.util.Utils
  * function).  This can be ensured by using the Python id() function and maintaining a reference
  * to the Python partitioning function so that its id() is not reused.
  */
+
+// TODO: Fix inline doc (if PR accepted)
```
mateiz, line 36: You should delete this in the final PR

mateiz:
Hey Andre, sorry for the delay, but I made a few comments on this. Do you mind updating it? Also, you'll have to send the PR against https://github.com/apache/incubator-spark/ now.

AndreSchumacher:
Hi Matei, no worries, I know you guys been busy with the release and all that. I will do the changes and resubmit to the new repository.

mateiz:
Closing this since it was sent to the Apache repo.

PR #921, opened 2013-09-11
Fix HDFS access bug with assembly build.

pwendell:
Due to this change in HDFS:
https://issues.apache.org/jira/browse/HADOOP-7549

there is a bug when using the new assembly builds with newer versions of Hadoop. The symptom is 
that any HDFS access results in an exception saying "No filesystem for scheme 'hdfs'". This adds a
merge strategy in the assembly build which fixes the problem by ensuring that if two jars expose
the manifest (in this case, hadoop-common and hadoop-hdfs), the assembly plugin concatenates them.

### DIFF
--- a/assembly/pom.xml
+++ b/assembly/pom.xml
@@ -106,6 +106,12 @@
               <transformers>
                 <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
                 <transformer implementation="org.apache.maven.plugins.shade.resource.AppendingTransformer">
+                  <resource>META-INF/services/org.apache.hadoop.fs.FileSystem</resource>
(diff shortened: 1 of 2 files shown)


mateiz:
Hey Patrick, make sure to add this to the Maven build as well (assembly/pom.xml in there). There's already some stuff with custom merge strategies there.

pwendell:
Ah good point, also in sbt I should coalesce this with the existing mergeStrategy stuff... I'll update

mateiz:
Cool, that sounds good. If you test it and it works with both SBT and Maven go for it.

pwendell:
I tested this on my machine and on EC2 with maven and sbt. Seems to work!

PR #923, opened 2013-09-11
fix run-example script

haoyuan:
_assembly_[0-9T].jar doesn't work for spark-examples-assembly-0.8.0-incubating.jar

### DIFF
--- a/run-example
+++ b/run-example
@@ -39,14 +39,14 @@ fi
 # to avoid the -sources and -doc packages that are built by publish-local.
 EXAMPLES_DIR="$FWDIR"/examples
 SPARK_EXAMPLES_JAR=""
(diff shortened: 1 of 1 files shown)


mateiz:
Hey HY, this fix is actually not perfect because it will find those -sources or -doc JARs that the comment warns about. A better fix is to change the regular expression to [0-9Tg]. Super ugly but that's what it takes!

mateiz:
Awesome, thanks for spotting this!

mateiz:
Actually I will make one more commit after this to fix it on Windows too.

mateiz:
Actually never mind, the Windows build doesn't do anything special here, so I'll skip it.

PR #929, opened 2013-09-13
Use different Hadoop version for YARN artifacts.

pwendell:
This uses a seperate Hadoop version for YARN artifact. This means when people link against
spark-yarn, things will resolve correctly.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -82,6 +82,7 @@
     <slf4j.version>1.7.2</slf4j.version>
     <log4j.version>1.2.17</log4j.version>
     <hadoop.version>1.0.4</hadoop.version>
+    <yarn.version>0.23.7</yarn.version>
     <hbase.version>0.94.6</hbase.version>
 
     <PermGen>64m</PermGen>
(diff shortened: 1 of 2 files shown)


pwendell:
Jenkins, test this please.

pwendell:
Okay I test this locally because Jenkins hasn't recovered from the UCB power outage yet :)

jey:
User applications don't link against `spark-yarn`, only against `spark-core`. The `spark-yarn` package is only used internally by the assembly that gets submitted to the YARN cluster.

pwendell:
Hey @jey Some folks actually -1'd the release because we didn't publish the spark-yarn artifact, so I think somehow someone is linking against this. I don't totally understand the use case though...

pwendell:
btw, does this interfere with the existing stuff you were working on? Or are you just suggesting it might not be necessary given the way `spark-yarn` is used.

mateiz:
The problem wasn't as much that we didn't publish spark-yarn, but that the version in its POM was wrong because it wasn't handled by the Apache release plugin. So we do need to do this. It doesn't matter whether users link to it.

jey:
OK. We should also update the docs to mention that YARN users now need to do `mvn -Phadoop2-yarn -Dhadoop.version=2.0.5-alpha -Dyarn.version=2.0.5-alpha package`

pwendell:
Hey @jey  good call I just opened this PR:
https://github.com/mesos/spark/pull/931

jey:
@pwendell, publishing it should cause no harm; just that it's not useful for anything except the `spark-assembly` target.

PR #930, opened 2013-09-15
Add mapPartitionsWithIndex

holdenk:
Add mapPartitionsWithIndex

### DIFF
--- a/core/src/main/scala/org/apache/spark/api/java/JavaRDDLike.scala
+++ b/core/src/main/scala/org/apache/spark/api/java/JavaRDDLike.scala
@@ -71,9 +71,10 @@ trait JavaRDDLike[T, This <: JavaRDDLike[T, This]] extends Serializable {
    * Return a new RDD by applying a function to each partition of this RDD, while tracking the index
    * of the original partition.
    */
-  def mapPartitionsWithIndex(f: JFunction2[Int, T, R],
+  def mapPartitionsWithIndex[R: ClassManifest](f: JFunction2[Int, java.util.Iterator[T], java.util.Iterator[R]],
 			     preservesPartitioning: Boolean = false): JavaRDD[R] =
-    new JavaRDD(MapPartitionsWithIndexRDD(this, sc.clean(f), preservesPartitioning))
+    new JavaRDD(rdd.mapPartitionsWithIndex(((a,b) => f(a,asJavaIterator(b))),
+					   preservesPartitioning))
 
   /**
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/api/java/JavaRDDLike.scala
```diff
@@ -68,6 +68,15 @@ trait JavaRDDLike[T, This <: JavaRDDLike[T, This]] extends Serializable {
     new JavaRDD(rdd.map(f)(f.returnType()))(f.returnType())
 
   /**
+   * Return a new RDD by applying a function to each partition of this RDD, while tracking the index
+   * of the original partition.
+   */
+  def mapPartitionsWithIndex[R: ClassManifest](f: JFunction2[Int, java.util.Iterator[T], java.util.Iterator[R]],
+			     preservesPartitioning: Boolean = false): JavaRDD[R] =
```
rxin: Thanks. Can we change this to indent 4 characters? 

```
def mapPartitionsWithIndex[R: ClassManifest](
    f: JFunction2[Int, java.util.Iterator[T], java.util.Iterator[R]],
    preservesPartitioning: Boolean = false): JavaRDD[R] =
```

holdenk:
Done :)

rxin:
Sorry there was a misunderstanding. We use 2 space indenting in general, but for function declarations in which parameters don't fit in a single line, we use 4 spaes.

See the example I gave.

holdenk:
Fixed again (but don't see the example?)

rxin:
I've merged this into asf spark branch-0.8.

core/src/main/scala/org/apache/spark/api/java/JavaRDDLike.scala
```diff
@@ -68,6 +68,16 @@ trait JavaRDDLike[T, This <: JavaRDDLike[T, This]] extends Serializable {
     new JavaRDD(rdd.map(f)(f.returnType()))(f.returnType())
 
   /**
+   * Return a new RDD by applying a function to each partition of this RDD, while tracking the index
+   * of the original partition.
+   */
+  def mapPartitionsWithIndex[R: ClassManifest](
```
JoshRosen, line 74: I'm think that declaring `[R:ClassManifest]` introduces an implicit ClassManifest argument to this function, which will be problematic when calling it from Java.  Can you fix this and add a unit test to JavaRDDSuite to verify that this method is callable from Java without having to pass the manifests?

PR #932, opened 2013-09-15
Bumping Mesos version to 0.13.0

pwendell:


pwendell:
Okay, I tested this locally.

PR #933, opened 2013-09-15
Fix typo in Maven build docs

jey:
Fixes a typo introduced in #931, where "-P" was written instead of "-D"


pwendell:
Ah thanks Jey! Matei and I both missed this... :0

PR #934, opened 2013-09-17
create key and value in correct scope for HadoopRdd

bunions1:


rxin:
This is actually a common optimization used in Hadoop - to reuse the key, value objects to reduce the number of temporary objects created.

bunions1:
Oh. Thanks for responding. I raised a jira (spark-910) that I guess can be disregarded. 

Can you tell me is this XmlInputFormat implementation wrong? Or am I using it wrong some how. 

hf = sc.hadoopFile("hdfs://namenod.local/something.xml", XmlInputFormat.class, LongWritable.class, Text.class);
hf.take(5);
====> returns same things 5 times

https://github.com/apache/mahout/blob/ad84344e4055b1e6adff5779339a33fa29e1265d/examples/src/main/java/org/apache/mahout/classifier/bayes/XmlInputFormat.java

bunions1:
I believe I've figured it out. I ran across this other implementation of XmlInputFormat in the mahout project.

https://github.com/apache/mahout/blob/d522863754c682dc4bdf19951ed000126cb2c57e/integration/src/main/java/org/apache/mahout/text/wikipedia/XmlInputFormat.java

It looks like it takes care to create new key and value objects where as the first one doesn't. I originally assumed there was a bug in spark since the InputFormat class came from mahout, but this seems to confirm the problem is with the mahout code.

rxin:
Thanks for updating it. Closing this one for now.

PR #935, opened 2013-09-17
job killing in Spark

rxin:
This subsumes pull request #665  by @harsha2010

Different from #665 , this PR is not using interrupts to kill tasks because interrupts would rely on the user code properly handling the interrupts. One big counter example is Hadoop client would mark a data node as dead when it sees an interrupt during reads. Instead, this pull requests add a new volatile variable "interrupted" to the TaskContext class, and rely on the upper level code to check that flag. This can be done by simply wrapping an existing iterator with an InterruptibleIterator.

Todos:
1. Use InterruptibleIterator in all "input" iterators, including the ones in ParallelCollectionsRDD, HadoopRDD, NewHadoopRDD, ShuffleFetcher. (It might be easier to create an InterruptibleRDD class and just make the various input RDDs the parent of this new RDD.)
2. Design an interface for the user to be able to kill jobs. One idea is to add a new submitJob function in DAGScheduler, and for that to return a Future-like object that has a "cancel" or "kill" method. Note that this can be reused in the future to implement progress reporting feature.
3. Unit tests.

### DIFF
--- a/core/src/main/scala/org/apache/spark/SparkContext.scala
+++ b/core/src/main/scala/org/apache/spark/SparkContext.scala
@@ -20,6 +20,7 @@ package org.apache.spark
 import java.io._
 import java.net.URI
 import java.util.Properties
+import java.util.concurrent.Future
 import java.util.concurrent.atomic.AtomicInteger
 
 import scala.collection.Map
@@ -810,6 +811,24 @@ class SparkContext(
       localProperties.get)
     logInfo("Job finished: " + callSite + ", took " + (System.nanoTime - start) / 1e9 + " s")
     result
+  }
+
+  def submitJob[T, U, R](
+      rdd: RDD[T],
+      processPartition: Iterator[T] => U,
+      partitionResultHandler: (Int, U) => Unit,
+      resultFunc: () => R): Future[R] =
+  {
+    val callSite = Utils.formatSparkCallSite
+    val waiter = dagScheduler.submitJob(
+      rdd,
+      (context: TaskContext, iter: Iterator[T]) => processPartition(iter),
+      0 until rdd.partitions.size,
+      callSite,
+      allowLocal = false,
+      partitionResultHandler,
+      null)
+    new FutureJob(waiter, resultFunc)
   }
 
   /**
--- a/core/src/main/scala/org/apache/spark/SparkContext.scala
+++ b/core/src/main/scala/org/apache/spark/SparkContext.scala
@@ -813,6 +813,13 @@ class SparkContext(
   }
 
   /**
+   * Kill a running job.
+   */
+  def killJob(jobId: Int) {
+    dagScheduler.killJob(jobId)
+  }
+
+  /**
    * Clean a closure to make it ready to serialized and send to tasks
    * (removes unreferenced variables in $outer's, updates REPL variables)
    */
--- a/core/src/main/scala/org/apache/spark/SparkContext.scala
+++ b/core/src/main/scala/org/apache/spark/SparkContext.scala
@@ -816,6 +816,7 @@ class SparkContext(
   def submitJob[T, U, R](
       rdd: RDD[T],
       processPartition: Iterator[T] => U,
+      partitions: Seq[Int],
       partitionResultHandler: (Int, U) => Unit,
       resultFunc: () => R): Future[R] =
   {
@@ -823,7 +824,7 @@ class SparkContext(
     val waiter = dagScheduler.submitJob(
       rdd,
       (context: TaskContext, iter: Iterator[T]) => processPartition(iter),
-      0 until rdd.partitions.size,
+      partitions,
       callSite,
       allowLocal = false,
       partitionResultHandler,
--- a/core/src/main/scala/org/apache/spark/SparkContext.scala
+++ b/core/src/main/scala/org/apache/spark/SparkContext.scala
@@ -839,6 +839,12 @@ class SparkContext(
     dagScheduler.killJob(jobId)
   }
 
+  def killAllJobs() {
+    dagScheduler.activeJobs.foreach { job =>
+      killJob(job.jobId)
+    }
+  }
+
   /**
    * Clean a closure to make it ready to serialized and send to tasks
    * (removes unreferenced variables in $outer's, updates REPL variables)
--- a/core/src/main/scala/org/apache/spark/executor/Executor.scala
+++ b/core/src/main/scala/org/apache/spark/executor/Executor.scala
@@ -160,6 +160,7 @@ private[spark] class Executor(
     @volatile private var task: Task[Any] = _
 
     def kill() {
+      logInfo("Executor is trying to kill task " + taskId)
       killed = true
       if (task != null) {
         task.kill()
@@ -188,7 +189,7 @@ private[spark] class Executor(
         // If this task has been killed before we deserialized it, let's quit now. Otherwise,
         // continue executing the task.
         if (killed) {
-          logInfo("Task " + taskId + " was killed before it had a chance to run.")
+          logInfo("Executor killed task " + taskId)
           execBackend.statusUpdate(taskId, TaskState.KILLED, ser.serialize(TaskKilled))
         }
 
@@ -203,6 +204,7 @@ private[spark] class Executor(
 
         // If the task has been killed, let's fail it.
         if (task.killed) {
+          logInfo("Executor killed task " + taskId)
           execBackend.statusUpdate(taskId, TaskState.KILLED, ser.serialize(TaskKilled))
         }
 
--- a/core/src/main/scala/org/apache/spark/executor/Executor.scala
+++ b/core/src/main/scala/org/apache/spark/executor/Executor.scala
@@ -17,7 +17,7 @@
 
 package org.apache.spark.executor
 
-import java.io.{File}
+import java.io.File
 import java.lang.management.ManagementFactory
 import java.nio.ByteBuffer
 import java.util.concurrent._
@@ -31,7 +31,8 @@ import org.apache.spark.util.Utils
 
 
 /**
- * The Mesos executor for Spark.
+ * The backend executor for Spark. The executor maintains a thread pool and uses it to execute
+ * tasks.
  */
 private[spark] class Executor(
     executorId: String,
@@ -101,18 +102,38 @@ private[spark] class Executor(
   val executorSource = new ExecutorSource(this, executorId)
 
   // Initialize Spark environment (using system properties read above)
-  val env = SparkEnv.createFromSystemProperties(executorId, slaveHostname, 0, false, false)
+  private val env = SparkEnv.createFromSystemProperties(executorId, slaveHostname, 0,
+    isDriver = false, isLocal = false)
   SparkEnv.set(env)
   env.metricsSystem.registerSource(executorSource)
 
-  private val akkaFrameSize = env.actorSystem.settings.config.getBytes("akka.remote.netty.message-frame-size")
+  private val akkaFrameSize = {
+    env.actorSystem.settings.config.getBytes("akka.remote.netty.message-frame-size")
+  }
 
   // Start worker thread pool
   val threadPool = new ThreadPoolExecutor(
     1, 128, 600, TimeUnit.SECONDS, new SynchronousQueue[Runnable])
 
+  // Maintains the list of running tasks.
+  private val runningTasks = new ConcurrentHashMap[Long, TaskRunner]
+
   def launchTask(context: ExecutorBackend, taskId: Long, serializedTask: ByteBuffer) {
-    threadPool.execute(new TaskRunner(context, taskId, serializedTask))
+    val task = new TaskRunner(context, taskId, serializedTask)
+    runningTasks.put(taskId, task)
+    threadPool.execute(task)
+  }
+
+  def killTask(taskId: Long) {
+    val task = runningTasks.get(taskId)
+    if (task != null) {
+      task.kill()
+      // We remove the task also in the finally block in TaskRunner.run.
+      // The reason we need to remove it here is because killTask might be called before the task
+      // is even launched, and never reaching that finally block. ConcurrentHashMap's remove is
+      // idempotent.
+      runningTasks.remove(taskId)
+    }
   }
 
   /** Get the Yarn approved local directories. */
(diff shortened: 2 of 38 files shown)


core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -812,6 +813,38 @@ class SparkContext(
     result
   }
 
+  def submitJob[T, U, R](
```
markhamstra, line 816: Certainly not unique to this method, but this is another example of something that is publicly accessible in a SparkContext that probably shouldn't be.  Is there any plan to provide a more restricted SparkContext API (for use, e.g., in the shell), or are we going to continue to assume that if we don't tell people about something, then they won't misuse it?

rxin, line 816: Actually why wouldn't we want this till be accessible? It is at the same
level as runJob.

markhamstra, line 816: That's true, and we certainly want custom RDD developers and the like to be able to access runJob etc.  My question, though, is whether we really want all users of, e.g., spark-shell to be able to call runJob, submitJob, etc. directly.  It's something of a question of whether spark-shell is targeted at developers (who want to be able to do anything in the repl that they could in standalone code), or whether it is targeted at end users who just want/need to be able to do the kinds of higher-level things with RDDs that we typically show in examples.  Maybe the issue isn't so much that, within SparkContext itself, there needs to be another access level for methods as it is that we could use another kind of interactive tool/UI (something like iPython Notebooks, perhaps...) that doesn't have full access to all the SparkContext methods that many users shouldn't really be messing with directly.

core/src/main/scala/org/apache/spark/scheduler/JobWaiter.scala
```diff
@@ -17,29 +17,40 @@
...
 
-  private var jobFinished = false          // Is the job as a whole finished (succeeded or failed)?
-  private var jobResult: JobResult = null  // If the job is finished, this will be its result
+  // Is the job as a whole finished (succeeded or failed)?
+  private var _jobFinished = totalTasks == 0
+
+  def jobFinished = _jobFinished
+
+  // If the job is finished, this will be its result
+  private var jobResult: JobResult = null
```
markhamstra: Our old friend the zero-partition RDD is getting you -- `jobResult` stays `null`, so `waiter.awaitResult() match {...}` fails in `DAGScheduler.runJob`.

Here's one way to fix it:

``` scala
  // If the job is finished, this will be its result
  private var jobResult: JobResult = if (_jobFinished) {
    JobSucceeded
  } else {
    JobFailed(new Exception, None)
  }
```

You could leave it `null` in the `else` branch, but better to have `jobResult` always be matchable.

rxin: Thanks for the catch. Actually I am going to remove this jobResult member variable and just return JobSucceeded in awaitResults if _jobFinished == true.

rxin: Ah never mind I can't do that because i need the exception ... ok

core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
```diff
@@ -262,32 +265,42 @@ class DAGScheduler(
...
 
   /**
-   * Returns (and does not submit) a JobSubmitted event suitable to run a given job, and a
-   * JobWaiter whose getResult() method will return the result of the job when it is complete.
-   *
-   * The job is assumed to have at least one partition; zero partition jobs should be handled
-   * without a JobSubmitted event.
+   * Submit a job to the job scheduler and get a JobWaiter object back. The JobWaiter object
+   * can be used to block until the the job finishes executing or can be used to kill the job.
+   * If the given RDD does not contain any partitions, the function returns None.
```
markhamstra: I don't understand this last sentence.

rxin: Yea it's an old comment. Need to change that.

core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala
```diff
@@ -330,19 +329,60 @@ class DAGScheduler(
...
   }
 
   /**
+   * Kill a job that is running or waiting in the queue.
+   */
+  def killJob(jobId: Int): Unit = this.synchronized {
+    activeJobs.find(job => job.jobId == jobId).foreach(job => killJob(job))
+
+    def killJob(job: ActiveJob): Unit = this.synchronized {
+      logInfo("Killing Job and cleaning up stages %d".format(job.jobId))
```
markhamstra: `"Killing job %d and cleaning up stages".format(job.jobId))`

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -101,18 +102,38 @@ private[spark] class Executor(
...
   // Start worker thread pool
   val threadPool = new ThreadPoolExecutor(
     1, 128, 600, TimeUnit.SECONDS, new SynchronousQueue[Runnable])
 
+  // Maintains the list of running tasks.
+  private val runningTasks = new ConcurrentHashMap[Long, TaskRunner]
+
   def launchTask(context: ExecutorBackend, taskId: Long, serializedTask: ByteBuffer) {
-    threadPool.execute(new TaskRunner(context, taskId, serializedTask))
+    val task = new TaskRunner(context, taskId, serializedTask)
```
markhamstra: Nit picking, but I'm not a big fan of your naming here since `task` isn't really a `Task`, but rather a `TaskRunner`.  Since it is being used in close proximity, a simple `t` or `tr` would be preferable to the somewhat misleading `task`.

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -101,18 +102,38 @@ private[spark] class Executor(
...
+    runningTasks.put(taskId, task)
+    threadPool.execute(task)
+  }
+
+  def killTask(taskId: Long) {
+    val task = runningTasks.get(taskId)
```
markhamstra: See prior naming nit pick.

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -124,15 +145,26 @@ private[spark] class Executor(
...
   }
 
   class TaskRunner(context: ExecutorBackend, taskId: Long, serializedTask: ByteBuffer)
     extends Runnable {
 
+    @volatile private var killed = false
+    @volatile private var task: Task[Any] = _
+
+    def kill() {
+      // Note that there is a tiny possibiliy of raising here.
```
markhamstra: possibility

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -124,15 +145,26 @@ private[spark] class Executor(
...
+    localDirs
   }
 
   class TaskRunner(context: ExecutorBackend, taskId: Long, serializedTask: ByteBuffer)
     extends Runnable {
 
+    @volatile private var killed = false
+    @volatile private var task: Task[Any] = _
+
+    def kill() {
```
markhamstra, line 162: Have you worked through the potential race conditions here and in `Task.kill`?  Just wondering whether they need to be atomic or whether nothing bad can happen if, e.g., `killed` gets set to `true` and then the thread is interrupted.

core/src/main/scala/org/apache/spark/executor/Executor.scala
```diff
@@ -150,22 +182,29 @@ private[spark] class Executor(
...
+
+        // If this task has been killed before we deserialized it, let's quit now. Otherwise,
+        // continue executing the task.
+        if (killed) {
+          context.statusUpdate(taskId, TaskState.KILLED, ser.serialize(""))
+        }
+
         attemptedTask = Some(task)
-        logInfo("Its epoch is " + task.epoch)
+        logDebug("Its epoch is " + task.epoch)
```
markhamstra, line 197: Not the most informative message -- needs to identify the task at issue more clearly than just "Its".

core/src/main/scala/org/apache/spark/rdd/RDD.scala
```diff
@@ -562,6 +563,15 @@ abstract class RDD[T: ClassManifest](
   }
 
   /**
+   * Return a future for retrieving the results of a collect in an asynchronous fashion.
+   */
+  def collectAsync(): Future[Seq[T]] = {
```
markhamstra, line 568: So do we now need async versions for all actions?  Do we want to maintain parallel sets of blocking/non-killable and non-blocking/killable actions, or should everything be refactored to use just async/non-blocking/killable actions?

core/src/main/scala/org/apache/spark/rdd/PairRDDFunctions.scala
```diff
@@ -85,17 +85,20 @@ class PairRDDFunctions[K: ClassManifest, V: ClassManifest](self: RDD[(K, V)])
...
+        .interruptible()
     } else {
       // Don't apply map-side combiner.
       // A sanity check to make sure mergeCombiners is not defined.
       assert(mergeCombiners == null)
       val values = new ShuffledRDD[K, V, (K, V)](self, partitioner).setSerializer(serializerClass)
       values.mapPartitions(aggregator.combineValuesByKey, preservesPartitioning = true)
+        .interruptible()
     }
   }
```
markhamstra, line 104: Which means that you need this in `PairRDDFunctionsSuite`:

```
-    assert(deps.size === 2) // ShuffledRDD, ParallelCollection
+    assert(deps.size === 3) // ShuffledRDD, ParallelCollection, InterruptibleRDD
```

rxin, line 104: yea i know. im intentionally not changing the tests until i am happy with my design

markhamstra, line 104: Heh, okay.  But so far, that's the only test you've broken.

mateiz:
Reynold, did you run the spark-perf tests on this? Would be good to check that it adds no overhead.

rxin:
Just ran some perf tests. Didn't notice any performance degradation with this turned on.

without job killing:

```
scheduling-throughput, --num-tasks=10000 --num-trials=10 --inter-trial-wait=30, 5.198, 0.312, 4.78, 5.198, 5.084
scala-agg-by-key, --num-trials=10 --inter-trial-wait=30 --num-partitions=200 --reduce-tasks=200 --random-seed=5 --persistent-type=memory  --num-records=100000000 --unique-keys=10000 --key-length=10 --unique-values=500000 --value-length=10, 2.728, 0.754, 2.595, 3.116, 2.633
scala-sort-by-key, --num-trials=10 --inter-trial-wait=30 --num-partitions=20 --reduce-tasks=20 --random-seed=5 --persistent-type=memory  --num-records=10000000 --unique-keys=1000 --key-length=10 --unique-values=50000 --value-length=10, 4.343, 0.223, 3.841, 4.343, 3.971
scala-count, --num-trials=10 --inter-trial-wait=30 --num-partitions=200 --reduce-tasks=200 --random-seed=5 --persistent-type=memory  --num-records=100000000 --unique-keys=10000 --key-length=10 --unique-values=500000 --value-length=10, 0.273, 0.062, 0.245, 0.398, 0.273
scala-count-w-fltr, --num-trials=10 --inter-trial-wait=30 --num-partitions=200 --reduce-tasks=200 --random-seed=5 --persistent-type=memory  --num-records=100000000 --unique-keys=10000 --key-length=10 --unique-values=500000 --value-length=10, 0.552, 0.049, 0.497, 0.643, 0.497
```

with job killing run 1

```
scheduling-throughput, --num-tasks=10000 --num-trials=10 --inter-trial-wait=30, 5.148, 1.739, 4.841, 5.134, 8.928
scala-agg-by-key, --num-trials=10 --inter-trial-wait=30 --num-partitions=200 --reduce-tasks=200 --random-seed=5 --persistent-type=memory  --num-records=100000000 --unique-keys=10000 --key-length=10 --unique-values=500000 --value-length=10, 2.727, 0.170, 2.569, 3.163, 2.569
scala-sort-by-key, --num-trials=10 --inter-trial-wait=30 --num-partitions=20 --reduce-tasks=20 --random-seed=5 --persistent-type=memory  --num-records=10000000 --unique-keys=1000 --key-length=10 --unique-values=50000 --value-length=10, 4.405, 0.217, 3.928, 4.425, 3.977
scala-count, --num-trials=10 --inter-trial-wait=30 --num-partitions=200 --reduce-tasks=200 --random-seed=5 --persistent-type=memory  --num-records=100000000 --unique-keys=10000 --key-length=10 --unique-values=500000 --value-length=10, 0.271, 0.058, 0.246, 0.384, 0.269
scala-count-w-fltr, --num-trials=10 --inter-trial-wait=30 --num-partitions=200 --reduce-tasks=200 --random-seed=5 --persistent-type=memory  --num-records=100000000 --unique-keys=10000 --key-length=10 --unique-values=500000 --value-length=10, 0.578, 0.032, 0.544, 0.635, 0.616
```

with job killing run 2

```
scheduling-throughput, --num-tasks=10000 --num-trials=10 --inter-trial-wait=30, 5.423, 2.570, 4.846, 5.423, 5.376
scala-agg-by-key, --num-trials=10 --inter-trial-wait=30 --num-partitions=200 --reduce-tasks=200 --random-seed=5 --persistent-type=memory  --num-records=100000000 --unique-keys=10000 --key-length=10 --unique-values=500000 --value-length=10, 2.765, 0.150, 2.714, 3.161, 2.738
scala-sort-by-key, --num-trials=10 --inter-trial-wait=30 --num-partitions=20 --reduce-tasks=20 --random-seed=5 --persistent-type=memory  --num-records=10000000 --unique-keys=1000 --key-length=10 --unique-values=50000 --value-length=10, 4.354, 0.230, 3.954, 4.511, 4.168
scala-count, --num-trials=10 --inter-trial-wait=30 --num-partitions=200 --reduce-tasks=200 --random-seed=5 --persistent-type=memory  --num-records=100000000 --unique-keys=10000 --key-length=10 --unique-values=500000 --value-length=10, 0.313, 0.078, 0.245, 0.512, 0.245
scala-count-w-fltr, --num-trials=10 --inter-trial-wait=30 --num-partitions=200 --reduce-tasks=200 --random-seed=5 --persistent-type=memory  --num-records=100000000 --unique-keys=10000 --key-length=10 --unique-values=500000 --value-length=10, 0.537, 0.041, 0.476, 0.608, 0.476
```

rxin:
Moving this one to https://github.com/apache/incubator-spark/pull/29

PR #936, opened 2013-09-18
Standalone Scheduler fault recovery

aarondav:
Implements a basic form of Standalone Scheduler fault recovery. In particular,
this allows faults (Master failures) to be manually recovered from by means of restarting the
Master process on the same machine. This is the majority of the code necessary
for general fault tolerance, which will first elect a leader and then recover
the Master state.

In order to enable fault recovery, the Master will persist a small amount of state related
to the registration of Workers and Applications to disk. If the Master is started and
sees that this state is still around, it will enter Recovery mode, during which time it
will not schedule any new Executors on Workers (but it does accept the registration of
new Clients and Workers).

At this point, the Master attempts to reconnect to all Workers and Client applications
that were registered at the time of failure. After confirming either the existence
or nonexistence of all such nodes (within a certain timeout), the Master will exit
Recovery mode and resume normal scheduling.

### DIFF
--- a/core/src/main/scala/org/apache/spark/deploy/ExecutorDescription.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/ExecutorDescription.scala
@@ -15,12 +15,15 @@
  * limitations under the License.
  */
 
-package org.apache.spark.deploy.master
+package org.apache.spark.deploy
 
-private[spark] object WorkerState
-  extends Enumeration("ALIVE", "DEAD", "DECOMMISSIONED", "UNKNOWN") {
+private[spark] class ExecutorDescription(
+    val appId: String,
+    val execId: Int,
+    val cores: Int,
+    val state: ExecutorState.Value)
+  extends Serializable {
 
-  type WorkerState = Value
-
-  val ALIVE, DEAD, DECOMMISSIONED, UNKNOWN = Value
+  override def toString: String =
+    "ExecutorState(appId=%s, execId=%d, cores=%d, state=%s)".format(appId, execId, cores, state)
 }
--- a/core/src/main/scala/org/apache/spark/deploy/master/ApplicationInfo.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/master/ApplicationInfo.scala
@@ -23,29 +23,52 @@ import akka.actor.ActorRef
 import scala.collection.mutable
 
 private[spark] class ApplicationInfo(
-    val startTime: Long,
-    val id: String,
-    val desc: ApplicationDescription,
-    val submitDate: Date,
-    val driver: ActorRef,
-    val appUiUrl: String)
-{
-  var state = ApplicationState.WAITING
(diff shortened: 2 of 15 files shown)


mateiz:

core/src/main/scala/org/apache/spark/deploy/master/ApplicationInfo.scala
```diff
@@ -23,29 +23,52 @@ import akka.actor.ActorRef
 import scala.collection.mutable
 
 private[spark] class ApplicationInfo(
-    val startTime: Long,
```
AndreSchumacher, line 26: Here and below there appear to be a few unnecessary formatting changes (indentation mostly), which clutter the diffs.

AndreSchumacher:
I am wondering whether it would make more sense to design master failover around some of the more established methods, e.g., Zookeeper, to provide persistent state information. That could simplify this design and avoid potential problems.. such as what happens when old masters get reconnected after timeouts.

pwendell:
@AndreSchumacher if you look at the patch it uses an abstracted persistence layer. The current implementation only has a disk based implementation, but the next step is to add one that uses Zookeeper.

core/src/main/scala/org/apache/spark/deploy/ExecutorDescription.scala
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.deploy
+
+private[spark] class ExecutorDescription(
```
hsaputra, line 20: I would like to ask for comment header to describe how this class is being used. It is easy for initial developer to know where and how the class is used but would be harder for new guys coming to trace the code.

core/src/main/scala/org/apache/spark/deploy/master/FileSystemPersistenceEngine.scala
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Stores data in a single on-disk directory with one file per application and worker.
+ * Files are deleted when applications and workers are removed.
+ *
+ * @param dir Directory to store files. Created if non-existent (but not recursively).
+ * @param serialization Used to serialize our objects.
+ */
+private[spark] class FileSystemPersistenceEngine(
+  val dir: String,
```
hsaputra, line 35: Style nit, the constructor parameters should be indented in 4 characters (http://docs.scala-lang.org/style/declarations.html)

aarondav:
Thanks for your comments -- I've incorporated them, but am going to merge this change with the ZooKeeper ones and put both onto the ASF github. Thus, closing this change.

PR #937, opened 2013-09-18
Fix PR926 local properties issues in Spark Streaming like scenarios

jerryshao:
[PR926](https://github.com/mesos/spark/pull/926) changes local properties from `DynamicVariable` to `ThreadLocal`,   which cannot pass properties set in parent thread to child thread. Scenarios like Spark Streaming which separate `runJob()` function in child thread will always get `null` properties and pass to DAGScheduler. So thread local properties should be inheritable, also should be isolated in different threads.

### DIFF
--- a/core/src/main/scala/org/apache/spark/SparkContext.scala
+++ b/core/src/main/scala/org/apache/spark/SparkContext.scala
@@ -275,7 +275,8 @@ class SparkContext(
     }
   }
 
-  def getLocalProperty(key: String): String = Option(localProperties.get).map(_.getProperty(key)).getOrElse(null)
+  def getLocalProperty(key: String): String =
+    Option(localProperties.get).map(_.getProperty(key)).getOrElse(null)
 
   /** Set a human readable description of the current job. */
   def setJobDescription(value: String) {
--- a/core/src/main/scala/org/apache/spark/SparkContext.scala
+++ b/core/src/main/scala/org/apache/spark/SparkContext.scala
@@ -256,7 +256,9 @@ class SparkContext(
   private[spark] var checkpointDir: Option[String] = None
 
   // Thread Local variable that can be used by users to pass information down the stack
-  private val localProperties = new ThreadLocal[Properties]
+  private val localProperties = new InheritableThreadLocal[Properties] {
+    override protected def childValue(parent: Properties): Properties = new Properties(parent)
+  }
 
   def initLocalProperties() {
     localProperties.set(new Properties())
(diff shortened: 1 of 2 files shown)


core/src/test/scala/org/apache/spark/ThreadingSuite.scala
```diff
@@ -149,4 +149,40 @@ class ThreadingSuite extends FunSuite with LocalSparkContext {
...
+          assert(sc.getLocalProperty("test") === "parent")
+          sc.setLocalProperty("test", i.toString)
+          assert(sc.getLocalProperty("test") === i.toString)
+        }
+      }
+    }
+
+    threads.foreach(_.start())
+    assert(sc.getLocalProperty("test") === "parent")
+    assert(sc.getLocalProperty("Foo") === null)
```
rxin, line 193: Don't we need a barrier here to properly test this?

core/src/test/scala/org/apache/spark/ThreadingSuite.scala
```diff
@@ -149,4 +149,40 @@ class ThreadingSuite extends FunSuite with LocalSparkContext {
...
+    threads.foreach(_.start())
+
+    assert(sc.getLocalProperty("test") === null)
+  }
+
+  test("set and get local properties in parent-children thread") {
+    sc = new SparkContext("local", "test")
+    sc.setLocalProperty("test", "parent")
+
+    val threads = (1 to 5).map{ i =>
```
rxin: Adding a space before {

core/src/test/scala/org/apache/spark/ThreadingSuite.scala
```diff
@@ -149,4 +149,40 @@ class ThreadingSuite extends FunSuite with LocalSparkContext {
       fail("One or more threads didn't see runningThreads = 4")
     }
   }
+
+  test("set local properties in different thread") {
+    sc = new SparkContext("local", "test")
+
+    val threads = (1 to 5).map{ i =>
```
rxin: space before {

core/src/main/scala/org/apache/spark/SparkContext.scala
```diff
@@ -273,6 +275,8 @@ class SparkContext(
     }
   }
 
+  def getLocalProperty(key: String): String = Option(localProperties.get).map(_.getProperty(key)).getOrElse(null)
```
rxin: Make sure the line does not exceed 100 chars.

kayousterhout:
Thanks for fixing this -- sorry for not realizing this use case!  The fix looks good.

jerryshao:
Hi @rxin ,thanks for your reviewing. I think it's more reasonable to add a barrier to unit test, I've already update the patch.

rxin:
Jenkins, test this please.

rxin:
Jenkins, retest this please.

rxin:
We should also merge this back into 0.8 branch once 0.8 is released.

rxin:
I've merged this into asf spark branch-0.8.

PR #938, opened 2013-09-19
Fix issue with spark_ec2 seeing empty security groups

aarondav:
Under unknown, but occasional, circumstances, reservation.groups is empty
despite reservation.instances each having groups. This means that the
spark_ec2 get_existing_clusters() method would fail to find any instances.
To fix it, we simply use the instances' groups as the source of truth.

Note that this is actually just a revival of PR #827, now that the issue
has been reproduced.

### DIFF
--- a/ec2/spark_ec2.py
+++ b/ec2/spark_ec2.py
@@ -364,12 +364,12 @@ def get_existing_cluster(conn, opts, cluster_name, die_on_error=True):
   slave_nodes = []
   for res in reservations:
     active = [i for i in res.instances if is_active(i)]
-    if len(active) > 0:
-      group_names = [g.name for g in res.groups]
+    for inst in active:
+      group_names = [g.name for g in inst.groups]
       if group_names == [cluster_name + "-master"]:
-        master_nodes += res.instances
+        master_nodes.append(inst)
       elif group_names == [cluster_name + "-slaves"]:
-        slave_nodes += res.instances
+        slave_nodes.append(inst)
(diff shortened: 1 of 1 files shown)


pwendell:
Thanks Aaron, also thanks @lucarosellini who originally reported this.

JoshRosen:
It sounds like this may fix [SPARK-749](https://spark-project.atlassian.net/browse/SPARK-749), a similar issue that I ran into once but had trouble reproducing.

lucarosellini:
@pwendell @ilikerps 
Using a patch very similar to what has been merged, in certain circumstances I got duplicate entries in "master_nodes" and "slave_nodes". 
That's why I had to add commit bf693d41ff87dd23565db64b3f23fe59164cd605 to the original pull request #827.
During the next days I'll try to reproduce the problem and I'll let you know.

aarondav:
@lucarosellini 
I believe you ran into that issue because you were appending res.instances for every instance -- that is, you were adding every instance in a reservation len(res.instances) times. We should now avoid this by simply appending "inst" instead of "res.instances".

aarondav:
@JoshRosen 
SPARK-749 definitely looks like the same issue, or at very least the same solution. I'll mark it as resolved if you have no objections.

pwendell:
Ya lets resolve SPARK-749 seems like the same thing.

lucarosellini:
@ilikerps yes, you're right, I missed that. Thank you.

PR #939, opened 2013-09-20
Support putting files/archives in the hadoop distributed cache for spark on yarn

tgravescs:
Add options for to put files/archives into the hadoop distributed cache. The options --files and --archives were added. They are very similar to the Hadoop options.  

I also added an option to control the replication factor of the files uploaded into the staging directory. This includes the spark jar, app jar, and any distributed cache files.

we now attempt to cleanup the staging directory. There are some cases that it doesn't handle but it will work if the applications finish nicely, which is at least an improvement. 

Question - Is there a reason the spark examples all do the System.exit() at the end?


tgravescs:
I don't think the DriverSuite test failure is from my changes.  When I run it manually it passes. Looks like it fails sometimes in the other builds?

rxin:
Closing this one since it has been submitted to the asf repo.

PR #940, opened 2013-09-20
After unit tests, clear port properties unconditionally

ankurdave:
In MapOutputTrackerSuite, the "remote fetch" test sets spark.driver.port and spark.hostPort, assuming that they will be cleared by LocalSparkContext. However, the test never sets `sc`, so it remains null, causing LocalSparkContext to skip clearing these properties. Subsequent tests therefore fail with java.net.BindException: "Address already in use".

This pull request makes LocalSparkContext clear the properties even if sc is null.


rxin:
Thanks!

PR #941, opened 2013-09-21
Add "org.apache." prefix to packages in spark-class

aarondav:
Lacking this, the if/case statements never trigger on Spark 0.8.0+, leading to lost java opts.

### DIFF
--- a/spark-class
+++ b/spark-class
@@ -37,7 +37,7 @@ fi
 
 # If this is a standalone cluster daemon, reset SPARK_JAVA_OPTS and SPARK_MEM to reasonable
 # values for that; it doesn't need a lot
-if [ "$1" = "spark.deploy.master.Master" -o "$1" = "spark.deploy.worker.Worker" ]; then
+if [ "$1" = "org.apache.spark.deploy.master.Master" -o "$1" = "org.apache.spark.deploy.worker.Worker" ]; then
   SPARK_MEM=${SPARK_DAEMON_MEMORY:-512m}
   SPARK_DAEMON_JAVA_OPTS="$SPARK_DAEMON_JAVA_OPTS -Dspark.akka.logLifecycleEvents=true"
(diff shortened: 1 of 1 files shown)


pwendell:
Jenkins, test this please.

pwendell:
This looks good to me.

rxin:
We should also merge this into branch-0.8 after 0.8 is released.

hsaputra:
@rxin,

Do you mean to 0.8 branch in ASF Git repo?

We should stop merge pull requests to mesos/spark github repo since we are having 0.8 out the door and also do final merges to ASF Git.

Maybe adding message/warning in the README.md file for the mesos/spark github repo to send pull request to ASF github mirror repo should be added.

rxin:
Yea - I will merge these PRs manually into ASF git.

One thing I realized is I can't close pull requests on the ASF git, and also it doesn't tell me if a PR merges cleanly or not (likely because I don't have write access to the repo). Is there anyway to fix that?

hsaputra:
Cool, and I see some already duplicate the pull requests to ASF github repo =)

Yeah, I believe for now we (or pull submitter) need to manually close them since we cannot really push commit to ASF git repo via github. I think Apache Cordova has the same problem since they also use github flow to pull changes.

rxin:
I don't have permission to close those PRs. How do I get that?

pwendell:
I think we can close them if we put text in the commit message: "Closes
#444". I just spoke with ASF infra and closing requests we don't merge is
not supported yet.

pwendell:
Ah I think it might even close just by virtue of merging it in.

rxin:
Yes, they close automatically if we merge. The question is for PRs that we
don't want to merge. Looks like there is no way to do that right now unless
we ask the submitter to close.

rxin:
This has also been merged into branch-0.8

PR #943, opened 2013-09-25
Scala 2.10 with akka 2.2

ScrapCodes:
This is a latest sync with master.

Not all warnings are fixed (This is intended, because it creates unnecessary overhead during merge. )

Failing tests:
`DriverSuite` only for cluster mode. (which is related to fixing of https://www.assembla.com/spaces/akka/tickets/3621 in akka.)

Fixed test.
`UISuite`  

None of those tests should affect the functionality.

review by @mateiz


rxin:
Thanks

PR #944, opened 2013-09-27
SPARK-900 Use coarser grained naming for metrics

GraceH:
SPARK-900 Use coarser grained naming for metrics. 
Now the new metric name is formatted as {XXX.YYY.ZZZ.COUNTER_UNIT},  XXX.YYY.ZZZ represents the group name, which can group several metrics under the same Ganglia view.


pwendell:
Hey Grace, can you close this and open it instead against the new repo?

PR #945, opened 2013-09-28
fix a bug of using wildcard in make-distribution.sh; ask sbt to check local maven repo in SparkBuild.scala

Du-Li:
fixed a wildcard bug in make-distribution.sh:
         move the wildcard out of the quotes

ask sbt to check local maven repo in SparkBuild.scala:
       sbt uses ivy which by default only checks ~/.ivy2. need to tell sbt to also check ~/.m2 because mesos make maven-install only publishes to ~/.m2


Du-Li:
Sorry for the mess. I was trying to create one pull request for each of the two latest fixes I made. But it somehow combined many irrelevant commits, which is not what I intended.

hsaputra:
Hi Du,

Would you mind filing the pull request to the Apache Spark repo instead

Thx,

Henry

Du-Li:
Sure. Will do today.

Du

Du-Li:
done

PR #946, opened 2013-09-30
Fixed non termination of Executor backend, when sc.stop is not called and system.exit instead.

ScrapCodes:
A simple fix suggestion.


mateiz:
Looks good, thanks.

